Issue #9, opened 2017-11-23
unify Scheme with Type


johnynek:
Scheme is basically a top-level, non-nested `forall` type. It is a bit ugly that we have to separate this out. It may be related to #5 and we can't easily remove Scheme until we address that.

johnynek:
#61 did this.

PR #29, opened 2018-09-19
Add TypeLambda to the Type AST

johnynek:
This is part of #9 but not the whole story.

It adds TypeLambda back to the Type AST, which is what Scheme really is. It does not yet remove Scheme since we are using in many places, but Scheme should be a synonym for Type.

This does not yet do anything in Infer with TypeLambda, which I think has to be done if we are going to handle unifying them at all. I think the unification should be obvious: you can only unify with a Var or another TypeLambda, and if you unify with the TypeLambda you replace the unbound parameter to a common value.

@snoble can you take a look and let me know what you think if you have time? I think you are more concerned with semantics and syntax than typechecking, but I'd be interested in your code review.


johnynek:
One thing that comes up here is that Arrow has to go. It should probably be replaced with `Bosatsu.Predef.Fn` which has kind `* -> * -> *`, which is to say `Arrow(a, b)` becomes `TypeApply(a, TypeApply(b, Declared(Fn)))`

Issue #6, opened 2017-11-23
actually use the def type annotations


johnynek:
currently, def annotations are totally ignored by inference.

To use them, we need a new Expr which is a type-annotated lambda and a type annotationed expr.

johnynek:
This is now easy to do, we just need to improve the translation from declaration to expression.

Issue #5, opened 2017-11-23
implement inference of rank-n types


johnynek:
https://www.microsoft.com/en-us/research/publication/practical-type-inference-for-arbitrary-rank-types/

We might want something like:

```
struct TypeEq[a, b](subst: forall f. f[a] -> f[b])
struct Monad[f](pure: forall a. a -> f[a], bind: forall a. f[a] -> (a -> f[b]) -> f[b])
```
this is a pretty expert thing, so a low priority, but currently, we don't have syntax for this and inference wouldn't work.

johnynek:
#61 merged support for this.

PR #32, opened 2018-09-20
Add some more unification tests

johnynek:
I want `forall a. Int` to unify with `Int` and other simplifications of type lambdas.


johnynek:
This moves us slightly towards:

https://github.com/johnynek/bosatsu/issues/5

We look at #6 and #5 together since the solution to the latter seems to be the former.

I don't think you could actually use a
```
struct Monad(pure: forall a. a -> f[a], bind: forall a, b. f[a] -> (a -> f[b]) -> f[b])
```
Because I'm not sure we would ever infer the right types, but that would be a good next exercise to check.

PR #34, opened 2018-09-23
Add an implementation of rank-n type inference

johnynek:
This is a direct port of the haskell code in the paper in #5 

This is not yet used in any bosatsu code. We will also need to follow the paper to expand how to handle matches, which are not included in the simple language used in the appendix of the paper.

Rather than using IO, I introduce a RefSpace monad to support mutable references, which their algorithm uses.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/rankn/Ref.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/rankn/Ref.scala
@@ -1,0 +1,84 @@
+package org.bykn.bosatsu.rankn
+import cats.Monad
+import cats.data.State
+import scala.collection.mutable.{Map => MutableMap}
+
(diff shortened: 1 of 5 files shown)


johnynek:
@snoble I just can't seem to think about anything else until we get the type inference on the level I want... :)

So, I guess I'm going to knock out support for rank-n types, which are needed (I think) to simulate typeclasses with structs like:
```
struct Monad(pure: forall a. a -> f[a], bind: forall a, b. f[a] -> (a -> f[b]) -> f[b])
```
If my understanding is correct, this is a rank2 type. We have an understood `forall f.` on the type, but the parameters have separate `forall` which we can't bubble up to the top.

It may not seem critical to be able to do this immediately, but I'm a bottom up kind of guy, and getting the inference in a good place is something I want to do.

PR #41, opened 2018-09-26
Normalization

snoble:
Probably needs a lot of clean up and splitting

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Main.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Main.scala
@@ -148,8 +148,9 @@ object MainCommand {
       val norm = Normalization(packs)
       norm.normalizeLast(mainPackage) match {
         case None => MainResult.Error(1, List("found no main expression"))
-        case Some((expr, scheme)) =>
-          MainResult.Success(List(s"$expr: $scheme"))
+        case Some((Left(_), _)) => MainResult.Error(1, List("found no main expression"))
+        case Some((Right(expr), scheme)) =>
+          MainResult.Success(List(s"${expr.mapTag(_ => ())}: $scheme"))
       }
     }
   }
--- a/core/src/main/scala/org/bykn/bosatsu/Main.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Main.scala
@@ -112,7 +112,7 @@ object MainCommand {
     def run() =
       blockingQueue(typeCheck(inputs, externals).flatMap { case (ext, packs, _) =>
         val ev = Evaluation(packs, Predef.jvmExternals ++ ext)
-        ev.evaluateLast(mainPackage) match {
+        ev.evaluateLast(mainPackage, Map()) match {
           case None => MainResult.Error(1, List("found no main expression"))
           case Some((eval, scheme, ne)) =>
             val res = eval.value
@@ -144,7 +144,7 @@ object MainCommand {
     def run() =
       blockingQueue(typeCheck(inputs, externals).flatMap { case (ext, packs, _) =>
         val ev = Evaluation(packs, Predef.jvmExternals ++ ext)
-        ev.evaluateLast(mainPackage) match {
+        ev.evaluateLast(mainPackage, Map()) match {
           case None => MainResult.Error(1, List("found no main expression"))
           case Some((eval, scheme, _)) =>
             val res = eval.value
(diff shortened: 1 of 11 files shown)


core/src/main/scala/org/bykn/bosatsu/Main.scala
```diff
@@ -44,18 +47,37 @@ object Std {
...
+  def waitReturn(a: Any): Any = {
+    println(s"a waitReturn ${a.toString}")
+    a
+  }
+
+  import java.net.Socket
+  import java.io.OutputStreamWriter
+
+  def expensiveFunction(a: Any, b: Any): Any = {
+    val socket = new Socket("127.0.0.1", 5555)
```
johnynek, line 66: what is this?

core/src/main/scala/org/bykn/bosatsu/NameKind.scala
```diff
@@ -1,52 +1,54 @@
 package org.bykn.bosatsu
 
-sealed abstract class NameKind
+sealed abstract class NameKind[T]
```
johnynek, line 3: can we send this as a PR by itself?

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,241 @@
+package org.bykn.bosatsu
+
+import cats.data.{NonEmptyList, State}
+import com.stripe.dagon.Memoize
+import cats.Id
+import cats.implicits._
+
+sealed abstract class NormalExpression {
```
johnynek, line 8: don't need the `{ }`

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,241 @@
...
+
+}
+
+object NormalExpression {
```
johnynek, line 12: this seems like it could be a separate PR that we could test independently, which would be nice.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,241 @@
...
+  case class App(fn: NormalExpression, arg: NormalExpression) extends NormalExpression
+  case class ExternalVar(pack: PackageName, defName: String) extends NormalExpression
+  case class Match(arg: NormalExpression, branches: NonEmptyList[(Int, NormalExpression)]) extends NormalExpression
+  case class LambdaVar(index: Int) extends NormalExpression
+  case class Lambda(expr: NormalExpression) extends NormalExpression
+  case class Struct(enum: Int, args: List[NormalExpression]) extends NormalExpression
+  case class Literal(lit: Lit) extends NormalExpression
```
johnynek, line 19: do we need a separate AST here or could we reuse `Expr[T]` with an appropriate `T` value?

project/Dependencies.scala
```diff
@@ -8,4 +8,5 @@ object Dependencies {
   lazy val paiges = "org.typelevel" %% "paiges-core" % "0.2.0"
   lazy val scalaTest = "org.scalatest" %% "scalatest" % "3.0.3"
   lazy val scalaCheck = "org.scalacheck" %% "scalacheck" % "1.13.5"
+  lazy val betterFiles = "com.github.pathikrit"  %% "better-files" % "3.5.0"
```
johnynek, line 11: do we really need this, it would be great to minimize the deps.

snoble:
Oh my gosh, I clearly need to setup my github notifications better because I didn't see these comments until today. I'm going to close this PR today because this is turning basically into a proof of concept and I'll re-implement on a modern version soon.

Reading through the comments though!

core/src/main/scala/org/bykn/bosatsu/Main.scala
snoble: Ug... Shouldn't have been part of the PR. It's just there for the demo

core/src/main/scala/org/bykn/bosatsu/NameKind.scala
snoble: If you would find this useful I can still do this

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: I tried so hard to make this work. The problem is that NormalExpression doesn't have a `Let` and it's `Lambda` doesn't have a variable name. And it introduces a `LambdaVar` for de Brujin notation. My fear is if I just have all of these in one AST and then just try to not let the `Let` creep into the `NormalExpression`s I won't have the compiler checking my work

Issue #15, opened 2017-11-23
Improve pattern matching


johnynek:
Currently the only kinds of pattern matches are a full list of all constructors with each variable bound or replaced with `_`.

It would be nice to support:
```
enum List:
  Empty
  NonEmpty(head: a, tail: List[a])

x = NonEmpty(1, NonEmpty(2, Empty))

match x:
  NonEmpty(a, NonEmpty(b, _)):
    f(a, b)
  _:
    0
```
So, this shows two features:
1. patterns inside the pattern.
2. a catch-all `_` pattern.

The only challenge here is totality checking, which may not be too much work actually.

An additional thing is to allow if statements:
```
match x:
  NonEmpty(x, _) if x > 0:
    foo(x)
  NonEmpty(x, _):
    bar(x)
  Empty:
    baz
```

johnynek:
It would be nice to support python 3 style list pattern matches, where `*a` matches a list of the needed size.

E.g.
```
x = [1, 2, 3]

match x:
  [a, *b, c]:
    # now a = 1, b = [2], c = 3
    foo(b)
  _:
    "no middle part"
```
I would probably rather a syntax like `[a, b..., c]` but since python 3 seems to have something close to this, I guess I shouldn't stray.

cc @non what do you think? Should I use the `*` syntax here?

johnynek:
actually, everything except the list pattern matching literal has been done already.

PR #42, opened 2018-09-27
Add Pattern to rank-n support

johnynek:
This continues following the paper into section 7 where patterns and matching is discussed.

Unfortunately the paper does not include inference rules for a match statement, and my current understanding is only enough to follow the directions in the paper, not prove results. So, I can't yet see the right way to extend the inference rules for the case of a match.

In the case of `if` we know the type of the argument, but in the case of `match` we have to infer the type of the argument.

In addition to the above difficulty, we also have the problem that parametric type constructors `struct Option[a](a: a)` cannot be supported currently because we have no way to remember if we have an `Option[Int]` or `Option[String]`. For this, we need, I think a TypeApply term in the rankn.Type ast.


johnynek:
relates to #15

PR #49, opened 2018-09-29
only one Pattern type

johnynek:
Pattern is pretty easy to unify the prior and rankn implementations.

Next is probably getting the rankn to use Expr instead of Term.


johnynek:
relates to #15

PR #51, opened 2018-09-30
Add type annotation to patterns

johnynek:
Moving towards replacing the old type inference, we need a way to fully annotate the types of the inferred expressions.

The old approach was to embed the annotation on each expression node on the tag. It is not 100% clear this was ever totally good enough for us.

We need to know the types during evaluation if we want to use compact representations of structs and enums, because we need to know what index a field name corresponds to, or what name a variant index refers to.

Section 8 of then rankn paper (see #5) discusses type directed translation. In this approach, when we infer, we also emit a new AST that is fully annotated such that inference of types is trivial. This is what we want to move towards, but to do this, it seemed likely we would need to implement type annotations in patterns, so I went ahead and did that.

This does complicate the Pattern type a bit, because at different phases, we have different information. On approach, which I'm still not sure is wrong, is to have different ASTs at different phases. We use this approach in a few places (e.g. Declaration vs Expr or TypeRef vs Type). This approach is nice when there is an injective mapping when we want to rule out some kinds of recursions in either the syntax or the internal representation, but it has the cost of code duplication. On the other hand, Pattern does not (yet) have the property that the syntactic and internal representations are different except the way we refer to type names (are they fully qualified or not).

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
@@ -63,12 +63,12 @@ sealed abstract class Declaration {
           Doc.char(q) + Doc.text(Parser.escape(Set(q), str)) + Doc.char(q)
       case Match(typeName, args) =>
         val pid = Document[Padding[Indented[Declaration]]]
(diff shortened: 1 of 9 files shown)


johnynek:
relates to #15

PR #66, opened 2018-10-11
Add foldLeft and Evaluation test

johnynek:
Somehow when we pass direct values which are external vs wrapping them in lambdas, the result is different.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Externals.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Externals.scala
@@ -28,22 +28,21 @@ sealed abstract class FfiCall {
     Eval.later {
       val cls = Class.forName(clsName)
       val args = FfiCall.getJavaType(t).toArray
-      //println(s"$this $t -> ${args.toList}")
       val m = cls.getMethod(parts.last, args.init :_*)
       val inst = instFn(cls)
 
       def invoke(tpe: rankn.Type, args: List[Any]): Any =
         tpe match {
           case rankn.Type.ForAll(_, t) => invoke(t, args)
-          case rankn.Type.Fun(_, tail) =>
+          case rankn.Type.Fun(a, tail) =>
             new Fn[Any, Any] {
+              val tpeStr = TypeRef.fromType(tpe).get.toDoc.render(80)
+              override def toString = s"$tpeStr, args: $args"
               def apply(x: Any) = {
-                println(s"x = $x")
                 invoke(tail, x :: args)
               }
             }
           case _ =>
-            println(s"$m.invoke(${args.reverse})")
             m.invoke(inst, args.reverse.toArray.asInstanceOf[Array[AnyRef]]: _*)
         }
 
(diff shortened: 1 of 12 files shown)


core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -76,10 +76,24 @@ x = Some(1)
...
+
+  test("do a fold") {
+    evalTest(
+      List("""
+package Foo
+
+three = NonEmptyList(1, NonEmptyList(2, EmptyList))
+
+sum0 = three.foldLeft(0, add)
+sum1 = three.foldLeft(0, \x, y -> add(x, y))
```
johnynek: this one is correct, the one above is returning `4` and it looks like `add` is working by applying the second argument twice.

core/src/main/scala/org/bykn/bosatsu/Externals.scala
```diff
@@ -61,13 +67,15 @@ object FfiCall {
   def getJavaType(t: rankn.Type): List[Class[_]] = {
     def loop(t: rankn.Type, top: Boolean): List[Class[_]] = {
       t match {
-        case t if t == rankn.Type.IntType => classOf[java.lang.Integer] :: Nil
-        case t if t == rankn.Type.BoolType => classOf[java.lang.Boolean] :: Nil
+        //case t if t == rankn.Type.IntType => classOf[java.lang.Integer] :: Nil
+        //case t if t == rankn.Type.BoolType => classOf[java.lang.Boolean] :: Nil
```
johnynek, line 71: we can revisit the typing of the FFI later, but it looks like if we don't keep things as anyref, then things will be tricky to get right in generic contexts since I think we are currently resolving the final types. So, `foldLeft` in some contexts is using `Int` all the way through.

core/src/main/scala/org/bykn/bosatsu/TypedExpr.scala
```diff
@@ -127,14 +158,70 @@ object TypedExpr {
...
+  def coerceFn(arg: Type, result: Type.Rho, coarg: Coerce, cores: Coerce): Coerce =
     new FunctionK[TypedExpr, TypedExpr] { self =>
-      def apply[A](expr: TypedExpr[A]) =
-        AnnotatedLambda("x", arg, cores(App(expr, coarg(Var("x", arg, expr.tag)), result, expr.tag)), expr.tag)
+      def apply[A](expr: TypedExpr[A]) = {
+        /**
+         * We have to be careful not to collide with the free vars in expr
+         */
+        val free = freeVars(expr :: Nil).toSet
+        val name = Type.allBinders.iterator.map(_.name).filterNot(free).next
```
johnynek, line 218: this was the bug: we can't blindly assign variable names because we may be shadowing if we do so (obviously). I copied the paper here without much thought.

I think the TypedExpr work is still somewhat weak.

core/src/main/scala/org/bykn/bosatsu/TypeRef.scala
```diff
@@ -86,6 +87,26 @@ object TypeRef {
 
   case class TypeLambda(params: NonEmptyList[TypeVar], in: TypeRef) extends TypeRef
 
+  def fromType(tpe: Type): Option[TypeRef] = {
```
johnynek: this can maybe better be:
```scala
def fromTypeA[F[_]: Apply](tpe: Type, orElse: F[TypeRef])(fn: Type.Const.Defined => F[TypeRef]): F[TypeRef]
```
then we can use it in more contexts.

PR #72, opened 2018-10-12
Add the Value ADT to make Evaluation safer

### LINKED ISSUE #69: add an ADT to Evaluation
I thought about having an ADT in Evaluation, but then didn't mostly due to some ill-conceived anxiety about performance.

Instead I am just returning `Any` but sometimes pattern matching. This is almost certainly a bad idea because we have given 0 attention to performance it without compilation it is likely to be very slow.

We could add something like:
```scala
sealed abstract class Value
object Value {
  sealed abstract class ProductValue extends Value
  case object UnitValue extends ProductValue
  case class ConsProduct(head: Value, tail: Eval[ProductValue]) extends ProductValue
  case class SumValue(variant: Int, data: Eval[Value]) extends Value
  // it would be nice to give FnValue a good equals method, which
  // we can do better than scala because we have the enviroment it captures and the ast.
  case class FnValue(toFunction: Value => Eval[Value]) extends Value
  case class ExternalValue(toAny: Any) extends Value
}
```

johnynek:
closes #69

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
@@ -7,13 +7,104 @@ import cats.implicits._
 
 import org.bykn.bosatsu.rankn.Type
 
+object Evaluation {
+  import Value._
+
+
+  /**
+   * If we later determine that this performance matters
+   * and this wrapping is hurting, we could replace
+   * Value with a less structured type and put
+   * all the reflection into unapply calls but keep
+   * most of the API
+   */
+  sealed abstract class Value {
+    def asFn: FnValue =
+      this match {
+        case f@FnValue(_) => f
+        case other => sys.error(s"invalid cast to Fn: $other")
+      }
+  }
+
+  object Value {
+    sealed abstract class ProductValue extends Value {
+      def toList: List[Value] =
+        this match {
+          case UnitValue => Nil
+          case ConsValue(head, tail) => head :: tail.toList
+        }
+    }
+
+    object ProductValue {
+      def fromList(ps: List[Value]): ProductValue =
+        ps match {
+          case Nil => UnitValue
+          case h :: tail => ConsValue(h, fromList(tail))
+        }
+    }
+
+    case object UnitValue extends ProductValue
+    case class ConsValue(head: Value, tail: ProductValue) extends ProductValue
+    case class SumValue(variant: Int, value: ProductValue) extends Value
+    case class FnValue(toFn: Value => Eval[Value]) extends Value {
+      def apply(a: Value): Eval[Value] = toFn(a)
+    }
+    case class ExternalValue(toAny: Any) extends Value
+
+    val False: Value = SumValue(0, UnitValue)
+    val True: Value = SumValue(1, UnitValue)
+
+    object VInt {
+      def apply(v: Int): Value = ExternalValue(java.lang.Integer.valueOf(v))
+      def unapply(v: Value): Option[Int] =
+        v match {
+          case ExternalValue(v: java.lang.Integer) => Some(v.intValue)
+          case _ => None
+        }
+    }
+
+    object Str {
+      def apply(str: String): Value = ExternalValue(str)
+      def unapply(v: Value): Option[String] =
+        v match {
+          case ExternalValue(str: String) => Some(str)
+          case _ => None
+        }
+    }
+
+    object VOption {
+      def unapply(v: Value): Option[Option[Value]] =
+        v match {
+          case UnitValue =>
+            Some(None)
+          case ConsValue(head, UnitValue) =>
+            Some(Some(head))
+          case _ => None
+        }
+    }
+
+    object VList {
+      def unapply(v: Value): Option[List[Value]] =
+        v match {
+          case SumValue(0, UnitValue) => Some(Nil)
+          case SumValue(1, ConsValue(head, ConsValue(rest, UnitValue))) =>
+            unapply(rest).map(head :: _)
+          case _ => None
+        }
+    }
+  }
+}
+
 case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
-  def evaluate(p: PackageName, varName: String): Option[(Eval[Any], Type)] =
+  import Evaluation.Value
+  import Value._
+
+  def evaluate(p: PackageName, varName: String): Option[(Eval[Value], Type)] =
     pm.toMap.get(p).map { pack =>
       eval((Package.asInferred(pack), Left(varName), Map.empty))
     }
 
-  def evaluateLast(p: PackageName): Option[(Eval[Any], Type)] =
+  def evaluateLast(p: PackageName): Option[(Eval[Value], Type)] =
     for {
       pack <- pm.toMap.get(p)
       (_, expr) <- pack.program.lets.lastOption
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -204,99 +312,86 @@ case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
...
-  private def constructor(c: ConstructorName, dt: rankn.DefinedType): Any = {
+  private def constructor(c: ConstructorName, dt: rankn.DefinedType): Eval[Value] = {
     val (enum, arity) = dt.constructors
       .toList
       .iterator
       .zipWithIndex
       .collectFirst { case ((ctor, params, resType), idx) if ctor == c => (idx, params.size) }
       .get // the ctor must be in the list or we wouldn't typecheck
 
+    val singleItemStruct = dt.constructors.lengthCompare(1) == 0
```
johnynek: move this function to DefinedType.

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -7,13 +7,104 @@ import cats.implicits._
...
+          case UnitValue =>
+            Some(None)
+          case ConsValue(head, UnitValue) =>
+            Some(Some(head))
+          case _ => None
+        }
+    }
+
+    object VList {
+      def unapply(v: Value): Option[List[Value]] =
```
johnynek, line 104: this should be a cons match: `Option[(Value, Value)]` to return head and tail, that is how we actually use it, in most cases.

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -26,26 +117,28 @@ case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
...
-      def toTest(a: Any): Test =
+//
+      def toTest(a: Value): Test =
         a match {
-          case (0, assertValue :: Nil) => Test.Assert(toBool(assertValue))
-          case (1, (label: String) :: test :: Nil) => Test.Label(label, toTest(test))
-          case (2, listOfTests :: Nil) => Test.TestList(toList(listOfTests))
+          case SumValue(0, ConsValue(assertValue, _)) =>
+            Test.Assert(toBool(assertValue))
+          case SumValue(1, ConsValue(Str(label), ConsValue(test, _))) =>
```
johnynek, line 142: use the cons match function we should add.

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -92,26 +185,41 @@ case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
             bindEnv(arg, (p, next) :: tail, acc)
           case (Pattern.PositionalStruct((pack, ctor), items), next) :: tail =>
             // let's see if this matches
-            arg match {
-              case (enumId: Int, params: List[Any]) =>
-                val cname = dt.constructors(enumId)._1
-                if (cname == ctor) {
+            val optParams =
+              if (dt.constructors.lengthCompare(1) == 0) {
```
johnynek: again here is the `isStruct` check.

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -150,22 +258,22 @@ case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
          val earg = recurse((p, Right(arg), env))._1
          (for {
            fn <- efn
-           afn = fn.asInstanceOf[Fn[Any, Any]] // safe because we typecheck
+           afn = fn.asInstanceOf[FnValue] // safe because we typecheck
```
johnynek: we have `fn.asFn`

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -150,22 +258,22 @@ case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
...
+         } yield res, resT)
        case a@AnnotatedLambda(name, argt, expr, _) =>
-         val fn = new Fn[Any, Any] {
-           override def toString = a.repr
-           def apply(x: Any) =
-             recurse((p, Right(expr), env + (name -> x)))._1.value
-         }
+         val fn =
+           FnValue { x =>
+             recurse((p, Right(expr), env + (name -> x)))._1
```
johnynek, line 285: if we separate this function do `(Package, Ref) => Map[String, Eval[Value]] => Eval[Value]` we can run the outer function before we get `x` then do less recomputation on each `x` value.

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -150,22 +258,22 @@ case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
...
+           }
          (Eval.now(fn), Type.Fun(argt, expr.getType))
        case Let(arg, e, in, _) =>
          (recurse((p, Right(e), env))._1.flatMap { ae =>
            recurse((p, Right(in), env + (arg -> ae)))._1
          }, in.getType)
-       case Literal(Lit.Integer(i), tpe, _) => (Eval.now(i), tpe)
-       case Literal(Lit.Str(str), tpe, _) => (Eval.now(str), tpe)
+       case Literal(Lit.Integer(i), tpe, _) => (Eval.now(VInt(i)), tpe)
+       case Literal(Lit.Str(str), tpe, _) => (Eval.now(Str(str)), tpe)
```
johnynek: we should have `Value.fromLit(l: Lit): Value`

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -204,99 +312,86 @@ case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
...
+    val singleItemStruct = dt.constructors.lengthCompare(1) == 0
+
     // TODO: this is a obviously terrible
     // the encoding is inefficient, the implementation is inefficient
-    def loop(param: Int, args: List[Any]): Any =
-      if (param == 0) (enum, args.reverse)
-      else new Fn[Any, Any] {
-        override def toString = s"Constructor($c, $dt)"
-        def apply(a: Any) = loop(param - 1, a :: args)
+    def loop(param: Int, args: List[Value]): Value =
```
johnynek, line 343: `@annotation.tailrec`

core/src/main/scala/org/bykn/bosatsu/Externals.scala
```diff
@@ -31,19 +33,16 @@ sealed abstract class FfiCall {
...
-              override def toString = s"$tpeStr, args: $args"
-              def apply(x: Any) = {
-                invoke(tail, x :: args)
-              }
+            Value.FnValue { x =>
+              Eval.always(invoke(tail, x :: args))
             }
           case _ =>
             m.invoke(inst, args.reverse.toArray.asInstanceOf[Array[AnyRef]]: _*)
+              .asInstanceOf[Value]
```
johnynek: maybe we should change the FFI to just require using `Evaluation.Value`, which removes these casts.

core/src/main/scala/org/bykn/bosatsu/Predef.scala
```diff
@@ -69,38 +68,40 @@ object Predef {
...
-    JInteger.valueOf(i(a).intValue * i(b).intValue)
+    VInt(i(a) * i(b))
 
   def eq_Int(a: Any, b: Any): Any =
-    if (i(a).intValue == i(b).intValue) True else False
+    if (i(a) == i(b)) True else False
 
   def foldLeft(list: Any, bv: Any, fn: Any): Any = {
-    val fnT = fn.asInstanceOf[Fn[Any, Fn[Any, Any]]]
+    val fnT = fn.asInstanceOf[FnValue]
```
johnynek: `.asFn`

core/src/main/scala/org/bykn/bosatsu/Predef.scala
```diff
@@ -69,38 +68,40 @@ object Predef {
...
     @annotation.tailrec
-    def loop(list: Any, bv: Any): Any =
+    def loop(list: Value, bv: Value): Value =
       list match {
-        case (0, _) =>
+        case SumValue(0, _) =>
```
johnynek: use a cons matcher to reuse code here.

core/src/main/scala/org/bykn/bosatsu/Predef.scala
```diff
@@ -69,38 +68,40 @@ object Predef {
...
-        case (1, head :: tail :: Nil) =>
-          val nextBv = fnT(bv)(head)
+        case SumValue(1, ConsValue(head, ConsValue(tail, UnitValue))) =>
+          val nextBv = fnT(bv).flatMap(_.asFn(head)).value
           loop(tail, nextBv)
         case _ => sys.error(s"unexpected: $list")
       }
 
-    loop(list, bv)
+    loop(list.asInstanceOf[Value], bv.asInstanceOf[Value])
```
johnynek: if the FFI api required Value, we wouldn't need these casts.

core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -50,7 +52,7 @@ class EvaluationTest extends FunSuite {
 package Foo
 
 x = 1
-"""), "Foo", 1)
+"""), "Foo", ExternalValue(1))
```
johnynek: use `Value.fromLit(Lit(1))` here instead.

core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -64,7 +66,7 @@ z = if x.eq_Int(1):
   "foo"
 else:
   "bar"
-"""), "Foo", "foo")
+"""), "Foo", ExternalValue("foo"))
```
johnynek: use `fromLit`

core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -120,7 +122,8 @@ package Foo
 struct Bar(a: Int)
 
 main = Bar(1)
-"""), "Foo", (0, List(1)))
+"""), "Foo",
+  ConsValue(VInt(1), UnitValue))
```
johnynek, line 126: use a Value List constructor.

PR #77, opened 2018-10-13
Fix the bug in if inference

### LINKED ISSUE #75: Bug in the if expression type inference
we have a typo and use the true branch twice here:

https://github.com/johnynek/bosatsu/blob/33892e70bc521193762c2b2efc650ce83b4a3729/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala#L567

We are not actually translating syntax into if currently (only into matches) so we can't hit this bug now. This is more reason to set up code coverage (#67).

To fix this we should first write a failing test to prevent similar recursions. Such a test would be something that has two different types in the if/else result, it should return the type of the true branch.

johnynek:
closes #75

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
@@ -600,7 +600,7 @@ object Infer {
                 tExp <- inferRho(ifTrue)
                 fExp <- inferRho(ifFalse)
                 rT = tExp.getType
-                rF = tExp.getType
+                rF = fExp.getType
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
```diff
@@ -600,7 +600,7 @@ object Infer {
                 tExp <- inferRho(ifTrue)
                 fExp <- inferRho(ifFalse)
                 rT = tExp.getType
-                rF = tExp.getType
+                rF = fExp.getType
```
johnynek, line 603: this was the bug.

Issue #84, opened 2018-10-15
type variables don't work with nested defs


johnynek:
see the TODO here:

https://github.com/johnynek/bosatsu/blob/858324e5f6fb9d8bc133299c4b88637e7f05460c/core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala#L187

johnynek:
not actually a bug (it seems).

PR #85, opened 2018-10-15
fix non-bug number 84

johnynek:
actually, #84 isn't a bug.

The error message isn't clear, but if you are unifying, for instance `a = List[a]`, then you need an infinite type, this does not unify.

I wrote the wrong type annotations in #84, but assumed it was a bug with the compiler.

Related to #4

### DIFF
--- a/core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
+++ b/core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
@@ -184,10 +184,7 @@ def reverse(ls):
 struct Pair(fst, sec)
 
 def zip(as: List[a], bs: List[b]) -> List[Pair[a, b]]:
-  # TODO if we write pair: Pair[a, b] below we get an internal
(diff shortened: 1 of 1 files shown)


core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -219,6 +215,33 @@ def eq_list(a, b, fn):
...
+package Foo
+
+def reverse(ls):
+  # note foldLeft is also built in, and not implementable
+  ls.foldLeft(EmptyList, \tail, h -> NonEmptyList(h, tail))
+
+struct Pair(fst, sec)
+
+def zip(as: List[a], bs: List[b]) -> List[Pair[a, b]]:
+  def cons(pair: Pair[List[Pair[a, b]], List[b]], item: a) -> Pair[List[Pair[a, b]], List[b]]:
```
johnynek, line 227: the type here is just really brutal, and I got it wrong.

PR #88, opened 2018-10-25
Use partial evaluation to go faster

johnynek:
This PR is based on a comment I heard in a talk on the Unity programming language. The idea is that an interpreter can partially evaluate a fair amount statically by currying the environment. By doing this, you can build and cache everything you need in one advance pass, then pass the environments along, which are the things that change and probably are not useful to cache.

This seems to cut the bazel bosatsu test time way down (maybe 3x faster, and note we are still paying JVM startup cost, so it may be more than that if we carefully account for the improvement).

cc @snoble if you have time to take a look at this approach.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
@@ -123,22 +123,103 @@ object Evaluation {
         }
     }
   }
+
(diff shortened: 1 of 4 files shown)


johnynek:
actually, after bazel is warmed up, this is more like 10x faster than it was before, and that still (I think) without worker support for bosatsu.

I think this solves the performance issues for the short term.

Issue #52, opened 2018-09-30
Document design principles, goals and non-goals.


johnynek:
One somewhat unspoken guiding principle of bosatsu is that implementing tooling, and understanding dependencies should be very easy to do.

This is why we have explicit imports and exports at the top of the file, and it should be possible with a very simple parser to read and correctly see the dependencies between bosatsu files. This should allow it to be very easy to integrate with build tooling, since all dependencies are declared.

This has some implications that currently, I do not think are fully fleshed out.

1. if a value has a type that refers to an imported package, when exporting that value, how do we communicate and exported dependency?
2. when resolving declarations to expressions, we currently only use the package name, but probably we should parse some kind of resolver out of the file first that can turn names in the files into their fully qualified names (either pointing externally or locally). In this way, the expression graph is always dealing with fully qualified names.
3. should we have syntax to allow users to refer to fully qualified names? I tend to say no. We should require import statements. It may seem that imports are duplicative, and they are somewhat: it is a tax to write external names or exported names twice, but that tax aids in allowing humans and tools to understand dependencies without running type inference.

PR #100, opened 2018-10-29
Create docs directory, start documentation

johnynek:
relates to #52

### DIFF
--- a/docs/language_guide.md
+++ b/docs/language_guide.md
@@ -1,0 +1,71 @@
+# The Bosatsu Programming Language
+
+## Design Philosophy and Goals
+Bosatsu has several main themes:
+
+1. correctness
+1. simplicity
+1. usability
+
+These are roughly in priority order: correctness is more important than simplicity: we don't want to sacrifice correctness for usability. Similarly, generally, we prefer simplicity to usability. Some features which can increase usability make a language considerably more complex. In general, we will be skeptical of such trades.
+
+The main user experience of the language is inspired by Python. It is a bit of a mystery that Python is currently one of the fastest growing languages as well as already being incredibly popular that it has not been more influential in language design. No other language in the top 20 most commonly used languages is very syntactically similar to Python. Put succinctly, we want Bosatsu to by a minimalistic statically typed Python.
+
+A second goal is that package relationships should be clear in Bosatsu. All external names must be explicitly imported, and all values are private unless exported. To see the dependencies between packages no type inference is required.
+
+## Problem domain for Bosatu
+In the current programming landscape, the vast majority of code is either in Turing complete languages or very anemic data-structure languages (JSON, YAML, TOML). Bosatsu is exploring an intermediate space. Bosatu is not designed as a general purpose language. Bosatsu in a total language: all functions return values. This rules out both infinite loops and non-explicit failure possibilities (exceptions, crashes, etc...). The goal of Bosatsu is somewhat similar to Google's [starlark](https://github.com/bazelbuild/starlark) language, a python like language for configuration and embedded applications. Starlark is used to configure bazel builds, but you can imagine configuring systems currently configured with yaml or json, or replacing general purpose languages in systems configured with those (e.g. Scala's sbt is configured with scala).
+
+## Differences with Starlark
+The biggest three differences between Bosatsu and Starlark is that Bosatsu is statically typed, all values are immutable, and there are no side-effects. The smaller differences come from adding features found in functional programming that Python lacks: a more powerful anonymous function syntax as well as the ability to define enums similar to Rust.
+
+# Language Features
+1. A static type system using a generalization of [Hindley-Milner (or Damas-Milner)](https://en.wikipedia.org/wiki/Hindley%E2%80%93Milner_type_system) published in [this paper by Simon Peyton Jones et. al.](https://www.microsoft.com/en-us/research/publication/practical-type-inference-for-arbitrary-rank-types/)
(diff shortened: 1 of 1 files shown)


docs/language_guide.md
```diff
@@ -0,0 +1,88 @@
...
+## Design Philosophy and Goals
+Bosatsu has several main themes:
+
+1. correctness
+1. simplicity
+1. usability
+
+These are roughly in priority order: correctness is more important than simplicity: we don't want to sacrifice correctness for usability. Similarly, generally, we prefer simplicity to usability. Some features which can increase usability make a language considerably more complex. In general, we will be skeptical of such trades.
+
+The main user experience of the language is inspired by Python. It is a bit of a mystery that Python is currently one of the fastest growing languages as well as already being incredibly popular that it has not been more influential in language design. No other language in the top 20 most commonly used languages is very syntactically similar to Python. Put succinctly, we want Bosatsu to by a minimalistic statically typed Python.
```
avibryant: The second sentence feels like it's missing the word "given" or something along those lines.

docs/language_guide.md
```diff
@@ -0,0 +1,88 @@
...
+
+A second goal is that package relationships should be clear in Bosatsu. All external names must be explicitly imported, and all values are private unless exported. To see the dependencies between packages no type inference is required.
+
+## Problem domain for Bosatu
+In the current programming landscape, the vast majority of code is either in Turing complete languages or very anemic data-structure languages (JSON, YAML, TOML). Bosatsu is exploring an intermediate space. Bosatu is not designed as a general purpose language. Bosatsu in a total language: all functions return values. This rules out both infinite loops and non-explicit failure possibilities (exceptions, crashes, etc...). The goal of Bosatsu is somewhat similar to Google's [starlark](https://github.com/bazelbuild/starlark) language, a python like language for configuration and embedded applications. Starlark is used to configure bazel builds, but you can imagine configuring systems currently configured with yaml or json, or replacing general purpose languages in systems configured with those (e.g. Scala's sbt is configured with scala).
```
avibryant: "Bosatsu in a total language" - change to "is"

avibryant: Inconsistent capitalization of language names (starlark/Starlark, python/Python)

docs/language_guide.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Differences with Starlark
+The biggest three differences between Bosatsu and Starlark is that Bosatsu is statically typed, all values are immutable, and there are no side-effects. The smaller differences come from adding features found in functional programming that Python lacks: a more powerful anonymous function syntax as well as the ability to define enums similar to Rust.
```
avibryant: "biggest three differences" ... "is" should be "are"

docs/language_guide.md
```diff
@@ -0,0 +1,88 @@
...
+1. data classes or named tuples, which we call structs following Rust
+1. sum types, which we call enums following Rust.
+1. a package system with explicit import and export syntax.
+1. a highly python-inspired syntax including function definition, literal integers and strings, literal lists, list comprehensions, tuples [(todo)](https://github.com/johnynek/bosatsu/issues/18) and (string key) dictionaries [(todo)](https://github.com/johnynek/bosatsu/issues/14). Generally, if a syntax is possible to support in the current type system with immutable values, we generally intend to support it.
+
+There are also some un-features, or items we don't currently and may never support
+1. There is no recursion at all. The name of the current function or let binding is not in scope as we are creating a value, so it is impossible to explicitly write recursion or infinite data structures.
```
avibryant, line 35: You do say "explicitly" here but I'm curious, is it possible to implement a fixed-point combinator (like the Y combinator) in bosatsu and still end up with recursion?

johnynek:
Thanks a ton for reading this and reviewing @avibryant ! I'll incorporate this feedback tonight.

docs/language_guide.md
johnynek: indeed, you are right.

```
package Y

struct W(fn: W[a, b] -> a -> b)

def call(w0, w1):
  match w0:
    W(fn): fn(w1)

def y(f):
  g = \w -> f(call(w, w))
  g(W(g))

def ltEqZero(i):
  match i.cmp_Int(0):
    GT: False
    _: True

fac = y(\f, i -> 1 if ltEqZero(i) else f(i).times(i))

main = fac(6)
```
type checks but seems to run forever.

I should fix the bug causing it to run forever, and make it not type check in that order. I think the issue is that we allow recursion in the definition of structs: `struct W(f: W[a, b] -> a -> b)`. While not an infinite value, this has a kind of an infinite type, which is giving rise to this. The fix, I think, is to ban any recursion in structs/enums, which seems obvious now, but didn't come up because they are not type checked (they are the axioms of the type system), and I didn't add any explicit checking to forbid them.

johnynek: I based this implementation on the scala implementation from: https://rosettacode.org/wiki/Y_combinator#Scala and if I change that one to:
```scala
def Y[A,B](f: (A=>B)=>(A=>B)) = {
  case class W(wf: W=>A=>B) {
    def apply(w: W) = wf(w)
  }
  val g: W=>A=>B = w => f(w(w)) // eagerly call f, but the original makes this lazy only calling on need
  g(W(g))
}
```
Then I can replicate the infinite loop. It shouldn't matter if we use strict or lazy evaluation in bosatsu (if it is free of recursion and is total), but that's not the case for this function. I think if I make interpretation lazy, this implementation would work, although, we want to ban it anyway.

PR #102, opened 2018-10-30
Ban Recursive Structs

johnynek:
@avibryant asked, and then I subsequently noticed that since we were allowing recursive (though not infinite) data types, that is enough to be able to put a type on a fixed point combinator, which allows it to be expressed and breaks totality.

Original comment here:
https://github.com/johnynek/bosatsu/pull/100#discussion_r228962619

The fix here is to add a check that types form a DAG. This makes it impossible to define the list type in bosatsu, but that isn't too much of a loss (since we add it with an external type). This also requires changing the built in testing framework from a tree to an optionally 3 level construct.

This is much more limited than before, but I think we have regained totality.

### DIFF
--- a/core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
+++ b/core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
@@ -54,7 +54,7 @@ class EvaluationTest extends FunSuite {
     }
   }
 
-  def evalFail(packages: List[String], mainPackS: String, extern: Externals = Externals.empty) = {
+  def evalFail(packages: List[String], mainPackS: String, extern: Externals = Externals.empty)(errFn: PartialFunction[PackageError, Unit]) = {
     val mainPack = PackageName.parse(mainPackS).get
 
     val parsed = packages.zipWithIndex.traverse { case (pack, i) =>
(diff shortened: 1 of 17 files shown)


core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -366,4 +366,30 @@ main = [1, 2]
...
+package Y
+
+struct W(fn: W[a, b] -> a -> b)
+
+def call(w0, w1):
+  match w0:
+    W(fn): trace("fn(w1)", fn(w1))
+
+def y(f):
+  g = \w -> \a -> trace("calling f", f(call(w, w), a))
```
johnynek, line 382: this does not actually run because we are not lazy enough.

I made some changes to evaluation to try to be lazier, but it didn't work. That's probably a future work: replace the use of cats.Eval in Evaluation with something that has clearer semantics. We likely either want strict or memoizing lazy, it is hard to reason about composing cats.Evals when the whole dag may be repeatedly recomputed.

johnynek:
Note, this was a mistake. We only need to ban recursion in types in contravariant positions, which indeed the y-combinator example used. It will be nice to allow recursive data types.

PR #103, opened 2018-11-04
Add TotalityCheck and tests

johnynek:
When we expanded the power of the patterns the naive exhaustivity checking was removed because it wouldn't work.

Since then, we have actually been rolling with no match exhaustivity checking.

This adds code to check all patterns are exhaustive. It *seems* that with the current pattern language, we actually don't need to typecheck to do totality checking, which is interesting. That might be a good property to maintain (and maybe is always, or almost always true...)

This turns out to be a bit tricky and I had to derive some formulas, which I just left in the comments. I have pretty good tests of the checker, but will see a better test when I wire it into the compiler.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
@@ -70,6 +70,64 @@ case class TotalityCheck(inEnv: TypeEnv) {
       case Right(_) :: _ => false
     }
 
+  private def difference0List(
+    lp: List[Either[Option[String], Pattern[Cons, Type]]],
+    rp: List[Either[Option[String], Pattern[Cons, Type]]]): Res[Patterns] = {
+    (lp, rp) match {
+      case (Nil, Nil) =>
+        // total overlap
+        Right(Nil)
+      case (Nil, Right(_) :: _) =>
+        // a list of 1 or more, can't match less
+        Right(ListPat(lp) :: Nil)
+      case (Nil, Left(_) :: tail) =>
+        // we can have zero or more, 1 or more clearly can't match:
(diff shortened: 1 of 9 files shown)


core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
```diff
@@ -0,0 +1,354 @@
...
+        case (PositionalStruct(ln, lps), PositionalStruct(rn, rps)) =>
+          if (ln == rn) {
+            val check = for {
+              _ <- checkArity(ln, lps.size, left)
+              _ <- checkArity(rn, rps.size, right)
+            } yield ()
+
+            type ResList[A] = Res[List[A]]
+            implicit val app = Applicative[Res].compose(Applicative[List])
+            check >>
```
johnynek, line 275: double checking here.

core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
```diff
@@ -0,0 +1,589 @@
...
+          case None => Left(NonEmptyList.of(UnknownConstructor(nm, right, inEnv)))
+          case Some(dt) =>
+            dt.constructors.traverse {
+              case (c, params, _) if (dt.packageName, c) == nm =>
+                /*
+                 * At each position we compute the difference with _
+                 * then make:
+                 * Struct(d1, _, _), Struct(_, d2, _), ...
+                 */
+                def poke[M[_]: Applicative, A](items: List[A])(fn: A => M[List[A]]): M[List[List[A]]] =
```
johnynek, line 185: TODO: as a follow up, this can be handled recursively: we already have handling below. I'm also suspect of this method.

Issue #104, opened 2018-11-05
rules for total functional programming


johnynek:
http://www.jucs.org/jucs_10_7/total_functional_programming/jucs_10_07_0751_0768_turner.pdf

This paper lays out three rules:

1. all cases must match #103 
2. recursion on types must be in covariant position #102 
3. recursion in functions must be on a syntactic subcomponent of a formal parameter #87 

It seems by following this paper, we could add a lot of power to bosatsu. We recently removed ANY recursion in structs, but the example we found to be problematic *was* a contravariant recursion. We could (and probably should) re-add recursion in types in the covariant positions.

Part 3 is less of a priority to me, but might not be too hard. Is covered by #87.

johnynek:
related twitter thread: https://twitter.com/mioalter/status/1060576185904640000

PR #107, opened 2018-11-09
Use match totality checking

johnynek:
put the work from #103 into the compiler to error on non total matches in code.


johnynek:
this is rule 1 from the paper of #104

PR #108, opened 2018-11-09
Bug with name shadowing of Predef

johnynek:
Rather than report an issue, I just figured I'd open a PR showing the error.

It seems we are partially shadowing, so either the constructors for matching, or the type name is not getting shadowed.


johnynek:
The issue was that we were assuming any imported name took precedence when converting the Statement to Expr but ignoring that a type might be defined in the current module.

Issue #19, opened 2017-11-23
destructuring assignment


johnynek:
currently, the only way to access items from a struct or enum is with a pattern match. This is a pain for big items.

For structs (single item enums) we should be able to do:
```
x = Foo(1, 2)
Foo(y, z) = x
```
and have that be the same as the match variant. This should only work for structs (or eventually tuples) to preserve totality.

johnynek:
This can be done with one of two ways: 1. a new let form that accepts a pattern, 2. translate the above syntax into a single branch match.

It seems like the latter is less work. If we do this, this is purely a change to the parser to support pattern lets in Declaration, and then when converting to Expr just convert that form into a match. Everything else should work if we do this.

This seems like a very nice change for usability, and especially as pattern matching gets more powerful.

johnynek:
Not clear if we should allow:
```
export [ a ]

Foo(a, _) = fn(bar)
```
at the top level.

I tend to think no, because it is junky, but it creates another difference between top level assignment and assignment inside expressions. For now, #112 does not support it. Only `x = y` type patterns are allowed at the top level.

I think we should also error on single arm matches that do no binding. Assuming they are total (or we would error), they aren't doing anything. We should probably warn or just error on them.

johnynek:
Reversing my stance on this after having written some code. It is pretty annoying to have different rules at the top level (even though structs and enums are only allowed at the top level). Working with tuples is pretty annoying at the top level because of it.

PR #112, opened 2018-11-10
Support destructuring assignment in Declaration

johnynek:
Still left to do:

1. generate patterns for parsing testing
2. support reading a pattern in the parser

part of #19 

Since we already have #107 this should just work once we support it in the parser.


johnynek:
the coverage only drops because we didn't yet exercise pattern assignments, which we know that. Merging for now.

PR #116, opened 2018-11-12
Parse pattern match assignment

johnynek:
This is the parsing side to #112 which closes out #19 

Since we already have totality checking, this is perfectly safe since it is just a match statement internally, and we already needed to check match statements.

Testing the parser changes here made me see I was misusing cuts, which should probably very rarely appear in combinators since you don't know how they might be used. The cuts were making it impossible to do a simple parsing of patterns as binding targets (on the left of an equals) since the basic approach uses backtracking.

The performance is not stellar, but we do through insanely brutal randomly generated code at the parser.

### DIFF
--- a/core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
+++ b/core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
@@ -28,7 +28,7 @@ class EvaluationTest extends FunSuite {
         errs.toList.foreach { p =>
           p.showContext.foreach(System.err.println)
         }
-        sys.error(errs.toString)
+        sys.error("failed to parse") //errs.toString)
     }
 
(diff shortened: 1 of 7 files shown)


core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -244,19 +244,16 @@ def zip(as: List[a], bs: List[b]) -> List[Pair[a, b]]:
       Pair(acc, [h, *tail]):
         Pair([Pair(item, h), *acc], tail)
 
-  rev = as.foldLeft(Pair([], bs), cons)
-  match rev:
-    Pair(res, _):
-      reverse(res)
+  Pair(res, _) = as.foldLeft(Pair([], bs), cons)
```
johnynek, line 247: this is really nice, in my opinion, and completely safe!

PR #126, opened 2018-12-02
Implement Union Patterns

### LINKED ISSUE #110: union patterns
we should allow:
```
enum Option: None, Some(get)

x = Some(1)

y = match x:
  None | Some(_): 42
```

or more interestingly:
```
enum Either: Left(l), Right(r)

def get(v):
  match v:
    Left(x) | Right(x): x
```
In the latter example, we ensure that all arms of the union have the binding names and that all the types unify.

This could also work with total match let bindings:
```
Left(x) | Right(x) = y
f(x)
```

johnynek:
closes #110 

Seems to be correct, but some of the more hard-core intersection tests I couldn't get to pass with unions. I think this is due to the fact that with unions we no longer have a normal form without expanding unions out, which could cause exponential blowup.

The more important tests seem to pass: the totality checks, and adding missing branches causes us to reach totality.

We should add more tests of evaluation, but we have some basic ones here.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
@@ -360,6 +360,9 @@ case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
           case (Pattern.Annotation(p, _), next) :: tail =>
             // TODO we may need to use the type here
             bindEnv(arg, (p, next) :: tail, acc)
+          case (Pattern.Union(h, t), next) :: tail =>
+            // we can just loop expanding these out:
+            bindEnv(arg, (h, next) :: t.toList.map((_, next)) ::: tail, acc)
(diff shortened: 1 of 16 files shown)


core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -360,6 +360,9 @@ case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
           case (Pattern.Annotation(p, _), next) :: tail =>
             // TODO we may need to use the type here
             bindEnv(arg, (p, next) :: tail, acc)
+          case (Pattern.Union(h, t), next) :: tail =>
+            // we can just loop expanding these out:
+            bindEnv(arg, (h, next) :: t.toList.map((_, next)) ::: tail, acc)
```
johnynek, line 365: Amazingly this is the only change you need to handle evaluation. Goes to show how much work the type system is vs evaluation.

PR #131, opened 2018-12-08
Add a failing test showing incorrect unification of generic

johnynek:
The issue is, if we have something we declare to be generic, we currently can unify it with a concrete type, which is not what we want that syntax to mean.

It seems like we should be using skolem variables here, but if we do that, predef fails to compile.


johnynek:
The issue with predef not compiling using skolem variables is that I was being to naive: we need to convert bound variables to skolems, replace them in the inner expression, then unskolemize at the end of inference. We have to do this both for AnnotatedLambda and Annotation.

PR #136, opened 2019-02-13
Second attempt at dealing with generic types in defs

johnynek:
After #131 was thought to be fixed:

I get:
```
[info] - test inference with some defined types *** FAILED *** (170 milliseconds)
[info]   in file: <unknown source>, package Test, type Opt does not unify with type $f
[info]   Region(414,418) (RankNInferTest.scala:98)  
```

In this example, it is trying to unify `Opt` with a skolem variable `f`.

It's not really clear what is going on, but I think the code should typecheck. 

The fix in #131 was not correct. Here we revert that and add a few more tests to exercise use of universal quantification (generics) in syntax.

I'm much more confident in the current approach.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
@@ -265,24 +265,16 @@ object Infer {
       }
 
     def makeGeneric[A](forAlls: List[Type.Var.Skolem], rho: TypedExpr.Rho[A]): Infer[TypedExpr[A]] =
-      NonEmptyList.fromList(forAlls) match {
-        case None =>
-          zonkTypedExpr(rho)
-        case Some(skols) =>
-          val used: Set[Type.Var.Bound] = Type.tyVarBinders(List(rho.getType))
-          val aligned = Type.alignBinders(skols, used)
-          val rho1 = substTyExpr(skols, aligned.map { case (_, n) => Type.TyVar(n) }, rho)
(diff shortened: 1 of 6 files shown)


johnynek:
looks like #131 was not really solved yet, the solution caused more errors. Added several new tests.

core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
```diff
@@ -468,13 +471,13 @@ object Infer {
         case (_, Type.TyVar(b@Type.Var.Bound(_))) =>
           fail(Error.UnexpectedBound(b, t1))
         // the only vars that should appear are skolem variables, we check here
-        case (Type.TyVar(v1), Type.TyVar(v2)) if v1 == v2 => pure(())
-        case (Type.TyMeta(m1), Type.TyMeta(m2)) if m1.id == m2.id => pure(())
+        case (Type.TyVar(v1), Type.TyVar(v2)) if v1 == v2 => unit
+        case (Type.TyMeta(m1), Type.TyMeta(m2)) if m1.id == m2.id => unit
         case (Type.TyMeta(m), tpe) => unifyVar(m, tpe, r1, r2)
-        case (tpe, Type.TyMeta(m)) => unifyVar(m, tpe, r1, r2)
+        case (tpe, Type.TyMeta(m)) => unifyVar(m, tpe, r2, r1)
```
johnynek, line 461: I noticed that the right hand side switches the order of `tpe, m` from the left, but not the regions.

PR #141, opened 2019-02-21
Add parser support for top level patterns

johnynek:
half of #140 

So far we are just handling parsing. A follow up will handle converting the pattern to a series of let bindings.

Issue #140, opened 2019-02-21
support top level destructuring


johnynek:
Related to #19 

Currently we allow top level binds, `x = 42` but not total matches: `Foo(x) = Foo(42)`. This is currently disallowed by a parse error, which is pretty bad.

But it actually makes some code involving tuples a lot less fun to write. It also makes the top level work differently than nesting, which I think makes things a bit harder to learn. I think we should allow top level total pattern matches.

johnynek:
This is complete.

PR #142, opened 2019-02-22
support evaluation of top level pattern matches

johnynek:
evaluation part of #140 follow on to #141 

I found a bug in evaluation that this triggered that requires a code branch to be lazy which was being strict. That was probably really hurting performance.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Pattern.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Pattern.scala
@@ -27,6 +27,10 @@ sealed abstract class Pattern[+N, +T] {
       case Pattern.Union(h, t) => Pattern.Union(h.mapType(fn), t.map(_.mapType(fn)))
     }
 
+  /**
+   * List all the names that are bound in Vars inside this pattern
+   * in the left to right order they are encountered, without any duplication
+   */
   def names: List[String] = {
     @annotation.tailrec
     def loop(stack: List[Pattern[N, T]], seen: Set[String], acc: List[String]): List[String] =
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/bykn/bosatsu/Program.scala
```diff
@@ -52,22 +55,50 @@ object Program {
...
+              val thisName = anonNames.next()
+              val v = Expr.Var(None, thisName, decl.tag)
+              ((thisName, decl) :: Nil, v)
+          }
+
+          val tail = complex.names.map { nm =>
+            val pat = complex.filterVars(Set(nm))
+            (nm, Expr.Match(left,
+              NonEmptyList.of((pat, Expr.Var(None, nm, decl.tag))), decl.tag))
+          }
```
johnynek, line 92: If tail is empty we still need to typecheck that the pattern is sane. We probably want to lint no-op pattern matches but we shouldn’t accept a bad match just because the code doesn’t bing any names.

core/src/main/scala/org/bykn/bosatsu/Pattern.scala
```diff
@@ -26,6 +26,53 @@ sealed abstract class Pattern[+N, +T] {
...
+    loop(this :: Nil, Set.empty, Nil)
+  }
+
+  /**
+   * Return the pattern with all the binding names removed
+   */
+  def unbind: Pattern[N, T] =
+    filterVars(Set.empty)
+
+  def filterVars(set: Set[String]): Pattern[N, T] =
```
johnynek: We should take a predicate here. We don’t need to have a finite set.

Issue #143, opened 2019-02-23
track variance of user defined types


johnynek:
more specific but related to #106

There is one primitive contravariant type: `a -> b` is covariant in the first parameter. But, structs and enums can become contravariant or invariant due to including such types:
```
struct Contra(fn: a -> Int)

struct Covariant(res: a, fn: Int -> (Int, a))

struct Invariant(arg: a, fn: a -> Int)
```

By tracking this variance we can correctly apply substitution and skolemization rules which we are currently specializing for function.

I *think* the current code treats all types other than function as invariant, which is safe, but rules out some programs which we should be able to accept.

johnynek:
Note, once we understand variance we can also allow recursion in covariant positions of type according to #104

johnynek:
Done.

Issue #106, opened 2018-11-08
open questions on well foundedness


johnynek:
The paper we implemented for type inference has no custom types.

It does have special handling for function types. I think the only special handling for function types is that they are contravariant in input and covariant in output.

Our current defined types, I believe as they are implemented, has every type being invariant in all type parameters.

However, allowing recursive types #102 as discussed in #104 allows recursion only in covariant positions. So, to support that, we need to understand variance of types (or at least some useful lower bound on which parameters are covariant). But, once we understand variance, we want to use it in substitution in inference, and not require strict unification.

johnynek:
Without thinking too much, it seems like using variance and custom types should happen in two places:

1. in skolemization, we skolemize through covariant parameters.
2. in substitution checking, we substitute using variance.

PR #144, opened 2019-02-23
Add variance tests, show some issues

johnynek:
The paper we derive the type system from does not directly address user defined types. So, there is a question of what special casing of the function types need to be done for user defined types.

This PR adds some tests that show two good statements (I think) that are ruled out, despite them being safe, which I think could be allowed if we tracked variance of structs (see #143 #106).

Basically, any test that passes with a function type should pass if we replace the function with a wrapper of function: `struct FnWrapper(fn: a -> b)`, but this test shows two examples of that not being true.

### DIFF
--- a/core/src/test/scala/org/bykn/bosatsu/rankn/RankNInferTest.scala
+++ b/core/src/test/scala/org/bykn/bosatsu/rankn/RankNInferTest.scala
@@ -600,6 +600,9 @@ struct Foo
 
 (idFoo: Foo -> Foo) = id
 
+(id2: Foo -> Foo) = \x -> x
+(idGen2: (forall a. a) -> Foo) = id2
+
 main = Foo
 """, "Foo")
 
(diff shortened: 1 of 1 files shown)


core/src/test/scala/org/bykn/bosatsu/rankn/RankNInferTest.scala
```diff
@@ -552,6 +552,214 @@ main = fst
...
+(bar2: forall a. Cont[a, Foo]) = Cont(\fn -> Bar)
+(baz2: forall a. Cont[a, Foo]) = Cont(\fn -> Baz)
+(bar3: Cont[forall a. a, Foo]) = Cont(\fn -> Bar)
+(baz3: Cont[forall a. a, Foo]) = Cont(\fn -> Baz)
+
+(bar41: Cont[Foo, Foo]) = bar1
+(bar42: Cont[Foo, Foo]) = bar2
+(baz41: Cont[Foo, Foo]) = baz1
+(baz42: Cont[Foo, Foo]) = baz2
+# Cont is covariant in a, this should be allowed
```
johnynek, line 659: If we track variance, we should be able to accept the following two lines.

core/src/test/scala/org/bykn/bosatsu/rankn/RankNInferTest.scala
```diff
@@ -552,6 +552,214 @@ main = fst
...
+main = Foo
+""")
+
+    parseProgram("""#
+struct Foo
+enum Opt: Nope, Yep(a)
+
+struct FnWrapper(fn: a -> b)
+
+# TODO: this should pass because FnWrapper is covariant, but skolemization ignores custom types
```
johnynek, line 723: we should be able to accept this line.

core/src/test/scala/org/bykn/bosatsu/rankn/RankNInferTest.scala
```diff
@@ -552,6 +552,214 @@ main = fst
...
+#(producer: FnWrapper[Foo, forall a. Opt[a]]) = FnWrapper(\x -> Nope)
+# in the covariant position, we can substitute
+#(producer1: FnWrapper[Foo, Opt[Foo]]) = producer
+(consumer: FnWrapper[Opt[Foo], Foo]) = FnWrapper(\x -> Foo)
+# in the contravariant position, we can generalize
+# TODO: this should pass because FnWrapper is contravariant in the first type
+# but subsCheck only knows about contravariance in the Fn type
+#(consumer1: FnWrapper[forall a. Opt[a], Foo]) = consumer
```
johnynek, line 731: we should be able to accept this line.

PR #149, opened 2019-02-26
Solve for variance

johnynek:
This gives an algorithm based on a few approaches:

0. any variable used as a type constructor is invariant: `f[a]` implies `f` and `a` are now invariant. We can't infer these in general, and making `f` and hence `a` invariant is the most restrictive, which should make it safe.
1. the variance algebra means that some formulas are directly solvable, `unknown = covariant` or `unknown = covariance + contravariant` cases. When you see a solution, substitute it into the other formulas and repeat until you have a stable set of formulas.
2. after that, set all variables to phantom, and consider the constraints as an iterative equation. Keep iterating until we reach a limit number of steps or we hit a stable point (input is output), which is clearly a legal solution of the variances.

This should unblock us re-enabling covariant recursive data types (see #104 )

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/VarianceFormula.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/VarianceFormula.scala
@@ -55,7 +55,8 @@ object VarianceFormula {
 
   object Identifier {
     private case class IdImpl(toLong: Long) extends Identifier
-    def nextIdent: State[SolutionState, Identifier] =
+
+    val nextIdent: State[SolutionState, Identifier] =
       State { ss: SolutionState =>
         require(ss.next.toLong < Long.MaxValue, "we shouldn't run long enough to overflow Long")
         (ss.copy(next = IdImpl(ss.next.toLong + 1L)), ss.next)
@@ -199,7 +200,36 @@ object VarianceFormula {
     def constrain(unknowns: SortedMap[(PackageName, TypeName), DefinedType[Unknown]], dt: DefinedType[Unknown]): State[SolutionState, Unit] = {
       import Type._
 
+      /*
+       * We assume all higher kinded type variables are invariant, which is a worst
+       * case assumption, although also meeting with expectations for a generic type
+       */
+      val unit: State[SolutionState, Unit] = State.pure(())
+      def constrainHKinParam(tpe: Type, umap: Map[Var, Unknown]): State[SolutionState, Unit] =
+        tpe match {
+          case TyApply(TyVar(v), right) =>
+            val left = umap.get(v) match {
+              case Some(u) =>
+                State.modify[SolutionState](_.constrain(u, Invariant.toF))
+              case None => unit
+            }
+            left *> constrainHKinParam(right, umap)
+          case TyApply(left, right) =>
+            constrainHKinParam(left, umap) *> constrainHKinParam(right, umap)
+          case TyConst(_) => unit
+          case TyVar(_) => unit // not in the apply position
+          case TyMeta(_) => unit // probably should be an error
+          case ForAll(bound, in) =>
+            // remove these bound variables:
+            constrainHKinParam(in, umap -- bound.toList)
+        }
+
       val umap: Map[Var, Unknown] = dt.annotatedTypeParams.toMap
+      val hks = dt.constructors.traverse_ { case (_, ps, _) =>
+        ps.traverse_ { case (_, tpe) =>
+          constrainHKinParam(tpe, umap)
+        }
+      }
 
       def constructorVariance(tpe: Type): Stream[VarianceFormula] =
         tpe match {
@@ -261,7 +291,7 @@ object VarianceFormula {
         loop(Set.empty, tpe)
       }
 
-      dt.annotatedTypeParams.traverse_ { case (b, u) =>
+      hks *> dt.annotatedTypeParams.traverse_ { case (b, u) =>
         dt.constructors.traverse_ { case (_, argType, _) =>
           argType.traverse_ { case (_, tpe) =>
             val formula = constrainTpe(b, u, tpe)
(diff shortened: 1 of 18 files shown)


core/src/main/scala/org/bykn/bosatsu/VarianceFormula.scala
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  case class SolutionState(
+    next: Identifier,
+    constraints: List[(VarianceFormula, VarianceFormula)],
+    solutions: Map[Unknown, Variance],
+    unknowns: Set[Unknown]) {
+
+
+    def isSolved: Boolean = unknowns.isEmpty
```
johnynek: should probably be `def isSolved == constraints.isEmpty`

core/src/main/scala/org/bykn/bosatsu/VarianceFormula.scala
```diff
@@ -0,0 +1,254 @@
...
+object VarianceFormula {
+  import Variance._
+
+  sealed abstract class Identifier {
+    def toLong: Long
+  }
+
+  object Identifier {
+    private case class IdImpl(toLong: Long) extends Identifier
+    def nextIdent: State[SolutionState, Identifier] =
```
johnynek: can be a val.

core/src/main/scala/org/bykn/bosatsu/VarianceFormula.scala
```diff
@@ -0,0 +1,254 @@
...
+    this match {
+      case Known(_) => true
+      case _ => false
+    }
+}
+
+object VarianceFormula {
+  import Variance._
+
+  sealed abstract class Identifier {
```
johnynek, line 52: make it only possible to create an initial ID, or get the next one in a state context.

core/src/main/scala/org/bykn/bosatsu/rankn/TypeEnv.scala
```diff
@@ -1,20 +1,26 @@
 package org.bykn.bosatsu.rankn
 
+import cats.{Applicative, Eval, Traverse}
 import org.bykn.bosatsu.{ConstructorName, TypeName, PackageName, ParamName}
+import scala.collection.immutable.SortedMap
+
+import cats.implicits._
+
+// TODO, we should be using SortedMap for reproducibility
```
johnynek: we have done this, we also needed it for a lawful traverse.

core/src/main/scala/org/bykn/bosatsu/VarianceFormula.scala
```diff
@@ -0,0 +1,298 @@
...
+        case None => right
+      }
+      copy(constraints = constraints.updated(left, newRight))
+    }
+
+    /**
+     * remove any constraints we can solve for. This is the main meat
+     * of the algorithm
+     */
+    def simplify: SolutionState = {
```
johnynek, line 96: this method is too large. Let's split out the two strategies (first remove trivial cases, then do iteration).

core/src/main/scala/org/bykn/bosatsu/VarianceFormula.scala
```diff
@@ -0,0 +1,298 @@
...
+      @annotation.tailrec
+      def loop(ss: SolutionState): SolutionState = {
+        // look for anything solved:
+        val solved = ss.constraints.iterator.collect { case (u, Known(v)) => (u, v) }.toMap
+        if (solved.isEmpty) {
```
johnynek, line 101: this is a junky branch

core/src/main/scala/org/bykn/bosatsu/Wrappers.scala
```diff
@@ -7,11 +7,20 @@ case class ConstructorName(asString: String)
 
 object ConstructorName {
   implicit val orderCN: Order[ConstructorName] = Order[String].contramap[ConstructorName](_.asString)
+  implicit val orderingCN: Ordering[ConstructorName] = orderCN.toOrdering
 }
 
 case class ParamName(asString: String)
 case class TypeName(asString: String)
 
+object TypeName {
```
johnynek, line 16: can be in another PR.

core/src/main/scala/org/bykn/bosatsu/rankn/Type.scala
```diff
@@ -138,15 +138,9 @@ object Type {
...
-    ts.foldLeft(List.empty[Type.Var]) { (acc, t) =>
-      go(t :: Nil, Set.empty, acc)
-    }
-    // reverse and distinct in one go
-    .foldLeft((Set.empty[Type.Var], List.empty[Type.Var])) {
-      case (res@(vset, vs), v) if vset(v) => res
-      case ((vset, vs), v) => (vset + v, v :: vs)
-    }
-    ._2
+    go(ts, Set.empty, Nil)
```
johnynek: should be in a PR on master with the tests that found this bug.

core/src/test/scala/org/bykn/bosatsu/rankn/TypeTest.scala
```diff
@@ -101,4 +101,15 @@ class TypeTest extends FunSuite {
     val fb = Type.ForAll(NonEmptyList.of(ba), Type.Fun(Type.TyVar(ba), Type.IntType))
     assert(Type.freeTyVars(fb :: Nil).isEmpty)
   }
+
+  test("Type.freeTyVars is left to right for functions") {
```
johnynek: pull this out to a PR on master.

core/src/main/scala/org/bykn/bosatsu/VarianceFormula.scala
```diff
@@ -0,0 +1,328 @@
...
+
+            thisDt.annotatedTypeParams.map(_._2).toStream
+          case TyApply(_, _) => sys.error(s"invariant violation: never called on TyApply: $tpe")
+          case TyVar(_) =>
+            // TODO: we assume the worst case: this is invariant, is this safe?
+            Stream.continually(Invariant.toF)
+          case TyMeta(_) =>
+            // TODO: error here:
+            Stream.continually(Invariant.toF)
+          case ForAll(_, _) => ???
```
johnynek: this is unimplemented. Write a test that causes this to fail, then fix it.

PR #150, opened 2019-02-26
Add Traverse instances for DefinedType and TypeEnv

johnynek:
Add traverse instances for DefinedType and TypeEnv.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/rankn/DefinedType.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/rankn/DefinedType.scala
@@ -1,6 +1,8 @@
 package org.bykn.bosatsu.rankn
 
+import cats.{Applicative, Eval, Traverse}
 import org.bykn.bosatsu.{ConstructorName, TypeName, PackageName, ParamName}
+import cats.implicits._
 
 case class DefinedType[+A](
   packageName: PackageName,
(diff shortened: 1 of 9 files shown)


core/src/main/scala/org/bykn/bosatsu/rankn/DefinedType.scala
```diff
@@ -52,4 +54,22 @@ object DefinedType {
...
+
+  implicit val definedTypeTraverse: Traverse[DefinedType] =
+    new Traverse[DefinedType] {
+      val listTup = Traverse[List].compose[(Type.Var.Bound, ?)]
+      def traverse[F[_]: Applicative, A, B](da: DefinedType[A])(fn: A => F[B]): F[DefinedType[B]] =
+        listTup.traverse(da.annotatedTypeParams)(fn).map { ap =>
+          da.copy(annotatedTypeParams = ap)
+        }
+
+      def foldRight[A, B](fa: DefinedType[A], b: Eval[B])(fn: (A, Eval[B]) => Eval[B]): Eval[B] =
```
johnynek, line 66: need test coverage for this.

core/src/main/scala/org/bykn/bosatsu/rankn/TypeEnv.scala
```diff
@@ -56,5 +60,42 @@ case class TypeEnv[+A](
...
+            fn(b, fa._2)
+        }
+
+      val cTrav = Traverse[SortedMap[(PackageName, ConstructorName), ?]]
+        .compose[(List[(ParamName, Type)], ?, Type)]
+        .compose[DefinedType]
+
+      val dtTrav = Traverse[SortedMap[(PackageName, TypeName), ?]].compose[DefinedType]
+      def traverse[F[_]: Applicative, A, B](da: TypeEnv[A])(fn: A => F[B]): F[TypeEnv[B]] =
+        (cTrav.traverse(da.constructors)(fn), dtTrav.traverse(da.definedTypes)(fn)).mapN { (cs, ds) =>
```
johnynek: this is wrong. We are traversing the DefinedTypes twice. This not what I expect. The maps are really just conveniences. Semantically, this is just a `List[DefinedType[A]]` with the map of values. The two maps are just cached representations of that list for faster access.

It would be nice to move away from a case class here and move to accessor functions to be clearer about what the contracts are rather than publicly sharing a full data structure.

PR #151, opened 2019-02-27
fix a bug in freeTyVars in the ordering of the vars

johnynek:
this turned up in #149 pulling out into its own PR.


core/src/test/scala/org/bykn/bosatsu/rankn/TypeTest.scala
```diff
@@ -101,4 +101,15 @@ class TypeTest extends FunSuite {
     val fb = Type.ForAll(NonEmptyList.of(ba), Type.Fun(Type.TyVar(ba), Type.IntType))
     assert(Type.freeTyVars(fb :: Nil).isEmpty)
   }
+
+  test("Type.freeTyVars is left to right for functions") {
```
johnynek, line 105: this test exposed the bug.

PR #157, opened 2019-03-03
Fix a bug around external types and variance inference

johnynek:
Assume that external structs are invariant (a worst case).

We could imagine adding syntax to denote variance, but following scala we can let the unannotated mean invariant, which shouldn't be backwards incompatible.


johnynek:
related to #149

Issue #148, opened 2019-02-24
Support covariant recursion in structs and enums


johnynek:
See rule 2 in #104 and #143 #145 #146 

After noting that allowing recursion of structs in a contravariant position, e.g.
```
struct Foo(fn: Foo -> a)
```
opens us up to infinite recursion, which is one of the main goals of bosatsu to disallow (so types are not all potentially hiding an infinite loop), we removed all recursion from structs.

This is very strict, and also prevents us from being able to write perfectly fine types, such as List:
```
enum List:
  EmptyList, NonEmptyList(head: a, tail: List[a])
```

The ability to support data types like this natively seems very desirable if it can be made safe.

The approach, I think, is to infer the variance of all type parameters in user defined structs, then verify that any recursion in types only happens via phantom or covariant paths.

johnynek:
this would undo the setback we saw in #102

johnynek:
also related to #101

johnynek:
Done.

PR #158, opened 2019-03-05
Allow basic restricted recursion in types.

johnynek:
This PR attempts to address #148 undoing #102.

The check currently is rather restrictive. The graph of types must:
1. form a DAG *excluding self references*.
2. for each type, all self referential loops must be via covariant or phantom variance paths.

This allows us to write all the kinds of types that I think we want, and are also safe: List, Tree, etc... while banning types that can be used to allow recursion, e.g. types associated with the y-combinator.

A follow up to this PR will be making List a normal type again, and not an external def, but I will keep the special python inspired syntax we have for lists.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/VarianceFormula.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/VarianceFormula.scala
@@ -333,10 +333,6 @@ object VarianceFormula {
       val unknowns = te.toList.filter(isUn)
       NonEmptyList.fromList(unknowns) match {
         case None =>
-          /*
-           * TODO: external defs have no constructors, we should assume their types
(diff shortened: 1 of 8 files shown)


core/src/main/scala/org/bykn/bosatsu/VarianceFormula.scala
```diff
@@ -333,10 +333,6 @@ object VarianceFormula {
       val unknowns = te.toList.filter(isUn)
       NonEmptyList.fromList(unknowns) match {
         case None =>
-          /*
-           * TODO: external defs have no constructors, we should assume their types
-           * are invariant, but this may be setting them to phantom, which is unsound
```
johnynek, line 338: this was fixed previously.

PR #159, opened 2019-03-07
Make List a native, non-external type again

johnynek:
This reverts the list change in #102 

now that we can support recursive types in covariant position, no need to have List as an external type.


johnynek:
made possible by #158

Issue #161, opened 2019-03-07
Add an Interface view of Packages


johnynek:
Currently, when we typecheck and render fully typed ASTs (TypedExpr), we require all the fully typed packages of all the imports.

We should be able to have an interface view that only lists the values and types exported by a package, without needing the fully typed ASTs.

The value of this is to cement in the design that we can do this separately, and as long as an interface does not change, there is no need to rerun a package typechecking.

This becomes really nice when coupled with hashing of the API or when used with build systems such as bazel that can make compilation only depend on the API artifacts, and not the fully typed ASTs.

For this, we would likely have a linking stage at the end that converts fully typed ASTs into some other output (JSON, Yaml, skylark, JVM class files, C, etc...)

johnynek:
done in #162 ... no serialization yet, but that is probably a subissue of #54

PR #162, opened 2019-03-07
Define Package.Interface to avoid using all of the package

johnynek:
relates to #161 

This weakens the view, in the types, of what inference sees to be clear you don't see the full bodies of the expressions you import.

This cleans up things a bit, and hopefully makes it easier to add separate compilation since now we could add not only parsed inputs but previously inferred interface files as well.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Package.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Package.scala
@@ -37,7 +37,8 @@ object Package {
   type PackageF2[A, B] = PackageF[A, A, B]
   type Parsed = Package[PackageName, Unit, Unit, Statement]
   type Resolved = FixPackage[Unit, Unit, (Statement, ImportMap[PackageName, Unit])]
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/bykn/bosatsu/Package.scala
```diff
@@ -37,7 +37,8 @@ object Package {
   type PackageF2[A, B] = PackageF[A, A, B]
   type Parsed = Package[PackageName, Unit, Unit, Statement]
   type Resolved = FixPackage[Unit, Unit, (Statement, ImportMap[PackageName, Unit])]
-  type Inferred = FixPackage[NonEmptyList[Referant[Variance]], Referant[Variance], Program[TypeEnv[Variance], TypedExpr[Declaration], Statement]]
+  type Interface = FixPackage[NonEmptyList[Referant[Variance]], Referant[Variance], Unit]
```
johnynek: we can probably weaken this more: I think we only need the package name and the exports. I don't think we need the imports at all.

PR #164, opened 2019-03-10
finish first draft of strutural recursion support

### LINKED ISSUE #87: consider supporting structural recursion
https://en.wikipedia.org/wiki/Structural_induction

We could implement both range and foldLeft if we allowed structural recursion: which is to say, we allow recursion in a certain case where we can prove it is total.

This may be tricky to implement and it adds considerable power to the language, but retains totality, which is very nice.

johnynek:
close #87 

This is a significant addition that allows limited recursion where we can prove it will terminate.

cc @puffnfresh @snoble @avibryant 

Comments/code review welcome. Putting up the PR so I can see where the test coverage is missing. Will be nice to do a follow up PR that leverages recursion more (e.g. rewrite fold, review some of the existing project euler examples).

### DIFF
--- a/core/src/main/resources/bosatsu/predef.bosatsu
+++ b/core/src/main/resources/bosatsu/predef.bosatsu
@@ -40,7 +40,10 @@ enum Bool:
 enum List:
   EmptyList, NonEmptyList(head: a, tail: List[a])
 
-external def foldLeft(lst: List[a], item: b, fn: b -> a -> b) -> b
+recursive def foldLeft(lst: List[a], item: b, fn: b -> a -> b) -> b:
+  recur lst:
+    EmptyList: item
+    NonEmptyList(head, tail): foldLeft(tail, fn(item, head), fn)
 
 def reverse_concat(front: List[a], back: List[a]) -> List[a]:
   foldLeft(front, back, \tail, h -> NonEmptyList(h, tail))
--- a/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
@@ -166,8 +166,11 @@ sealed abstract class Declaration {
           loop(p).map(_ => decl)
         case Var(name) =>
           Expr.Var(None, name, decl)
-        case Match(kind, arg, branches) =>
-          // TODO handle recursive matches
+        case Match(_, arg, branches) =>
+          /*
+           * The recursion kind is only there for DefRecursionCheck, once
+           * that passes, the expr only cares if lets are recursive or not
+           */
           val expBranches = branches.get.map { case (pat, oidecl) =>
             val decl = oidecl.get
             val newPattern = unTuplePattern(pat, nameToType, nameToCons)
--- a/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
@@ -73,7 +73,7 @@ sealed abstract class Declaration {
       case Lambda(args, body) =>
         Doc.char('\\') + Doc.intercalate(Doc.text(", "), args.toList.map(Doc.text _)) + Doc.text(" -> ") + body.toDoc
       case Literal(lit) => Document[Lit].document(lit)
-      case Match(typeName, args) =>
+      case Match(kind, typeName, args) =>
         val pid = Document[OptIndent[Declaration]]
 
         implicit val patDoc: Document[(Pattern[Option[String], TypeRef], OptIndent[Declaration])] =
@@ -85,8 +85,12 @@ sealed abstract class Declaration {
           Document.instance { ts => Doc.intercalate(Doc.line, ts.toList.map(Document[T].document _)) }
 
         val piPat = Document[OptIndent[NonEmptyList[(Pattern[Option[String], TypeRef], OptIndent[Declaration])]]]
+        val kindDoc = kind match {
+          case RecursionKind.NonRecursive => Doc.text("match ")
+          case RecursionKind.Recursive => Doc.text("recur ")
+        }
         // TODO this isn't quite right
-        Doc.text("match ") + typeName.toDoc + Doc.char(':') + args.sepDoc +
+        kindDoc + typeName.toDoc + Doc.char(':') + args.sepDoc +
           piPat.document(args)
       case Parens(p) =>
         Doc.char('(') + p.toDoc + Doc.char(')')
@@ -121,7 +125,7 @@ sealed abstract class Declaration {
         case Binding(BindingStatement(pat, value, Padding(_, rest))) =>
           pat match {
             case Pattern.Var(arg) =>
-              Expr.Let(arg, loop(value), loop(rest), decl)
+              Expr.Let(arg, loop(value), loop(rest), RecursionKind.NonRecursive, decl)
             case pat =>
               val newPattern = unTuplePattern(pat, nameToType, nameToCons)
               val res = loop(rest)
@@ -132,13 +136,13 @@ sealed abstract class Declaration {
           loop(decl).map(_ => decl)
         case Constructor(name) =>
           Expr.Var(None, name, decl)
-        case DefFn(defstmt@DefStatement(_, _, _, _)) =>
+        case DefFn(defstmt@DefStatement(kind, _, _, _, _)) =>
           val (bodyExpr, inExpr) = defstmt.result match {
             case (oaBody, Padding(_, in)) =>
               (loop(oaBody.get), loop(in))
           }
           val lambda = defstmt.toLambdaExpr(bodyExpr, decl)(_.toType(nameToType))
-          Expr.Let(defstmt.name, lambda, inExpr, decl)
+          Expr.Let(defstmt.name, lambda, inExpr, recursive = kind, decl)
         case IfElse(ifCases, elseCase) =>
           def ifExpr(cond: Expr[Declaration], ifTrue: Expr[Declaration], ifFalse: Expr[Declaration]): Expr[Declaration] =
             Expr.If(cond, ifTrue, ifFalse, decl)
@@ -162,7 +166,8 @@ sealed abstract class Declaration {
           loop(p).map(_ => decl)
         case Var(name) =>
           Expr.Var(None, name, decl)
-        case Match(arg, branches) =>
+        case Match(kind, arg, branches) =>
+          // TODO handle recursive matches
           val expBranches = branches.get.map { case (pat, oidecl) =>
             val decl = oidecl.get
             val newPattern = unTuplePattern(pat, nameToType, nameToCons)
@@ -270,7 +275,9 @@ object Declaration {
     elseCase: OptIndent[Declaration])(implicit val region: Region) extends Declaration
   case class Lambda(args: NonEmptyList[String], body: Declaration)(implicit val region: Region) extends Declaration
   case class Literal(lit: Lit)(implicit val region: Region) extends Declaration
-  case class Match(arg: Declaration,
+  case class Match(
+    kind: RecursionKind,
+    arg: Declaration,
     cases: OptIndent[NonEmptyList[(Pattern[Option[String], TypeRef], OptIndent[Declaration])]])(
     implicit val region: Region) extends Declaration
   case class Parens(of: Declaration)(implicit val region: Region) extends Declaration
@@ -281,6 +288,9 @@ object Declaration {
    * This represents the list construction language
    */
   case class ListDecl(list: ListLang[Declaration])(implicit val region: Region) extends Declaration
+
+  val matchKindParser: P[RecursionKind] =
+    (P("match").map(_ => RecursionKind.NonRecursive) | P("recur").map(_ => RecursionKind.Recursive))
 
   /**
    * A pattern can also be a declaration in some cases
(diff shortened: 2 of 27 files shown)


core/src/main/scala/org/bykn/bosatsu/Declaration.scala
```diff
@@ -162,7 +166,8 @@ sealed abstract class Declaration {
           loop(p).map(_ => decl)
         case Var(name) =>
           Expr.Var(None, name, decl)
-        case Match(arg, branches) =>
+        case Match(kind, arg, branches) =>
+          // TODO handle recursive matches
```
johnynek: I think we actually don't need to worry about the recur vs match here. We only need to know that this for doing check that the recursion is allowed. Once we see that it is allowed, it is a normal match.

core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
```diff
@@ -0,0 +1,251 @@
...
+import cats.data.{NonEmptyList, Validated, ValidatedNel}
+
+import cats.implicits._
+
+/**
+ * Recursion in bosatsu is only allowed on a substructural match
+ * of one of the parameters to the def. This strict rule, along
+ * with strictly finite data, ensures that all recursion terminates
+ *
+ * The rules are as follows:
```
johnynek, line 12: I would especially appreciate reviews to this, and any tests anyone would like to see added.

core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
```diff
@@ -0,0 +1,251 @@
...
+                (f.traverse_(checkDecl(state, _)))
+          }
+      }
+    }
+
+    /*
+     * Binds are not allowed to be recursive, only defs, so here we just make sure
+     * none of the free variables of the pattern are used in decl
+     */
+    // def checkBind(pat: Pattern[Option[String], TypeRef], decl: Declaration): Res =
```
johnynek: remove this code.

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -239,7 +248,7 @@ case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
   def evaluateLast(p: PackageName): Option[(Eval[Value], Type)] =
     for {
       pack <- pm.toMap.get(p)
-      (_, expr) <- pack.program.lets.lastOption
+      (_, rec, expr) <- pack.program.lets.lastOption
```
johnynek: ignoring the rec here. Probably isn't a problem since we can only have recursive functions and we don't evaluate main modules which have fn type, but still, we are ignoring it, and that should be fixed.

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -199,6 +199,15 @@ object Evaluation {
     def unreachable: Scoped =
       const(Eval.later(sys.error("unreachable reached")))
 
+    def recursive(name: String, item: => Scoped): Scoped = {
```
johnynek: I don't think item actually has to be call-by-name. I think the laziness we need is the inner code recursively sets up the environment.

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -467,8 +476,13 @@ case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
          val inner = recurse((p, Right(expr)))._1
 
          inner.asLambda(name)
-       case Let(arg, e, in, _) =>
-         val eres = recurse((p, Right(e)))._1
+       case Let(arg, e, in, rec, _) =>
+         val eres =
+           if (rec.isRecursive) {
+             Scoped.recursive(arg,
```
johnynek: if recursive isn't call by name, we cal pull the recurse call up above and not repeat ourselves.

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -514,7 +528,8 @@ case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
             // this isn't great, but since we fully compile, even when
             // we don't use a branch, we hit this now
             (Scoped.unreachable, Type.IntType)
-          case Some(NameKind.Let(expr)) =>
+          case Some(NameKind.Let(recursive, expr)) =>
+            // TODO handle recursion
```
johnynek: write a test that this fails, and fix. I think it should use the same Scope.recursive as above.

core/src/main/scala/org/bykn/bosatsu/Package.scala
```diff
@@ -152,6 +152,14 @@ object Package {
           .leftMap { badPaths =>
             badPaths.map(PackageError.CircularType(p, _))
           }
+      /*
+       * Check that all recursion is allowable
+       */
+      val defRecursionCheck: ValidatedNel[PackageError, Unit] =
```
johnynek, line 158: we aren't actually using this... :/ nice work man... Add a test that shows the omission, then fix.

core/src/main/scala/org/bykn/bosatsu/Pattern.scala
```diff
@@ -54,6 +54,36 @@ sealed abstract class Pattern[+N, +T] {
     loop(this :: Nil, Set.empty, Nil)
   }
 
+  /**
+   * List all the names that strictly smaller than anything that would match this pattern
+   * e.g. a top level var, would not be returned
```
johnynek, line 59: exercise directly this more, the current property check is pretty weak.

core/src/main/scala/org/bykn/bosatsu/Pattern.scala
```diff
@@ -54,6 +54,36 @@ sealed abstract class Pattern[+N, +T] {
...
+    @annotation.tailrec
+    def loop(stack: List[(Pattern[N, T], Boolean)], seen: Set[String], acc: List[String]): List[String] =
+      stack match {
+        case Nil => acc.reverse
+        case ((Pattern.WildCard, _) | (Pattern.Literal(_), _)) :: tail => loop(tail, seen, acc)
+        case (Pattern.Var(v), isTop) :: tail =>
+          if (seen(v) || isTop) loop(tail, seen, acc)
+          else loop(tail, seen + v, v :: acc)
+        case (Pattern.ListPat(Left(Some(_)) :: Nil), true) :: tail =>
+            // this is a total match at the top level, not a substructue
```
johnynek, line 74: substructure is misspelled.

core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
```diff
@@ -595,12 +596,29 @@ object Infer {
...
+                  for {
+                    typedRhs <- inferSigma(rhs)
+                    varT = typedRhs.getType
+                    typedBody <- typeCheckRho(body, expect)
+                  } yield TypedExpr.Let(name, typedRhs, typedBody, isRecursive, tag)
+                }
+              }
+          }
+          else {
+            for {
```
johnynek: this block is repeated, let's move it up.

core/src/main/resources/bosatsu/predef.bosatsu
```diff
@@ -40,7 +40,10 @@ enum Bool:
 enum List:
   EmptyList, NonEmptyList(head: a, tail: List[a])
 
-external def foldLeft(lst: List[a], item: b, fn: b -> a -> b) -> b
```
johnynek, line 43: this is the motivation for all this work....

johnynek:
Unfortunately, making foldLeft native, we are now in a situation that performance matters again. Two of the project euler tests don't finish in time now.

I'm worried the evaluation story is far too naive at the moment. We may be recomputing a lot now. Using cats.Eval is probably a really bad default.

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -349,7 +364,7 @@ case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
                         splice match {
                           case None => (acc1, next)
                           case Some(nm) =>
-                            val rest = Eval.later(VList(spliceVals.reverse))
+                            val rest = Eval.now(VList(spliceVals.reverse))
```
snoble, line 367: Is this related to structural recursion support. Was it an attempt at improving performance?

core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
```diff
@@ -0,0 +1,284 @@
...
+import cats.implicits._
+
+/**
+ * Recursion in bosatsu is only allowed on a substructural match
+ * of one of the parameters to the def. This strict rule, along
+ * with strictly finite data, ensures that all recursion terminates
+ *
+ * The rules are as follows:
+ * 0. recursive defs may not be shadowed. This makes checking for legal recursion easier
+ * 1. until we reach a recur match, we cannot call a recursive def
```
snoble, line 14: What is the expected behaviour if you have a variable name in an outer scope that is shadowed by the recursive def, and it is referenced before the recur match? My expectation is these checks will say this rule is broken by that reference.

core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
```diff
@@ -0,0 +1,284 @@
...
+ * 2. a recur match must occur on one of the literal parameters to the recursive def
+ * 3. inside each branch of the recur match, we may only recur on on substructures in the match
```
snoble: s/on on/on/

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
johnynek: part of the hacks to improve performance (still not great, but I did get the tests to pass).

core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
johnynek: Can you write an example, and I will add a test. You mean:
```
foo = 1
recursive def foo(lst):
  fn = foo
  recur lst:
    []: 0
    [_, *tail]: foo(tail).add(1)
```
something like that?

I think this won't work since that will look like accessing foo, which is in scope, but not accessible.

core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
johnynek: thanks! fixed.

core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
johnynek: added a test, let me know if that captured your idea.

snoble: Yeah, this is exactly what I was curious about.

The other question that occurs to me is what happens when the `recur` statement isn't the return statement. I haven't had a chance to try for myself yet.

johnynek:
more comments always welcome!

core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
johnynek: That’s a good example to add. It should work, although it won’t be tail recursion.

avibryant:
Why do you need the `recursive` annotation on the `def`? Isn't the presence of `recur` unambiguous?

johnynek:
1. I don’t mind if people write a bit extra in recursive definitions. They should be a last resort, not something to reach for.

2. If I don’t write recursive, the current checking algorithm has to back track. For instance shadowing a function name is legal, but shadowing a recursive function is not legal.

3. Typing needs to know if a binding is recursive or not. Currently the check and typing are independent so I can accumulate errors from both. If we don’t require the recursive annotation on def then recursion checking must happen before typing so we can see if the def is in scope for itself or not.

What do you think about those answers?

avibryant:
It seems like (1) is what dominates. If you want to introduce friction to make people reach less often for recursive definitions (which makes sense), then yeah, leave in the annotation. 2 and 3 are less compelling but that's moot.

johnynek:
After having written more of these tests, and making the rules more strict, if you don't do any let bindings before the recur it feels repetitive to write:
```
recursive def foo(x):
  recur x:
    ...
```

Maybe @avibryant is right that we shouldn't have both. The only thing I'm worried about is that both the def and the recur are somewhat magical here: inside a recursive def there is no shadowing of the foo name and no use of it other than to call foo.

However, in a non-recursive def if foo was previously bound, you see that:
```
foo = \x -> x.add(1)
def foo(x)
  x1 = foo(x)
  x1.add(1)
```
that is legal, and is the same as `foo = \x -> x.add(1).add(1)` Looking at that, I'm pretty sure that should not be legal code, or at least it is very confusing.

Maybe that's the problem: maybe we should use the strict rules around def names everywhere: no shadowing of self, no use of self except to call. And then we could infer if a def is recursive simply by the existence of a self call inside a recur.

I like harmonizing the rules about shadowing def names for sure. Having those be context dependent is confusin. Also, the above definitely looks like recursion, so that's bad: you have code that you have to be an expert to see it isn't recursion, but also isn't doing what you would guess.

johnynek:
@snoble also very interested in your feedback on the above.

snoble:
Generalizing the shadow rule to all functions just feels right. I think most users would be surprised that a reference to the function name would actually reach into the outer scope.

My only concern with removing the redundancy is I worry about the clarity around the multivariable function case where you're only allowed to recurse on a single variable. eg, consider

```
def foo(a,b,c):
  case []: ...
  case [head, *tail]: foo(tail, tail, tail)
```
That should be legal but it's clearly ambiguous.

This feels like the wrong syntax, but this would disambiguate:
```
def foo(a,b,c) recurses on b:
  case []: ...
  case [head, *tail]: foo(tail, tail, tail)
```

johnynek:
we could keep `recur` as required to actually enable the ability to call a recursive function.
```
def foo(a, b, c):
  recur b:
    []: ...
    [h, *t]: foo(t, t, t)
```

snoble:
That feels very python

PR #168, opened 2019-03-14
Fill out recursion tests a bit

snoble:


core/src/test/scala/org/bykn/bosatsu/DefRecursionCheckTest.scala
```diff
@@ -214,6 +214,74 @@ recursive def len(lst):
...
+  match lstA:
+    []: []
+    [headA, *tailA]: recur lstB:
+      []: []
+      [headB, *tailB]: [(headA, headB)].concat(zip(tailA, tailB))
+""")
+    allowed("""#
+recursive def zip(lstA, lstB):
+  recur lstA:
+    []: []
```
snoble, line 281: Are we sure the type is imputed correctly here? It seems like bosatsu might not know how to parameterize the list type

johnynek:
Looks like you got hit by one of the sometimes flakey totality checks (has to do with unifying list unions).

Thanks, this looks great.

I think bosatsu can infer all these types, but the recursion check isn’t type checking. If we add an Eval test with the same code it will type check and run or RankNInferTest can do a test just of inference.

I’ll merge this (after I rerun the flakey test). We can follow up with more tests later.

Thank you so much!

snoble:
I think you may need an extra rule. Consider

```
recursive def foo(lstA, lstB):
  x = recur lstA:
    []: 1
    [a, •as]: foo(as, [1,*lstB])
  y = recur lstB:
    []: 2
    [b, *bs]: foo([1, *lstA], bs)
  x.plus(y)
```

Consider
```
foo([1,1], [1]) = foo([1], [1,1]).plus(...)
foo([1],[1,1]) = foo(...).plus(foo([1,1], [1]))
```
I think you need to add the rule that all `recur` statements must be on the same variable

johnynek:
Interesting example... that should be an infinite loop now. I'll write that test...

PR #170, opened 2019-03-14
Add some more recursive evaluation tests

johnynek:
These are just to test the full stack of recursion checking, inference and evaluation.

cc @snoble who suggested some of these.

### DIFF
--- a/core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
+++ b/core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
@@ -389,15 +389,13 @@ three = [0, 1]
 # exercise the built-in range function (not implementable in bosatsu)
 threer = range(3)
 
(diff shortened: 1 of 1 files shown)


core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -389,15 +389,13 @@ three = [0, 1]
...
-def zip(as: List[a], bs: List[b]) -> List[(a, b)]:
-  def cons(pair, item):
-    match pair:
-      (acc, []): (acc, [])
-      (acc, [h, *tail]):
-        ([(item, h), *acc], tail)
-
-  (res, _) = as.foldLeft(([], bs), cons)
-  reverse(res)
+recursive def zip(as, bs):
```
johnynek, line 392: note we can infer the type of zip without any annotation, or with as below.

PR #171, opened 2019-03-14
Fix infinite loop snoble found

johnynek:
Never let it be said that I let bugs in totality checking linger!

https://github.com/johnynek/bosatsu/pull/168#issuecomment-472724055

@snoble noticed that we are missing a requirement on recursion: that you cannot recurse on different variables. I added the test he suggested (which did infinitely loop for me when I checked it in Eval, and the test initially failed to disallow).

I'm very grateful to you @snoble for thinking about this, and noticing this problem.

### DIFF
--- a/core/src/test/scala/org/bykn/bosatsu/DefRecursionCheckTest.scala
+++ b/core/src/test/scala/org/bykn/bosatsu/DefRecursionCheckTest.scala
@@ -171,6 +171,15 @@ recursive def len(lst):
       # shadowing len is not okay
       12
 """)
+    disallowed("""#
+recursive def len(lst):
+  def len_helper(l, r):
+    len = l.add(r)
+    len
+  recur lst:
+    []: 0
+    [_, *t]: len_helper(len(t), 1)
+""")
   }
 
   test("infinite loop isn't okay") {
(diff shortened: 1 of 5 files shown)


core/src/test/scala/org/bykn/bosatsu/DefRecursionCheckTest.scala
```diff
@@ -253,7 +270,7 @@ recursive def foo(lst):
   a.plus(b)
 """)
 
-    allowed("""#
+    disallowed("""#
```
johnynek, line 294: @snoble note: your example is no longer legal, which I think is what we want.

johnynek:
@snoble question: should we just limit a single recur match? I don't see why you ever need two. If you are matching on a single variable. Maybe that rule is simpler since it subsumes that all recurs must be on the same variable.

I'm wary the more I think of it, of allowing a pattern that seems unnecessary and encourages writing non-tailrecursive functions.

johnynek:
The more i think, the more I think requiring exactly 1 recur, and requiring at least one branch to make a recursive call is the way to go.

You can always weaken rules later if safe, it is better to start with stricter rules.

snoble:
Yeah, the only reason to have it now (that i can think of) is it might help with how you organize your cute, and it's hard to think of a good use of that

snoble:
Great to see this fix get in so quickly!!

PR #178, opened 2019-03-17
Use partial evaluation in pattern matching

johnynek:
This is a follow up to #88 . In that PR, we didn't implement that partial evaluation strategy in matching, so matches were still pretty expensive.

This PR fully inspects the patterns once for the code path, and generates the functions to transform then environment one time. This should help in complex match situations, and at least harmonizes the approach.

PS: I think this partial evaluation is closer to real compilation, so it is helpful to see an example of pattern matching implemented in this way.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
@@ -371,24 +371,13 @@ case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
                 }
               case Left(splice) :: Nil =>
                 // this is the common and easy case: a total match of the tail
+                // we don't need to match on it being a list, because we have
+                // already type checked
                 splice match {
                   case Some(nm) =>
-                    { (arg, acc) =>
-                      arg match {
-                        case VList(_) =>
-                          // now bind the rest into splice:
-                          val rest = Eval.now(arg)
-                          Some(acc.updated(nm, rest))
-                        case _ => None
-                      }
-                    }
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -320,106 +320,140 @@ case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
...
+                // we reverse the tails, do the match, and take the rest into
+                // the splice
+                val revPat = Pattern.ListPat(ptail.reverse)
+                val fnMatchTail = maybeBind(revPat)
+
+                { (arg, acc) =>
+                  arg match {
+                    case VList(asList) =>
+                      // we only allow one splice, so we assume the rest of the patterns
+                      val (revArgTail, spliceVals) = asList.reverse.splitAt(ptail.size)
```
johnynek: we could move `ptail.size` out of the lambda.

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -320,106 +320,140 @@ case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
...
                     sys.error(s"ill typed in match (${ctor.asString}${items.mkString}): $ts\n\n$other")
                 }
               }
-              else {
+            }
+            else {
+              { (arg: Value, acc: Env) =>
                 arg match {
                   case SumValue(enumId, v) =>
                     val cname = dt.constructors(enumId)._1
```
johnynek: can compute the index of the cname in advance.

PR #185, opened 2019-03-21
Add uncurry2 and uncurry3

snoble:
Just PRing 2 and 3 to make sure this looks right before typing out the rest

### DIFF
--- a/core/src/main/resources/bosatsu/predef.bosatsu
+++ b/core/src/main/resources/bosatsu/predef.bosatsu
@@ -30,6 +30,8 @@ export [
   sub,
   times,
   trace,
+  uncurry2,
+  uncurry3
 ]
 
 struct Unit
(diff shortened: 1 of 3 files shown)


core/src/main/resources/bosatsu/predef.bosatsu
```diff
@@ -67,6 +69,16 @@ def concat(front: List[a], back: List[a]) -> List[a]:
...
     _: reverse_concat(reverse(front), back)
 
+def uncurry2(f: t1 -> t2 -> r) -> (t1, t2) -> r:
+  \tuple ->
+    (x1, x2) = tuple
+    f(x1, x2)
+
+def uncurry3(f: t1 -> t2 -> t3 -> r) -> (t1, t2, t3) -> r:
+  \tuple ->
+    (x1, x2, x3) = tuple
```
johnynek, line 79: I want to support patterns in lambdas so this can become `\(x1, x2, x3) -> f(x1, x2, x3)`

Issue #14, opened 2017-11-23
add a Dict[String, a] literal


johnynek:
would be nice to have:
```
{"a": 1, "b": c, ...}
```
to return a hash or sorted Dict type.

johnynek:
Since I think it is unlikely I will ever add typeclasses directly to bosatsu, a fully generic encoding of a mapping type would require dictionary passing, either when constructing the dict or when accessing the dict.

The alternative is to fix the key type to for instance String. Since this is purely a syntactic trick, I think fixing key type to string is the way to go. We can still implement fully generic maps with dictionary passing if needed.

As an example of dictionary passing:
```
struct Order(compare: a -> a -> Comparison)

# e.g. an AVLTree,
enum Map: ...

# the empty map embeds the Order
def empty_Map(ord: Order[k]) -> forall v. Map[k, v]:
 ...
def add_Map(m: Map[k, v], k, v) -> Map[k, v]: ...

def get_Map(m: Map[k, v], k) -> Option[v]:
```

A crazier syntax could be passing the dictionary optionally to the literal:
```
str_map = {"one": 1, "two": 2, "four": 4}
int_map = {1: "one", 2: "two", 3: "three"} @ intOrd
```
I don't love this idea since it basically sneaks in typeclasses, but only for a few cases: string gets a free ordering, and dict has a way to pass in typeclasses, but they don't exist elsewhere in the language.

By sticking with only allowing string keys, we also leave the door open to changing our minds about typeclasses later and still string key dicts would work.

Given the intended use cases, I tend to think the dict should always be sorted by key, so the order in which things are computed or added is not observable at the output (which we often want in configurations). If the user intends to track order, they should use a list.

avibryant:
+1 to your reasoning above.

PR #187, opened 2019-03-22
Add Dict fns to Predef

johnynek:
This is the follow up to #167 that adds the methods.

The final step is to expand out literal dicts into calls to these methods.

part of #14

### DIFF
--- a/core/src/main/resources/bosatsu/predef.bosatsu
+++ b/core/src/main/resources/bosatsu/predef.bosatsu
@@ -18,7 +18,6 @@ export [
   add_key,
   cmp_Int,
   concat,
-  dict_to_List,
   div,
   empty_Dict,
   eq_Int,
@@ -27,6 +26,7 @@ export [
   gcd_Int,
   get_key,
   int_loop,
+  items,
   map_List,
   mod_Int,
   range,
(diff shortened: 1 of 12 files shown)


core/src/main/resources/bosatsu/predef.bosatsu
```diff
@@ -88,6 +97,15 @@ enum Option:
...
   Some(v)
 
+struct Order(fn: a -> a -> Comparison)
+
+external struct Dict[k, v]
+external def empty_Dict(comp: Order[k]) -> forall v. Dict[k, v]
+external def add_key(dict: Dict[k, v], key: k, value: v) -> Dict[k, v]
+external def get_key(dict: Dict[k, v], key: k) -> Option[v]
+external def remove_key(dict: Dict[k, v], key: k) -> Dict[k, v]
+external def dict_to_List(dict: Dict[k, v]) -> List[(k, v)]
```
johnynek: this should probably be called items_Dict

core/src/main/resources/bosatsu/predef.bosatsu
```diff
@@ -88,6 +97,15 @@ enum Option:
   None
   Some(v)
 
+struct Order(fn: a -> a -> Comparison)
```
johnynek, line 100: not really sure if we should name this type or not, but you can write combinators on it, so I tend to say yes.

Eventually we could optimize 1 item structs to not box. I think that's a good idea for sure for a compiled backend.

johnynek:
I'm kind of mixing two concepts here... I'm a bad human being...

I improve the error messages, but then remove some duplication in that code, but then realized it wasn't tested super well...

So:

1. we add the dict methods
2. we improve the error reporting which I used to debug adding these methods
3. improve the tests around conversion to and from Type to TypeRef, which I should have done a long time ago because it is annoying to see Predef::Fn, Predef::Tuple2, Predef::Unit in type error messages.

Issue #70, opened 2018-10-11
add an ability to escape to allow any unicode string to be an identifier


johnynek:
scala has a nice feature that you can use backticks to get any identifier to be legal. This is nice especially if we use bosatsu as a configuraton language and we want to use structs to generate yaml/json keys. We might need a key that is not a legal name in bosatsu.

```
val `val` = 3
```

More precisely, we could have:
```
struct Foo(`def :` bar)
```
so we can support any possible needed identifier.

johnynek:
The motivation to do this is to make Json generation from structs able to support existing json formats that may require characters or strings that are illegal bosatsu identifiers.

The narrowest way to support that is to only allow this to be used in struct and enum definitions, and anything implied by that (e.g. if we did name-based pattern matching or name-based applies to create those types). We use the types to guide Json generation, but the type names don't show up, so that seems to still satisy the use case if we have this limitiation.

I'm not sure I see the need to support beyond that. If we did, it should probably be only where a lower case identifier is currently used. That is a superset of the above. If we do this, that means patterns probably also need to track, and remember if backtick is used syntactically. Since we want to be able to re-render.

We could forget if we have a backtick, and just normalize down anything that isn't a valid unescaped indentifier.

johnynek:
There is subtlety here: if you escape a name you could legally use, is it distict:

e.g.

```
`foo` = 3
fn(foo)
```
Does that call `fn(3)`? I can see arguments on both sides. One argument is that there are just separate namespaces. The other, is there is only one namespace, but `` just lets you use names that are otherwise illegal to parse.

I tend to prefer the one-big-namespace interpretation, but that does mean that `Foo` can now create a let binding that shadows the constructor `Foo`.

Issue #12, opened 2017-11-23
unify operator overloading


johnynek:
a syntax like:
```
operator def (a: Int) + (b: Int) -> Int:
  ...
```
could allow us to unify unary and binary operators. We might not want many operators which have not been popular in most programming languages. That said, it is ugly to have some exceptions to rules and it makes things harder to learn.

johnynek:
actually, out of band, @avibryant suggested reusing the import binding syntax:

```
package Foo/Bar

import Baz [ foo as operator / ]
import Bosatsu/Predef [ plus_Int as operator + ]

```

So this introduces `operator` and probably `prefix operator` and `postfix operator` inside the import statement.

It also highlights we need to be able to do rebinding of Bosatsu/Predef.

johnynek:
@snoble thoughts on this?

johnynek:
The challenge here is that if we allow totally unbounded syntax as operators, how do we parse them? Maybe there is something elegant, but I don't know what that might be.

Alternatively, we could have a fixed universe of possible operators, and then use the binding syntax we describe here. In this way, the parser does not need to know about what operators are in scope to parse, but of course we could have a name error later.

johnynek:
A simplifying requirement would be that all operators are postfix, and due to currying, they are all functions on one variable. So `1 +` is parsed as `OperatorApply(Literal(1), Operator("+"))`

This means the few prefix operators can't be easily supported, but I don't want to go crazy with operator syntax. Again a core goal is you can understand any file by reading it.

If we force all operators to be functions, (maybe curried) then I wonder if we can just use regular syntax to define them, the difference is, they have different call syntax.

```
external def +(left: Int, right: Int) -> Int
```
then we could `import Bosatsu/Predef [ + as plus ]` if we wanted to opt out.

This is somewhat of a reversal of the idea that operators are second class and that we wanted to tax the writer by explaining them at the top of each file.

what do you think @snoble ?

snoble:
This feels really compelling. Just thinking of some edge cases. Can an operator still be treated as a regular variable? Can you do f(+)? Can you have multicharacter operators (eg >> would be nice for composition)? If so, how do you prioritize how multiple characters are combined?

Can you use a let to assign a function to an operator?

johnynek:
Very good questions.

Maybe there could be some notation to treat an operator like a regular value. Without that notation, then we treat it as autoapplying.

```
operator(+) = plus

z = 1 + 2

[...]

PR #192, opened 2019-03-25
Introduce Identifier remove a lot of Strings

johnynek:
This was not a fun refactor, but it was real techdebt.

Previously, we were using `String` in many cases for identifiers. It is hard to refactor such code because we don't know which of those are referring to binding names, type names or Constructors.

This introduces Identifier which is intended to keep track of this and be clear that we are never recreating those strings after parsing.

This is related to #12 and #70 since both of those will need to touch how identifiers are delt with.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
@@ -153,7 +153,7 @@ object DefRecursionCheck {
      */
     def checkForIllegalBinds[A](
       state: State,
-      bs: Iterable[Identifier.Bindable],
+      bs: Iterable[Bindable],
       decl: Declaration)(next: ValidatedNel[RecursionError, A]): ValidatedNel[RecursionError, A] =
       state.outerDefNames match {
         case Nil=> next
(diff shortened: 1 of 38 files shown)


core/src/main/scala/org/bykn/bosatsu/Package.scala
```diff
@@ -95,12 +97,12 @@ object Package {
...
     stmt: Statement):
-      ValidatedNel[PackageError, (TypeEnv[Variance], List[(String, RecursionKind, TypedExpr[Declaration])])] = {
+      ValidatedNel[PackageError, (TypeEnv[Variance], List[(Bindable, RecursionKind, TypedExpr[Declaration])])] = {
 
-    val importedTypes: Map[String, (PackageName, String)] =
+    val importedTypes: Map[Identifier, (PackageName, TypeName)] =
       Referant.importedTypes(imps)
 
-    val resolveImportedCons: Map[String, (PackageName, ConstructorName)] =
+    val resolveImportedCons: Map[Identifier, (PackageName, Constructor)] =
```
johnynek, line 105: Seems the key here should be Constructor.

core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
```diff
@@ -151,7 +153,7 @@ object DefRecursionCheck {
      */
     def checkForIllegalBinds[A](
       state: State,
-      bs: Iterable[String],
+      bs: Iterable[Identifier.Bindable],
```
johnynek: Bindable has been imported.

core/src/main/scala/org/bykn/bosatsu/Package.scala
johnynek: this isn't easy actually, imports don't know currently if the are bindable or constructors. I'm going to leave it as Identifier, which is at least an improvement.

PR #194, opened 2019-03-26
Add backtick escaping to allow any string as a binding

### LINKED ISSUE #70: add an ability to escape to allow any unicode string to be an identifier

johnynek:
close #70 

This is pretty easy after the #192 cleanup.

The JSON test shows the motivation: make objects with keys which are keywords in bosatsu.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/StringUtil.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/StringUtil.scala
@@ -42,10 +42,9 @@ abstract class GenericStringUtil {
     // We can ignore escaping the opposite character used for the string
     // x isn't escaped anyway and is kind of a hack here
     val ignoreEscape = if (quoteChar == '\'') '"' else if (quoteChar == '"') '\'' else 'x'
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/bykn/bosatsu/StringUtil.scala
```diff
@@ -42,9 +42,10 @@ abstract class GenericStringUtil {
     // We can ignore escaping the opposite character used for the string
     // x isn't escaped anyway and is kind of a hack here
     val ignoreEscape = if (quoteChar == '\'') '"' else if (quoteChar == '"') '\'' else 'x'
+    val tab = if (encodeTable.contains(quoteChar)) encodeTable else encodeTable.updated(quoteChar, s"\\$quoteChar")
```
johnynek: previously the ` character was not in the escape table, which means, internal versions were left unescaped. We need to ensure the quote character is in the replacement table.

PR #198, opened 2019-03-27
Add basic operator support

johnynek:
Implement the proposal from #12 

- [x] make Gen emit operators both in Var and applying form
- [x] consider a more constrained Declaration than Apply, which can't prove it has the right shape.
- [x] use operators in more tests and examples
- [x] exercise imports of operators in tests
- [x] implement prioritization so that `*` is applied before `+` and `&` is applied before `|` and `^`.

cc @snoble

### DIFF
--- a/core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
+++ b/core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
@@ -1134,4 +1134,15 @@ main = Foo(1, `package`, 3, 4)
       ("def" -> Json.JNumberStr("4")))
     ))
   }
+
+  test("test operators") {
+    evalTest(List("""
+package A
+
+operator + = add
+operator * = times
+
+main = 1 + 2 * 3
+"""), "A", VInt(7))
+  }
 }
--- a/core/src/test/scala/org/bykn/bosatsu/ParserTest.scala
+++ b/core/src/test/scala/org/bykn/bosatsu/ParserTest.scala
@@ -7,6 +7,7 @@ import org.scalatest.FunSuite
 import org.scalatest.prop.PropertyChecks.{ forAll, PropertyCheckConfiguration }
 import org.typelevel.paiges.{Doc, Document}
 
+import cats.implicits._
 import fastparse.all._
 import Parser.Indy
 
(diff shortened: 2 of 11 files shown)


core/src/test/scala/org/bykn/bosatsu/ParserTest.scala
```diff
@@ -372,6 +375,37 @@ class ParserTest extends FunSuite {
...
+
+/**
+ * This is a separate class since some of these are very slow
+ */
+class SyntaxParseTest extends ParserTestBase {
+
+  implicit val generatorDrivenConfig =
+    //PropertyCheckConfiguration(minSuccessful = 5000)
+    //PropertyCheckConfiguration(minSuccessful = 300)
+    PropertyCheckConfiguration(minSuccessful = 5)
```
johnynek: turn this back up to 300.

core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -1134,4 +1134,15 @@ main = Foo(1, `package`, 3, 4)
...
     ))
   }
+
+  test("test operators") {
+    evalTest(List("""
+package A
+
+operator + = add
+operator * = times
+
```
johnynek, line 1096: add tests for `def operator +(foo)`

johnynek:
@snoble check out these examples:

https://github.com/johnynek/bosatsu/pull/198/files#diff-754cdb31b141ef4ceaa2dfc0e731d063R61

What do you think?

I'm pretty happy with this...

I ulitimately did not make `.` an operator. The reason why is that we want `a.f` to be higher priority than `g(x)` so when we write `a.f(x) == (a.f)(x)`. So that mean function application is lower than `.`, but function application is not an operator (it has a start and end, and multiple arguments). So, the easiest thing was to pull `.` and applying out.

I do allow `.` in an operator like `=` if there are 2 or more characters.

Also, operators that start with `.` are higher priority than ones that don't, so you could build your own `.`-like operators that are multiple characters long.

Of course `>>=` is allowed since it has more than 1 character and can't be confused with `=` which is used for let binding. :)

snoble:
Nice! The examples are extremely satisfying

snoble:
Do you have a test for left to right precedence when other rules are equal?

johnynek:
@snoble I have added the tests you suggested. Let me know if you have any other concerns. I think operators are ready to go in this PR.

snoble:
I think it looks amazing!

Issue #118, opened 2018-11-13
make a list of normalization rules


johnynek:
We should make a doc listing all normalizations we can do:

1. we can inline lets (form of beta reduction, see below).
2. remove unused lets
3. push down a let to the smallest scope needed.
4. sort lets by string when at equivalent scope
5. replace total single arm matches with lets `x = (1, 2)` followed by `(y, z) = x` could be `y = 1` and `z = 2`, similarly with structs. Could be a similar normalization that desugars matches into a series of if/else statements in a normal order.
6. alpha conversion: rename let bindings/fn params in a canonical order.
7. beta reduction: `(\x -> g(x))(7) => g(7)`. Put another way, App on Lambda is the same as a Let.
8. eta conversion: `\x -> g(x) => g`

@snoble are there others?

snoble:
you can sometimes reduce a `match`. Eg if after other reductions you find you're matching on a call to a struct constructor

johnynek:
If you have `let x = y in f`, and y is cheap to compute (Var, Literal) or only referenced once, we could replace `x` with `y` in `f`.

johnynek:
this could be useful: https://www.microsoft.com/en-us/research/wp-content/uploads/1996/05/float.pdf

johnynek:
from the above paper:
```
let x = let y = z in z1
in w
```
to 
```
let y = z
in let x = z1
in w
```
also:
`case (let a = b in c) of ...` to let a = b in case c of ...`

also, lets above a match where they aren't needed on all branches can be moved inside:
```
x = y
match z:
  b1: r1
  b2: r2
```
becomes:
```
match z:
  b1:
    x = y
    r1
  b2:
    x = y
    r2
```
but of course we don't put any lets that have been shadowed or are unused.

Another note: currently normalization ignores the environment:
```
def normalize[A](te: TypedExpr[A]): Option[TypedExpr[A]]
```
but it could be:
```
def normalize[A](env: Map[Identifier, TypedExpr[A]], expr: TypedExpr[A]): Option[TypedExpr[A]]
```
where we can look up, rather than having to look down, not sure if this matters for expressions, but for whole programs it allows us to see other top-level bindings which we should aggressively inline.

johnynek:
if we have
```
match Cons(a1, a2, ... an):
   Cons(x1, ... xn): fn
```
even if there are more branches to the match, we can statically compute which one is right, then keep only that branch and:
```
x1 = a1
x2 = a2
...
[...]

PR #200, opened 2019-03-27
Draft of Normal Expression tags

snoble:
This is a first pass at adding a Normal Expression tag. Currently that tag is created at the beginning of evaluation and then never used. This is almost certainly the wrong time to normalize but it was convenient for the PR draft to see that this does not break any of the current evaluation tests.

The goal is to have a tag for each expression that describes the computation for the expression without and `Let`s and the remaining variable names (used in matches and lambdas) to be replaced by De Bruijn indices. We would also like to apply eta and beta regularization but that isn't in this PR.

This is useful for our project's use case to introduce caching into the evaluation process. If this is useful for bosatsu in general it would be lovely to have this merged in, after review and more tests. But I also understand if this isn't useful for the core project, or if this isn't the approach you would like to take.

Would love to hear your thoughts.


core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,303 @@
+package org.bykn.bosatsu
+
+import cats.data.{NonEmptyList, State}
+import cats.implicits._
+import rankn._
+
+import Identifier.Constructor
+
+sealed abstract class NormalExpression {
+  val maxLambdaVar: Option[Int]
```
snoble: it's useful to have a `maxLambda` to know when you have unbound lambdas inside your expression. e.g. when you're defining an expression inside of a lambda definition

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,303 @@
...
+  case class Literal(lit: Lit) extends NormalExpression {
+    val maxLambdaVar = None
+  }
+  case class Recursion(lambda: NormalExpression) {
+    val maxLambdaVar = None
+  }
+}
+
+object Normalization {
+  case class NormalExpressionTag(ne: NormalExpression, children: Set[NormalExpression])
```
snoble, line 83: `children` is useful to know in caching for when an expression has the same normal expression as one of the expressions it links to. In this case we would only save the value from the linked to expression in the cache so that we don't end up with circular evaluations

core/src/main/scala/org/bykn/bosatsu/TypedExpr.scala
```diff
@@ -20,6 +20,7 @@ sealed abstract class TypedExpr[T] {
    * for each expression
    *
    */
+  def updatedTag(t: T): TypedExpr[T]
```
snoble: This is always implemented using `copy`. There is surely a more direct way to implement this but I couldn't figure out how

johnynek:
Super excited! Will read tonight!

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -276,10 +276,11 @@ object Evaluation {
 
 }
 
-case class Evaluation(pm: PackageMap.Inferred, externals: Externals) {
+case class Evaluation(inferredPm: PackageMap.Inferred, externals: Externals) {
   import Evaluation.{Value, Scoped, Env}
   import Value._
 
+  val pm = NormalizePackageMap(inferredPm).normalizePackageMap
```
johnynek: nit: if we require NormalizedPackages before Evaluation, I'd rather the constructor require it and then make an apply method on Evaluation to pass it in. This is a minor issue though.

core/src/main/scala/org/bykn/bosatsu/NameKind.scala
```diff
@@ -2,38 +2,38 @@ package org.bykn.bosatsu
 
 import Identifier.Bindable
 
-sealed abstract class NameKind
+sealed abstract class NameKind[T]
```
johnynek: the changes in this file are a good change and can be PR'ed and merged independently. I like to look out for anything that is small and easy to merge and understand atomically.

johnynek: I think if you make `+T`, so T is covariant, then only `Let[T]` has a type parameter, the others can extend `NameKind[Nothing]` which may make some of the changes go away.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
johnynek: Can you add a comment to the meaning of None in this case? Also nit: can we make this a `def`. Note in scala you can override a `def` with a `val`. Unititalized vals can cause confusing errors, so I really only use abstract vals if I absolutely must (i.e. I am referencing a path dependent type of the val, and that can't be done on a `def`).

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,303 @@
...
+    branches: NonEmptyList[(Pattern[(PackageName, Constructor), Type], NormalExpression)])
+  extends NormalExpression {
+    val maxLambdaVar =
+      (arg.maxLambdaVar.toList ++ branches.toList.flatMap(_._2.maxLambdaVar))
+        .reduceLeftOption(Math.max)
+  }
+  case class LambdaVar(index: Int) extends NormalExpression {
+    val maxLambdaVar = Some(index)
+  }
+  case class Lambda(expr: NormalExpression) extends NormalExpression {
```
johnynek, line 52: a comment here as to how this gets away without having an identifier that is bound by a call to the lambda would help me.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,303 @@
...
+    val maxLambdaVar = None
+  }
+  case class Match(arg: NormalExpression,
+    branches: NonEmptyList[(Pattern[(PackageName, Constructor), Type], NormalExpression)])
+  extends NormalExpression {
+    val maxLambdaVar =
+      (arg.maxLambdaVar.toList ++ branches.toList.flatMap(_._2.maxLambdaVar))
+        .reduceLeftOption(Math.max)
+  }
+  case class LambdaVar(index: Int) extends NormalExpression {
```
johnynek, line 39: the purpose of this would be helpful, since this is a difference from the Expr and TypeExpr graph.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,303 @@
...
+    val maxLambdaVar = Some(index)
+  }
+  case class Lambda(expr: NormalExpression) extends NormalExpression {
+    val maxLambdaVar = expr.maxLambdaVar.map(_ - 1).filter(_ >= 0)
+  }
+  case class Struct(enum: Int, args: List[NormalExpression])
```
johnynek: is this a constructor function?

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
johnynek: what is a child in this context?

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,303 @@
...
+object Normalization {
+  case class NormalExpressionTag(ne: NormalExpression, children: Set[NormalExpression])
+}
+
+case class NormalizePackageMap(pm: PackageMap.Inferred) {
+  import Normalization._
+  import TypedExpr._
+
+  val normalizePackageMap: PackageMap.Normalized = 
+    PackageMap(
```
johnynek: I might expand this out a bit:
```scala
val packs = pm.toMap.toList
val normAll = packs.traverse { case (name, pack) =>
  normalizePackage(name, pack)
    .map((name, _))
}
PackageMap(normAll.runA(Map.empty).value.toMap)
```
I think is the same as this.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,303 @@
...
+
+  def normalizeExpr(expr: TypedExpr[Declaration], env: Env, p: Package.Inferred): State[
+    Map[(PackageName, Identifier), ResultingRef],
+    TypedExpr[(Declaration, NormalExpressionTag)]] =
+      expr match {
+        case a@Annotation(_, _, _) => normalizeAnotation(a, env, p)
+        case g@Generic(_, _, _) => normalizeGeneric(g, env, p)
+        case v@Var(_, _, _, _) => normalizeVar(v, env, p)
+        case al@AnnotatedLambda(_, _, _, _) => normalizeAnnotatedLambda(al, env, p).map(_.asInstanceOf[TypedExpr[(Declaration, NormalExpressionTag)]])
+        case a@App(_, _, _, _) => normalizeApp(a, env, p).map(_.asInstanceOf[TypedExpr[(Declaration, NormalExpressionTag)]])
```
johnynek: I wonder how hard it would be to remove these casts. What are the blockers to that?

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,303 @@
...
+        case l@Let(_, _, _, _, _) => normalizeLet(l, env, p)
+        case l@Literal(_, _, _) => normalizeLiteral(l, env, p)
+        case m@Match(_, _, _) => normalizeMatch(m, env, p)
+      }
+
+  private def combineWithChildren(nt: NormalExpressionTag) = nt.children + nt.ne
+
+  def normalizeAnotation(a: Annotation[Declaration], env: Env, p: Package.Inferred): State[
```
johnynek: `Anotation` is spelled wrong.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,303 @@
...
+  TypedExpr[(Declaration, NormalExpressionTag)]] = env._1.get(v.name) match {
+    case None => for {
+      ne <- norm((p, Left((v.name, v.tag)), env))
+      neTag = getTag(ne)._2
+    } yield v.copy(tag=(v.tag, neTag))
+    case Some(neTag) =>
+      State.pure(v.copy(tag=(v.tag, neTag)))
+  }
+
+  def normalizeAnnotatedLambda(al: AnnotatedLambda[Declaration], env: Env, p: Package.Inferred): State[
```
johnynek: I think a `type NormState[A] = State[Map[(PackageName, Identifier), ResultingRef, A]` in the `object Normalization` would help cut down on the type verbosity here.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,303 @@
...
+      neTag = term.tag._2
+      newNeTag = neTag.copy(children = (neTag.children + neTag.ne))
+      tag = (a.tag, newNeTag)
+    }  yield a.copy(term=term, tag=tag)
+
+  def normalizeGeneric(g: Generic[Declaration], env: Env, p: Package.Inferred): State[
+  Map[(PackageName, Identifier), ResultingRef],
+  TypedExpr[(Declaration, NormalExpressionTag)]] =
+    for {
+      in <- normalizeExpr(g.in, env, p)
```
johnynek: this is a single .map. I'd probably write it that way:
```scala
normalizeExpr(g.in, env, p).map { in =>
 ... 
}
```

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,303 @@
...
+      neTag = in.tag._2
+      newNeTag = neTag.copy(children = neTag.children + neTag.ne)
+      tag = (g.tag, newNeTag)
+    } yield g.copy(in=in, tag=tag)
+
+  def normalizeVar(v: Var[Declaration], env: Env, p: Package.Inferred): State[
+  Map[(PackageName, Identifier), ResultingRef],
```
johnynek: nit, I'd rather indent 2 beyond the def if you need to wrap.

core/src/main/scala/org/bykn/bosatsu/Package.scala
```diff
@@ -43,7 +43,9 @@ object Package {
   type Parsed = Package[PackageName, Unit, Unit, Statement]
   type Resolved = FixPackage[Unit, Unit, (Statement, ImportMap[PackageName, Unit])]
   type Interface = FixPackage[Nothing, Referant[Variance], Unit]
-  type Inferred = Package[Interface, NonEmptyList[Referant[Variance]], Referant[Variance], Program[TypeEnv[Variance], TypedExpr[Declaration], Statement]]
+  type Typed[T] = Package[Interface, NonEmptyList[Referant[Variance]], Referant[Variance], Program[TypeEnv[Variance], TypedExpr[T], Statement]]
```
johnynek: this is a nice change and can go in easily in a separate PR.

core/src/main/scala/org/bykn/bosatsu/PackageMap.scala
```diff
@@ -52,7 +52,24 @@ object PackageMap {
   type MapF2[A, B] = MapF3[A, A, B]
   type ParsedImp = PackageMap[PackageName, Unit, Unit, (Statement, ImportMap[PackageName, Unit])]
   type Resolved = MapF2[Unit, (Statement, ImportMap[PackageName, Unit])]
-  type Inferred = PackageMap[Package.Interface, NonEmptyList[Referant[Variance]], Referant[Variance], Program[TypeEnv[Variance], TypedExpr[Declaration], Statement]]
+  type Typed[T] = PackageMap[
```
johnynek: good change. PS: this type is just wild isn't it? I get pretty baffled messing around in this code due to how deep the types go. That said, having one set of values that go through different stages enforced by the types has been pretty cool.

core/src/main/scala/org/bykn/bosatsu/PackageMap.scala
```diff
@@ -52,7 +52,24 @@ object PackageMap {
...
+      Statement
+    ]
+  ]
+  
+  type Inferred = Typed[Declaration]
+
+  type Normalized = Typed[(Declaration, Normalization.NormalExpressionTag)]
+
+  def normalizePackages[A, B, C, D](map: Inferred): Normalized = {
+    PackageMap.empty
```
johnynek: this looks suspect... Can we add a comment or leave it as `???`

core/src/main/scala/org/bykn/bosatsu/PackageMap.scala
```diff
@@ -52,7 +52,24 @@ object PackageMap {
...
+    Program[
+      TypeEnv[Variance],
+      TypedExpr[T],
+      Statement
+    ]
+  ]
+  
+  type Inferred = Typed[Declaration]
+
+  type Normalized = Typed[(Declaration, Normalization.NormalExpressionTag)]
```
johnynek: Maybe this should be a type defined in `object Normalization`. I like to avoid circular dependencies where possible, and I think putting this here makes PackageMap depend on Normalization, and also vice-versa, but maybe I'm wrong.

core/src/main/scala/org/bykn/bosatsu/TypedExpr.scala
johnynek: I don't know of one. But Nit: can we implement this with a pattern match rather than a method to be consistent with the other code?

This can be a PR that can be merged immediately.

core/src/test/scala/org/bykn/bosatsu/NormalizationTest.scala
```diff
@@ -0,0 +1,79 @@
+package org.bykn.bosatsu
+
+import cats.data.{ Validated, ValidatedNel}
+import fastparse.all._
+import org.scalatest.FunSuite
+
+class NormalizationTest extends FunSuite {
+
+  def resolveThenInfer(ps: Iterable[Package.Parsed]): ValidatedNel[PackageError, PackageMap.Inferred] =
```
johnynek: I have a `object TestUtils` that I have been trying to centralize utilities for testing to make it easier to get going in tests. I think we should see what is there and if these aren't, move them there. I let some of the test code get out of control a bit...

core/src/test/scala/org/bykn/bosatsu/NormalizationTest.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  test("simple package normalizes") {
+    val p1 = parse(
+"""
+package Willem/Foo
+
+struct Pair(first, second)
+
+def bar(x):
+  baz = \y -> Pair(x, y)
```
johnynek: the bug you hit with `def baz(y): Pair(x, y)` should be fixed now, BTW. I added that as a test case.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,303 @@
...
+    case Right(te) => te.tag
+    case Left((_, t)) => t
+  }
+
+  private type Ref[T] =
+    Either[(Identifier, T), TypedExpr[T]]
+
+  private type SourceRef = Ref[Declaration]
+  private type ResultingRef = Ref[(Declaration,  NormalExpressionTag)]
+  private type Env = (Map[Identifier, NormalExpressionTag], List[Option[Identifier]])
```
johnynek: can we comment the meaning of `List[Option[Identifier]]`? I don't know what the None means, nor why there is a List there.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
johnynek: Is this basically a subgraph where `ne` points to all `children` forming an equivalence set?

If so, I wonder if another representation might be:
```
type NormTag = Long
type TagsToExpr = Map[NormTag, Set[NormalExpression]]
type ExprToTag = Map[NormalExpression, NormTag] // we may not need this since you are putting it in the tag on the TypedExpr?
```

johnynek: Thinking more, I wonder if we could get to a world where we make the `NormTag` a `Sha256HashValue` or something, where it is a hashed version of everything the expression depends on.

So, then, two expressions are the same (assuming no collisions in sha256) if they share the same value. That would be really cool since we could serialize these Normalization Tags and load them back up. You've probably thought about the pros and cons here. What do you think?

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: I'm also 90% sure I can get rid of the `Option`. This was just because when `Identifier`s were just `String` and when you cared about them being Bindable there were `Option[String]`. Now we just have `Bindable` as being explicit.

I'll add a comment but I'll also explain here. This is here for the de bruijn indexing substitutions. The context is the `Env` is only here to produce a `NormalExpression` given a `TypedExpr`. When an identifier is encountered it looks it up in `_1` of the `Env`. Sometimes that will map to `NormalExpression.LambdaVar(2)` for example, eg when you're referencing the identifier of the argument of a lambda.

But, when you define a new lambda inside (which you do all the time because of alpha regularization) all the `LambdaVar` indices have to be incremented. This is where the `List` comes in. It is just a `List` of identifiers of `Lambda` args. By appending to it each time we start a new `Lambda` we have exactly the de bruijn index as the list index.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: Yeah... I think this is exactly right. For my purposes I was just using the case class instances as keys in a Map. The hashing would have to happen after the reductions of course, but that's an implementation detail.

That said, let me devil's advocate the other side.
a) (and I think this is a good argument for now) it's nice to have them unsha'd as we're trying to understand them and figure out how to use them. Eventually we can have tests where we take the sha of what were expecting. But while we're learning it's really nice for the failing test to give us what we expect
b) I think this is a little sillier of an argument, but in theory the normal expressions contain a bunch of static analysis that may be useful. For example, you could parse the normal expression at evaluation time and quickly see if a `Let` variable is used and if not just not evaluate. This is actually right in the `maxLambda` I think. The cool thing about this is it's a completely static decision so you don't need a `Promise` or a lazy `Eval`. You just skip it all together.

So my proposal would be not to sha these for now but set a calendar date to re-evaluate

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: `None` is for when a `NormalExpression` does not contain any expressions. eg with a `LambdaVar` and a `Literal`. All other numbers actually contain meaning that isn't communicated by this. `None` is really `-Infinity`

snoble:
> This is super interesting! I really want to merge this or something like it.
> 
> If you have time, could we pull out the easy parts first maybe into one PR. Maybe everything other than the Normalization file and the test? Then we could have two PRs: one that make the other changes (updateTag, the new `type` aliases, etc... then one branching off that with the normalization algorithm?

Yeah! I'll do this right now. Should be super uncontentious since it'll just generalize a bunch of things

 
> I'm still wrapping my head around how normalization works, and I don't fully understand the differences with the TypedExpr dag yet.
> 
> I'm also interested to hear your thoughts on where alpha transforming (which maybe you are doing here: renaming the internal let bindings in a constant order), and beta and eta reduction should all go. Seems very related to what you want here.
> 
> Lastly, I made this issue to track related things: #118

You're asking exactly the right questions. In essence all this level of normalization is doing is creating expressions that have forced alpha reduction because there are literally no `Identifier`s allowed in the type. Why not just re-use `Expr` or `TypedExpr`? Because it's hard to enforce the no `Identifier` in the type system if you do. And you end up introducing things like `LambdaVar` which would mean you would have to `case` over it in evaluation with an error that says it never happens.

The one exception to the above is `Pattern` because they are so much richer than they used to be. But this shouldn't be hard to eliminate `Identifier` from `Pattern` for the use of `NormalExpression`s.

I have beta reduction code but I didn't submit it because it's no longer current (mostly because of changes how `Match` works). But it'll be fun to re-write and is surprisingly easy. I was doing normal-order reduction because of the warnings in https://en.wikipedia.org/wiki/Beta_normal_form but I suspect bosatsu's normality constraints makes that unnecessary. But it also works fine.

One question would be to apply beta reduction to recursive functions. My instinct is to say no because it means you can write small programs that would take hours to normalize. But maybe there are times you want that?

I haven't done any work on eta reduction but I suspect maxLambda will help.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: > what is a child in this context?

`children` is a flattened set of all the normalexpressions of all the child expressions (eg, any expressions linked to by this expression). It's useful for cases like

```
def foo(a):
  bar(a)
```
Since after reduction `foo` and `bar` have the same normal expression, if `foo` ends up getting stored in the cache then it is a function that passes `a` to a function that it will look up in the cache and get itself :scream:. This could really be replaced with a boolean.

Can you explain those `type`s more?

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
snoble: since Evaluation is now agnostic tags I'll normalize in `Main` and pass it in

core/src/main/scala/org/bykn/bosatsu/NameKind.scala
snoble: oops, missed this. I can go back and fix this

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: The main purpose is it allows you to be agnostic to identifiers chosen by the user that have no impact on the result of the expression. But you can accomplish this with variable renaming. It has the added value of being more agnostic to context in the normal expression.

Consider
```
foos.map(\foo -> (foo, bars.map(\bar -> bar.add(1))))
```
If you were to just do variable substitution you'd end up with something like
```
foos.map(\x0 -> (x0, bars.map(\x1 -> x1.add(1))))
```

but note that `bars.map(\bar -> bar.add(1))` could be its own expression without any unbound variables. But if you were normalize you would have
```
bars.map(\x0 -> x0.add(1))
```
Which is not what we saw in the first normalization. With de bruijn notation (I'm going to make up a bosatsu syntax that does not exist to accomplish this) we get
```
foos.map(-> (λ0, bars.map(-> λ0.add(1))))
```
and
```
bars.map(-> λ0.add(1))
```
I don't have a use for this but I could imagine someone wanting to parse the normal expressions and this would allow one more place where you can prove equality

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: Yeah. Well, it is an applied constructor function

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: There is a place where I call `normalizeApp` and I want to then deconstruct the resulting `App` to get out the normal expression so the resulting type is `TypedExpr.App` and not `TypedExpr`. I think there is a way I could fix the signature of `normalizeExpr` to accept extensions of `TypedExpr`. But I couldn't figure out the syntax.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: will do... I need a better vim linter (it also gets confused when a def doesn't have `{}`s)

core/src/main/scala/org/bykn/bosatsu/PackageMap.scala
snoble: heh, something I have to remind myself when confronted with it is it's all just rewriting types. It's still just the same case class

core/src/main/scala/org/bykn/bosatsu/PackageMap.scala
snoble: oops, gone

core/src/main/scala/org/bykn/bosatsu/PackageMap.scala
snoble: I'm sure it does. I'll move it

core/src/test/scala/org/bykn/bosatsu/NormalizationTest.scala
snoble: cool. I'm going to try to fill out the tests this weekend

core/src/test/scala/org/bykn/bosatsu/NormalizationTest.scala
snoble: oh right... we'll delete this

snoble:
As a note for myself I need to move the type in `PackageMap` and write tests (which will surely expose some bugs)

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,397 @@
...
+
+  private def normalOrderReduction(expr: NormalExpression): NormalExpression = {
+    import NormalExpression._
+    val nextExpr = expr match {
+      case App(Lambda(nextExpr), arg) => {
+        applyLambdaSubstituion(nextExpr, Some(arg), 0)
+      }
+      case App(fn, arg) => {
+        val nextFn = normalOrderReduction(fn)
+        App(nextFn, arg)
```
johnynek: why don't we need to apply to arg?

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,397 @@
...
+      }
+      /*case Match(struct@Struct(enum, args), branches) => { TODO: If s matches a pattern we should perform lambdaSubstitution
+        Match(struct, branches)
+      }*/
+      case Recursion(Lambda(innerExpr)) if(innerExpr.maxLambdaVar.map(_ < 0).getOrElse(false)) => {
```
johnynek: nit: space after if, but actually no parens needed in case matches:

```scala
if innerExpr.maxLambdaVar.exists(_ < 0) =>
```
what is a negative index? Is it special? Can we remind the reader here with a comment?

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,397 @@
...
+        applyLambdaSubstituion(innerExpr, None, 0)
+      }
+      case Lambda(App(innerExpr, LambdaVar(0))) => {
+        innerExpr
```
johnynek: this is eta conversion, right? Can we add a comment to help me and other keep them straight.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,397 @@
...
+      case al @ App(Lambda(_), _) => normalOrderReduction(al)
+      case App(fn, arg) =>
+        normalOrderReduction(fn) match {
+          case l @ Lambda(_) => normalOrderReduction(App(l, arg))
+          case nfn @ _       => App(nfn, normalOrderReduction(arg))
+        }
+      case extVar @ ExternalVar(_, _) => extVar
+      case Match(arg, branches) =>
+        normalOrderReduction(arg) match {
+          // case s @ Struct(_, _) => normalOrderReduction(Match(s, branches)) TODO: Check if we know enough about s to validate it matches a pattern. If so, recurse.
```
johnynek: you can see code like this in Evaluation for how to do this.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,397 @@
...
+    NormState[TypedExpr[(Declaration, NormalExpressionTag)]] = {
+      val (nextEnv, neWrapper) = l.recursive match {
+        case RecursionKind.Recursive => {
+          val lambdaVars = Some(l.arg) :: env._2
+          val nextEnv = (env._1 ++ lambdaVars.zipWithIndex
+            .collect { case (Some(n), i) => (n, i) }
+            .toMap
+            .mapValues(idx => NormalExpressionTag(NormalExpression.LambdaVar(idx), Set[NormalExpression]())),
+            lambdaVars)
+          val neWrapper = {ne: NormalExpression => NormalExpression.Recursion(NormalExpression.Lambda(ne)).asInstanceOf[NormalExpression]}
```
johnynek: can we remove the cast? I think by adding a `: NormalExpression` is all that is needed (ascribing a type, vs casting).

core/src/test/scala/org/bykn/bosatsu/NormalizationTest.scala
```diff
@@ -0,0 +1,140 @@
...
+      List("""
+package Lambda/Identity
+
+def foo(x):
+  y = x
+  y
+out = foo
+"""
+      ), "Lambda/Identity", NormalExpressionTag(
+        Lambda(LambdaVar(0)), Set(LambdaVar(0))
```
johnynek, line 122: this is really cool that it can see that this is the same as `\x -> x`

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,397 @@
...
+        .iterator
+        .zipWithIndex
+        .collectFirst { case ((ctor, params, resType), idx) if ctor == c => (idx, params.size) }
+        .get
+
+      def loop(params: Int, expr: NormalExpression): NormalExpression =
+        if (params == 0) expr
+        else loop(params - 1, NormalExpression.Lambda(expr))
+
+        loop(arity, NormalExpression.Struct(enum, ((arity - 1) to 0 by -1).map(NormalExpression.LambdaVar(_)).toList))
```
johnynek, line 651: the order of the indices looks backwards to me, but I'm very likely wrong. Is it easy to add a test for this (apologies if I missed it), since `struct Foo(x, y)` is like `Foo = \x -> \y -> Foo(x, y)`

Which, by the way, do we have eta conversion for constructor functions? Seems like we need that too.

snoble:
I will definitely be able to hit that target. It's nice so much is being hit already
image

core/src/test/scala/org/bykn/bosatsu/NormalizationTest.scala
snoble: We're just getting started with stuff like that! :p

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: Ohhh... I was busy looking at covariant and contravariant type variables. That makes way more sense

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: This is a good question and I should at minimum add a comment (although it would be nice to have the structure of the code enforce this). The algorithm goes, normalize the head, normalize the tail recursively, check the head again for normalization opportunities, if they exist then repeat.

If I understand correctly this should result in it always normalizing (but this might not be necessary given the type restrictions) https://en.wikipedia.org/wiki/Lambda_calculus#Optimal_reduction

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: This is a great question and i will add a comment. You can't have a negative index (and I thiiiiink I could prove that with the type system. Just need to give some care in `lambdaSubstitution` where there is a decrement... but I think that can be shown to be non-negative. But `maxLambdaVar` can. That's because the `maxLambdaVar` of a `Lambda` is the decrement of its expression. If an expression has a negative `maxLambdaVar` then all the variables are bounded internally and wrapping with an `App(Lambda` will have no effect (it'll just thunk).

In the `App(Lambda` case it's captured with `beta` reduction. But I've got this `Recursion` expression for a solution to recursion.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: and a test to make sure it works

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: thank you!!!

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: yeah... that needs to be checked. I keep straining to wrap my head around which way the indexes go that which probably means it needs to be refactored (but at minimum tested)

One way to think about it is that the constructor uses the outer most Lambda first (ie the inner most App applies to the outer most Lambda). So the first argument of the struct will have the highest `LambdaVar` index... I think?

Note that constructor functions are just `Lambda`s that return a `Struct`. So beta reduction should handle that. But I need tests (and to actually use the beta reduction). I realized the other day from something you said that we could eliminate `Struct` and just have special constructor function expression. The latter feels a little more correct, but it feels harder to implement with taking advantage of reductions of `Match`. I'll go forward this way, get the tests working, then we can refactor with tests keeping us safe.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,666 @@
...
+   * ref: https://en.wikipedia.org/wiki/De_Bruijn_index
+   */
+  case class Lambda(expr: NormalExpression) extends NormalExpression {
+    val maxLambdaVar = expr.maxLambdaVar.map(_ - 1)
+  }
+  case class Struct(enum: Int, args: List[NormalExpression]) extends NormalExpression {
+    val maxLambdaVar = args.flatMap(_.maxLambdaVar).reduceLeftOption(Math.max)
+  }
+  case class Literal(lit: Lit) extends NormalExpression {
+    val maxLambdaVar = None
```
johnynek: nit: better as `def maxLambdaVar = None`

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,666 @@
...
+  }
+  case class Recursion(lambda: NormalExpression) extends NormalExpression {
+    val maxLambdaVar = lambda.maxLambdaVar
```
johnynek: nit: better as `def maxLambdaVar = lambda.maxLambdaVar` no need to have a field on this.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,666 @@
...
+  }
+}
+
+sealed abstract class NormalPattern {}
```
johnynek, line 66: nit: don't need `{}`

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
johnynek: I think we should schedule a video chat to talk about this. I'm still not sure I understand.

a type like `(A, Set[A])` looks like a node and a set of edges. So, a collection of those can be thought of as a graph. But is it a directed or undirected graph? The types don't say, but there might be the invariant that if edge `a -> b` exists so will `b -> a`.

Also, we can imagine another representation of a graph: `List[(A, A)]` which is the list of all edges. Lastly, I *think* the graphs we care about here are actually cliques: clusters of equivalent expressions, for those kinds of graphs, we can represent them as `(Set[A], Id)` some id type along with the set of nodes in the clique.

I still don't currently understand why we would choose one representation over the other. I think a comment would help clarify for me.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,666 @@
...
+
+  sealed trait PatternMatch[+A]
+  case class Matches[A](env: A) extends PatternMatch[A]
+  case object NoMatch extends PatternMatch[Nothing]
+  case object NotProvable extends PatternMatch[Nothing]
+
+  val noop: (NormalExpression, PatternEnv) => PatternMatch[PatternEnv] = { (_, _) => Matches(Map()) }
+  val neverMatch: (NormalExpression, PatternEnv) => PatternMatch[Nothing] = { (_, _) => NoMatch }
+
+  def structListAsList(ne: NormalExpression): PatternMatch[List[NormalExpression]] = {
```
johnynek, line 97: can you comment on the meaning of this? Is this just the analog of the Value type for checking the, representation of Lists?

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,666 @@
...
+  case class NormalExpressionTag(ne: NormalExpression, children: Set[NormalExpression])
+  type NormalizedPM = PackageMap.Typed[(Declaration, Normalization.NormalExpressionTag)]
+  type NormalizedPac = Package.Typed[(Declaration, Normalization.NormalExpressionTag)]
+
+  type PatternEnv = Map[Int, NormalExpression]
+
+  sealed trait PatternMatch[+A]
+  case class Matches[A](env: A) extends PatternMatch[A]
+  case object NoMatch extends PatternMatch[Nothing]
+  case object NotProvable extends PatternMatch[Nothing]
```
johnynek, line 92: what does does this branch mean? Is it kind of ill-typed, or unknown in some way? It would seem that well-typed programs either match or not.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,666 @@
...
+  }
+
+  def findMatch(m: NormalExpression.Match) =
+    m.branches.collectFirst(Function.unlift( { case (pat, result) =>
+      maybeBind(pat).apply(m.arg, Map()) match {
+        case Matches(env) => Some(Some((pat, env, result)))
+        case NotProvable => Some(None)
+        case NoMatch => None
+      }
+    })).get // Totallity of matches should ensure this will always find something unless something has gone terribly wrong
```
johnynek, line 282: nit: totality is misspelled.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,666 @@
...
+
+  def solveMatch(env: PatternEnv, result: NormalExpression) =
+    ((env.size - 1) to 0 by -1).toList.map(env.get(_).get) // If this exceptions then somehow we didn't get enough names in the env
```
johnynek: why the .toList? seems not needed. I think `.iterator` would be slightly better for perf so we wouldn't reallocate a copy in the map.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,666 @@
...
+ 
+  def normalOrderReduction(expr: NormalExpression): NormalExpression = {
+    import NormalExpression._
+    val nextExpr = expr match {
+      case App(Lambda(nextExpr), arg) => {
+        applyLambdaSubstituion(nextExpr, Some(arg), 0)
+      }
+      case m@Match(struct@Struct(enum, args), branches) =>
+        findMatch(m) match {
+          case None => m
```
johnynek, line 323: this shouldn't really happen, right?

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,666 @@
...
+    })).get // Totallity of matches should ensure this will always find something unless something has gone terribly wrong
+
+  def solveMatch(env: PatternEnv, result: NormalExpression) =
+    ((env.size - 1) to 0 by -1).toList.map(env.get(_).get) // If this exceptions then somehow we didn't get enough names in the env
+      .foldLeft(result) { case (ne, arg) => NormalExpression.App(ne, arg) }
+ 
+  def normalOrderReduction(expr: NormalExpression): NormalExpression = {
+    import NormalExpression._
+    val nextExpr = expr match {
+      case App(Lambda(nextExpr), arg) => {
```
johnynek: nit: we usually avoid `{}` in the case matches. They aren't needed. Just indentation is enough

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,666 @@
...
+      case Match(arg, branches) =>
+        val nextMatch = Match(normalOrderReduction(arg), branches)
+        findMatch(nextMatch) match {
+          case None => nextMatch
+          case Some(_) => normalOrderReduction(nextMatch)
+        }
+      case lv @ LambdaVar(_)  => lv
+      case Lambda(expr)       =>
+        normalOrderReduction(expr) match {
+          case a@App(innerExpr, LambdaVar(0)) => normalOrderReduction(Lambda(a))
```
johnynek: this looks like it can be an infinite loop. Can you comment why it can't be? e.g. if you have `a == App(_, LambdaVar(0))` why does normalOrderReduction(a) not return the same? Is this just hard wiring knowledge of eta conversion here?

Like, why can `normalOrderReduction` ever return something that could be eta converted? That seems like that shouldn't happen.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
johnynek: this is beta conversion right? If so, can we comment here just to keep it straight?

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,666 @@
...
+        }
+      case Recursion(Lambda(innerExpr)) if(innerExpr.maxLambdaVar.map(_ < 0).getOrElse(true)) => {
+        applyLambdaSubstituion(innerExpr, None, 0)
+      }
+      case Lambda(App(innerExpr, LambdaVar(0))) => {
+        innerExpr
+      }
+      case _ => expr
+    }
+    val res = nextExpr match {
```
johnynek: why the two steps? why can't we express in a single loop? Can you comment?

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -0,0 +1,666 @@
...
+        val nextMatch = Match(normalOrderReduction(arg), branches)
+        findMatch(nextMatch) match {
+          case None => nextMatch
+          case Some(_) => normalOrderReduction(nextMatch)
+        }
+      case lv @ LambdaVar(_)  => lv
+      case Lambda(expr)       =>
+        normalOrderReduction(expr) match {
+          case a@App(innerExpr, LambdaVar(0)) => normalOrderReduction(Lambda(a))
+          case na @ _ => Lambda(na)
```
johnynek: it seems to me this should be safe:
```
val exp1 = normalOrderReduction(exp)
if (exp1 != exp) normalOrderReduction(Lambda(exp1))
else {
  Lambda(exp)
}
```
Not clear to me why that doesn't work.

core/src/test/scala/org/bykn/bosatsu/NormalizationTest.scala
```diff
@@ -0,0 +1,242 @@
...
+      )
+    )
+    normalTagTest(
+      List("""
+package Lambda/Always
+
+out = \x -> \y -> x
+"""
+      ), "Lambda/Always", NormalExpressionTag(
+        Lambda(Lambda(LambdaVar(1))), Set(Lambda(LambdaVar(1)), LambdaVar(1))
```
johnynek, line 86: maybe make a function to build these:
```
def mkTag(ne: NormalExpression) = NormalExpressionTag(ne, Set(ne))
```
so it is clearer the part we are testing?

Ahh, I think this relates to my misunderstanding of the meaning of the Set. So, the right hand side is the set of all strict sub-expressions, so, which I guess we hope to form a lattice. Is that right?

johnynek:
BTW: this page is helpful:

https://codecov.io/gh/johnynek/bosatsu/pull/200/diff?src=pr&el=tree#diff-Y29yZS9zcmMvbWFpbi9zY2FsYS9vcmcvYnlrbi9ib3NhdHN1L05vcm1hbGl6YXRpb24uc2NhbGE=

you can see what is currently untested. Maybe we can just write tests to trigger enough of those to get Normalization up to 85% or so? Then we can merge and iterate.

johnynek:
question: it seems to me we want to lift all lets out of a recursion, but pull all lets into a lambda. This is vague intuition about making things easy to compile, since recursions we will try to compile into loops (if they recur in tail position), whereas lambdas we want to minimize creating closures.

How does that intuition sound to you? How does it relate to normalization? Are these distinct concerns? I must admit, I don't understand this code well enough now to see how lets are handled at all. I need to look into that (are they all converted into lambda + app, in which case they just ride on the normalization of those rules?, if so, what about recursion, since I think we want to minimize the scope of the recursion).

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: oh, I thought the `def` was only important on the abstract class. I am totally prematurely optimizing here though so I'll just make it a `def` instead of over thinking it

core/src/test/scala/org/bykn/bosatsu/NormalizationTest.scala
snoble: Totally happy to have a video chat

They are supposed to be a flat set of all descendants, so perhaps `children` is a bad name. Not that a NormalExpression does not have `children` themselves. But `NormalExpressionTag` does and that's on the tag to these `TypedExpr`.

I could actually remove these sets for now. I found them helpful in the past but they also seem to be a sledge hammer to a problem that might have another solution.

What they solve is in caching if you have
```
def bar(x):
  x.add(2)

def foo(x):
  bar(x)
```
Then you have an `AnnotatedLambda` for `foo` and for `bar`. After normalizations both of these expressions have the same `NormalExpression`. But it is important to note that `NormalExpressions` are just descriptors of the final outcome, the `TypedExpr`s themselves do not undergo normalization. So what happens if we add some caching to the `Evaluation` code. Both the `TypedExpr` for `foo` and `bar` have the same `NormalExpression` so if we use this as a key only one of them can be in the cache.

But the evaluation for a `AnnotatedLambda` is just a `Fn` that calls back into the evaluator with an `Env` and a `TypedExpr` in the closure.

What happened with my implementation is that `foo` would go into the cache, when it was called it would come to a call to `bar`, which it would look up in the cache, and find `foo` again.

So my hammer was to introduce these `children` and if a `TypedExpr` had its `ne` as a member of its `children` we would not write it to the cache

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: I keep convincing myself that I need this but then when I go to write out the explanation I convince myself of the opposite. I should probably construct a proof one way or the other.

The intuition is that I'm looking for the opportunity to reduction the head without reducing the children first (which would be applicative order reduction I think). So first you look for an opportunity to reduce the head. But the new head could be anything, so you have to match through all expression types again.

This gets you to the second step, which is to reduce the children only if their are no opportunities on the head. The second step consists of,
  1) checking for a head reduction opportunity. Recurs if it exists
  2) fully reduce the children
  3) check if the children reduction created a new head reduction opportunity. Recurs if it exists

(I should add the above as a comment)

All this said, a lot of this code feels like exactly what dagon does

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: I think it does? Let me get a few more tests then I can give it a try

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: This is the most common branch. Remember we're dealing with NormalExpressions and not actual values. This code is just trying to figure out if there's enough information in the NormalExpression to pick a branch (with just static analysis).

So if you are matching on anything other than a `Struct` expression then you'll like end up here. It's a little misleading in the tests because virtually everything reduces unless we're in a lambda or using an external

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: huh, yeah. That's clearly right. Don't know what I was thinking

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: It can. It's ugly code because of my `Option[Option[(...)]]` type which is hard to read. The `collectFirst` will go until it finds a `Matches` or a `NotProvable`. If it finds a `NotProvable` it return `None`. Basically `NotProvable` is saying "No head reduction opportunities here"

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: It shouldn't infinite loop in that case because of
```scala
      case Lambda(App(innerExpr, LambdaVar(0))) => {
        innerExpr
      }
```
above which will unwrap to the `_`.

The second question is harder for me to answer. I'll try to come up with an example

snoble:
> question: it seems to me we want to lift all lets out of a recursion, but pull all lets into a lambda. This is vague intuition about making things easy to compile, since recursions we will try to compile into loops (if they recur in tail position), whereas lambdas we want to minimize creating closures.
> 
> How does that intuition sound to you? How does it relate to normalization? Are these distinct concerns? I must admit, I don't understand this code well enough now to see how lets are handled at all. I need to look into that (are they all converted into lambda + app, in which case they just ride on the normalization of those rules?, if so, what about recursion, since I think we want to minimize the scope of the recursion).

My instincts are a little different in that I think I'd lift all Lets with unbound variables out of the lambdas, recursive or otherwise. I think the normalization tags could help in the process of this in that they know which expressions depend on unbound variables and which don't.

That said, these NormalExpressions shouldn't get affected because you're not changing any semantics outcomes of the expressions, just the manner in which they are computed

snoble:
> BTW: this page is helpful:
> 
> https://codecov.io/gh/johnynek/bosatsu/pull/200/diff?src=pr&el=tree#diff-Y29yZS9zcmMvbWFpbi9zY2FsYS9vcmcvYnlrbi9ib3NhdHN1L05vcm1hbGl6YXRpb24uc2NhbGE=
> 
> you can see what is currently untested. Maybe we can just write tests to trigger enough of those to get Normalization up to 85% or so? Then we can merge and iterate.

This is super helpful. If I can get to 85% that should knock out most of the bugs. And if a few more make it in it should be fine because this code isn't actually used by anything.

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: Here's an example
```
Lambda(App(
  ExternalVar(
    PackageName(NonEmptyList.fromList(List("Extern","Simple")).get),Identifier.Name("foo"
  ),
  App(Lambda(LambdaVar(0), LambdaVar(0)))
))
```
The tail will get beta'd to `LambdaVar(0)` so this will get eta'd down to just the `ExternVar`

snoble:
image
one more line of code!

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: I think that's right. It's for when you have a `NormalExpression` and you're trying to see if it pattern matches to a list pattern, so you attempt to unwrap the `Struct` into a list. This only works if the normal expression is fully represented as a list (eg `Struct(1, List(a, Struct(1, List(b, Struct(0, List())))))`)

Note we only use this for when we have a `splice` followed by a pattern because we need to do that reversing of the tail trick

snoble:
I mean... technically it's a pass
image

johnynek:
Nice work!!! I'll make one more pass over it, but expect to merge soon.

snoble:
One test I'd love to add, but I don't know if it's possible, is for a program that passes type checking that
a) adding the normal tags never crashes
b) if you map the normal tag away you get back something that is equal to the original packageMap (I don't want to be accidentally changing any expressions)

core/src/test/scala/org/bykn/bosatsu/NormalizationTest.scala
snoble: Oh... I think I realize a way I'm still being unclear in my description. "They are supposed to be a flat set of all descendants" means it is the set of `NormalExpression`s for all the strict-sub `TypedExpr`s of the `TypedExpr` being tagged

snoble:
Something I keep thinking about with `normalOrderReduction` is it's just a DAG rewrite. I've not explored dagon well enough to know if that's how this should be done

johnynek:
This is a huge work and great test coverage. I think I understand it now. Kudos to you!

I’m happy to merge now. You keep sending PRs? Should we merge and continue in follow ups?

snoble:
Thank you! Yeah, totally happy to iterate and keep PRing. I'm sure there are many ways for this code to continue to improve

johnynek:
BTW: I didn't get the Unprovable idea until you replied to my comment: in for instance a lambda, you might know nothing about a variable until later.

It would be interesting to see if we can use this code, or this idea to optimize the expressions before we evaluate them.

snoble:
Yeah that's interesting. In a sense reduction is just a form of evaluation with incomplete information. So in the extreme you could just keep looping back and forth between some sort normalization and acquiring runtime information (which I think is just evaluating externals) until you have no more unknowns and can't normalize anymore. I suspect this wouldn't perform particularly well. But there's probably something short of that which would be valuable

PR #202, opened 2019-03-28
Generalize some things to get normalization in

snoble:


snoble:
Just pulled over from https://github.com/johnynek/bosatsu/pull/200

I think I've updated for all the relevant comments and this should be ready to go

johnynek:
This is great! Thanks.

I'm not too concerned with the minor drop in code coverage here since I'm sure will test updatedTag soon enough.

Issue #196, opened 2019-03-26
Using the type system to represent a recordset


wvanbergen:
I am trying to represent a recordset / query result using the type system.

- It will have 0 or more rows.
- It will have 0 or more columns.
- Values in the same column should always have the same type. Or, phrased alternatively, every record should conform to the same type.

The goal is to provide a library of functions to help dealing with these resultsets: e.g. remove a field, calculate a new field, join or union two resultsets, etc.

#### First attempt using lists of lists

The first attempt to describe this is using a List of List of optional values

```
enum Value:
  IntValue(i: Int)
  StringValue(s: String)
  
struct Resultset(records: List[List[Option[Value]]])

resultset = Resultset([
  [Some(StringValue("foo")), Some(IntValue(123)), None],
  [Some(StringValue("bar")), Some(IntValue(456)), Some(IntValue(789))]
])
```

This works, but there are definitely cons to this approach:
- Excessive wrapping/boxing of values which makes it harder to do anything with it. Generally we would not define a resultset statically like this (it would come from a DB). But any resultset manipulation will require extremely cumbersome pattern matching, and handling of cases that we know will never occur. 
- We don't actually use the type system but work around it. As a result, it doesn't actually enforce things like every column having the same type, or that every row has the same number of fields. 
- The standard JSON encoding of this data structure is kind of iffy.

#### Second attempt attempt using lists of tuples

Rather than using lists for the individual rows, we could use tuples. This will enforce that every row will have the same number of fields, and that every column will have the same type.

```
struct Resultset(records: List[a])
resultset = Resultset([
  # A minor variation would be to use a struct for records, rather than a tuple
  ("foo", 123, None), 
  ("bar", 456, Some(789))
])
```

Unfortunately, this doesn't really work for our use case, because:
- It's basically impossible to build generic functions that work on the Resultset type, because the type variable is completely undefined.
- The type gets "fixed" the first time we use it, which means all the other resultsets in the same script will have to use the exact same record type, which generally is not the case in practice.

### Questions

- Am I missing something? Is it possible to use richer types than `List[List[Option[Value]]]`?
- Could certain language feature additions make this better?

snoble:
It feels like what would help would be being able to put extra restrictions on `a` in `struct Resultset(records: List[a])`. Like if you could say that `a` is always of a tuple of Ints and Strings. Then you could have `addField(resultset: Resultset[a], f: a -> b): Resultset[Tuple[b, *a]]`

I think type classes would help?

[...]

PR #204, opened 2019-03-29
Issue with metavariables not fully inferred

johnynek:
I added a requirement (which I thought was there) that metas never escape. It's not cool to have that requirement, but it does show that there are a few places with this bug currently.

this was discovered in #196 

I guess we are missing a call to quantify/zonk...

However, if that were the case, those shoudn't cause anything to not typecheck. I think this is also related to recursive methods. I think all the current examples of failures are recursive methods.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
@@ -945,7 +945,12 @@ object Infer {
 
       // todo this should be a law...
     res.map { te =>
-      scala.Predef.require(Type.freeTyVars(te.getType :: Nil).isEmpty, te.getType.toString)
+      val tp = te.getType
+      lazy val teStr = TypeRef.fromTypes(None, tp :: Nil)(tp).toDoc.render(80)
+      scala.Predef.require(Type.freeTyVars(tp :: Nil).isEmpty, s"illegal inferred type: $teStr")
+
+      scala.Predef.require(Type.innerMetas(tp).isEmpty,
+        s"illegal inferred type: $teStr")
       te
     }
   }
--- a/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
@@ -430,7 +430,7 @@ object Infer {
       ty2 match {
         case Type.TyMeta(m2) =>
           readMeta(m2).flatMap {
-            case Some(ty2) => unify(Type.TyMeta(m), ty2, right, left)
+            case Some(ty2) => unify(Type.TyMeta(m), ty2, left, right)
             case None => writeMeta(m, ty2)
           }
         case nonMeta =>
--- a/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
@@ -3,6 +3,8 @@ import cats.Monad
 import cats.data.{NonEmptyList, Writer}
 import cats.implicits._
 
+import scala.collection.immutable.SortedSet
+
 import org.bykn.bosatsu.{
   Expr,
   HasRegion,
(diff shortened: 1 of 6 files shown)


johnynek:
cc @snoble

johnynek:
So, this turned out to be the same issue as #183, but at the top level. Unfortunately top level package names are handled on a separate path from internal let bindings. Maybe that can be fixed. The bug was in both places.

We need to unify top level recursive meta variables with themselves.

Interestingly, when we do this we see that the fix in #183 was incomplete: we were letting the metavariable stick around in the environment after we had solved for it.

The bug is fixed here. No old code is ruled out (unless the bazel tests fail), but it does seem that the function you wanted to write is not a legal type because it is an infinite type.

I think we still have a puzzler on our hands. I'll take a look tomorrow to see if I can think of a good way to model this.

johnynek:
note the code coverage isn't great because I improved the data store in type errors, but we don't exercise all the type errors in tests. We should do that and print the messages and make sure they are good.

core/src/main/scala/org/bykn/bosatsu/rankn/Type.scala
```diff
@@ -298,17 +297,25 @@ object Type {
...
-  def metaTvs(s: List[Type]): Set[Meta] = {
+  object Meta {
+    implicit val orderingMeta: Ordering[Meta] =
+      Ordering.by { m: Meta => m.id }
+  }
+
+  /**
+   * Final the set of all of Metas inside the list of given types
+   */
+  def metaTvs(s: List[Type]): SortedSet[Meta] = {
```
johnynek, line 308: the order of these variables is used when we create the forAll. Using a sorted set makes sure that is repeatable. We should really only use a Set when it is used a membership check, and never enumerated.

core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
```diff
@@ -605,9 +604,10 @@ object Infer {
                   for {
                     // the type variable needs to be unified with varT
                     // note, varT could be a sigma type, it is not a Tau or Rho
-                    typedRhs <- inferSigmaMeta(rhs, Some((rhsTpe, region(rhs))))
+                    typedRhs <- inferSigmaMeta(rhs, Some((name, rhsTpe, region(rhs))))
                     varT = typedRhs.getType
-                    typedBody <- typeCheckRho(body, expect)
+                    // we need to overwrite the metavariable now with the full type
```
johnynek, line 609: this is an important fix.

core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
```diff
@@ -862,20 +862,27 @@ object Infer {
...
       inferSigmaMeta(e, None)
 
-    def inferSigmaMeta[A: HasRegion](e: Expr[A], meta: Option[(Type.TyMeta, Region)]): Infer[TypedExpr[A]] =
+    def inferSigmaMeta[A: HasRegion](e: Expr[A], meta: Option[(Identifier, Type.TyMeta, Region)]): Infer[TypedExpr[A]] = {
+      def unifySelf(tpe: Type): Infer[Map[Name, Type]] =
+        meta match {
+          case None => getEnv
+          case Some((nm, m, r)) =>
+            (unify(tpe, m, region(e), r) *> getEnv).map { envTys =>
+              // we have to remove the recursive binding from the environment
```
johnynek, line 871: this is an important fix.

core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
```diff
@@ -423,21 +425,18 @@ object Infer {
...
-    def occursCheckErr(m: Type.Meta, t: Type): Infer[Unit] =
-      fail(Error.UnexpectedMeta(m, t))
-
     // invariant the flexible type variable tv1 is not bound
     def unifyUnboundVar(m: Type.Meta, ty2: Type.Tau, left: Region, right: Region): Infer[Unit] =
       ty2 match {
         case Type.TyMeta(m2) =>
           readMeta(m2).flatMap {
-            case Some(ty2) => unify(Type.TyMeta(m), ty2, left, right)
+            case Some(ty2) => unify(Type.TyMeta(m), ty2, right, left)
```
johnynek: This needs to be reverted.

PR #208, opened 2019-03-30
Add support for total patterns in lambdas

### LINKED ISSUE #188: support patterns in lambdas
We should support irrefutable patterns in lambdas:
```
f = \(x: Int) -> x.add(1)

f = \(x, y) -> x.add(y)

const = \_ -> 42

struct Order(fn: a -> a -> Comparison)
f = \Order(fn), lst -> lst.sort(fn)
```

This would have made the curry examples slightly cleaner: #185

johnynek:
close #188 

This will be pretty nice in a lot of cases, is easy to support.

What do you think @snoble ?

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Expr.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Expr.scala
@@ -182,18 +182,26 @@ object Expr {
     body: Expr[A],
     outer: A): Expr[A] = {
 
+    def makeBindBody(matchPat: Pattern[(PackageName, Constructor), rankn.Type]): (Bindable, Expr[A]) =
+      // We don't need to worry about shadowing here
+      // because we immediately match the pattern but still this is ugly
+      matchPat match {
+        case Pattern.Var(arg) =>
+          (arg, body)
+        case _ =>
+          val anonBind: Bindable = Identifier.Name("$anon") // TODO we should have better ways to gensym
+          val matchBody: Expr[A] =
+            Match(Var(None, anonBind, outer), NonEmptyList.of((matchPat, body)), outer)
+          (anonBind, matchBody)
+      }
+
     args match {
-      case NonEmptyList(Pattern.Var(arg), Nil) =>
-        Expr.Lambda(arg, body, outer)
-      case NonEmptyList(Pattern.Annotation(Pattern.Var(arg), tpe), Nil) =>
-        Expr.AnnotatedLambda(arg, tpe, body, outer)
+      case NonEmptyList(Pattern.Annotation(pat, tpe), Nil) =>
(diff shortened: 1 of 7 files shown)


core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -1086,4 +1086,33 @@ main = Foo(1, `package`, 3, 4)
...
+inc = \Foo(x) -> x.add(1)
+
+test0 = Assertion(inc(Foo(1)).eq_Int(2), "inc(Foo(1)) == 2")
+
+enum FooBar: F(x), B(x)
+
+inc2 = \F(x) | B(x), Foo(y) -> x.add(y)
+test1 = Assertion(inc2(F(1), Foo(1)).eq_Int(2), "inc2(F(1), Foo(1)) == 2")
+test2 = Assertion(inc2(B(1), Foo(1)).eq_Int(2), "inc2(B(1), Foo(1)) == 2")
+
```
snoble, line 1113: This is really cool!

Maybe add a couple tests for a multi-parameter struct and a nested pattern? Like `inc3 = \(F(x) | B(x), Foo(y)) -> x.add(y)` for example

core/src/main/scala/org/bykn/bosatsu/Expr.scala
```diff
@@ -177,5 +177,27 @@ object Expr {
         buildLambda(NonEmptyList.of(arg), body1, outer)
     }
 
+  def buildPatternLambda[A](
```
snoble, line 180: Looks like `buildLambda` is only used by `DefStatement` now. So not quite ready to delete it... but soon maybe?

johnynek: this is interesting. Should we allow patterns in defs?
```
def getBar(Bar(x)): x
```
doesn't look much like python, but if you want to go nuts, Haskell has something like this that allows multiple branches:
```
enum Either: Left(a), Right(b)

def map(Right(a), fn): Right(fn(a))
def map(left @ Left(_), _): left
```
I'm not crazy about that syntax. I think a match would be better.

snoble: My first thought when I see this is someone could get the impression that bosatsu has function overloading. And I think a match is always more clear

But for the non enum case this seems really natural and would be confusing if it were absent I think

johnynek: Consistency is a powerful argument. Okay. I think I agree with your take: yes to patterns in defs, no to multi defs for enums. So harmony with lambdas.

PR #216, opened 2019-04-03
Support patterns in def args

### LINKED ISSUE #211: add support for patterns in defs
related to #188 which is complete.

Allow things like:

```
struct Foo(x)

def useFoo(Foo(x: Int)):
  x.add(1)
```
this harmonizes with the lambda syntax which would be allowed:
```
useFoo = \Foo(x: Int) -> x.add(1)
```

cc @snoble

### LINKED ISSUE #212: support type annotation patterns at the top-level outside matches
In a match, you have to parenthesize a type annotation:
```
match x:
  (y: Int): y
```
because otherwise the parsing is ambiguous. In pattern contexts outside of a match, there is no ambiguity:

```
x: Int = 4
\x: Int -> x.add(2)
```
But currently, we still require `(x: Int)` in such cases. We should be able to parse matches differently. This is related to: #211 if we don't solve this, then we can't unify with defs, because they already allowed `x: Int` patterns.

I tried a quick stab at this, but the parsing didn't work, so I must have goofed some precedence thing up.

johnynek:
close #211 

which seems like we need to have to be consistent between lambdas and defs, and to emphasize the importance of patterns in bosatsu.

It seems like this *should* close #212 but for a reason I don't yet understand, I can't yet parse `\x: Int -> x` we need to have `\(x: Int) -> x` despite bare type annotations working for defs. I'll debug that in a separate PR.

cc @snoble

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
@@ -55,7 +55,7 @@ sealed abstract class Declaration {
       case Comment(c) =>
         CommentStatement.document[Padding[Declaration]].document(c)
       case DefFn(d) =>
-        val pairDoc: Document[(OptIndent[Declaration], Padding[Declaration])] =
+        implicit val pairDoc: Document[(OptIndent[Declaration], Padding[Declaration])] =
           Document.instance {
             case (fnBody, letBody) =>
                 fnBody.sepDoc +
@@ -63,7 +63,7 @@ sealed abstract class Declaration {
                 Doc.line +
                 Document[Padding[Declaration]].document(letBody)
           }
-        DefStatement.document(pairDoc).document(d)
+        DefStatement.document[Pattern.Parsed, (OptIndent[Declaration], Padding[Declaration])].document(d)
       case IfElse(ifCases, elseCase) =>
         // TODO, we could make this ternary if it is small enough
         def checkBody(cb: (Declaration, OptIndent[Declaration])): Doc = {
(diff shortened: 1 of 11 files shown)


core/src/main/scala/org/bykn/bosatsu/Declaration.scala
```diff
@@ -545,7 +534,7 @@ object Declaration {
   }
 
   val lambdaP: Indy[Lambda] = {
-    val params = Indy.lift(P("\\" ~/ maybeSpace ~ Pattern.parser.nonEmptyList))
+    val params = Indy.lift(P("\\" ~/ maybeSpace ~ Pattern.bindParser.nonEmptyList))
```
johnynek, line 537: this should allow `\x: Int -> x` to parse, but for some reason it doesn't seem to in my example in EvaluationTest.

core/src/main/scala/org/bykn/bosatsu/Pattern.scala
```diff
@@ -324,72 +324,80 @@ object Pattern {
         Doc.intercalate(Doc.text(" | "), (head :: rest.toList).map(doc(_)))
     }
 
-  lazy val parser: P[Pattern[Option[Constructor], TypeRef]] = {
+  val pwild = P("_").map(_ => WildCard)
```
johnynek: should be private.

core/src/main/scala/org/bykn/bosatsu/DefStatement.scala
```diff
@@ -1,66 +1,64 @@
...
-  def map[A](fn: T => A): DefStatement[A] =
-    copy(result = fn(result))
   /**
    * This ignores the name completely and just returns the lambda expression here
    *
    * This could be a traversal: TypeRef => F[rankn.Type] for an Applicative F[_]
    */
-  def toLambdaExpr[A](resultExpr: Expr[A], tag: A)(trFn: TypeRef => rankn.Type): Expr[A] = {
-    val unTypedBody = resultExpr
+  def toLambdaExpr[C](resultExpr: B => Expr[C],
```
johnynek, line 22: Maybe move this to Declaration and remove all the generic types here.

core/src/main/scala/org/bykn/bosatsu/Declaration.scala
johnynek: I wonder if somehow this `~/` cut join is failing somehow. I've been confused by this before. My mental model suggests this is fine, but it should also work if we remove it, so let's see if that fixes.

johnynek:
The issue with `\x: Int -> x` is that `Int -> x` parses as a type. So, we need to write `(x: Int)` to disambiguate.

In scala, `{ x: Int => Int => x(1) }` doesn't parse, so you need to write `{x: (Int => Int) => x(1) }` Maybe we could do something similar to require a `(Int -> Int)` to be treated as a type at the top level in lambdas.

Seems like something not to sweat too much for now except that the default seems backwards: we are assuming a `Int -> x` type vs assuming `Int` type returning `x`.

PR #222, opened 2019-04-13
Fix exception in printing UnknownImportPackage

### LINKED ISSUE #219: missing transitive dependencies in json generation gives bad error
if you use the bosatsu_json target, but don't include a needed transitive dependency, you get a bad error message:

```
Exception in thread "main" java.util.NoSuchElementException: key not found: PackageName(NonEmptyList(DictTools))
        at scala.collection.MapLike$class.default(MapLike.scala:228)
        at scala.collection.AbstractMap.default(Map.scala:59)
        at scala.collection.MapLike$class.apply(MapLike.scala:141)
        at scala.collection.AbstractMap.apply(Map.scala:59)
        at org.bykn.bosatsu.PackageError$UnknownImportPackage.message(PackageMap.scala:271)
        at org.bykn.bosatsu.MainCommand$$anonfun$fromPackageError$1.apply(Main.scala:222)
        at org.bykn.bosatsu.MainCommand$$anonfun$fromPackageError$1.apply(Main.scala:222)
        at cats.data.NonEmptyList.map(NonEmptyList.scala:76)
        at org.bykn.bosatsu.MainCommand$.fromPackageError(Main.scala:222)
        at org.bykn.bosatsu.MainCommand$$anonfun$typeCheck$1.apply(Main.scala:96)
        at org.bykn.bosatsu.MainCommand$$anonfun$typeCheck$1.apply(Main.scala:91)
        at org.bykn.bosatsu.MainResult$Success.flatMap(Main.scala:60)
        at org.bykn.bosatsu.MainCommand$.typeCheck(Main.scala:91)
        at org.bykn.bosatsu.MainCommand$TypeCheck.run(Main.scala:134)
        at org.bykn.bosatsu.Main$.main(Main.scala:283)
        at org.bykn.bosatsu.Main.main(Main.scala)
```

We should either correctly report the name of the missing transitive dependency, or include all transitives.

johnynek:
close #219 

I reproduced the error, which was reversing the parsed package and the missing package. I also generalized and moved some code from Main into PackageMap so it was cleaner to reuse.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Main.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Main.scala
@@ -84,7 +84,7 @@ sealed abstract class MainCommand {
 object MainCommand {
 
   def typeCheck(inputs: NonEmptyList[Path], externals: List[Path]): MainResult[(Externals, PackageMap.Inferred, List[(Path, PackageName)])] = {
-    val ins = parseInputs(inputs)
+    val ins = PackageMap.parseInputs(inputs)
     val exts = readExternals(externals)
 
     toResult(ins.product(exts))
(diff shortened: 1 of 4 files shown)


johnynek:
This increases test coverage 1.13% by testing message creation that we weren’t exercising before.

The diff itself isn’t tested to 85% but that’s partially because Main is tested by the bazel tests which aren’t included in coverage stats here.

Issue #225, opened 2019-04-14
Use the variance of custom types in Inference


johnynek:
Currently, Inference has a couple of places where we special case for function types, which are really encoding rules about variance. We should remove the special casing and just let the fact that functions are contravariant in the first parameter and covariant in the second do the work.

relates to #143 #160

johnynek:
I'm closing this for now. The code in #226 isn't perfect, since it can be overly conservative when dealing with unknown higher kinded types, but that does not seem easily solved so I will wait until we have bug reports and real cases we expect to typecheck and reopen a new issue then.

Issue #209, opened 2019-03-31
Type inference fails for `forall` types


snoble:
This might be asking too much of type inference... but figured I'd make an issue anyhow

```
package Example/Wrap

def wrap(x):
  Some(x)

def doToSomething(fn: forall a. a -> b[a]):
  (fn(1), fn("two"))

out = doToSomething(wrap)
```

is a valid program. But if we do
```
package Example/Wrap

def wrap(x):
  Some(x)

def doToSomething(fn):
  (fn(1), fn("two"))

out = doToSomething(wrap)
```

we get

```
in file: ../bosatsu_examples/wraptest.bosatsu, package Example/Wrap, type Bosatsu/Predef::Int does not unify with type Bosatsu/Predef::String
5|
6|def doToSomething(fn):
7|  (fn(1), fn("two"))
               ^^^^^
```

The work around? In these cases, just write the type

johnynek:
This is a complex type to infer. Thanks for the examples. The paper discusses some similar problems. For instance the algorithm will not infer, IIRC a rank-N type even though it is a legal type.

That said, the paper didn’t contemplate user defined types. And this is an interaction there I think. So it may be possible to change the algorirthm to improve cases like this without regressing in other areas.

johnynek:
I wonder if this is related to this comment:

https://github.com/johnynek/bosatsu/issues/106#issuecomment-470354751

We don't use variance information in custom type inference, but we do for functions. Basically, the contravariance of the input and covariance of the output is known.

I think this code: https://github.com/johnynek/bosatsu/blob/b81e52372d7940377ca3e5ea54cc96103d9ed961/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala#L384

probably needs to know about `Apply(t1, t2)` types and it needs to know the variance of `t1` in that case. We have already solved variance, and I think we are passing it in, but it is currently being ignored in the inference for substitution.

The second thing, for a covariant type, `t[forall a. a]` is the same as `forall a. t[a]`. We do this normalization in functions here:
https://github.com/johnynek/bosatsu/blob/b81e52372d7940377ca3e5ea54cc96103d9ed961/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala#L289

I think we should be doing that for all Apply types for their covariant parts.

It would be good to come up with some minimal test cases we *think* should pass and get those to pass.

This #144 had some cases I commented out:
[...]

PR #226, opened 2019-04-14
Use variance in skolemization and substitution

johnynek:
part of #225 

This seems to solve the open problems. This approach exposes an issue in this approach: if we have unknown types in an apply, we should really introduce variables to track what final variance is substituted in. This does not do that, and instead falls back on a worse case of invariance.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
@@ -550,7 +550,7 @@ object Infer {
         case (Type.TyApply(a1, b1), Type.TyApply(a2, b2)) =>
           unify(a1, a2, r1, r2) *> unify(b1, b2, r1, r2)
         case (Type.TyConst(c1), Type.TyConst(c2)) if c1 == c2 => unit
-        // these shouldn't be reachable since we have Tau types, but they are reachable
+        // these shouldn't be reachable since we have Tau types, but may be reachable
         // case (Type.ForAll(_, _), _) =>
         //   sys.error(s"expected tau type: $t1, $t2")
         // case (_, Type.ForAll(_, _)) =>
--- a/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
@@ -440,6 +440,42 @@ object Infer {
             case (a2, r2) =>
               subsCheckFn(a1, r1, a2, r2, left, right)
           }
+        case (rho1, Type.TyApply(l2, r2)) =>
+          unifyTyApp(rho1, left, right).flatMap {
+            case (l1, r1) =>
+              // TODO since either side may have metas, we should check both
+              val check2 = varianceOf(l2).flatMap {
+                case Variance.Covariant =>
+                  subsCheck(r1, r2, left, right).void
+                case Variance.Contravariant =>
+                  subsCheck(r2, r1, right, left).void
+                case Variance.Phantom =>
+                  // this doesn't matter
+                  pure(())
+                case Variance.Invariant =>
+                  unify(r1, r2, left, right).void
+              }
+              // should we coerce to t2? Seems like... but copying previous code
+              (subsCheck(l1, l2, left, right) *> check2).as(TypedExpr.coerceRho(t))
+          }
+        case (Type.TyApply(l1, r1), rho2) =>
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
```diff
@@ -268,6 +291,30 @@ object Infer {
           (bound *> zonkTypedExpr(rho)).map(TypedExpr.forAll(aligned.map(_._2), _))
       }
 
+    def varianceOf(t: Type): Infer[Variance] = {
+      import Type._
+      def revVariances(vs: Map[Type.Const.Defined, List[Variance]], t: Type): List[Variance] =
```
johnynek: they are not reversed, I think the name should change.

core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
```diff
@@ -448,7 +550,13 @@ object Infer {
         case (Type.TyApply(a1, b1), Type.TyApply(a2, b2)) =>
           unify(a1, a2, r1, r2) *> unify(b1, b2, r1, r2)
         case (Type.TyConst(c1), Type.TyConst(c2)) if c1 == c2 => unit
-        case (left, right) => fail(Error.NotUnifiable(left, right, r1, r2))
+        // these shouldn't be reachable since we have Tau types, but may be reachable
```
johnynek, line 558: this no longer fails with this PR, so the issue here is likely due to TyApply, and properly handling it fixes things.

johnynek:
@snoble if you have any minimal examples around the issue #209 that would be great.

This change, while not complete as I have commented, was enough to pass some tests I previously commented out. It would be nice to find some failing examples that should pass.

snoble:
They don't seem to quite work yet (but the first might be because I'm doing something wrong or there's another bug)
```scala
  test("function with quantified type") {
    evalTest(
      List("""
package Wrap

def wrap(x):
  Some(x)

out = (wrap(1), wrap("two"))
"""), "Wrap", ProductValue.fromList(List(VOption.some(VInt(1)), VOption.some(Str("two"))))
    )
    evalTest(
      List("""
package Wrap

def wrap(x):
  Some(x)

def doSomething(fn):
  (doSomething(1), doSomething("two"))

out = doSomething(wrap)
"""), "Wrap", ProductValue.fromList(List(VOption.some(VInt(1)), VOption.some(Str("two"))))
    )
  }
```

snoble:
The first one will run with `master` but has a strange reason for not passing 
```
[info]   java.lang.AssertionError: assertion failed: failed: for type: (Bosatsu/Predef::Option[Bosatsu/Predef::Int], Bosatsu/Predef::Option[Bosatsu/Predef::String]), ConsValue(SumValue(1,ConsValue(ExternalValue(1),UnitValue)),ConsValue(ConsValue(SumValue(1,ConsValue(ExternalValue(two),UnitValue)),ConsValue(UnitValue,UnitValue)),UnitValue)) != ConsValue(SumValue(1,ConsValue(ExternalValue(1),UnitValue)),ConsValue(SumValue(1,ConsValue(ExternalValue(two),UnitValue)),UnitValue))
```

I don't know why there's a `ConsValue(UnitValue,UnitValue)` in there. But it's surely unrelated to this.

The second one fails on this branch with

```
[info]   Bosatsu/Predef::Int (Region(73,78)) cannot be unified with Bosatsu/Predef::String (Region(73,78))
```

johnynek:
@snoble I actually don't think this algorithm claims to solve the second case looking more at it. The paper is explicit that some complex types it will infer more simply in a way that might not unify, and that the programmer needs to annotate for some more complex types.

In this example, it seems to me there is a tension between allowing legal code vs ruling out likely errors. Your example seems pretty likely to me to actually be an error, and inferring a crazy type might really confuse a lot of programmers.

It seems more and more clear to me we need type systems *AND* ml-based tools for code development. We need the ML to help guide the rules based approach in order to learn likely bugs vs legal code.

I've been wanting to try to use sequence learning + a compiler to do a GAN-type approach: the ML tries to generate a sequence that can pass the compiler, e.g. https://tabnine.com/ combine that with learning sequences that will typecheck.

Anyway.... this PR seems like an improvement since it passes several commented out tests, so that seems good.

snoble:
Makes sense. For what it's worth this is also what elm does when trying to infer the type for the argument of doSomething

PR #227, opened 2019-04-16
Add functionality to hash normal expression tags

snoble:
I had a little free time today so I thought I'd just type out the simplest way to do this. Needs tests still


core/src/main/scala/org/bykn/bosatsu/TypedExpr.scala
```diff
@@ -101,6 +101,37 @@ sealed abstract class TypedExpr[T] {
         (expr.traverseType(fn), tbranch).mapN(Match(_, _, tag))
     }
 
+  def traverseTag[F[_]: Applicative, S](fn: T => F[S]): F[TypedExpr[S]] =
```
johnynek: this is the normal `Traverse[TypedExpr]` instance, which I used to have, but I removed since I wasn't using it.

Can we instead make this:
```scala
implicit val traverseTypedExpr: Traverse[TypedExpr] =
  new Traverse....
```
we can actually use cats-laws to test that, but it is a bit of a pain since we have to generate random TypedExpr. Maybe easier to test it by calling it in your code.

snoble: Totally!

snoble:
Still needs tests

johnynek:
Thanks, this looks good.

One way to add tests might be to use `checkLast`: https://github.com/johnynek/bosatsu/blob/master/core/src/test/scala/org/bykn/bosatsu/TestUtils.scala#L49

to traverse several expressions and do something like, count them (using `Writer[Int, ?]` monad, or even just `Id[?]` monad and assert the it is identity.

snoble:
@johnynek: I added `serialize` for NormalExpressions with a hashing function parameter. I'm not sure that's actually a good idea, as opposed to just hashing the once at the end.

I also added some tests

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -63,7 +84,29 @@ object NormalExpression {
   }
 }
 
-sealed abstract class NormalPattern {}
+sealed abstract class NormalPattern {
+  def escapeString(unescaped: String) = unescaped.replaceAll("""([\\'])""", """\\$1""")
```
johnynek: there is a well tested string escaping code here:

https://github.com/johnynek/bosatsu/blob/master/core/src/main/scala/org/bykn/bosatsu/StringUtil.scala#L41

Is this doing something else special? Actually, can we comment what it is doing?

core/src/test/scala/org/bykn/bosatsu/TestUtils.scala
```diff
@@ -155,18 +156,33 @@ object TestUtils {
...
       (for {
         pack <- normPackMap.toMap.get(mainPack)
-        (name, rec, expr) <- pack.program.lets.lastOption
+        exprs <- pack.program.lets.traverse {case (_, rec, expr) =>
+          expr.traverse[Option, (Declaration, Normalization.NormalExpressionTag, String)]{
+            case (d, nt) => Some((d, nt, nt.ne.serialize(x => x)))
+          }
+        }
+        expr <- exprs.lastOption
+        fleft = exprs.foldLeft(0) {case (n, expr) => expr.foldLeft(n) { case (m, _) => m +1 }}
```
johnynek: I think `Foldable[List].compose[TypedExpr].size(exprs)` is the size as this.

core/src/test/scala/org/bykn/bosatsu/TestUtils.scala
```diff
@@ -155,18 +156,33 @@ object TestUtils {
       val normPackMap = NormalizePackageMap(infPackMap).normalizePackageMap
       (for {
         pack <- normPackMap.toMap.get(mainPack)
-        (name, rec, expr) <- pack.program.lets.lastOption
+        exprs <- pack.program.lets.traverse {case (_, rec, expr) =>
+          expr.traverse[Option, (Declaration, Normalization.NormalExpressionTag, String)]{
+            case (d, nt) => Some((d, nt, nt.ne.serialize(x => x)))
+          }
+        }
+        expr <- exprs.lastOption
```
johnynek, line 162: this is a bit confusing because `expr` is shadowed below. Can we move this below fleft and fright?

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -398,6 +443,16 @@ case class NormalizePackageMap(pm: PackageMap.Inferred) {
     PackageMap(normAll.run(Map()).value._2.toMap)
   }
 
+  def hashKey[T](fn: NormalExpression => T): PackageMap.Typed[(Declaration, ExpressionKeyTag[T])] = {
+    val lst = normalizePackageMap.toMap.toList
+      .map { case (name, pack) => (name, pack.copy(program = pack.program.copy(
+        lets = pack.program.lets.map { case (name, recursive, expr) => (name, recursive, expr.traverse[Id, (Declaration, ExpressionKeyTag[T])] {
```
johnynek: this is a bit dense. Can we expand it into a few lines with a few `val =`? I'd rather use a bit more vertical than horizontal space here.

core/src/main/scala/org/bykn/bosatsu/TypedExpr.scala
```diff
@@ -139,6 +138,93 @@ object TypedExpr {
     }
   }
 
+  implicit val traverseTypedExpr: Traverse[TypedExpr] = new Traverse[TypedExpr] {
```
johnynek, line 141: sorry, I think I caused a merge conflict here... I added `Functor[TypedExpr]` which I needed, but traverse is a super set of that, so if you merge master, you can do so by either removing my `Functor` or copying the `def map` method into your Traverse with a `override def map` (it is *slightly* more efficient to manually implement map than do the default implementation which uses `traverse` with `Id[A]`

snoble: oh nice, easy enough

core/src/test/scala/org/bykn/bosatsu/TestUtils.scala
snoble: sure. I can also rename `expr` above to just `e`

core/src/main/scala/org/bykn/bosatsu/Normalization.scala
snoble: I'm definitely happier to use something well tested than a regex hack. This is just attempting to escape `'` to `\'` and also escape `\` to `\\`

core/src/test/scala/org/bykn/bosatsu/TestUtils.scala
snoble: oh nice, I'll use that

snoble: I couldn't figure out how to use `Foldable`'s `compose` here instead of `List`'s. So I just `map`d and `sum`d

snoble: I realized I don't want to sum/fold/compose the lists. I just want to exercise the TypedExpr folding code

snoble:
Coverage hit!

snoble:
any more thoughts?

PR #229, opened 2019-04-21
Reduce use of reflection, add sbt-assembly to make assembly images

johnynek:
I used this to make a native-image with graal:

```
sbt core/assembly
native-image --static -jar ~/oss/bosatsu/core/target/scala-2.12/bosatsu-core-assembly-0.1.0-SNAPSHOT.jar bosatsu
```
that does it...
you need to add the bin directory of:

https://github.com/oracle/graal/releases

to your path, but that takes a few minutes.

With the changes to not use reflection in predef, we can run bosatsu (which doesn't use reflective FFI calls), in a native-image.

cc @snoble


snoble:
Nice! Do you think it's worth adding those commands to the README or would anyone who needs this already know how?

johnynek:
okay, with this:
```
time ./bosatsu write-json --input test_workspace/bazel_deps_api.bosatsu --input test_workspace/gen
_deps.bosatsu --input test_workspace/dicttools.bosatsu --main GenDeps --output deps.json
```
We get:
```
real    0m1.129s
user    0m1.019s
sys     0m0.111s
```

what do you think @snoble? 1sec for compilation and json generation seems pretty good to me. :)

johnynek:
compare to using the jvm:
```
oscarboykin@penguin:~/oss/bosatsu$ time java -jar core/target/scala-2.12/bosatsu-core-assembly-0.1.0-SNAPSHOT.jar write-json --input test_workspace/bazel_deps_api.bosatsu --input test_workspace/gen_deps.bosatsu --input test_workspace/dicttools.bosatsu --main GenDeps --output deps.json

real    0m3.537s
user    0m6.400s
sys     0m0.491s
```

johnynek:
the generated static image is 15MB, but that doesn't seem unreasonable to me.

johnynek:
This fails coverage because reflection is untested. I’ll add a test reflective call so it doesn’t break.

johnynek:
addresses #230

PR #231, opened 2019-04-22
Add two data structure examples: Queue, TreeList

johnynek:
This adds two implementations of potentially useful libraries as bosatsu examples, and immutable Queue and TreeList both originally described by Chris Okasaki.

This is the first "library code" I've ever actually written in bosatsu, and I really liked it. I think there are still patterns to be learned in the language, but I'm pretty happy living without typeclasses actually and just manually passing functions around. In this example I wrote `eq_Queue` and `eq_TreeList` which each themselves take functions to check equality on items, and returns a function to check equality on the data structure. This is exactly writing the `Eq` instances by hand, but it really doesn't seem very painful in this case to do so.

I am exploring a kind of strange style lately: actively using shadowing in order to prevent the possibility of accidentally using the wrong variable later. Some think shadowing is bad, but one upside of shadowing means the original value isn't able to be used, which can be a similar benefit to linear types: you can't accidentally reuse the wrong one later. Maybe it will be a stylistic dead-end, but I played with it a bit here.

cc @snoble if you want to do any style review.

### DIFF
--- a/test_workspace/BUILD
+++ b/test_workspace/BUILD
@@ -74,3 +74,25 @@ bosatsu_json(
     srcs = ["gen_deps.bosatsu"],
     deps = [":bazel_deps_api", ":dict_tools"],
     package = "GenDeps")
+
+bosatsu_library(
+    name = "queue",
+    srcs = ["Queue.bosatsu"],
+    )
+
+# todo we need to be able to test a library without recompiling it
+bosatsu_test(
+    name = "queue_test",
+    srcs = ["Queue.bosatsu"],
+    )
+
+bosatsu_library(
+    name = "treelist",
+    srcs = ["TreeList.bosatsu"],
+    )
+
+# todo we need to be able to test a library without recompiling it
+bosatsu_test(
+    name = "treelist_test",
+    srcs = ["TreeList.bosatsu"],
+    )
--- a/test_workspace/Queue.bosatsu
+++ b/test_workspace/Queue.bosatsu
@@ -1,8 +1,8 @@
 package Queue
 
 export [ Queue,
-  empty_Queue, fold_Queue, push, pop, pop_value, pop_Queue, reverse_Queue, eq_Queue,
-  queue_to_List,
+  empty_Queue, fold_Queue, push, unpush, pop_value, pop, reverse_Queue, eq_Queue,
+  to_List, from_List
 ]
 
 struct Queue(front: List[a], back: List[a])
@@ -12,14 +12,13 @@ empty_Queue: forall a. Queue[a] = Queue([], [])
 # convenient local alias
 empty = empty_Queue
 
-def queue_from_List(list: List[a]) -> Queue[a]:
+def from_List(list: List[a]) -> Queue[a]:
   Queue(list, [])
 
-def push(queue: Queue[a], item: a) -> Queue[a]:
-  Queue(f, b) = queue
+def push(Queue(f, b): Queue[a], item: a) -> Queue[a]:
   Queue(f, [item, *b])
 
-def pop(queue: Queue[a]) -> Option[(a, Queue[a])]:
+def unpush(queue: Queue[a]) -> Option[(a, Queue[a])]:
   match queue:
     Queue([h, *t], b): Some((h, Queue(t, b)))
     Queue([], b):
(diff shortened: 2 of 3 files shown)


test_workspace/BUILD
```diff
@@ -74,3 +74,25 @@ bosatsu_json(
...
+
+bosatsu_library(
+    name = "treelist",
+    srcs = ["TreeList.bosatsu"],
+    )
+
+# todo we need to be able to test a library without recompiling it
+bosatsu_test(
+    name = "treelist_test",
+    srcs = ["TreeList.bosatsu"],
```
snoble, line 97: nit: trailing commas a inconsistent with above

test_workspace/Queue.bosatsu
```diff
@@ -0,0 +1,107 @@
...
+########
+## Tests below
+########
+
+def eq_Opt(fn, a, b):
+  match (a, b):
+    (None, None): True
+    (None, Some(_)): False
+    (Some(_), None): False
+    (Some(a), Some(b)): fn(a, b)
```
snoble, line 68: I find this a little verbose... I suspect it would be a little more clear if you did
```
def eq_Opt(fn, a, b):
  match (a, b):
    (None, None): True
    (Some(a), Some(b)): fn(a, b)
    _: False
```

test_workspace/Queue.bosatsu
```diff
@@ -0,0 +1,107 @@
...
+    Some((_, queue)): queue
+    None: empty
+
+def fold_Queue(queue: Queue[a], init: b, fn: b -> a -> b) -> b:
+  Queue(f, b) = queue
+  front = f.foldLeft(init, fn)
+  b.reverse.foldLeft(front, fn)
+
+def reverse_Queue(queue: Queue[a]) -> Queue[a]:
+  Queue(f, b) = queue
```
snoble: I kind of like the idiom of having this in the param declaration. Maybe it's too dense? It just feels like less note keeping code to me

test_workspace/Queue.bosatsu
snoble: (oh, this is just for testing)

test_workspace/Queue.bosatsu
```diff
@@ -0,0 +1,107 @@
...
+        []: None
+        [h, *t]: Some((h, Queue(t, [])))
+
+def pop_value(queue: Queue[a]) -> Option[a]:
+  match pop(queue):
+    Some((a, _)): Some(a)
+    None: None
+
+# drop an item off and return the rest, or empty
+def pop_Queue(queue: Queue[a]) -> Queue[a]:
```
snoble: it kind of feels like your convention is inconsistent here. In other places the `_Type` suffix means "this is the version of this function for `Type`" but here you're saying "this returns the `Queue` side of the tuple". Maybe the convention isn't that big of a deal

test_workspace/TreeList.bosatsu
```diff
@@ -0,0 +1,152 @@
...
+
+def head(tl: TreeList[a]) -> Option[a]:
+  match decons(tl):
+    Some((h, _)): Some(h)
+    None: None
+
+operator + = add
+operator - = sub
+
+#def int_loop(intValue: Int, state: a, fn: Int -> a -> (Int, a)) -> a
```
snoble: what does `#def` mean?

test_workspace/TreeList.bosatsu
```diff
@@ -0,0 +1,152 @@
...
+          else: (idx - 1, (rest, None))
+        [Branch(s, _, t1, t2), *rest]:
+          # we can either skip just one, or skip the entire Tree
+          match cmp_Int(idx, s):
+            LT | EQ: (idx - 1, ([t1, t2, *rest], None))
+            _: (idx - s, (rest, None))
+  )
+  item
+
+def from_List(list: List[a]) -> TreeList[a]:
```
snoble, line 82: the `Queue` equivalent you called `queue_from_List`

snoble:
I think there are some opportunities to start to come up with some naming convention rules but this really readable

test_workspace/TreeList.bosatsu
snoble: Oh... this is just a comment so the signature is handy

johnynek: Yep. I should remove.

johnynek:
These are really great comments. I appreciate them.

I've tried to address them. Let me know what you think. I agree having a naming convention (and style guide) would be good.

I am now leaning to using the shortest clear names local to a package, and then use the aliasing imports as needed to disambiguate, rather than preemptively try to give longer names we hope won't collide. What do you think of that guidance?

test_workspace/BUILD
johnynek: bazel generally suggests trailing commas, so I will leave these, but I agree it would be nice to run buildifier (the bazel formatter).

snoble:
Nice. Yeah, the advice of do the convenient thing for yourself locally seems like it has all the advantages.

Issue #241, opened 2019-05-11
Strictness mode as declared in the file


johnynek:
One thing I’ve been thinking is how annoying it is in many languages to run with all warnings as failure while developing.

Since this setting is usually a compiler flag, it becomes a pain to selectively enable it.

We could instead be like Perl and have a `use strict` declaration at the top of a file if we want to be as strict as possible (no unused let bindings, for example).

This suggests a principle for bosatsu: semantics of what compiles is never controlled by compiler flags, only what is written in the source code.

What do you think @snoble ?

snoble:
This would be so good and feels like the right philosophy. You can tell if a file will work or not based entirely with what's in the file, with out having to worry about the environment (as much as possible). This seems consistent with how imports have to be explicit

PR #242, opened 2019-05-13
Allow struct and enum to have explicit type arguments

### LINKED ISSUE #83: allow explicitly listing type params when naming types in enums/structs
the bug in #82 was that we didn't name the type parameters, but we discovered them in the opposite order you would expect.

In that example, no type parameters were listed, but if we had done `struct Pair(fst: a, snd: b)` the result would have been the same.

We should allow `struct Pair[a, b](fst: a, snd: b)` and do a name check that all the types must be fully specified in such a case.

johnynek:
close #83 

This is pretty easy it turns out, but since we have no errors allowed going from syntax into types, it means we have coerce anything into something meaningful. Two error conditions that are ignored currently:

1. duplicating type parameters shouldn't be allowed (similarly with value parameters)
2. I think we want type parameters to be a superset of the free type parameters in the args.

I think the right way to deal with these two cases are lints that run on the source code that give warnings or errors (depending on the package's source defined strictness, see #241 ), but we want to give the user as much feedback as possible, so it is good to be able to run the typechecker on the sanest interpretation of the input.

What do you think @snoble ?

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
@@ -49,13 +49,13 @@ object DefRecursionCheck {
         checkStatement(cs.on.padded)
       case Def(defn) =>
         checkDef(TopLevel, defn) *> checkStatement(defn.result._2.padded)
-      case Struct(_, _, rest) =>
+      case Struct(_, _, _, rest) =>
         checkStatement(rest.padded)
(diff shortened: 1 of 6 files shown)


snoble:
Whoa Monad structs! This looks so good.

A checker of lint rules seems like it would just be generally useful. Maybe the rules you described could be the equivalent of predef imports (ie non-optional) but later you could extend the system to run user defined checks?

snoble:
This definitely motivates me to try a type safe resultset again once this is in

PR #251, opened 2019-05-15
Snoble/resultset

snoble:
This might be fun to have as an example. But mostly creating the PR for visibility

test currently passes

### DIFF
--- a/test_workspace/recordset.bosatsu
+++ b/test_workspace/recordset.bosatsu
@@ -1,0 +1,42 @@
+package Example/Resultset2
+
+def reorder(row: t[a, b], f: t[a, b] -> s[a, b]) -> s[a, b]:
+  f(row)
+
+struct A[a,b](x1: a, x2: b, x3: a)
+struct B[a,b](x1: a, x2: a, x3: b)
+
+x = A("a", True, "c")
+y = A(1, 2, 3)
+
+def swap_last_two(A(first, second, third)):
+  B(first, third, second)
+
+x2 = x.reorder(swap_last_two)
+y2 = y.reorder(swap_last_two)
+
+def and(x, y):
+  match x:
+    True: y
+    False: False
+
+operator && = and
+
+def equals(compare, x, y):
+  match compare(x,y):
+    EQ: True
+    _: False
+
+def cmp_Bool(x, y):
+  match (x, y):
+    (True, False): GT
+    (False, True): LT
+    _: EQ
+
+def compare_triples((compare1, compare2), B(x1, x2, x3), B(y1, y2, y3)):
+  compare1.equals(x1, y1) && compare1.equals(x2, y2) && compare2.equals(x3, y3)
+
+tests = Test("reordering",
+  [Assertion((string_Order_fn, cmp_Bool).compare_triples(x2, B("a", "c", True)), "letter swap"),
+  Assertion((cmp_Int, cmp_Int).compare_triples(y2, B(1, 3, 2)), "number swap")]
+)
--- a/test_workspace/recordset.bosatsu
+++ b/test_workspace/recordset.bosatsu
@@ -10,18 +10,25 @@ struct RecordSet[shape](fields: shape[Field[String], Field[Int], Field[Bool]], r
 def restructure(RecordSet(fields, rows), f: forall string,int,bool. shape1[string, int, bool] -> shape2[string, int, bool]) -> RecordSet[shape2]:
   RecordSet(f(fields), rows.map_List(f))
 
-struct Source[string,int,bool](s1: string, s2: int, s3: bool)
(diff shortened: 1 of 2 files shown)


johnynek:
This is great! I’d love to merge it as an example and test of the compiler.

johnynek:
Please remove the draft tag when you are ready!

snoble:
Oh sure. I'll clean out the weeds and remove the draft tag. There's a bunch of other approaches I'm going to try still but that's kind of orthogonal

test_workspace/recordset.bosatsu
```diff
@@ -0,0 +1,61 @@
...
+  [Source("a", 1, True)]
+)
+
+def swap_last_two(Source(first, second, third)):
+  Target(first, third, second)
+
+rs2 = rs.restructure(swap_last_two)
+
+def and(x, y):
+  match x:
```
johnynek: small note: `def and(x, y): y if x else False` is a one liner for this, but ultimately the same code.

snoble: nice!

snoble:
I'll just remove the tag and continue to fiddle. Happy to take comments or to merge whenever (I'll just make another PR if it gets more interesting)

snoble:
One nice to have is it would be nice to be able to write
```
result_rs = rs
  .restructure(\Shape1(a,b,c) -> Shape2(a,c,b))
  .add_int_field("Second Int", \Shape2(_,_,x) -> x.add(4), \Shape2(a,b,c), d -> Shape3(a,b,c,d))
```

snoble:
Another puzzle I can't seem to solve is it would be nice to not have to define all these struct types. Any thoughts on what else would be possible?

snoble:
Something like this would be nice
```
struct PairIntShape[string,int,bool](int: int, rest: shape[string, int, bool])
def makePairIntShape(rest: shape[string,int,bool], i: int) -> PairIntShape[string,int,bool]:
  PairIntShape(i, rest)

def add_int_field(RecordSet(fields, rows): RecordSet[shape1], name: String, f: shape1[String, Int, Bool] -> Int) -> RecordSet[PairIntShape]:
  RecordSet(makePairIntShape(fields, Field(name, RVInt)), rows.map_List(\row -> makePairIntShape(row, f(row))))


struct Shape1[string,int,bool](f1: string, f2: int, f3: bool)
struct Shape2[string,int,bool](f1: string, f2: bool, f3: int)
struct Shape3[string,int,bool](f1: string, f2: bool, f3: int, f4: int)

rs = RecordSet(
  Shape1(Field("String", RVString), Field("Int", RVInt), Field("Bool", RVBool)),
  [Shape1("a", 1, True)]
)

rs0 = rs.restructure(\Shape1(a,b,c) -> Shape2(a,c,b))
rs1 = rs0.add_int_field("Second Int", \Shape2(_,_,x) -> x.add(4))
rs2 = rs1.restructure(\PairIntShape(i, Shape2(a,b,c)) -> Shape3(a,b,c,i))
```

The problem is that my shapes have to take exactly 3 type variables, but `PairIntShape` has this `shape` variable in there. It feels like if I could have a syntax like
```
struct PairIntShape[string,int,bool][shape](int: int, rest: shape[string, int, bool])
```
where the 2nd block would be inferred if not provided.... starts to get pretty complicated though

johnynek:
Interesting suggestions. Would you be willing to edit them to include some sketch as to what goes wrong when you try what you want now?

I *thought* currying of type parameters should just work since internally like there is only 1 function type, there is only 1 type application type, so `f[a][b]` at the use site should be the same as `f[a, b]`, and I think you should be able to leave them off. You can (currently) omit a type parameter at definition site and all the implicit ones come after the explicit ones. I'm not sure if we want that long team, but maybe it is useful as you mention here. It would be interesting to try and see if that works.

snoble:
I don't know if I understand it enough to write out a sketch but let me write out what I've observed.

My use case is I have these `shape` types that 3 type parameters. The idea is that they define how a row is shaped and lets me have a row that is populated with `Fields`.

I actually just got it to work while trying to get all the errors. But I still don't fully understand what I did.
```
struct PairIntShape[shape,string,int,bool](int: int, rest: shape[string, int, bool])
def makePairIntShape(rest: shape[string,int,bool], i: int) -> PairIntShape[shape,string,int,bool]:
  PairIntShape(i, rest)

def add_int_field(RecordSet(fields, rows): RecordSet[shape1], name: String, f: shape1[String, Int, Bool] -> Int) -> RecordSet[PairIntShape[shape1]]:
  RecordSet(makePairIntShape(fields, Field(name, RVInt)), rows.map_List(\row -> makePairIntShape(row, f(row))))
```

As I look at this I guess this is just the power of currying. `PairIntShape[shape1]` becomes a type that takes 3 parameters.

This is definitely an interesting experiment

PR #256, opened 2019-05-16
Go back to PairShape

snoble:
one last tweak. I may have gone too far.

this example would be nicer with type aliases. But that might just mean I'm doing this the wrong way


johnynek:
I have thought about type aliases but so far have hoped we could get by without them using single item structs. The downside to that is the wrapping and unwrapping. A compiler can make single item structs the same as a type alias and remove that at compile time, but syntactically it is still there...

type aliases shouldn't be hard to implement since they are only syntactic, and can be erased, I think, before typechecking, but I guess it does complicate error messages in that case because the types reported in error messages wouldn't be the aliases....

snoble:
My interest in type aliases in this case is I think it would have let me use TupleCons with the type signature I wanted and then I might have been able to use the tuple syntax. But I think I'd still need to make a new type alias everytime restructure was called, so it's not really that appealing.

Feels like there are a lot of things still to try within the existing language feature set

snoble:
The other thing I found myself wondering about was if there was a way to destructure when you don't know the name of the struct

PR #260, opened 2019-05-19
Put external defs in Program

johnynek:
To serialize compiled packages, we don't want to serialize the source ADTs (I don't think). But currently, evaluation actually reaches all the way back to the source to get the list of external defs. If we need them, we should explicitly put them on Program. That's what this PR does.


johnynek:
this is needed to support #54

PR #262, opened 2019-05-21
improvements to the proto deserialization code

johnynek:
This is trying to make the code more regular in advance of doing serialization of compiled packages which need to use the DAG approach used by `Type` for both `Pattern` and `TypedExpr`. Pattern depends on Strings and Types. TypedExpr depends on Strings, Types and Patterns.


johnynek:
part of #54

PR #268, opened 2019-05-29
Proto support for compiled packages

johnynek:
part of #54 

This is the last bit of proto serialization: compiled packages with all the ASTs inside suitable for evaluation or code generation.

There is not yet any testing, nor is it fully complete.


johnynek:
This seems to be functional... I'll follow up next with changes to the Main compiler tool to use the serialized code in incremental compilation!

snoble:
Congratulations 🎉🎉

johnynek:
@snoble thanks man... this has been pretty painful to get it all working.... hopefully it will actually be useful.

PR #269, opened 2019-05-30
Fix forall arg in function

### LINKED ISSUE #267: When using forall in the type of a parameter of a returned function other types get confused
I've tried to simplify this a bunch. It's possible this could be related to #264, or I might have just come across this at the same time.
```
package SubsumeTest

def restructure(input: String, f: String -> TupleCons[String, TupleCons[(forall ss. List[ss] -> Int) -> Int, Unit]]) -> String:
  (output, _) = f(input)
  output
```
results in the error
```
in file: /home/steven/Code/bosatsu2/test_workspace/subsume_test_case.bosatsu, package SubsumeTest, type ?a
2|
3|def restructure(input: String, f: String -> TupleCons[String, TupleCons[(forall ss. List[ss] -> Int) -> Int, Unit]]) -> String:
4|  (output, _) = f(input)
                  ^^^^^^^^

does not subsume type forall ss. Bosatsu/Predef::List[ss] -> Bosatsu/Predef::Int
3|def restructure(input: String, f: String -> TupleCons[String, TupleCons[(forall ss. List[ss] -> Int) -> Int, Unit]]) -> String:
4|  (output, _) = f(input)
5|  output
    ^^^^^^
```

It looks like `output` is getting assigned the function that `f` returns instead of the string

snoble:
Trying to fix #267 

This doesn't quite work yet. somehow ended up with `PackageName(NonEmptyList(Test)),TypeName(Constructor(Int))` in one of the test results which makes no sense.

Also regressed on another test.

Finally I suspect this doesn't work if the `forall` is nested inside a `struct` in an arg. I think the real answer has to do with the variance of meta. But this is sort of informative at least

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
@@ -437,29 +437,17 @@ object Infer {
         case (fa@Type.ForAll(_, _), rho) =>
           // Rule SPEC
           instantiate(fa).flatMap(subsCheckRho(_, rho, left, right))
-        case (Type.Fun(fa@Type.ForAll(vars, _), r1), rho2) =>
-          unifyFnForAllArg(rho2, vars, right, left).flatMap { case (a2, r2) =>
-            instantiate(fa).flatMap { instA1 =>
-              subsCheckFn(instA1, r1, a2, r2, left, right)
-            }
-          }
-        case (rho, Type.Fun(fa@Type.ForAll(vars, _), r1)) =>
-          unifyFnForAllArg(rho, vars, right, left).flatMap { case (a2, r2) =>
-            instantiate(fa).flatMap { instA1 =>
-              subsCheckFn(a2, r2, instA1, r1, left, right)
-            }
-          }
         case (rho1, Type.Fun(a2, r2)) =>
           // Rule FUN
           unifyFn(rho1, left, right).flatMap {
             case (a1, r1) =>
-              subsCheckFn(a1, r1, a2, r2, left, right)
+              subsCheckFn(a2, r2, a1, r1, left, right)
           }
         case (Type.Fun(a1, r1), rho2) =>
           // Rule FUN
           unifyFn(rho2, right, left).flatMap {
             case (a2, r2) =>
-              subsCheckFn(a1, r1, a2, r2, left, right)
+              subsCheckFn(a2, r2, a1, r1, left, right)
           }
         case (rho1, Type.TyApply(l2, r2)) =>
           unifyTyApp(rho1, left, right).flatMap {
@@ -519,16 +507,6 @@ object Infer {
             argT <- newMetaType
             resT <- newMetaType
             _ <- unify(tau, Type.Fun(argT, resT), fnRegion, evidenceRegion)
-          } yield (argT, resT)
-      }
-    def unifyFnForAllArg(fnType: Type, vars: NonEmptyList[Type.Var.Bound], fnRegion: Region, evidenceRegion: Region): Infer[(Type, Type)] =
-      fnType match {
-        case Type.Fun(Type.ForAll(vars2, arg), res) => pure((arg, res))
-        case tau =>
-          for {
-            argT <- newMetaType
-            resT <- newMetaType
-            _ <- unify(tau, Type.Fun(Type.ForAll(vars, argT), resT), fnRegion, evidenceRegion)
           } yield (argT, resT)
       }
 
--- a/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
@@ -441,13 +441,13 @@ object Infer {
           // Rule FUN
           unifyFn(rho1, left, right).flatMap {
             case (a1, r1) =>
-              subsCheckFn(a2, r2, a1, r1, left, right)
+              subsCheckFn(a1, r1, a2, r2, left, right)
           }
         case (Type.Fun(a1, r1), rho2) =>
           // Rule FUN
           unifyFn(rho2, right, left).flatMap {
             case (a2, r2) =>
-              subsCheckFn(a2, r2, a1, r1, left, right)
+              subsCheckFn(a1, r1, a2, r2, left, right)
           }
         case (rho1, Type.TyApply(l2, r2)) =>
           unifyTyApp(rho1, left, right).flatMap {
@@ -669,25 +669,25 @@ object Infer {
                 _ <- infer.set((Type.Fun(varT, bodyT), region(term)))
               } yield TypedExpr.AnnotatedLambda(name, varT, typedBody, tag)
           }
-        case AnnotatedLambda(name, tpe, result, tag) =>
+        case AnnotatedLambda(name, annotatedArgTy, result, tag) =>
           expect match {
             case Expected.Check((expTy, rr)) =>
               for {
                 vb <- unifyFn(expTy, rr, region(term))
-                (varT, bodyT) = vb
-                typedBody <- extendEnv(name, varT) {
+                (argT, resT) = vb
+                typedBody <- extendEnv(name, argT) {
                     // TODO we are ignoring the result of subsCheck here
                     // should we be coercing a var?
-                    subsCheck(tpe, varT, region(term), rr) *>
-                      checkRho(result, bodyT)
+                    subsCheck(argT, annotatedArgTy, region(term), rr) *>
+                      checkRho(result, resT)
                   }
-              } yield TypedExpr.AnnotatedLambda(name, varT /* or tpe? */, typedBody, tag)
+              } yield TypedExpr.AnnotatedLambda(name, argT /* or tpe? */, typedBody, tag)
             case infer@Expected.Inf(_) =>
               for { // TODO do we need to narrow or instantiate tpe?
-                typedBody <- extendEnv(name, tpe)(inferRho(result))
-                bodyT = typedBody.getType
-                _ <- infer.set((Type.Fun(tpe, bodyT), region(term)))
-              } yield TypedExpr.AnnotatedLambda(name, tpe, typedBody, tag)
+                typedBody <- extendEnv(name, annotatedArgTy)(inferRho(result))
+                resT = typedBody.getType
+                _ <- infer.set((Type.Fun(annotatedArgTy, resT), region(term)))
+              } yield TypedExpr.AnnotatedLambda(name, annotatedArgTy, typedBody, tag)
           }
         case Let(name, rhs, body, isRecursive, tag) =>
           if (isRecursive.isRecursive) {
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
```diff
@@ -509,6 +521,16 @@ object Infer {
             _ <- unify(tau, Type.Fun(argT, resT), fnRegion, evidenceRegion)
           } yield (argT, resT)
       }
+    def unifyFnForAllArg(fnType: Type, vars: NonEmptyList[Type.Var.Bound], fnRegion: Region, evidenceRegion: Region): Infer[(Type, Type)] =
+      fnType match {
+        case Type.Fun(Type.ForAll(vars2, arg), res) => pure((arg, res))
```
snoble: I should probably look at the `vars2` and compare to `vars`

But as I say, i don't think this is the right approach. Just a step

core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
```diff
@@ -437,6 +437,18 @@ object Infer {
         case (fa@Type.ForAll(_, _), rho) =>
           // Rule SPEC
           instantiate(fa).flatMap(subsCheckRho(_, rho, left, right))
+        case (Type.Fun(fa@Type.ForAll(vars, _), r1), rho2) =>
+          unifyFnForAllArg(rho2, vars, right, left).flatMap { case (a2, r2) =>
+            instantiate(fa).flatMap { instA1 =>
+              subsCheckFn(instA1, r1, a2, r2, left, right)
```
snoble: `for` was giving me grief so I just `flatMap`d and moved on

snoble:
Yeah, this does not work when I nest. But the tests are good

snoble:
This version is very suggestive. It's breaking another test (haven't narrowed down what part of `substitution works correctly`). And the new tests that break are because of this bizarre `TyConst(Defined(PackageName(NonEmptyList(Test)),TypeName(Constructor(Int)))))))` again.

But the fact that all that changed was the ordering of the variables to subsume seems to suggest there's somewhere that subsumption checking is happening in the wrong direction (which goes with your observation that this is probably a Covariance/Contravariance confusion)

johnynek:
The `Int` thing is a known issue in RankNInfer test, sorry... :( the predef is not set up correctly, so your naming of `Int` gets named as an `Int` in that package.

Also, the substitution tests were some I thought of to exercise substitution, some could even be wrong. I'm looking to see which one fails with this change.

Looking at the PR code now...

Thanks so much for working on this!

johnynek:
The test that is failing is this one:

https://github.com/johnynek/bosatsu/blob/master/core/src/test/scala/org/bykn/bosatsu/rankn/RankNInferTest.scala#L599

this looks wrong to me now:
```
struct Foo
(id: forall a. a -> Foo) = \x -> Foo
(idFoo: Foo -> Foo) = id
(id2: Foo -> Foo) = \x -> x
(idGen2: (forall a. a) -> Foo) = id2
main = Foo
```

`idGen2` can't be `(forall a. a) -> Foo` because I could pass it for instance `0: Int`, but it is really identity, so it would return `0` and not an instance of `Foo`....

That looks to me like, your fix has found a previous bug... :)

johnynek:
So, in the paper:
https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/putting.pdf

page 77:
```
subsCheckRho rho1 (Fun a2 r2) -- Rule FUN
  = do { (a1,r1) <- unifyFun rho1; subsCheckFun a1 r1 a2 r2 }

subsCheckRho (Fun a1 r1) rho2 -- Rule FUN
  = do { (a2,r2) <- unifyFun rho2; subsCheckFun a1 r1 a2 r2 }

subsCheckFun :: Sigma -> Rho -> Sigma -> Rho -> Tc ()
subsCheckFun a1 r1 a2 r2
  = do { subsCheck a2 a1 ; subsCheckRho r1 r2 }
```

Isn't that the current code? If so, I'm still trying to see if the paper wrong, or if there is an issue elsewhere.

snoble:
Yeah, I couldn't find anywhere that your code departed from the paper so there may be a flaw there. There is zipped up source for their code so it would be fun to read the struct free version.

snoble:
There is the place where subsCheckFun swaps a1 and a2 to subsCheck because of covariance. It's possible your variance code swaps it back, but I looked and I couldn't find anything like that

johnynek:
I think the bug is here:

paper page 54:
```
tcRho (ALam var var_ty body) (Check exp_ty)
= do { (arg_ty, body_ty) <- unifyFun exp_ty
; subsCheck arg_ty var_ty
; extendVarEnv var var_ty (checkRho body body_ty) }
```
but the code in master:
https://github.com/johnynek/bosatsu/blob/master/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala#L672

it looks to me like we have it reversed:
```scala
        case AnnotatedLambda(name, tpe, result, tag) =>
          expect match {
            case Expected.Check((expTy, rr)) =>
              for {
                vb <- unifyFn(expTy, rr, region(term))
                (varT, bodyT) = vb
                typedBody <- extendEnv(name, varT) {
                    // TODO we are ignoring the result of subsCheck here
                    // should we be coercing a var?
                    subsCheck(tpe, varT, region(term), rr) *>
                      checkRho(result, bodyT)
                  }
              } yield TypedExpr.AnnotatedLambda(name, varT /* or tpe? */, typedBody, tag)
```
see we are doing: `subsCheck(tpe, varT)` and `tpe` is type of the annotation, but they have `subsCheck arg_ty var_ty`, an `var_ty` was the type in the annotated lambda. So that is backwards, it seems to me.

If we reverse that `subsCheck` will that also fix the other cases? Seems like your change is effectively reversing a subsCheck, but maybe this is the bug?

snoble:
Ah yeah, that definitely looks right. Nice find!

johnynek:
I was wrong about my comment here:
https://github.com/johnynek/bosatsu/pull/269#issuecomment-497581020

I now believe that test again... We should want it to pass.

`(forall a. a)` is bottom type. If you had one, you could instantiate anything. It is therefore (if the type system is sound) uninhabited. So, `(forall a. a) -> Foo` is a function you can't call. You should be able to not call any function, and due to contravariance, since `Foo -> Foo` is contravariant in the first parameter, we would expect `(forall a. a) -> Foo` to substitute.

It will be interesting to see what fixing the last bug we identified in AnnotatedLambda support does.

snoble:
Ok, yeah.  `(forall a. a) -> Foo` is less polymorphic than `Foo -> Foo`

poking a bit now

snoble:
That puts us back to our previous state. But I have a better thread now. Something must be calling `subsCheckRho` backwards presumably.

johnynek:
@snoble looking at the error, are we back to this code:

https://github.com/johnynek/bosatsu/blob/master/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala#L626

maybe just putting a bunch of println's there would give us some clue...

Like, we are type checking, but somehow a type is making it out that shouldn't....

johnynek:
It would be great if we could find a test that fails with the previous bug... It's concerning that nothing we had showed that bug... :(

snoble:
Yeah, it's always that subsCheck code... and it just seems so right that somewhere upstream the variables are being flipped

I had previously very slowly worked back through stack traces to find out how we got there. I'm currently trying to think of a way to that by sneaking in a stack into the Infer monad

snoble:
> It would be great if we could find a test that fails with the previous bug... It's concerning that nothing we had showed that bug... :(

So when would that happen... when we have a `typeCheckRho` on an `AnnotatedLambda` with an `Expected.Check`. Should be manageable

snoble:
So `typeCheckRho` with an `App` does a `checkSigma` on the arg. Which will get us back to `typeCheckRho` in a few hops. And if the `arg` is an `AnnotatedLambda` then there we are.

I think I can devise that

core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
```diff
@@ -669,25 +669,25 @@ object Infer {
...
           expect match {
             case Expected.Check((expTy, rr)) =>
               for {
                 vb <- unifyFn(expTy, rr, region(term))
-                (varT, bodyT) = vb
-                typedBody <- extendEnv(name, varT) {
+                (argT, resT) = vb
+                typedBody <- extendEnv(name, argT) {
                     // TODO we are ignoring the result of subsCheck here
                     // should we be coercing a var?
```
snoble, line 679: This comment just means the `coerce` is being ignored? if the `subsCheck` fails then it all fails, right?

johnynek: that's right. We are only using it to check for failure.

snoble:
I now only have one failing test, which is a new test. But the fix diverges from the paper. Basically I started wondering why you would use `subsCheckRho` for a subsume test for the right side of a function but `subsCheck` for the left side. `subsCheckRho` basically handles all of the weird things you might find in a type.

Once I started wondering that I saw it everywhere. I'm not sure how to consider this solution and get more or less confident. Certainly the fact that a test is still failing isn't good.

johnynek:
See page 48:
```
After recursively calling subsCheck, we must check the side condition a 6∈ ftv(σ) for
rule skol, namely that the skolemised variables skol_tvs are not free in sigma1.
You might wonder how this could possibly be the case, since skol_tvs are freshly
made, but the recursive call to subsCheck might have bound a meta type variable in
sigma1 to one of the skolems. That is why we wait until after the call to subsCheck
before making the test. For example, consider the term:

\x. (x :: (forall a. a->a))

Rule annot will invoke a subsumption check that tries to confirm that the type of
the body of the annotated term (x in this case) is at least as polymorphic as the
type signature ∀a.a → a. By this time, x will be in the environment with type τ ,
a meta type variable, so we end up checking this judgement
τ ≤ ∀a.a → a
We skolemise ∀a.a → a to get b → b (where b is the fresh skolem constant), and
then unify, which binds τ to the type b → b. But rule skol requires that the skolem
constant b not be free in the type on the left of the ≤. It wasn’t to begin with, but
after the unification it may be! The function getFreeTyVars finds the free TyVars
of its argument, which are precisely the skolem constants. Like getMetaTyVars,
getFreeTyVars takes account of the substitution, which is why it has a monadic
type.
```
This seems that banning inferring types like this is expected, doesn't it?

johnynek:
So, re-reading the paper, page 63, section 7.4, it appears this limitation was actually always in the paper and they don't permit in the paper's algorithm polytypes as type arguments to constructors. So, while you can create them in this code, you can never use them currently because no accessing of them typechecks.
```
In the preceding discussion, we have implicitly assumed that data types can only
be instantiated with monomorphic types. For example, consider:
data Tree a = Leaf a | Branch (Tree a) (Tree a)
One can construct values of types such as (Tree Int) or (Tree (Tree Int)), but
what about Tree (∀a.a → Int)? More generally, in a type, can the argument of a
type constructor be a σ-type, or must it be a τ -type? Well, the constructor Leaf is a
polymorphic function of type (∀a.a → Tree a), and our restriction to predicativity
therefore requires that we instantiate Leaf only at a τ -type (Section 3.4). So the
simplest solution is to require that type constructors are parameterised only by
monotypes. Then, just as instSigma instantiates a polymorphic function with fresh
meta (mono-)type variables, so instDataCon instantiates the data constructor’s
type with fresh meta (mono-)type variable.
```

snoble:
> This seems that banning inferring types like this is expected, doesn't it?

Yeah, I think this is right

PR #270, opened 2019-06-01
Make Type.Rho a real type, not an alias

johnynek:
related to #267 and #269 

Before this PR, `Rho` was a type alias for `Type` as in the paper. Scala's subtyping allows us to define it as a real type, and let the compiler check some of the invariants we claim. Doing this didn't seem to uncover any new bugs, but seems valuable for safety.

There are two other invariants not checked:
1. Type.Tau: no ForAll anywhere. For instance `(forall a. a) -> Foo` is a Rho type, but not a Tau type.
2. weak-prenex form: this means we have moved all the ForAlls up as far as possible using covariance.

While doing this, I added a variant of the tests in #269. I think at least part of our issue is around pattern matches. Currently, we would do a subsCheck eventually on a meta variable, against an annotated `forall` type, and that will always fail if the type. By contrast, Expr.Annotation leads to a checkSigma, which will skolemize the forall and has a separate code path that is not subsCheck. This seems related to the issue since all the forms in #269 are using pattern matching under the hood.

cc @snoble

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
@@ -345,7 +345,7 @@ object Infer {
           for {
             sks1 <- tvs.traverse(newSkolemTyVar)
             sksT = sks1.map(Type.TyVar(_))
-            sks2ty <- skolemize(substTy(tvs, sksT, ty))
+            sks2ty <- skolemize(substTyRho(tvs, sksT)(ty))
             (sks2, ty2) = sks2ty
           } yield (sks1.toList ::: sks2, ty2)
         case Type.TyApply(left, right) =>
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
```diff
@@ -802,7 +885,11 @@ object Infer {
         case GenPattern.WildCard => Infer.pure((pat, Nil))
         case GenPattern.Literal(lit) =>
           val tpe = Type.getTypeOf(lit)
-          instSigma(tpe, sigma, reg).as((pat, Nil))
+          val check = sigma match {
```
johnynek, line 888: instSigma was only being called once, but here it is clearly on a Tau type because literals are fully known. I inlined it.

PR #275, opened 2019-06-06
Snoble/recordset

snoble:
I figure I'll keep PRing this example into this repo because it seems useful to see how the language can be used. But I should make a separate repo with its own dependent bazel build.

One thing that popped up is there might be a bug where you can export constructors but not import them? I get a parse error on the parentheses.

Something else that comes up is `ps(c, ps(b, ps(a, s("Plus 2".int_field( \r -> r.get(b).add(2) ))))))` is pretty meh for syntax. Note this pattern is the same for constructing tuples, but I'm doing more than just creating a structure so I can't just use tuples, even if there were a way to make the types work.

My dream would be if I could do something like `record[c, b, a, "Plus 2".int_field( \r -> r.get(b).add(2) )]` and somehow define `record[...,...]` as "apply `s` to the last term and nest applications of `ps` on through the rest of the terms in reverse order". You could then define `[]` for `List` and `()` for `TupleCons` this way too. But that might be a bit much for a fairly narrow use case.

Alternatively, if I could fold through a tuple I think that might unlock something. But it's a weird meaning of `fold` because the accumulator keeps changing type.

### DIFF
--- a/test_workspace/recordset.bosatsu
+++ b/test_workspace/recordset.bosatsu
@@ -1,17 +1,17 @@
 package RecordSet
 
-enum RowEntry:
-  REBool(value: Bool)
-  REInt(value: Int)
-  REString(value: String)
+enum RowEntry[w]:
+  REBool(value: w[Bool])
+  REInt(value: w[Int])
+  REString(value: w[String])
 
-struct RecordField[t](name: String, toValue: t -> RowEntry)
+struct RecordField[t](name: String, to_entry: forall w. w[t] -> RowEntry[w])
 struct RecordValue[t](value: t)
 
-struct RecordSet[shape](fields: shape[RecordField], rows: List[shape[RecordValue]])
+struct RecordSet[shape](fields: shape[RecordField], rows: List[shape[RecordValue]], to_row: forall w. shape[w] -> List[RowEntry[w]])
 
-def restructure(RecordSet(fields, rows), f: forall w. shape1[w] -> shape2[w]) -> RecordSet[shape2]:
-  RecordSet(f(fields), rows.map_List(f))
+def restructure(RecordSet(fields, rows), f: forall w. shape1[w] -> shape2[w], to_row: forall w. shape2[w] -> List[RowEntry[w]]) -> RecordSet[shape2]:
+  RecordSet(f(fields), rows.map_List(f), to_row)
 
 struct IntShape[w](int: w[Int])
 struct PairShape[shape1,shape2,w](left: shape1[w], right: shape2[w])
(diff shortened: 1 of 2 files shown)


test_workspace/recordset.bosatsu
```diff
@@ -1,37 +1,83 @@
...
+def create_field(rf: RecordField[t], fn: shape[RecordValue] -> t):
+  RecordGetter(\sh -> rf, \sh -> RecordValue(fn(sh)))
+
+def list_of_rows(RecordSet(fields, rows, getters, traverse, record_to_list): RecordSet[shape]):
+  def getter_to_row_entry(row: shape[RecordValue]):
+    (result_fn: forall tt. RecordGetter[shape, tt] -> RecordRowEntry[RecordValue, tt]) = \RecordGetter(get_field, get_value) ->
+      RecordField(name, to_entry) = get_field(fields)
+      RecordRowEntry(to_entry(get_value(row)))
+    result_fn
+  rows.map_List(\row -> record_to_list(getters.traverse(getter_to_row_entry(row))))
```
johnynek, line 37: you might consider `[record_to_list(getters.traverse(getter_to_row_entry(row))) for row in rows]` this will compile to the same code you have here but may be a bit prettier.... up to you.

test_workspace/recordset.bosatsu
```diff
@@ -1,37 +1,83 @@
...
+  RecordSet(reshaperF(fields), rows.map_List(reshaperV), new_getters, traverse, record_to_list)
 
-def add_int_field(RecordSet(fields, rows): RecordSet[shape1], name: String, f: shape1[String, Int, Bool] -> Int) -> RecordSet[PairShape[shape1, IntShape]]:
-  RecordSet(make_pair_int_shape(fields, Field(name, RVInt)), rows.map_List(\row -> make_pair_int_shape(row, f(row))))
+def concat_records(RecordSet(fields, rows, getters, traverse, record_to_list), more_rows):
+  RecordSet(fields, rows.concat(more_rows), getters, traverse, record_to_list)
 
-##################################################
+struct NilShape[w]
+struct PS[t,rest,w](left: w[t], right: rest[w])
```
johnynek, line 54: dude, you are a type maniac.... these are pretty wild types.... certainly stress testing things here. :)

test_workspace/recordset.bosatsu
```diff
@@ -60,12 +106,26 @@ def equal_List(is_equal, l1, l2):
...
-  compare1.equals(x1, y1) && compare2.equals(x2, y2) && compare3.equals(x3, y3) && compare3.equals(x4, y4)
+def equal_RowEntry(re1, re2):
+  match (re1, re2):
+    (REBool(RecordValue(x1)), REBool(RecordValue(x2))): cmp_Bool.equals(x1, x2)
+    (REInt(RecordValue(x1)), REInt(RecordValue(x2))): cmp_Int.equals(x1, x2)
+    (REString(RecordValue(x1)), REString(RecordValue(x2))): string_Order_fn.equals(x1, x2)
+    _: False
+
+def equal_rows(row1, row2):
+  equal_RowEntry.equal_List(row1, row2)
```
johnynek: `equal_rows = equal_List(equal_RowEntry)` would also work.... but maybe this is clearer.

johnynek:
Happy to merge this when you are ready (also, I can give you commit access if I haven't already if you would like).

test_workspace/recordset.bosatsu
snoble: Honestly, I think this is the first time I've ever felt I've understood a type system enough to be creative with it.

snoble:
I'll get those two suggestions added before merging. I'd feel very complemented to have commit access.

test_workspace/recordset.bosatsu
snoble: this change introduces the error
```
failure to get type: TyApply(TyVar(Bound(a)),TyConst(Defined(PackageName(NonEmptyList(RecordSet, Library)),TypeName(Constructor(RecordValue)))))
```

interestingly it only occurs when I run the tests and not when I typecheck the file

snoble:
ready to be merged

test_workspace/recordset.bosatsu
johnynek: That error message happens here:

https://github.com/johnynek/bosatsu/blob/master/core/src/main/scala/org/bykn/bosatsu/Evaluation.scala#L346

That looks like a bug of some kind. Testing uses evaluation, typechecking does not, and this bug looks like it happens with evaluation.

johnynek:
Thanks!

I also invited you as a collaborator. Thanks for your contributions!

snoble:
Amazing, thank you

PR #279, opened 2019-06-23
Fix bugs uncovered with a stricter checks

johnynek:
when starting to use the proto output I discovered new lurking bugs. The proto cannot serialize metavars or skolems, which should never appear in fully compiled code. Unfortunately that wasn't the case.

The top-level types for lets in the `TypedExpr` were correct, but inside there could still be meta-vars or skolems that shouldn't have been there. The full typing is something the paper is a bit short on and I made some mistakes implementing it.

I still have some questions, but it seems at least the bugs that were present are gone.

After this is merged, I should be able to send the PR which enables incremental compilation using the protobuf outputs.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/TypedExpr.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/TypedExpr.scala
@@ -341,7 +341,17 @@ object TypedExpr {
       case App(fnT, arg, tpe, tag) =>
         App(recur(fnT), recur(arg), tpe, tag)
       case Let(b, e, in, r, t) =>
-        if (b == name) te // shadow
+        if (b == name) {
+          if (r.isRecursive) {
+            // in this case, b is in scope for e
+            // so it shadows a the previous definition
+            te // shadow
+          } else {
+            // then b is not in scope for e
+            // but b does shadow inside `in`
+            Let(b, recur(e), in, r, t)
+          }
+        }
         else Let(b, recur(e), recur(in), r, t)
       case lit@Literal(_, _, _) => lit
       case Match(arg, branches, tag) =>
--- a/core/src/main/scala/org/bykn/bosatsu/TypedExpr.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/TypedExpr.scala
@@ -238,30 +238,40 @@ object TypedExpr {
   }
 
   type Coerce = FunctionK[TypedExpr, TypedExpr]
+
   def coerceRho(tpe: Type.Rho): Coerce =
-    new FunctionK[TypedExpr, TypedExpr] { self =>
-      def apply[A](expr: TypedExpr[A]) =
-        expr match {
-          case Annotation(t, _, _) => self(t)
-          case Generic(params, expr, tag) =>
-            // This definitely feels wrong,
-            // but without this, I don't see what else we can do
-            Generic(params, self(expr), tag)
-          case Var(p, name, _, t) => Var(p, name, tpe, t)
-          case AnnotatedLambda(arg, argT, res, tag) =>
-            // only some coercions would make sense here
-            // how to handle?
-            // one way out could be to return a type to Annotation
-            // and just wrap it in this case, could it be that simple?
-            Annotation(expr, tpe, expr.tag)
-          case App(fn, arg, _, tag) =>
-            // do we need to coerce fn into the right shape?
-            App(fn, arg, tpe, tag)
-          case Let(arg, argE, in, rec, tag) =>
-            Let(arg, argE, self(in), rec, tag)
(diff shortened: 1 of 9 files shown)


core/src/test/scala/org/bykn/bosatsu/NormalizationTest.scala
```diff
@@ -178,7 +176,7 @@ out=match Pair(1, "two"):
 
 """
       ), "Match/Structs",
-      Struct(0,List(Literal(Integer(BigInteger.valueOf(3))), Literal(Str("two")), Literal(Integer(BigInteger.valueOf(1)))))
+      Struct(0,List(Literal(Lit(3)), Literal(Lit(1)), Literal(Lit(1))))
```
johnynek: this looks really suspect to me. I looks like the `two` has been replaced by `1`... the tests pass though.

@snoble can you take a look at this? If pattern matching were broken, I would assume at least one of the evalueation tests or the existing bosatsu code that is tested would hit it.

core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -1251,4 +1251,25 @@ main = match Bar(a):
...
+  test("pattern example from pair to triple") {
+    runBosatsuTest(
+      List("""
+package A
+
+struct Pair(f, s)
+struct Trip(f, s, t)
+
+Trip(a, b, c) = match Pair(1, "two"):
+  Pair(f, s): Trip(3, s, f)
```
johnynek, line 1263: this seems like your example, and the evaluation codes seems to execute it correctly. What do you think @snoble am I misunderstanding Normalization?

core/src/main/scala/org/bykn/bosatsu/TypedExpr.scala
```diff
@@ -310,23 +320,65 @@ object TypedExpr {
...
+        else {
+          // no shadow
+          AnnotatedLambda(b, tpe, recur(expr), tag)
+        }
+      case Var(None, nm, _, tag) if nm == name => Var(None, name, tpe, tag)
+      case v@Var(_, _, _, _) => v
+      case App(fnT, arg, tpe, tag) =>
+        App(recur(fnT), recur(arg), tpe, tag)
+      case Let(b, e, in, r, t) =>
+        if (b == name) te // shadow
```
johnynek: this isn't quite right... if `r` is not recursive, we should unconditionally recurse on `e`.

core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
snoble: Oops, I need to clean up my notifications so I see these things. Looking

snoble: Yeah, this definitely looks like what Normalization _should_ be doing. It could be the constructor code or the pattern matching code (in Normalization)

core/src/test/scala/org/bykn/bosatsu/NormalizationTest.scala
snoble: Weird, it's the constructor code that seems to be causing the issue. Getting closer

snoble: Interesting, the constructor is fine. It's the reduction of `App(Lambda`

snoble: Something weird is happening with a bunch of variables named `a` and `b` showing up in the environment. I don't know where they are coming from yet

snoble: Ohhhh I see... I'm handling shadowing wrong of lambda variables. I'm seeing an environment like this 
`env: (Map(Name(a) -> ExpressionKeyTag(LambdaVar(2),Set()), Name(b) -> ExpressionKeyTag(LambdaVar(0),Set())),List(Name(b), Name(a), Name(a)))`
But `Name(a)` should point to `LambdaVar(1)` not `LambdaVar(2)` in the map.

I just need to remember where the `zip` that does that is and fix that

snoble:
I thought I had a fix but that seems to not have quite done it. I'll take more time tomorrow.

snoble:
ok... I think that actually did solve it. Just need to understand why that first match test has changed.

And I need to clean up the code and more explicitly test. But that can go in a separate pr.

johnynek:
@snoble will you be able to fix the new test that is broken now?

johnynek:
I fixed the tests.

snoble:
Sorry for the bugs. Thank you for catching them!

PR #281, opened 2019-06-23
Enable incremental compilation in bazel rules

johnynek:
This uses the work in #54 to incrementally compile `bosatsu_library` and avoid typechecking dependencies for given targets. bosatsu_test now reuses the library parsing and typechecking.

this uncovered the bugs in #279


johnynek:
note, this coverage decline is actually false, we exercise the Main.scala code in the bazel rules, but we don't merge that coverage into the main reports.

johnynek:
going to merge this when it goes green.... end of a long road.

It's pretty great to have incremental compilation working!

My next steps I think are:

1. documenting using bosatsu to generate json
2. try to write more libraries
3. use graal native-image to generate the bazel rule so the we don't pay the 1.5s JVM startup and don't need to worry about workers
4. add optimization/normalization rules that transform TypedExpr to TypedExpr.
5. add more tests of coercion in TypedExpr. It's a weaker kind of type inference, but I don't think it is solved: you know a type and cast it down to a narrower type.
6. scala API for embedding in scala programs directly without having to go through JSON -> circe, or whatever.

johnynek:
The coverage issue is due to Main not having coverage in tests, but is tested by the bosatsu bazel rules.

PR #297, opened 2019-07-22
Add Bosatsu/Nothing, showing how to make a Nothing or Void type

johnynek:
In a non-total language, we would need the ability to make enums with no variants.

In a total language, `forall a. (a -> a) -> a` is uninhabited, so we can use a struct containing one as a synonym for Nothing. If we ever had one, we could produce any value.

### DIFF
--- a/test_workspace/BUILD
+++ b/test_workspace/BUILD
@@ -103,3 +103,8 @@ bosatsu_test(
     srcs = ["recordset.bosatsu"],
     packages = ["RecordSet/Library"],
     )
+
+bosatsu_library(
+    name = "nothing",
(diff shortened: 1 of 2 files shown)


johnynek:
@snoble check this out. This is a neat trick. I had previously thought about encoding `Nothing` by using `external struct Nothing` and `external def impossible(n: Nothing) -> a` but actually, we don't need it to be external.

johnynek:
actually, we can simplify, indeed `forall a. (a -> a) -> a` is uninhabited, as are many types, but it is simpler to just make `Nothing` an alias for `forall a. a` which is also uninhabited.

Issue #289, opened 2019-07-14
use names on constructors


johnynek:
related to #240 

Currently, we have syntax like:
```
struct Foo(bar: Int)
```
But `bar` is never used anywhere in the code later. There is no way to do so.

Goals:

1. use the names at construction time
2. use the names when pattern matching to give "getters" for values.
3. minimize unfamiliarity (primarily Python, but possibly Rust for inspiration).

We explored some syntax in #240 for matching, but I think matching and creation should be designed together.

There is a related question: can use also support named arguments to `def` function values? This is very hard to do for general values that may be functions, so this would introduce a distinction between let bindings that point to functions vs functions. This is likely similar to a distinction between using a constructor with names vs assigning a constructor to a let binding and using it as a function.

Here are some decision axes:

## Creation looks like python/function calls
Python-like syntax for creation:
```python
x = Foo(bar = 3)
```
If we use that creation syntax, I think (see arguments in #240) that the pattern match version should be:
```
x = Foo(bar = 3)
Foo(bar = y) = x
# or
y = match x:
  Foo(bar = z): z

# now y = 3
```
A different variant of this is `Foo(bar: 3)`. It has something going for it, but does seem to mix metaphors of function calls and dict creation. I can see an argument that we should either be very close to python, or significantly different.

## Creation looks like Rust/Haskell
Rust/Haskell like record creation:
```
x = Foo { bar = 3 }
Foo { bar = y } = x 
#Colon syntax more like dicts:
x = Foo { bar: 3 }
Foo { bar: y } = x
```
One benefit, or issue, with this approach is that it doesn't invite the idea that function calls would work the same. A challenge to function calls supporting names is that we can't easily support named functions through let bindings:
```
def f(arg: Int): Int = arg

g = f
# this seems tractable: we can look up the name in the environment
f(arg = 3)
# this does not seem tractable since the names would become part of the type of the binding and that would seriously change inference and how we think about function types:
g(arg = 3)
```
If we use `Foo { bar = 3 }` or `Foo { bar: 3 }` it doesn't invite the idea that `f { arg = 3 }` or `f { arg: 3 }` would work since that syntactically looks so strange. Next, `Foo` is kind of like a dictionary, which we have literals for, so giving a name to them and reusing the syntax looks reasonable to me.

The main downside to this approach is that `Foo(3)` and `Foo { bar: 3 }` are the same, but syntactically they look different. `Foo(bar = 3)` seems to make that more similar.

[...]

PR #301, opened 2019-07-27
Add struct/enum record syntax

johnynek:
This PR adds the parsing of creating structs and enums using record syntax.

The variant in #289 I went with was the dict-like syntax:
```
struct Foo(bar: Int)

f = Foo { bar: 3 }

# like rust we can also do:
struct Person(name: String, age: Int)

name = "Alice"
age = 37
p = Person { name, age }
```
I didn't implement pattern matching yet, but that would work like:

```
Person { name: nm, ... } = p
# or use name as the binding value:
Person { name, ... } = p
```

The analogy was suggested by @avibryant : you can make values with tuple-like or dict-like syntax (but not a hybrid of the two).

Lastly, I've decided to make one minor change: pattern matches on structs/enums must cover all the fields or explicitly add `...` to say you don't care about the rest of the fields. To preserve the duality between matching and creation, I may remove automatic currying of tuple like constructors:
```
struct Foo(bar, baz)

# currently this works, but is suprising
# x is the same as x = \z -> Foo(12, z)
x = Foo(12)
```
When I'm done with this record project, the above will become a compilation error and require explicit `\baz -> Foo(12, baz)` to get this result.

This does open the question: is automatic currying too strange in the rest of the code? Syntactically it it is hard to avoid it now since there is only one function type: `a -> b`. So once we put a function in a binding, we can't check arity, so to be consistent, we allow currying of any function. But constructors have capital names which cannot be bound except in type definitions, so we can syntactically check arity for constructors.

cc @non who also recommended this syntax.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
@@ -118,8 +118,8 @@ sealed abstract class Declaration {
 
       case RecordConstructor(name, args) =>
         val argDoc = Doc.char('{') +
-          Doc.intercalate(Doc.char(',') + Doc.line,
-          args.toList.map(_.toDoc)).grouped.nested(2) + Doc.char('}')
+          Doc.intercalate(Doc.char(',') + Doc.space,
+          args.toList.map(_.toDoc)) + Doc.char('}')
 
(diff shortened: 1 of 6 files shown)


johnynek:
some travis issue failed the CI. Let's see if #302 can help.

Issue #311, opened 2019-08-03
bug with incorrect shadowing of top-level bindings in Evaluation


johnynek:
below we exposed a serious bug in evaluation. if we make a closure over a top-level let binding and match on it like:
```
a = 42
f2 = match a:
  11: 10000
   x: x
```

we should expect f2 is 42, but if we do `eq_Int(f2, 11)` this winds up creating internal lambdas: `(\a -> eq_Int(f2, a))(11)` which overwrites the environment for f2.

I'm not 100% what is going on, but I think the problem is f2 is not carrying with it the fact that a is its environment so when it is evaluated in the context of `\a -> ...` the value for a overwrites.
 
I think this is due to the way we handle falling back to the top level...

```
runBosatsuTest(List("""package A

a = 42
f2 = match a:
  11: 12
   x: x

tests = Test("test record",
  [
    Assertion(f2.eq_Int(11), "f2 == 42"),
  ])
"""), "A", 1)
```

PR #312, opened 2019-08-03
Rewrite records into tuple-like syntax

### LINKED ISSUE #308: Add pattern linting/translation phase to remove record notation
Add a translation of Declaration to Declaration and Statement to Statement which translates all Patterns to not have any Record Syntax matches and to make sure all constructor patterns match the arity of the constructor or use the `...` (see #306 ).

The goal of this is to add a purely syntactic transformation which we call in Package just before inference.

We may want to use `Ior` here to do something like `def translate(stmt: Statement): Ior[PatternErrors, Statement]` so we can accumulate both errors in this translation as well as errors in type inference.

An open question is how tight to enforce removal of record syntax in the types. A trick would be to parameterize `Statement` and `Declaration` on a type that represents where we are in the translation state. While this seems appealing, it is also a lot of work and since we only cross this phase once, of dubious benefit.

The alternative is to simply put `sys.error("internal error, invariant violated")` when we see the unexpected patterns or record declarations in the translation to Expr. Once we reach Expr, which is where Infer and code generation lives, it won't matter anyway.

As nice as type-enforced correctness is, it seems like a heavy lift with little actual added safety in this case, so I tend to think I will go with the error-based approach. Note: it won't be a perf hit since it will be an error on a pattern match which we already have to do, but we just are ignoring that we have a branch that we could statically prove we will never hit.

johnynek:
close #308 

Translate record syntax, so we don't need to change anything about typechecking or evaluation.

I did hit a new bug here filed at #311

- [x] remove cruft from code

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Package.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Package.scala
@@ -144,14 +144,12 @@ object Package {
       ValidatedNel[PackageError,
       Program[TypeEnv[Variance], TypedExpr[Declaration], Statement]] = {
 
-    val Program(parsedTypeEnv, lets, extDefs, _) = {
+    val Program((importedTypeEnv, parsedTypeEnv), lets, extDefs, _) = {
       // here we make a pass to get all the local names
       val localDefs = Statement.definitionsOf(stmt)
       val srcConv = SourceConverter(p, imps, localDefs)
       srcConv.toProgram(stmt)
     }
-
-    val importedTypeEnv = Referant.importedTypeEnv(imps)(_.name)
 
     val inferVarianceParsed: Either[PackageError, ParsedTypeEnv[Variance]] =
       VarianceFormula.solve(importedTypeEnv, parsedTypeEnv.allDefinedTypes)
--- a/core/src/main/scala/org/bykn/bosatsu/Package.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Package.scala
@@ -1,7 +1,7 @@
 package org.bykn.bosatsu
 
-import cats.Functor
-import cats.data.{ValidatedNel, Validated, NonEmptyList}
+import cats.{Functor, Semigroup}
+import cats.data.{Ior, ValidatedNel, Validated, NonEmptyList}
 import cats.implicits._
 import fastparse.all._
 import org.typelevel.paiges.{Doc, Document}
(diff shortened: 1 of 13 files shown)


core/src/main/scala/org/bykn/bosatsu/Package.scala
```diff
@@ -146,50 +144,20 @@ object Package {
...
-        }) // name to cons
-
-      srcConv.toProgram(p, stmt)
+      val srcConv = SourceConverter(p, imps, localDefs)
+      try {
+        srcConv.toProgram(stmt)
+      }
+      catch {
+        case err: SourceConverter.Error =>
+          // TODO, make toProgram return an Ior
```
johnynek: this should be done before merging.

core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
```diff
@@ -402,11 +419,25 @@ final class SourceConverter(
     }
   }
 
+  private def assertResult[A](a: Result[A]): A =
```
johnynek, line 427: remove this after we push Result through in more places.

core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
```diff
@@ -423,24 +454,125 @@ final class SourceConverter(
...
+                  SourceConverter
+                    .addError(res0,
+                      SourceConverter.UnexpectedField(nm, pat, extra, params.map(_._1), region))
+              }
+            case None =>
+              SourceConverter.failure(SourceConverter.UnknownConstructor(nm, pat, region))
+          }
+        }
+    }(SourceConverter.parallelIor) // use the parallel, not the default Applicative which is Monadic
+    .map(_.mapType(toType))
```
johnynek, line 570: we should use a `Result` type in toType as well so we record any errors we get looking up the type.

core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -1410,4 +1410,55 @@ tests = Test("reordering",
 )
 """), "RecordSet/Library", 1)
   }
+
+  test("record patterns") {
+    // TODO:
```
johnynek: remove this inline bug report now that #311 has been filed.

core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -1410,4 +1410,55 @@ tests = Test("reordering",
...
+
+// tests = Test("test record",
+  // [
+    // Assertion(f2.eq_Int(11), "f2 == 42"),
+  // ])
+// """), "A", 1)
+
+    runBosatsuTest(
+      List("""
+package A
```
johnynek, line 1417: Add more tests:

1. destructuring
2. creation with record
3. negative tests for incorrect match size
4. negative tests for unknown fields

core/src/test/scala/org/bykn/bosatsu/TestUtils.scala
```diff
@@ -78,6 +70,7 @@ object TestUtils {
   def evalTestMode[A](packages: List[String], mainPackS: String, expected: EvaluationMode[A], extern: Externals = Externals.empty) = {
     def inferredHandler(packMap: PackageMap.Inferred, mainPack: PackageName): Assertion = {
       val ev = Evaluation(packMap, Predef.jvmExternals ++ extern)
+      //if (mainPackS == "A") println(ev.repr)
```
johnynek, line 73: remove

johnynek:
note the coverage is too low especially the lack of testing the error paths.

core/src/main/scala/org/bykn/bosatsu/Pattern.scala
```diff
@@ -155,19 +144,22 @@ object Pattern {
...
+      }
+      object FieldKind {
+        final case class Explicit(field: Bindable) extends FieldKind
+        // an implicit field can only be associated with a Var of
+        // the same name
+        final case class Implicit(field: Bindable) extends FieldKind
+      }
       final case object TupleLike extends Style
       // represents the fields like: Foo { bar: x, age }
       // the None means we expect the corresponding argument to
```
johnynek: this comment is out of date.

core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
```diff
@@ -402,11 +419,25 @@ final class SourceConverter(
...
+      case Ior.Right(a) => a
+      case Ior.Both(err, a) =>
+        // TODO: push result all the way to the top
+        throw err.head
+      case Ior.Left(err) =>
+        // TODO: push result all the way to the top
+        throw err.head
+    }
+
+  private def unTuplePattern(pat: Pattern.Parsed, region: Region): Pattern[(PackageName, Constructor), rankn.Type] =
```
johnynek, line 438: this is big enough for now, we can follow up removing this and only using the Result type.

PR #315, opened 2019-08-04
Use Result in more of SourceConverter

### LINKED ISSUE #314: use SourceConverter.Result internally to propagate all errors
currently we are only keeping the first error.

We need to use an applicative style to translate `Declaration` to `Result[Expr[Declaration]]`

johnynek:
close #314

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
@@ -225,23 +225,24 @@ final class SourceConverter(
                 "flat_map_List"
             }
             val opExpr: Expr[Declaration] = Expr.Var(pn, Identifier.Name(opName), l)
-            val resR = apply(res.value)
             val resExpr: Result[Expr[Declaration]] =
               filter match {
-                case None => resR
+                case None => apply(res.value)
                 case Some(cond) =>
+                  // To do filters, we lift all results into lists,
+                  // so single items must be made singleton lists
                   val empty: Expr[Declaration] =
                     Expr.Var(pn, Identifier.Constructor("EmptyList"), cond)
-                  // we need theReturn
                   val ressing = res match {
                     case SpliceOrItem.Item(r) =>
-                      resR.map { ritem =>
+                      // here we lift the result into a a singleton list
+                      apply(r).map { ritem =>
                         Expr.App(
                           Expr.App(Expr.Var(pn, Identifier.Constructor("NonEmptyList"), r), ritem, r),
                           empty,
                           r)
                       }
-                    case SpliceOrItem.Splice(_) => resR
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
```diff
@@ -58,48 +58,59 @@ final class SourceConverter(
...
+        apply(decl).map(_.as(decl))
       case DefFn(defstmt@DefStatement(_, _, _, _)) =>
         val inExpr = defstmt.result match {
           case (_, Padding(_, in)) => apply(in)
         }
-        val lambda = defstmt.toLambdaExpr({ res => apply(res._1.get) }, decl)(
-          unTuplePattern(_, decl.region), toType(_))
+        // TODO
+        val lambda = defstmt.toLambdaExprF({ res => apply(res._1.get) }, success(decl))(
+          unTuplePattern(_, decl.region), { t => success(toType(t)) })
```
johnynek, line 108: we should lift `toType` into `Result` by checking the visible types, and either return `Ior.Right` or `Ior.Both` to signal failure but also check to see if we can find any other errors.

core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
```diff
@@ -207,33 +225,38 @@ final class SourceConverter(
...
-                }
+            val resExpr: Result[Expr[Declaration]] =
+              (res, filter) match {
+                case (itOrSp, None) =>
+                  apply(itOrSp.value)
+                case (itOrSp, Some(cond)) =>
+                  val empty: Expr[Declaration] =
+                    Expr.Var(pn, Identifier.Constructor("EmptyList"), cond)
+                  // we need theReturn
+                  val r = itOrSp.value
```
johnynek: `apply(itOrSp.value)` appears in both branches. Might be clearer to only match on filter (and a bit more efficient).

core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
```diff
@@ -653,62 +669,63 @@ final class SourceConverter(
...
-              }
-              val freeVars = rankn.Type.freeTyVars(tpe :: Nil)
-              // these vars were parsed so they are never skolem vars
-              val freeBound = freeVars.map {
-                case b@rankn.Type.Var.Bound(_) => b
-                case s@rankn.Type.Var.Skolem(_, _) => sys.error(s"invariant violation: parsed a skolem var: $s")
-              }
-              val maybeForAll = rankn.Type.forAll(freeBound, tpe)
-              (binds, (name, maybeForAll) :: exts)
+      stmts.toList.traverse {
```
johnynek: move the `toList` after and we get the filtering effect

PR #319, opened 2019-08-06
Fix the scoping issue with top-level lets

### LINKED ISSUE #311: bug with incorrect shadowing of top-level bindings in Evaluation

johnynek:
close #311

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
@@ -233,8 +233,8 @@ object Evaluation {
      * scope: that which is viewable at the top level
      */
     def withConstantEnv(env: => Env): Scoped = {
-      lazy val computedEnv = env
-      fromFn { _ => inEnv(computedEnv) }
+      lazy val computed = inEnv(env)
+      fromFn { _ => computed }
     }
   }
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -230,6 +227,15 @@ object Evaluation {
...
       }
+
+    /**
+     * Top level Scoped objects have a fixed and known
+     * scope: that which is viewable at the top level
+     */
+    def withConstantEnv(env: => Env): Scoped = {
+      lazy val computedEnv = env
+      fromFn { _ => inEnv(computedEnv) }
+    }
```
johnynek, line 238: We could put the `inEnv(env)` call in the lazy val once.

PR #332, opened 2019-08-13
Fix issue 328, evaluation/typechecking inconsistency

### LINKED ISSUE #328: inconsistency in typechecking and evaluation
In typechecking, we go through the top-level definitions in order. This ensures there are no dags and matches the way a person might go through it.

When we evaluate, we take more of a graph model, and work from the bottom up. This creates in inconsistency where things typecheck, but fail at runtime, but other kinds of bugs are possible.

We should be clear what the semantics are. There are at least two questions:

1. if we shadow a Predef function later in the file, which version do we see before then? It seems like there are only two answers to this: a) error on a predef shadow, b) we see predef version. I prefer the second answer.
2. should we allow shadowing at the top level? We do in local defs, and the type-checking allows it, but such shadows are sources of bugs now.

The semantics of typechecking, to me, seem more like a human would expect. The most direct fix is probably to make those semantics the language semantics and fix evaluation.

johnynek:
fix #328 

This does to by explicitly evaluating the file in the order we type-check it, with the correct environment at each point.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
@@ -241,13 +241,12 @@ object Evaluation {
         }
       }
 
-    def recursive(name: Bindable, item: Scoped): Scoped = {
+    def recursive(name: Bindable, item: Scoped): Scoped =
       fromFn { env =>
         lazy val env1: Map[Identifier, Eval[Value]] =
           env.updated(name, Eval.defer(item.inEnv(env1)).memoize)
         item.inEnv(env1)
       }
-    }
 
     private case class Let(name: Bindable, arg: Scoped, in: Scoped) extends Scoped {
       def inEnv(env: Env) = {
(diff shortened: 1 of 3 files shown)


johnynek:
@snoble if you have time, take a look at this...

This removes using NameKind from Evaluation, and just directly matches the type-checking semantics with the evaluation: namely: type constructors can't be shadowed, we consider external defs first, then finally lets.

This also passes over the package one time to build a lazy graph of what is to be computed, rather than doing so many lookups (I think). It may be faster, but I haven't benchmarked.

core/src/main/scala/org/bykn/bosatsu/rankn/TypeEnv.scala
```diff
@@ -87,10 +87,11 @@ class TypeEnv[+A] private (
    * annotated
    */
   def addExternalValue(pack: PackageName, name: Identifier, t: Type): TypeEnv[A] =
-    new TypeEnv(constructors = constructors, definedTypes = definedTypes, values = values.updated((pack, name), t))
+    new TypeEnv(
+      constructors = constructors,
+      definedTypes = definedTypes,
+      values = values.updated((pack, name), t))
 
-  // TODO to support parameter named patterns we'd need to know the
```
johnynek, line 92: this todo is already done.

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -284,18 +270,108 @@ case class Evaluation[T](pm: PackageMap.Typed[T], externals: Externals) {
...
+      .filter(_.packageName == p.name)
+      .flatMap { dt =>
+        dt.constructors.iterator.map { case (cn, _, _) =>
+          (cn, constructor(cn, dt))
+        }
+      }
+      .toMap
+  }
+
+  private def addLet(p: Package.Typed[T], env: Env, let: (Bindable, RecursionKind, TypedExpr[T])): Env = {
```
johnynek: p is unused.

PR #333, opened 2019-08-13
Solve imports from predef.bosatsu exports

snoble:
I saw this TODO so I figured I'd do it quickly. There might be a smarter way to dedup but I'd suspect that wouldn't affect performance in a measurable way.


johnynek:
Thanks, this is great!

Issue #90, opened 2018-10-26
better errors in parser


johnynek:
The parser needs better error messages. When we fail to parse e.g. missing `)` it does not give very good messages.

One interesting law: if we make a random package and then modify a random character in it, we should either continue to parse (it was a legal change) or we should point to the position where we made the change.

This is probably hard to pass that law, but it would be awesome if we could.

johnynek:
I'm pretty sure the built-in cons-list structure of Statement/Declaration is a disaster for parsing.

If you have:
```
x = 1
y = 4
{'x' : 'x' : x }
```
it is the dict that doesn't parse, but since it doesn't parse, the lets don't parse, so the error you get is that `x=1` can't parse. This is pretty terrible.

A minimal solution would to change the parser to be something like `P[List[A] => List[A]]` where we give a function that if you pass the tail, it can make the list, so basically we just do `{ t => h :: t }` in that function. Then we make a `P[List[Declaration => Declaration]]`, finally we foldRight on that pushing the final declaration/statement into it.

johnynek:
Here's another bad example:
```
5|export [ BinNat(), toInt, , toNat, toBinNat, next, add_BinNat, times2 ]
  ^
```
Note, we should point to the `, ,` part, not to the `export`.

johnynek:
can always be better but added the notion of hints and a few PRs: #1651 #1683

PR #354, opened 2019-09-01
Remove the built-in List structure of Statement

johnynek:
This refactors a design I've not been happy with: Statement is an ADT with a built in linked list. This creates parsing complications, and also weird ambiguities in that we can never deal with a single statement (but we kind of pretend to by dealing with the head).

This fixes this.


johnynek:
this relates to the comment on the original design flaw here: https://github.com/johnynek/bosatsu/issues/90#issuecomment-480527578

PR #358, opened 2019-09-04
Improve Record Declaration Parsing

### LINKED ISSUE #357: parsing is wrong with bindings and unions
If we have
```
x | y = ...
```
We could parse that as `x | (y = ...` or as `(x | y) = ...`

We mean the latter, but the parser will parse as the former now. This is related to the backtracking issue #344 

An overhaul probably separates bind-like declarations: 
```
x = y
z

def x:
  y
z
```
both of these forms need to be at the indentation level, not at any random position. One idea would be to subclass Declaration to separate the bindings (and comments) out.

Another approach is to make `=` parsing more like an operator and get the association correct (it should bind weaker than |).

Currently things like
```
{} | x = y
z
```
will parse, but it shouldn't we should get an error that `{} | x` isn't a pattern.

johnynek:
close #357

This introduces `Declaration.NonBinding` which prevents us from writing code like:
```
x = y = z
f(y)
f(x)
```
which would have confusingly been the same as:
```
f(f(z))
```

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
@@ -422,12 +422,12 @@ object Declaration {
       case _ => None
     }
 
-  private val restP: Indy[Padding[Declaration]] =
+  private def restP(parser: Indy[Declaration]): Indy[Padding[Declaration]] =
     (Indy.parseIndent *> parser).mapF(Padding.parser(_))
 
   // This is something we check after variables
-  private lazy val bindingOp: Indy[(Pattern.Parsed, Region) => Binding] = {
-    BindingStatement.bindingParser[Pattern.Parsed, Padding[Declaration]](nonBindingParser <* Indy.lift(toEOL), restP)
+  private def bindingOp(nonBindingParser: Indy[NonBinding], p: Indy[Declaration]): Indy[(Pattern.Parsed, Region) => Binding] = {
+    BindingStatement.bindingParser[Pattern.Parsed, Padding[Declaration]](nonBindingParser <* Indy.lift(toEOL), restP(p))
       .region
       .map { case (region, fn) =>
         { (pat: Pattern.Parsed, r: Region) => Binding(fn(pat))(r + region) }
@@ -441,7 +441,7 @@ object Declaration {
 
   def defP(parser: Indy[Declaration]): Indy[DefFn] = {
     val restParser: Indy[(OptIndent[Declaration], Padding[Declaration])] =
-      OptIndent.indy(parser).product(Indy.lift(toEOL) *> restP)
+      OptIndent.indy(parser).product(Indy.lift(toEOL) *> restP(parser))
 
     restParser.mapF { rp =>
       DefStatement.parser(Pattern.bindParser, maybeSpace ~ rp)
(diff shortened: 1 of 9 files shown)


core/src/main/scala/org/bykn/bosatsu/Declaration.scala
```diff
@@ -416,20 +426,20 @@ object Declaration {
     (Indy.parseIndent *> parser).mapF(Padding.parser(_))
 
   // This is something we check after variables
-  private val bindingOp: Indy[(Pattern.Parsed, Region) => Binding] = {
-    BindingStatement.bindingParser[Pattern.Parsed, Padding[Declaration]](parser <* Indy.lift(toEOL), restP)
+  private lazy val bindingOp: Indy[(Pattern.Parsed, Region) => Binding] = {
+    BindingStatement.bindingParser[Pattern.Parsed, Padding[Declaration]](nonBindingParser <* Indy.lift(toEOL), restP)
```
johnynek: this is reaching all the way to the end. This isn't good. We should stop hiding these loops and make them explicit.

core/src/main/scala/org/bykn/bosatsu/Declaration.scala
```diff
@@ -440,10 +450,10 @@ object Declaration {
     }
   }
 
-  def ifElseP(expr: Indy[Declaration]): Indy[IfElse] = {
+  def ifElseP(arg: P[NonBinding], expr: Indy[Declaration]): Indy[IfElse] = {
 
-    def ifelif(str: String): Indy[(Declaration, OptIndent[Declaration])] =
-      Indy.block(Indy.lift(P(str ~ spaces ~ maybeSpace)) *> expr, expr)
+    def ifelif(str: String): Indy[(NonBinding, OptIndent[Declaration])] =
+      Indy.block(Indy.lift(P(str) ~ spaces ~ maybeSpace ~ arg), expr)
```
johnynek: why `spaces ~ maybeSpace`? What is going on there? Seems like `P(str ~ spaces ~ arg)` would do it.

core/src/main/scala/org/bykn/bosatsu/Declaration.scala
```diff
@@ -469,57 +479,69 @@ object Declaration {
...
 
     Indy.blockLike(params, parser, P(maybeSpace ~ "->"))
       .region
       .map { case (r, (args, body)) => Lambda(args, body.get)(r) }
   }
 
-  def matchP(expr: Indy[Declaration]): Indy[Match] = {
-    val withTrailing = expr <* Indy.lift(maybeSpace)
-    val branch = Indy.block(Indy.lift(Pattern.matchParser), withTrailing)
+  def matchP(arg: P[NonBinding], expr: Indy[Declaration]): Indy[Match] = {
```
johnynek: should be `Indy[NonBinding]` since we are not applying the indentation, and wrapped `Parens` nodes would need to know the indentation.

core/src/main/scala/org/bykn/bosatsu/Declaration.scala
```diff
@@ -645,25 +658,25 @@ object Declaration {
...
 
         // This already parses as many as it can, so we don't need repFn
         val form = ops.map { fn =>
 
-          { d: Declaration => convert(fn(d)) }
+          { d: NonBinding => convert(fn(d)) }
         }
 
-        maybeAp(applied, form)
+        NoCut(maybeAp(applied, form))
```
johnynek: this `NoCut` is due to the use of cuts inside operator parsing. This is due to ambiguity between `Foo(x) | Bar(y)` it could be an `operator |` or it could be a pattern. When we hit `Foo(x) | Bar(_, x)` we want to fail to parse that as a value because need to see that as one atomic pattern.

core/src/main/scala/org/bykn/bosatsu/Declaration.scala
```diff
@@ -673,10 +686,40 @@ object Declaration {
...
+           * will fail.
+           *
+           * If we parse a declaration first, if we see = we need to convert
+           * to pattern. If we parse a pattern, but it was actually a declaration, we need
+           * to convert there. This code tries to parse as a declaration first, then converts
+           * it to pattern if we see an =
+           */
+          decOrBind(finalNonBind, indent) |
+          //patternLike |
+          //recordConstructorP(indent, recurse) |
```
johnynek: remove the comments.

PR #359, opened 2019-09-07
Fix 265, parse type annotations

### LINKED ISSUE #265: Parse error when declaring type in assignment within function
```
def foo(x: String):
  i: Int = 2
  i
```

produces
```
failed to parse test_workspace/recordset.bosatsu at line 23, column 4
######
21|
22|def foo(x: String):
23|  i: Int = 2
      ^
```
But
```
def foo(x: String):
  (i): Int = 2
  i
```
is fine.
As is
```
i: Int = 2
```
and
```
def foo(x: String):
  (i: Int, _) = (2, 3)
  i
```

johnynek:
close #265 

This turned out to be much more complex than I thought.

The key issue is that this creates a lot of ambiguity. consider `if x:` what comes next? if we expect `x: type` it fails to parse, which when done wrong makes us fall back to parsing `if` as a variable.

or consider in `match x:` what should come next? Lastly, dictionaries use the colon as well.

My solution is to make type annotation unusable in those locations without a `(x: type)` nesting. I think this is a fine solution

I think this allows us to use type annotations all the places we expect.

cc @snoble please reopen #265 if you see this in the future.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
@@ -668,7 +668,7 @@ object Declaration {
         if (pm == ParseMode.BranchArg) applied
         else {
           val an: P[NonBinding => NonBinding] =
-            P(":" ~ maybeSpace ~ TypeRef.parser)
+            P(maybeSpace ~ ":" ~ maybeSpace ~ TypeRef.parser)
               .region
               .map { case (r, tpe) =>
                 { nb: NonBinding => Annotation(nb, tpe)(nb.region + r) }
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/bykn/bosatsu/Declaration.scala
```diff
@@ -626,18 +653,29 @@ object Declaration {
...
           }
 
         (allNonBind ~ repFn(dotApply | applySuffix)).map { case (a, f) => f(a) }
       }
+      // lower priority than calls is type annotation
+      val annotated: P[NonBinding] = {
+        if (pm == ParseMode.BranchArg) applied
+        else {
+          val an: P[NonBinding => NonBinding] =
+            P(":" ~ maybeSpace ~ TypeRef.parser)
```
johnynek: We should allow maybeSpace before the colon.

johnynek: We should be able to cut after the colon also.

PR #361, opened 2019-09-07
Refactor skolemize freevar

### LINKED ISSUE #321: handling of free type parameters in defs and lets is restrictive
I noted in #320 this issue.

When typechecking, if we see a free type variable, we make that a forall parameter. The issue is, we do this at the top level let bindings:
```
def foo(x):
  def bar(y: a): y
  ...
```
here, we would expect that `a` is a universal quantifier in `bar` only, not all of foo. But the way we handle it now, we look for all internal free type variables, and generalize at the top level.

You can work around this by not building internal defs, but that limits their use with closures.

A better approach is to improve the approach for translation of this syntax into `Expr` and properly scoping the universal quantification.

johnynek:
this was an attempt to fix #321 but I couldn't yet get there.

My idea was to not fully skolemize all free variables, but instead look for the lambda patterns and only skolemize the lambdas. This makes it much more complex to see when variables escape. I wonder if this is a misfeature...

Alternatives:

1. only do this for top level declarations.
2. require type annotations like `def foo[a](x: a)` and only quantify over the declared variables.

It seems like we should be able to infer which are free, but more inference is also a double edged sword: it makes more programs type-check that the user might not intend. I'm not sure where the balance is in this case.

skolemizing all free variables for each top-level let is certainly simpler.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Expr.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Expr.scala
@@ -6,7 +6,7 @@ package org.bykn.bosatsu
  */
 
 import cats.implicits._
-import cats.data.{Writer, NonEmptyList}
+import cats.data.{Chain, Writer, NonEmptyList}
 import cats.{Applicative, Eval, Traverse}
 import scala.collection.immutable.SortedSet
 import org.bykn.bosatsu.rankn.Type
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/bykn/bosatsu/Expr.scala
```diff
@@ -100,6 +101,44 @@ object Expr {
...
+   *
+   * def foo(x: a): x
+   *
+   * We handle this by converting a to a skolem variable,
+   * running inference, then quantifying over that skolem
+   * variable.
+   */
+  def skolemizeFreeVars[F[_]: Applicative, A](expr: Expr[A])(newSkolemTyVar: Type.Var => F[Type.Var.Skolem]): Option[F[(NonEmptyList[Type.Var.Skolem], Expr[A])]] = {
+    val w = Expr.traverseType[A, Writer[List[Type.Var.Bound], ?]](expr, { t =>
+      val frees = Type.freeBoundTyVars(t :: Nil)
```
johnynek: Use Chain rather than List for these writer cases.

PR #364, opened 2019-09-10
Add a normalize method to TypedExpr

johnynek:
related to #118 and #200


core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -264,11 +264,6 @@ case class Evaluation[T](pm: PackageMap.Typed[T], externals: Externals) {
     tpe: Type,
     branches: NonEmptyList[(Pattern[(PackageName, Constructor), Type], TypedExpr[T])],
     recurse: Ref => Scoped): (Value, Env) => Eval[Value] = {
-      val dtConst@Type.TyConst(Type.Const.Defined(pn0, tn)) =
-        Type.rootConst(tpe).getOrElse(sys.error(s"failure to get type: $tpe")) // this is safe because it has type checked
-
-      val packageForType = pm.toMap(pn0)
```
johnynek: this was unused and can fail due to the sys error.

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -508,28 +508,12 @@ case class Evaluation[T](pm: PackageMap.Typed[T], externals: Externals) {
...
-
-             efn.applyArg(earg)
-         }
+         val efn = recurse(fn)
+         val earg = recurse(arg)
+
+         efn.applyArg(earg)
        case AnnotatedLambda(name, _, expr, _) =>
-         expr match {
-           case App(f, Var(None, name2, _, _), _, _) if name2 == name =>
```
johnynek, line 526: this rule is too simplistic... the TypedExpr.normalization code has a better implementation that (hopefully) is bug-free.

The problem is `\x -> f(x)(x)` you can't replace this with `f(x)` since now `x` is unbound.

johnynek:
cc @snoble here are some interesting optimizations/normalizations... I think I can go further but these seem good enough to enable all the time.

PR #366, opened 2019-09-16
Add a Javascript API and simple elm app using it

johnynek:
This makes a simple in-browser version of the compiler.

The UI/UX is terrible (see below). It starts off way too small and it the error messages are very difficult to read but I will leave that to a future work.

Screenshot 2019-09-15 at 14 27 12

cc @snoble

### DIFF
--- a/elmui/src/Main.elm
+++ b/elmui/src/Main.elm
@@ -1,9 +1,9 @@
 port module Main exposing (..)
 
 import Browser
-import Html exposing (Html, Attribute, button, div, input, text, textarea)
+import Html exposing (Attribute, Html, button, div, input, text, textarea)
 import Html.Attributes exposing (..)
-import Html.Events exposing (onInput, onClick)
+import Html.Events exposing (onClick, onInput)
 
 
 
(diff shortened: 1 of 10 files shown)


elmui/src/Main.elm
```diff
@@ -0,0 +1,62 @@
...
+
+main =
+  Browser.element { init = init, subscriptions = subs, update = update, view = view  }
+
+-- MODEL
+
+type alias Model =
+  { code : String,
+    result: String
+  }
```
snoble: Did you run this through elm-format?

elmui/src/Main.elm
```diff
@@ -0,0 +1,62 @@
+port module Main exposing (..)
+
+import Browser
+import Html exposing (Html, Attribute, button, div, input, text, textarea)
+import Html.Attributes exposing (..)
```
snoble, line 5: You could tighten up `..`

elmui/src/Main.elm
johnynek: I didn't. I just downloaded it and did it now. Thanks for looking at this and the suggestion!

elmui/src/Main.elm
johnynek: I'm going to leave it for now since this is such a skeleton. I'll hack on it more. I'm thinking of trying: http://elm-bootstrap.info/ to make it fancier. Have you tried that library?

What do you recommend for better looking UIs?

snoble: That's what I've used. In particular Bootstrap.Grid and Bootstrap.Tab were helpful.

johnynek:
This slightly lowers coverage, but I’m going to merge since I’m still iterating on JsApi.

Issue #318, opened 2019-08-04
add a non-type-checked evaluation mode


johnynek:
We could also run bosatsu without any typechecking since we convert to Expr before typechecking. We could just put in incorrect types and pass to Evaluation which doesn't use types.

This could be useful for teaching syntax before getting into typechecking since the errors would be more about data than about types, which can be harder to grasp.

Also, since types are always optional, without typechecking we basically get a kind of a functional python language which could be a nice language for learning about programming with immutable values.

johnynek:
We can't do this anymore since Matchless runs on TypedExpr. Also, maybe not a good idea.

PR #372, opened 2019-10-05
Simplify evaluation of match

johnynek:
I hit a bug in #364 that exposed that we are not using type information at all in matching, contrary to what the code was implying (it was accessing it, but ignoring it).

Removing it removes one case where we had sys.error, which is valuable.


johnynek:
this seems to open the door to solving #318 since evaluation shouldn't require type information now (in fact, it should probably depend only on `Expr[A]` not `TypedExpr[A]`.

PR #375, opened 2019-10-19
Support implicit package names

### LINKED ISSUE #10: Add implicit package names
for many DSLs we might want to have implicit package names, either based on directory name (a-la bazel's BUILD files) or by full file name.

This could be an option of a tool to automatically assign packages relative to a bosatsu root.

johnynek:
closes #10 

This allows the option of passing a list of package_roots, which if there is no explicit `package` in a file, we check if the given path is in the root in the order they are given.

This allows us to have implicit package names in the web UI, but could also be very useful in applications like bazel where you don't want to have to declare a redundant package name.

In Path based examples, we pass the relative path to `toLowerCase.capitalize` in an attempt to make it a valid package name.

### DIFF
--- a/cli/src/main/scala/org/bykn/bosatsu/Main.scala
+++ b/cli/src/main/scala/org/bykn/bosatsu/Main.scala
@@ -91,6 +91,34 @@ object PathModule extends MainModule[IO] {
 
         (ifres *> out)
     }
+
+  def pathPackage(roots: List[Path], packFile: Path): Option[PackageName] = {
+    import scala.collection.JavaConverters._
+
+    def getP(p: Path): Option[PackageName] = {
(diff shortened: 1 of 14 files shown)


cli/src/main/scala/org/bykn/bosatsu/Main.scala
```diff
@@ -91,6 +91,35 @@ object PathModule extends MainModule[IO] {
 
         (ifres *> out)
     }
+
+  def pathPackage(roots: List[Path], packFile: Path): Option[PackageName] = {
+    import scala.collection.JavaConverters._
+
+    def getP(p: Path): Option[PackageName] = {
```
johnynek, line 98: We should remove any file extension that exists.

Issue #344, opened 2019-08-18
Reduce/Remove backtracking in parser


johnynek:
see #342 for more discussion.

We should try to remove backtracking as much as possible. Cases where we currently have it:

1. binding parsing in Declaration. There is an ambiguity in some cases between declarations and patterns that can appear on the left.
2. ternary if parsing in declaration vs filtering in list comprehension (latter only has if, no else)

There are probably other cases.

johnynek:
If we had `Declaration.fromPattern(p: Pattern.Parsed): Option[Declaration]` then we could remove the backtracking in pattern parsing.

This would probably also improve performance.

To implement this function, we'd need regions on Patterns because Declarations have them. see #132

PR #381, opened 2019-10-31
Improve parsing errors somewhat

johnynek:
related to #90 

I hit a case where the parse error was pointing well before the real error.


johnynek:
relates to #344

PR #387, opened 2019-11-10
Improve Evaluation

### LINKED ISSUE #331: Optimize Nat-like data types
see: https://github.com/johnynek/bosatsu/pull/327/files#diff-2326946f266223993c3f0489f4c2304dR137

in #327

Nat like data types are very nice for iterative recursions. The problem is they take O(N) space to store a number N. An idea (I think Idris does) is for the compiler to identify Nat-like data types: `enum T: Z, S(x: T)` and represent them as integers. The only thing you can do with enums is match them, so the match on a type like this just becomes an `if (x == 0) ... else ` check. With this optimization, Nat-like types take O(log N) space

A related optimization is to represent fixed enums: `enum A: B, C, D` as a bounded integer (if 2, you can use a Boolean, if <= 256 we can use a Byte, ...). A generalization of this idea could encompass both ideas.

johnynek:
close #331 

This optimizes evaluation for a few cases:
1. unit like types can do less work
2. enums that are like ints are memoized for fewer allocations
3. when we have a struct match that doesn't do any binding (`Foo(_, _)`) we simpilfy things
4. we optimize the representation of Nat-like types so they are stored by a BigInteger, not a linear-sized linked list
5. new-types don't add a layer of wrapping.

cc @snoble

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
@@ -71,9 +71,11 @@ object Evaluation {
 
     def recursive(name: Bindable, item: Scoped): Scoped =
       fromFn { env =>
+        lazy val res: Eval[Value] =
+          Eval.defer(item.inEnv(env1)).memoize
         lazy val env1: Map[Identifier, Eval[Value]] =
-          env.updated(name, Eval.defer(item.inEnv(env1)).memoize)
-        item.inEnv(env1)
+          env.updated(name, res)
+        res
       }
 
     private case class Let(name: Bindable, arg: Scoped, in: Scoped) extends Scoped {
--- a/core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
@@ -258,6 +258,25 @@ case class Evaluation[T](pm: PackageMap.Typed[T], externals: Externals) {
       }
     }
 
+  private[this] val someTrue: Option[Boolean] = Some(true)
+  private[this] val someFalse: Option[Boolean] = Some(false)
+  // return None if the type isn't Nat-like (Zero, Succ(n))
+  // return Some(true) if Zero comes first
+  // return Some(false) if Zero is second
+  private def getNatLikeFirst(dt: DefinedType[Any]): Option[Boolean] =
+    if (dt.annotatedTypeParams.isEmpty && (dt.constructors.lengthCompare(2) == 0)) {
+      dt.constructors match {
+        case (_, Nil, _) :: (_, (_, t) :: Nil, _) :: _ =>
+          if (t == dt.toTypeTyConst) someTrue
+          else None
+        case (_, (_, t) :: Nil, _) :: (_, Nil, _) :: _ =>
+          if (t == dt.toTypeTyConst) someFalse
+          else None
+        case _ => None
+      }
+    }
+    else None
+
   private type Ref = TypedExpr[T]
 
   private def evalBranch(
@@ -369,9 +388,9 @@ case class Evaluation[T](pm: PackageMap.Typed[T], externals: Externals) {
                 }
             }
           case Pattern.Annotation(p, _) =>
-            // TODO we may need to use the type here
+            // we don't need types to match patterns once we have fully resolved constructors
             maybeBind(p)
-          case Pattern.Union(h, t) =>
+          case u@Pattern.Union(_, _) =>
             // we can just loop expanding these out:
             def loop(ps: List[Pattern[(PackageName, Constructor), Type]]): (Value, Env) => Option[Env] =
               ps match {
@@ -381,13 +400,12 @@ case class Evaluation[T](pm: PackageMap.Typed[T], externals: Externals) {
                   val fnt = loop(tail)
 
                   { (arg, acc) =>
-                    fnh(arg, acc) match {
-                      case None => fnt(arg, acc)
-                      case some => some
-                    }
+                    val rh = fnh(arg, acc)
+                    if (rh.isDefined) rh
+                    else fnt(arg, acc)
                   }
               }
-            loop(h :: t.toList)
+            loop(Pattern.flatten(u).toList)
           case Pattern.PositionalStruct(pc@(_, ctor), items) =>
             /*
              * The type in question is not the outer dt, but the type associated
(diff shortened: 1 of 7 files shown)


core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -258,6 +267,25 @@ case class Evaluation[T](pm: PackageMap.Typed[T], externals: Externals) {
       }
     }
 
+  private[this] val someTrue: Option[Boolean] = Some(true)
+  private[this] val someFalse: Option[Boolean] = Some(false)
```
snoble, line 271: What are the benefits of re-using the same instances rather than just constructing in the function? Maybe worth commenting on since it's a bit surprising.

snoble: also, you could rename `someTrue` to something like `zeroComesFirst` and `someFalse` to `zeroIsSecond`

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -396,6 +437,8 @@ case class Evaluation[T](pm: PackageMap.Typed[T], externals: Externals) {
             val dt = definedForCons(pc)
             val itemFns = items.map(maybeBind(_))
 
+            val itemsWild = items.forall(isTotalNonBinding(_))
+
             def processArgs(as: List[Value], acc: Env): Option[Env] = {
               // manually write out foldM hoping for performance improvements
```
snoble, line 443: I know this isn't new but it occurred to me it might help readability to have the non-optimal `fold` as a comment

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -396,6 +437,8 @@ case class Evaluation[T](pm: PackageMap.Typed[T], externals: Externals) {
             val dt = definedForCons(pc)
             val itemFns = items.map(maybeBind(_))
 
+            val itemsWild = items.forall(isTotalNonBinding(_))
```
snoble, line 440: do you need `(_)` here?

core/src/main/scala/org/bykn/bosatsu/rankn/DefinedType.scala
```diff
@@ -23,15 +23,26 @@ case class DefinedType[+A](
    * A type with exactly one constructor is a struct
    */
   def isStruct: Boolean = constructors.lengthCompare(1) == 0
+
+  /**
+   * A newtype is just a wrapper for another type.
+   * It could be removed statically from the program
+   */
+  def isNewType: Boolean =
```
snoble, line 31: Could this be a `val` (or `lazy val`) instead of `def`?

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
johnynek: Allocate once rather than once per call to the method.

core/src/main/scala/org/bykn/bosatsu/rankn/DefinedType.scala
johnynek: It could be. Might as well. Val is probably better.

core/src/main/scala/org/bykn/bosatsu/rankn/DefinedType.scala
```diff
@@ -23,15 +23,26 @@ case class DefinedType[+A](
...
+    constructors match {
+      case (_, _ :: Nil, _) :: Nil => true
```
snoble, line 33: I've seen pattern matches like `(_, _ :: Nil, _)` a few times in this code and it makes me wonder if `(Constructor, List[(Bindable, Type)], Type)` is a little too complicated of a tuple. That it might be a bit more self documenting if you had a case class with named methods.

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -415,39 +458,96 @@ case class Evaluation[T](pm: PackageMap.Typed[T], externals: Externals) {
...
-                    // $COVERAGE-OFF$this should be unreachable
-                    val itemStr = items.mkString("(", ", ", ")")
-                    sys.error(s"ill typed in match (${ctor.asString}$itemStr\n\n$other\n\nenv: $acc")
-                    // $COVERAGE-ON$
+            if (dt.isNewType) {
+              // we erase this entirely, and just match the underlying item
+              itemFns.head
+            }
+            else if (dt.isStruct) {
+              if (itemsWild) noop
```
snoble, line 466: possible further optimization is `itemsWild` could be a list of newly bound identifiers and you could check if they're actually referenced

johnynek: I’d rather do this optimization in the compiler when we normalize expressions.

I can see an argument for this being a compiler error actually.

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -415,39 +458,96 @@ case class Evaluation[T](pm: PackageMap.Typed[T], externals: Externals) {
...
+                        case other =>
+                          // $COVERAGE-OFF$this should be unreachable
+                          val itemStr = items.mkString("(", ", ", ")")
+                          sys.error(s"ill typed in match (${ctor.asString}$itemStr\n\n$other\n\nenv: $acc")
+                          // $COVERAGE-ON$
+                      }
+                    }
+                  else {
+                    // We are expecting non-zero
+                    val succMatch = itemFns.head
```
snoble, line 509: You use `itemFns.head` twice. Is it possible to get the type system to ensure this is non-empty? Maybe this changes too many things but perhaps a `Pattern.PositionalStruct` with no items should be handled separately?

snoble:
This is really cool!

2 general thoughts on things to consider in the future:

1. I think this could be generalized to any enum where all the constructors had either zero args or one arg of a self contained type. eg `enum Nat2: Zero1, Zero2, Succ1(prev: Nat2), Succ2(prev: Nat2)`. Instead of having this encoded as a `BigInteger` you'd encode it has `(Int, List[(Int, BigInteger)])` where the `Int` in the `List` would be the index of the Succ constructor and the `Int` in the outer tuple would be the index of the `Zero`. I can't figure out if this models anything though so it might be a useless generalization.

2. It feels a little odd having this in Evaluation and not having an Optimization step where this would live. I don't thiiiink any of this is dependent on input data. I think this would be desirable eventually because it makes it easier to re-implement the evaluator to be run in other environments.

johnynek:
regarding 1: I think you can represent all the types you describe as a single integer. regarding 2. I agree it would be unfortunate to have to rewrite this step in every back end, but right now there is only evaluation anyway, and the internal lambda calculus has no way to represent this optimization. So, maybe we need another lower intermediate representation that is close to machine code to do that.

I took your suggestion about removing the tuple 3 to represent constructor functions, but the diff got really big, so I will make that a follow up PR. The other changes also got swept into that.

Thanks for the review!

PR #389, opened 2019-11-12
Make ConstructorFn a type

johnynek:
addresses review comments @snoble left in #387

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Normalization.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Normalization.scala
@@ -587,7 +587,7 @@ case class NormalizePackageMap(pm: PackageMap.Inferred) {
         case Pattern.Annotation(p, tpe) => loop(p)
         case Pattern.PositionalStruct(pc@(_, ctor), params) =>
           val dt = definedForCons(pc)
-          val name = if (dt.isStruct) None else Some(dt.constructors.map(_._1).indexOf(ctor))
+          val name = if (dt.isStruct) None else Some(dt.constructors.map(_.name).indexOf(ctor))
           NormalPattern.PositionalStruct(name, params.map(loop(_)))
         case Pattern.Union(h, t) => NormalPattern.Union(loop(h), t.map(loop(_)))
       }
@@ -705,7 +705,7 @@ case class NormalizePackageMap(pm: PackageMap.Inferred) {
         .toList
         .iterator
         .zipWithIndex
-        .collectFirst { case ((ctor, params, resType), idx) if ctor == c => (idx, params.size) }
+        .collectFirst { case (cf, idx) if cf.name == c => (idx, cf.args.size) }
         .get
 
       def loop(params: Int, expr: NormalExpression): NormalExpression =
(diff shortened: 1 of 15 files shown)


core/src/main/scala/org/bykn/bosatsu/Normalization.scala
```diff
@@ -587,7 +587,7 @@ case class NormalizePackageMap(pm: PackageMap.Inferred) {
         case Pattern.Annotation(p, tpe) => loop(p)
         case Pattern.PositionalStruct(pc@(_, ctor), params) =>
           val dt = definedForCons(pc)
-          val name = if (dt.isStruct) None else Some(dt.constructors.map(_._1).indexOf(ctor))
+          val name = if (dt.isStruct) None else Some(dt.constructors.map(_.name).indexOf(ctor))
```
johnynek: `indexWhere(_.name == ctor)`

core/src/main/scala/org/bykn/bosatsu/rankn/DefinedType.scala
```diff
@@ -22,15 +20,15 @@ final case class DefinedType[+A](
...
    */
-  def isStruct: Boolean = constructors.lengthCompare(1) == 0
+  val isStruct: Boolean = constructors.lengthCompare(1) == 0
 
   /**
    * A newtype is just a wrapper for another type.
    * It could be removed statically from the program
    */
-  def isNewType: Boolean =
+  val isNewType: Boolean =
```
snoble, line 29: I wonder if it might be nice to make this an `Option[ConstructorFn]`? So if `constructors` has the properties you want then it'll be returned for you to use.

core/src/main/scala/org/bykn/bosatsu/rankn/DefinedType.scala
```diff
@@ -44,29 +42,37 @@ final case class DefinedType[+A](
...
+  object NatLike {
+    final case object Not extends NatLike
+    final case class Is(val zeroFirst: Boolean) extends NatLike {
+      @inline def zeroLast: Boolean = !zeroFirst
+      def idxIsZero(idx: Int): Boolean =
+        (zeroFirst && (idx == 0)) || (zeroLast && (idx == 1))
+    }
+    val ZeroFirst: Is = Is(true)
+    val ZeroLast: Is = Is(false)
+  }
```
snoble, line 71: I'm going to have to remember this pattern. It's really cool.

snoble:
This looks like it must have been super annoying to do but I think the result looks like a really nice readability improvement. :+1:

core/src/main/scala/org/bykn/bosatsu/rankn/DefinedType.scala
johnynek: I'm going to punt on this. I think what we really want is something like:

```scala
def asNewType: Option[NewTypeConstructorFn]
```
which has exactly one constructor function. To do that without duplication you might have:
```scala
case class ConstructorFn[F[_]](name: Constructor, args: F[(Bindable, Type)], fnType: Type)
```
Then we can have `ConstructorFn[List]` for the generic case or `ConstructorFn[cats.Id]` for the new-type case....

But at the end of the day, you are guarding against a small set of bugs that we already have tests for and complicating all uses of the code. So, I think the cost/benefit trade-off is a bit weak at the moment.

PR #392, opened 2019-11-17
Improve the usability of the CLI

johnynek:
This was plane code done to improve the usability of the CLI:

1. compiling bosatsu with native-image makes the CLI much nicer to use, so it would be nice not to have to use bazel for basic things, without compromising the correctness of the bazel operation.
2. for a python-like experience, it is nice to run a file, and have a `--search` mode for to automatically look for files based on package name. `--search` should probably be the default when you give a package_root, so I think that should be changed to `--no-search` to disable.
3. colorization flag should work for `test`.
4. test should find the last test value, not the last value. This is more general, so shouldn't break anything. By doing this you can do both tests and json for a single file.
5. `write-json` should have an optional file output. Without a file output, write to standard out
6. don't duplicate the test suite summary if there is only one suite and all the items are in that suite.
7. add a script to build the native-image bosatsu, add a script to run the assembly

After you build the native-image, you can do, for example
```
./bosatsu test --package_root ex --search --test_file ex/Nat.bosatsu
```
And it will instantly compile and run the tests in ex/Nat.bosatsu (including compiling any dependencies that are placed in the correct locations `package Foo/Bar` should be at `root/Foo/Bar.bosatsu`

cc @snoble take a look at this.... it's amazing how much more fun it is.

### DIFF
--- a/bench/src/main/scala/org/bykn/bosatsu/TestBench.scala
+++ b/bench/src/main/scala/org/bykn/bosatsu/TestBench.scala
@@ -21,7 +21,7 @@ class TestBench {
       case Validated.Valid(vs) => vs
       case Validated.Invalid(errs) =>
         errs.toList.foreach { p =>
-          p.showContext(LocationMap.Colorize.none).foreach(System.err.println)
+          p.showContext(LocationMap.Colorize.None).foreach(System.err.println)
         }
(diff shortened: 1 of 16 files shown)


johnynek:
This really increased the coverage (the main function is covered, but by bazel build which isn't instrumented). Going to merge this.

Issue #394, opened 2019-11-17
improve CLI


johnynek:
see #392 

One obvious thing I haven't done much work on is that it is much easier to get started if bosatsu CLI is usable without other tools or writing build files. I still believe that bazel integration keeps you honest in that it makes sure it is easy to do tooling, but it should also be easy to do json generation with modules and tests just using `bosatsu`.

Features missing:

1. ability to add all sources in a directory, both recursively and not. I think it is better to be explicit than allow the arg for including sources to automatically guess. Maybe `--input_dir` and `--input_all_under` or something.
2. Similarly for including a set of inferfaces or compiled files.
3. ability to print dependencies of files (what is the file dag, what is the package dag, what packages are in which files).
4. ability to inspect proto files: are they interfaces, compiled files, ability to combine and split them out, ability to make an empty interface file (useful when writing compositions).

johnynek:
pretty print an interface to see all full types of all the values inside.

PR #397, opened 2019-11-20
Add support for input directories in addition to files

johnynek:
This adds support for both 1-level of recursion on directories and infinite recursion on directories for inputs. This is designed to make using the compiler stand-alone nicer and workable for most users.

This is a follow up to #394


johnynek:
The coverage has increased, the CLI stuff is least coveraged by measured coverage so as long as we are still improving, I think that's fine.

PR #398, opened 2019-11-20
Run type inference in parallel

johnynek:
One nice benefit to a strict functional style is that is it fairly easy to add parallelism. This PR runs the typechecking of a complete PackageMap in parallel using the default ExecutionContext.global, which will use as many cores as the machine has.

This gives about a 2x speed up when running all the tests as a batch in the native-image CLI (meaning we can parse and run all tests in wall-clock time of 600ms on my machine).

A future work could allow the user to control the number of threads in the threadpool or disable parallelism entirely (but that is left to a future work as needed).

This also removes the dependency on dagon which we were only using to get memoize and since we have need of custom variants of that, might as well use our own code for that.


core/src/main/scala/org/bykn/bosatsu/DirectEC.scala
```diff
@@ -0,0 +1,11 @@
+package org.bykn.bosatsu
+
+import scala.concurrent.ExecutionContext
+
+object DirectEC {
+  implicit val directEC: ExecutionContext =
+      new ExecutionContext {
```
johnynek: Weird indentation.

core/.jvm/src/main/scala/org/bykn/bosatsu/Par.scala
```diff
@@ -0,0 +1,37 @@
...
+import scala.concurrent.duration.Duration
+
+/**
+ * This is an abstraction to handle parallel computation, not effectful
+ * computation. It is used in places where we have parallelism in expensive
+ * operations. Since scalajs cannot handle this, we use conditional build
+ * to replace the scalajs with just running directly
+ */
+object Par {
+  type F[+A] = Future[A]
```
johnynek: This doesn’t need to be covariant and Id isn’t so we should mismatch.

core/src/main/scala/org/bykn/bosatsu/Declaration.scala
```diff
@@ -598,7 +598,7 @@ object Declaration {
    * we also parse Bind, Def, Comment
    */
   private[this] val parserCache: ((ParseMode, String)) => P[Declaration] =
-    Memoize.function[(ParseMode, String), P[Declaration]] { case ((pm, indent), rec) =>
+    Memoize.memoizeDagHashed[(ParseMode, String), P[Declaration]] { case ((pm, indent), rec) =>
```
johnynek: This has to be thread safe because the parseCache is a static value and exposed by the Indy.

(bugs with this can be seen my doing parsing in parallel, which isn’t really needed).

core/src/main/scala/org/bykn/bosatsu/graph/Memoize.scala
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * This memoizes using a hash map in a non-threadsafe manner
+   * this throws if you don't have a dag
+   */
+  def memoizeDagHashed[A, B](fn: (A, A => B) => B): A => B = {
+    // this should be volatile if fn uses threads, but
+    // we are only using it below and can see that isn't the case
+    var cache = Map.empty[A, Option[B]]
+
+    new Function[A, B] { self =>
```
johnynek, line 39: Can share an implementation that has a private constructor for the initial map.

core/src/main/scala/org/bykn/bosatsu/Evaluation.scala
```diff
@@ -623,7 +623,7 @@ case class Evaluation[T](pm: PackageMap.Typed[T], externals: Externals) {
    * that names resolve
    */
   private[this] val eval: Ref => Scoped =
-    Memoize.function[Ref, Scoped] {
+    Memoize.memoizeDagHashed[Ref, Scoped] {
```
johnynek: this needs to be thread safe because it is on the object and can be seen by many threads.

PR #402, opened 2019-11-22
Add some List functions

johnynek:
Adds some standard list functions to exercise the language and explore patterns.

I like the approach of using:
```
def zip_tests:
  t1 = match ...

```
to do three things:
1. use a def with no args to set up a limited scope.
2. inside make a TestSuite to group the assertions
3. use match to get easier writing of tests.


johnynek:
really frustrating how small changes in coverage due to scalacheck cause the coverage to fail.

PR #422, opened 2019-11-24
Make unused bindings an error

### LINKED ISSUE #419: Error on unused bindings
One principle bosatsu has is that compiler flags should not change how a file compiles (package_root violates that but let’s set that aside).

In a pure language a binding that is never used is always useless. It is either done for documentation/expository purposes or it is an error. Since documentation can be done with either comments or unnamed bindings:
```
_: T = foo(a)
```
That can be legal and do real work at typechecking but be 100% removed before execution.

This also improves normalization because we error on unused bindings.

Note, this should also apply to patterns.

johnynek:
close #419 

This makes unused bindings an error. This is purely a change to make it easier to write correct code.

We cannot remove the unused let normalization because other changes in the graph can make a let unused. This is only a naive check that a variable is literally never used. It may still be removable by optimizations and normalizations.

Our goal here is simply to error on cases where trivial syntactic analysis we should something is unused.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/UnusedLetCheck.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/UnusedLetCheck.scala
@@ -4,78 +4,77 @@ import cats.Applicative
 import cats.data.{Chain, Validated, ValidatedNec, Writer, NonEmptyChain}
 import cats.implicits._
 
+import Expr._
 import Identifier.Bindable
 
 object UnusedLetCheck {
-  def check[A: HasRegion](e: Expr[A]): ValidatedNec[(Bindable, Region), Unit] = {
-    import Expr._
 
-    val ap = Applicative[Writer[Chain[(Bindable, Region)], ?]]
-    val empty: Writer[Chain[(Bindable, Region)], Set[Bindable]] = ap.pure(Set.empty)
+  private[this] val ap = Applicative[Writer[Chain[(Bindable, Region)], ?]]
+  private[this] val empty: Writer[Chain[(Bindable, Region)], Set[Bindable]] = ap.pure(Set.empty)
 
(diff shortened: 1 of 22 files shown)


test_workspace/BinNat.bosatsu
```diff
@@ -137,7 +137,6 @@ def fold_left_BinNat(fn: a -> BinNat -> a, init: a, cnt: BinNat) -> a:
 
 # fibonacci using the fuel pattern
 def fib(b: BinNat) -> BinNat:
-    bnat = toNat(b)
```
johnynek, line 140: this was a real waste in a non-lazy language. Yeah, for checks!

core/src/main/scala/org/bykn/bosatsu/UnusedLetCheck.scala
```diff
@@ -0,0 +1,80 @@
...
+import Identifier.Bindable
+
+object UnusedLetCheck {
+
+  private[this] val ap = Applicative[Writer[Chain[(Bindable, Region)], ?]]
+  private[this] val empty: Writer[Chain[(Bindable, Region)], Set[Bindable]] = ap.pure(Set.empty)
+
+  private[this] def checkArg(arg: Bindable, reg: => Region, w: Writer[Chain[(Bindable, Region)], Set[Bindable]]) =
+    w.flatMap { free =>
+      if (free(arg)) ap.pure(free - arg)
```
johnynek: This is not correctly handling shadowing. Just because you are binding X here, it doesn't mean it wasn't already bound. So, `free - arg` isn't correct, indeed, we unless we are tracking top down scope at the same time, this won't work.

johnynek: we need to add a test that exercises this failing to catch an unused variable, and show that it is broken before continuing. Because on second though, since we are returning the free variables, we do know for sure that arg isn't free if we are binding it, so this may actually be correct.

PR #424, opened 2019-11-26
Allow newline after =, allow any file trailing whitespace

johnynek:
I accidentally had trailing whitespace in a file and got a parse error, which was not great. Trailing whitespace isn't awesome, but it shouldn't cause a parse failure.

Also, this allows us to use `=` to make a block simliar to `:` so you can do:
```
x = Foo(some, really, long, call)

# instead:
x = 
    Foo(
       some,
       really,
       long,
       call,
    )
```
as long as there is indentation after the `=` on the next line.


johnynek:
related to #86 #90

Issue #412, opened 2019-11-23
support reading json as well as writing


johnynek:
right now, the `write-json` mode converts a type `T` to Json by some rules and failing for function types if they are found.

We could also imagine a mode that takes `A -> B` and accepts json, converts it to `A` runs the expression, and then emits `B` as json.

This allows bosatsu CLI to a json transformation engine and dramatically expands the power of the json mode at relatively little cost.

PR #425, opened 2019-11-27
Support reading json to Values

johnynek:
This is part of #412 

This improves the existing code but also does decoding and adds tests.

This is not yet wired into the CLI.


johnynek:
still need to fix Json -> Value using the same strategy the latest commit uses to handle circular types.

PR #427, opened 2019-11-28
Support applying functions as json -> json transformations.

### LINKED ISSUE #412: support reading json as well as writing

johnynek:
close #412

### DIFF
--- a/bosatsuj
+++ b/bosatsuj
@@ -1,4 +1,4 @@
 #!/bin/sh
 
 # make sure to run sbt cli/assembly first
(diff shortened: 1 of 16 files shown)


johnynek:
follow on to #425

johnynek:
we're increasing coverage and the uncovered stuff seems almost completely in error messages (which is where most of our lack of coverage is).

PR #428, opened 2019-11-30
Support for string interpolation and matching

### LINKED ISSUE #426: Support string interpolation
Here is the python version:

https://www.python.org/dev/peps/pep-0498/

To me a more minimal thing would be to make `${x}` in a string always splice in the expression x which should be a string.

Similarly, we could do pattern matching:
```
match “hello bosatsu”:
  “hello ${name}”: “goodbye ${name}”
   _: “who knows”
```

If you want to use a literal `${` you would need `\${`

An alternative is to allow just `$name` if there is only a binding name at the expense of having to always escape $ not just when followed by { — to me that seems like a bad trade if we want always on interpolation.

The other bummer is that without typeclasses we will need to manually convert to strings. But that’s the typeclasses-free lifestyle.

johnynek:
closes #426 

Since pattern and declaration parsing are related, it is not easy to add part of this, so this PR adds everything, expressions for interpolation as well as pattern matching for pattern strings.

Along the way, I improved my understanding of the set algebra of sequences so I think we can leverage this to check totality when we have multiple variable patterns in lists as well, which is probably complex to reason about, but at least doesn't create a difference with string patterns.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
@@ -113,7 +113,21 @@ sealed abstract class Declaration {
       case TupleCons(items) =>
         Doc.char('(') + Doc.intercalate(Doc.text(", "),
           items.map(_.toDoc)) + Doc.char(')')
-      case Var(name) => Document[Identifier].document(name)
+      case Var(name) =>
+        Document[Identifier].document(name)
+
+      case StringDecl(parts) =>
+        val useDouble = parts.exists {
+          case Right((_, str)) => str.contains('\'') && !str.contains('"')
+          case Left(_) => false
+        }
+        val q = if (useDouble) '"' else '\''
+        val inner = Doc.intercalate(Doc.empty,
+          parts.toList.map {
+            case Right((_, str)) => Doc.text(StringUtil.escape(q, str))
+            case Left(decl) => Doc.text("${") + decl.toDoc + Doc.char('}')
+          })
+        Doc.char(q) + inner + Doc.char(q)
 
       case ListDecl(list) =>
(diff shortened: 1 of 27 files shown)


core/src/main/scala/org/bykn/bosatsu/Declaration.scala
```diff
@@ -368,6 +444,10 @@ object Declaration {
    * Foo { bar: 12 }
    */
   case class RecordConstructor(cons: Constructor, arg: NonEmptyList[RecordArg])(implicit val region: Region) extends NonBinding
+  /**
+   * This represents interpolated strings
+   */
+  case class StringDecl(items: NonEmptyList[Either[NonBinding, (Region, String)]])(implicit val region: Region) extends NonBinding
```
johnynek, line 455: we have been using the trick that extra param groups on case classes don't count on `equals`. Embedding this region here on the string complicates things quite a bit...

Of course, maybe reusing Literal would sidestep it, but in any case I added replaceRegion to help make equality ignoring regions.

core/src/main/scala/org/bykn/bosatsu/Declaration.scala
```diff
@@ -395,6 +475,8 @@ object Declaration {
           Pattern.StructKind.Named(nm, Pattern.StructKind.Style.TupleLike), Nil))
       case Var(v: Bindable) => Some(Pattern.Var(v))
       case Literal(lit) => Some(Pattern.Literal(lit))
+      case StringDecl(NonEmptyList(Right((_, s)), Nil)) =>
+        Some(Pattern.Literal(Lit.Str(s)))
```
johnynek, line 484: we need to be more careful, if we generate strings with patterns inside, they could get parsed as a string via Pattern parsing, so we probably need Pattern support to pass the tests.

PR #436, opened 2019-12-03
Improve pattern checking

### LINKED ISSUE #433: error on unreachable code in matches
we are not currently erroring if you have unreachable code:
```
match x:
  _: "foo"
  _: "bar"
```
you can clearly never reach `"bar"` but that is not an error now.

johnynek:
close #433 

also closes a TODO to catch some bad patterns.

Previously, we were repeatedly validating patterns as we were computing the intersections and differences. This make a change to do one pass to validate and check everything, then compute the missing and unreachable patterns after that. This should be both more efficient as well as catching errors we were not catching before.

### DIFF
--- a/test_workspace/TreeList.bosatsu
+++ b/test_workspace/TreeList.bosatsu
@@ -41,7 +41,6 @@ def cons(head: a, TreeList(tail): TreeList[a]) -> TreeList[a]:
       # illegal state, we can just return empty
       empty
     [branch @ Branch(_, _, _, _)]: TreeList([Single(head), branch])
-    [Branch(_, _, _, _), Single(_), *_]: empty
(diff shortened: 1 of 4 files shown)


test_workspace/TreeList.bosatsu
```diff
@@ -41,7 +41,6 @@ def cons(head: a, TreeList(tail): TreeList[a]) -> TreeList[a]:
       # illegal state, we can just return empty
       empty
     [branch @ Branch(_, _, _, _)]: TreeList([Single(head), branch])
-    [Branch(_, _, _, _), Single(_), *_]: empty
```
johnynek, line 44: real duplicate match found!

Issue #396, opened 2019-11-18
`matches` syntax idea


johnynek:
We don't have type-classes in bosatsu, so it hasn't been clear how to implement a convenient equality that is also sound.

One idea would be to leverage pattern matching. Since pattern matching is such a heavy part of bosatsu we could say that:

```
x == y
```
is the same as:
```
match x:
  y: True
  _: False
```
or y is not a valid pattern with no bindings, then:
```
match y:
  x: True
  _: False
```
if neither x nor y are valid patterns with no bindings in them, this is a compiler error.

This would cover a ton of the common cases: matching numbers, matching strings, matching literal lists, matching structs/enums with 0 args, or simple args.

The annoying thing is that it wouldn't support something like:
```
x = Foo(1, 2)
y = Foo(1, 2)
x == y
```
so, it breaks the substitution principle there. We could potentially address this somewhat by looking in the score and doing partial evaluation as much as we can until we have one side or another as a valid binding-free pattern, but that's more handwavy and could result in weird cases where small changes to the code make it not compile for reasons that are not very clear.

If we could solve this, it would be very nice.

An alternate idea would be to add a new keyword, `matches` which allows you to write:
`x matches Foo(1, 2)`

which is more clear: the right hand side must be a valid pattern with no bindings in it, and if x matches it, it is true. This is maybe the better, if more verbose solution.

Any ideas @snoble

johnynek:
We could also extend this to support:
```
x matches Foo(y, 2) if f(y)
```
if we also allowed `if` in patterns this would work pretty well. One complication, the if expression, following python is `x if y else z` so, if `x = quux matches Foo(_, _)` then we have:
```
quux matches Foo(_, _) if y else z
```
and this either should fail to parse, or has to use backtracking to see that the `else` following the `if` implies that the if isn't part of the pattern. But it also complicates the scoping...

Maybe the simple variant should be done first the pattern can't have any binding and no `if`.

johnynek:
a related idea would be a syntax for partial functions:
```
\if Foo(x) -> Some(x)
```
where `\if` is a function from `a -> Option[b]` If the pattern does not match, return None, else return what the result is.

johnynek:
see #406 for an example:
```
    def lt(x, h):
        match to_Fn(x, h):
            LT: True
            _: False
```
which could be `to_Fn(x, h) matches LT`
greater than or equal to could be `to_Fn(x, h) matches (GT | EQ)`

[...]

PR #437, opened 2019-12-04
Allow a new form of apply

johnynek:
related to #396 

It occurred to me if we had the `\if Foo(x) -> fn(x)` syntax for making partial functions, then we could generate `Option` values with:

```
x.(\if Foo(y) -> fn(y)): Option[a]
```

maybe the syntax is too dense, but `|>` type compositions which do `x |> f = f(x)` which is what we have `x.f` doing, but we only allow names for `f` not any other form. Since `.(f)` is unambiguous to parse, we could allow more which can be useful in some cases...

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
@@ -765,8 +765,10 @@ object Declaration {
       val applied: P[NonBinding] = {
         val params = recNonBind.parensLines1Cut
         // here we are using . syntax foo.bar(1, 2)
+        // we also allow foo.(anyExpression)(1, 2)
+        val fn = varP | (recNonBind.parensCut)
         val dotApply: P[NonBinding => NonBinding] =
-          P("." ~ varP ~ params.?)
+          P("." ~/ fn ~ params.?)
             .region
             .map { case (r2, (fn, argsOpt)) =>
               val args = argsOpt.fold(List.empty[NonBinding])(_.toList)
--- a/core/src/main/scala/org/bykn/bosatsu/Parser.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Parser.scala
@@ -323,6 +323,9 @@ object Parser {
     def parens: P[T] =
       wrappedSpace("(", ")")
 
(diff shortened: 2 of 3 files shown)


avibryant:
Why not make this `x |> \if Foo(y) -> fn(y)`? The dot syntax seems overloaded here to me.

johnynek:
@avibryant you could also define such an operator:

```
def operator |>(x, f): f(x)
```
but the dot syntax isn't overloaded, it is what it always meant. Note, I only changed the parsing so that `x.(g)` would parse for any `g` previously, the parser would only parse a name after a `.`.

The fact that you can use it like `x.(\if Foo(x) -> f(x))` is a side effect of not making the `.` limited. Maybe (as I mentioned) stylistically it isn't great, but I don't really like banning it in the parser... but maybe i shouldn't merge this if we are sure that thngs like `foo.bar.(\x -> baz(g, x))` won't be useful.

avibryant:
@johnynek ok, I understand now. I find it a pretty unwieldy syntax and have a hard time imagining using it; the x.f sugar is nice when f is a name but since it's purely sugar anyway I don't see a huge argument for making it maximally general sugar. But YMMV.

johnynek:
you are probably right.

Note, I just checked this to be sure:
```
def operator |>(x, f): f(x)                                            

x = 1

y = x |> add(1)
```
and it works as expected, so maybe this is a bone-headed idea to support syntax that is pretty clearly terrible.

johnynek:
in light of #452 I think this is worth experimenting with to see how it fields to give us access to struct fields more conveniently:

```
struct Foo(x, y)

f = Foo("string", 1)

process_foo(f).(\Foo { x, ... } -> x ).use_str
```

avibryant:
@johnynek that's reasonably convincing. It reminds me a bit of Ruby's `tap` method (except that tap is all about side effects, and this is about pattern matching instead). Lets you keep a long method chain going instead of breakout out temp vars etc.

johnynek:
I'm so starved for reviews, I'll count that as a +1.

;)

Thanks for thinking about this, Avi. I really appreciate you thinking about this with me.

PR #446, opened 2019-12-05
Improve string intersection and difference

johnynek:
this improves string intersection and difference.

We use the same trick we use in lists: reverse the patterns if both end in literals so we can make progress there.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/SimpleStringPattern.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/SimpleStringPattern.scala
@@ -2,14 +2,22 @@ package org.bykn.bosatsu
 
 object SimpleStringPattern {
   sealed trait Pattern {
-    def matchesAny: Boolean = {
-      val list = toList
-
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/bykn/bosatsu/SimpleStringPattern.scala
```diff
@@ -166,9 +212,13 @@ object SimpleStringPattern {
         // we handled this above
         intersection(p2, p1)
       case (Cat(Var(_) | Wildcard, t1), Cat(Var(_) | Wildcard, t2)) =>
-        val left = intersection(t1, p2)
-        val right = intersection(t2, p1)
-        (left ::: right).map(Cat(Wildcard, _))
+        val t1norm = t1.unname.normalize
+        if (t1norm == t2.unname.normalize) (t1norm.prependWild :: Nil)
```
johnynek: We could preserve the Var name if both start with the same variable.

PR #454, opened 2019-12-15
Unify the list and string pattern matching algorithms

johnynek:
This is a large refactor which wound up reworking totality checking, string pattern matching, list pattern matching, and evaluation.

It was pretty tricky to find a design that could share so much code. The motivation is obvious: make it more likely we find any bugs by using the same code in many places.

By using this new code, we remove the limitation previously on patterns that there could only be one glob. So we can write:
```
def exists(list):
  match list:
    [*_, True, *_]: True
    _: False
```

cc @snoble


core/src/main/scala/org/bykn/bosatsu/Pattern.scala
```diff
@@ -349,41 +368,128 @@ object Pattern {
...
+      ListPat.toNamedSeqPattern(this)
+
+    lazy val toSeqPattern: SeqPattern[Pattern[N, T]] = toNamedSeqPattern.unname
   }
   case class Annotation[N, T](pattern: Pattern[N, T], tpe: T) extends Pattern[N, T]
   case class PositionalStruct[N, T](name: N, params: List[Pattern[N, T]]) extends Pattern[N, T]
   case class Union[N, T](head: Pattern[N, T], rest: NonEmptyList[Pattern[N, T]]) extends Pattern[N, T]
 
+  object ListPat {
+    def fromSeqPattern[N, T](sp: SeqPattern[Pattern[N, T]]) = {
```
johnynek: this has no return type.

johnynek:
```
[info] - (a - b) n c = (a n c) - (b n c) *** FAILED *** (70 milliseconds)

[info]   TestFailedException was thrown during property evaluation.

[info]     Message: List(Cat(AnyElem,Cat(Lit(1),Cat(Lit(0),Cat(Lit(0),Empty))))) did not equal List()

[info]     Location: (SetOpsLaws.scala:207)

[info]     Occurred when passed generated values (

[info]       arg0 = Cat(Wildcard,Cat(Lit(0),Empty)),

[info]       arg1 = Cat(AnyElem,Cat(Lit(1),Cat(AnyElem,Cat(Lit(0),Empty)))),

[info]       arg2 = Cat(AnyElem,Cat(Lit(1),Cat(Lit(0),Cat(Lit(0),Empty))))

[info]     )
```

PR #455, opened 2019-12-15
Implement SeqPattern to unify list and string patterns

johnynek:
This is part 1 of #454


johnynek:
despite my policy on small PRs, this major refactor would be very hard to stage that way as it touches so much of the code.

Going to close and focus on #454 since this is hardly smaller.

Issue #434, opened 2019-12-03
optimize tail recursion in evaluation


johnynek:
Any real compiler will need to compile tail recursion into a loop. To get ready for that, we should implement detection of this and explicit generation of loops in evaluation rather than relying on Eval to be stack safe.

We will probably eventually have another intermediate representation that includes loops so we can reuse this work (such a layer could also have a more realistic machine model for pattern matching, e.g. only if/else and checks on integer values).

PR #463, opened 2019-12-22
Make FnValue Value to Value not Eval to Eval

johnynek:
Since we have solved #434 we don't need Eval to make any loop stack safe. More over, by forcing us to use that more, we work any bugs out of that system before we make a backend that would almost certainly want to not put trampolines everywhere.

There is still some laziness here in the environment since setting up a recursive value is using it. That could still probably be removed or reduced for more performance improvements.

benchmarks show 80% faster on the most time consuming benchmark.
```
[info] Benchmark          Mode  Cnt      Score      Error  Units    
[info] TestBench.bench0  thrpt    3  13190.738 ±  463.814  ops/s    
[info] TestBench.bench1  thrpt    3  10599.481 ± 5759.104  ops/s    
[info] TestBench.bench2  thrpt    3      2.706 ±    0.574  ops/s    
     

[info] Benchmark          Mode  Cnt      Score      Error  Units
[info] TestBench.bench0  thrpt    3  14192.600 ± 3202.505  ops/s
[info] TestBench.bench1  thrpt    3  12457.443 ±  387.050  ops/s
[info] TestBench.bench2  thrpt    3      4.867 ±    0.511  ops/s

```

### DIFF
--- a/bench/src/main/scala/org/bykn/bosatsu/TestBench.scala
+++ b/bench/src/main/scala/org/bykn/bosatsu/TestBench.scala
@@ -95,7 +95,7 @@ def max_of(n, fn):
 # return the first defined value from largest to smallest
 # of the given function, if it is defined
 def first_of(n, fn):
-  int_loop(n, None, \i, res ->
+  int_loop(n, None, \i, _ ->
     match fn(i):
       None: (i - 1, None)
       nonNone: (0, nonNone))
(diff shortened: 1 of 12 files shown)


johnynek:
the native image tool can test everything in the test repo in:
```
real    0m1.110s
user    0m1.710s
sys     0m0.110s
```
vs
```
real    0m1.625s
user    0m2.214s
sys     0m0.101s
```
before this patch. Since that includes parsing and typechecking, that's pretty good to cut off 0.5s of realtime.

PR #467, opened 2019-12-23
implement matches syntax

### LINKED ISSUE #396: `matches` syntax idea

johnynek:
close #396 

take a look @avibryant @snoble if you have time.

I like this syntax and think it is pretty useful as you can see in the examples I changed.

The fact that rust recently added a macro for this seals the deal for me since I didn't know about that when I was thinking about this. The fact that it was independently considered by the rust folks seems like good evidence.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Declaration.scala
@@ -105,6 +105,15 @@ sealed abstract class Declaration {
         // TODO this isn't quite right
         kindDoc + typeName.toDoc + Doc.char(':') + args.sepDoc +
           piPat.document(args)
+      case m@Matches(arg, p) =>
+        val da = arg match {
+          // matches binds tighter than all these
+          case Lambda(_, _) | IfElse(_, _) | ApplyOp(_, _, _) | Match(_, _, _) =>
+            Parens(arg)(m.region).toDoc
+          case _ =>
+            arg.toDoc
+        }
+        da + Doc.text(" matches ") + Document[Pattern.Parsed].document(p)
       case Parens(p) =>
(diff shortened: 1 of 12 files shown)


avibryant:
Yeah, this is great. Strong +1.

johnynek:
What’s amazing is that the PR is nearly met zero change. The lines saved in the examples is already enough (nearly) to save the code needed to implement and add generative testing. That’s a pretty good endorsement. If a syntax can even save total lines in the compiler and its examples it probably pays its way.

PR #468, opened 2019-12-25
Handle both struct and list form of list patterns

johnynek:
Since Lists are first class types, we have an issue where we have both the literal list an named struct form of a pattern.

This PR handles that case.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
@@ -95,7 +95,7 @@ final class SourceConverter(
               }
             case pat =>
               // TODO: we need the region on the pattern...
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
```diff
@@ -544,14 +544,61 @@ final class SourceConverter(
...
+            // this pattern shows we have a list of something, but we don't know what
+            // changing to _ would allow more things to typecheck, which we can't do
+            // and we can't annotate because we don't know the type of the list
+            Pattern.ListPat(parts)
+          }
+          else {
+            // we are already in the tail of a list, so we can just put n here
+            Pattern.Var(n)
+          }
+        case (Pattern.ListPart.WildList | Pattern.ListPart.NamedList(_)) :: _ =>
```
johnynek, line 589: We can convert `[*_, _]` to `NonEmptyList(_, _)`

PR #488, opened 2020-07-31
An Execution for NormalExpressions and a Webserver that uses it for a Notebook like interface

snoble:
This is massive and can certianly be broken up... and needs a lot more testing. But I figured I'd make a PR to start to organize what steps are left. I think one possibility is to move the `ws` build into its own repo.

@johnynek would you be up for reading what's here and discussing what you'd be interested in merging?

Ultimately there are a few big pieces of work here that relate but can be pulled apart.

- An executor for NormalExpression.
- A webserver that uses this executor to provide a notebook like interface.
- A modification to Predef to allow for external functions to be passed the NormalExpression for a value instead of the Value itself when applied.
- Some visualization methods that operate by being passed a NormalExpression, putting the value of the expression into a cache, and returning an opaque type that gets rendered with the expressions and a cache key.

### DIFF
--- a/core/src/main/resources/bosatsu/predef.bosatsu
+++ b/core/src/main/resources/bosatsu/predef.bosatsu
@@ -42,10 +42,9 @@ export [
   trace,
   uncurry2,
   uncurry3,
-  text,
+  markdown,
   header,
   list_vis,
-  make_vis,
 ]
 
 struct Unit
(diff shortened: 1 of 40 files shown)


core/src/main/resources/bosatsu/predef.bosatsu
```diff
@@ -42,6 +42,9 @@ export [
   trace,
   uncurry2,
   uncurry3,
+  markdown,
+  header,
+  list_vis,
```
johnynek, line 47: small comment, I'd really like to keep the predef as minimal as possible (even now I think it could be too big).
 
I think these functions should be in a separate module (which could be configured to be auto imported by your notebook system).

snoble: I think I just struggled how to have a second .bosatsu file make it into the build. But I should definitely go back and fix that.

PR #490, opened 2020-08-02
Move top level name uniquing to SourceConverter

johnynek:
This fixes fully resolving names when possible. Now, the only free variables are those introduced by lambda bindings, as it should be. Otherwise they are Globals fully resolved to a package.

todo: leverage the invariant that all top level lets have no free variables in Evaluation.

By not treating global names like closures, we can more efficiently compile them since we statically know which name they are referring to and don't need to do any dynamic lookup.


core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
```diff
@@ -1011,6 +1133,105 @@ object SourceConverter {
...
+              }
+              else {
+                NonEmptyList.fromList(tail) match {
+                  case Some(netail) =>
+                    // this b is different from name, but may reference it
+                    val d1 = fn(d)
+                    renameUntilNext(name, netail, (b, r, d1) :: acc)(fn)
+                  case None =>
+                    // head is the last, and the last is never renamed
+                    NonEmptyList(head, acc).reverse
```
johnynek: This branch can't happen because it means we are renaming but we never hit the bound name. That implies the name isn't actually duplicated. We should throw here that an invariant has been violated.

Note the code coverage does not reach here.

core/src/main/scala/org/bykn/bosatsu/Declaration.scala
```diff
@@ -188,7 +191,7 @@ sealed abstract class Declaration {
         case Lambda(args, body) =>
           val bound1 = bound ++ args.toList.flatMap(_.names)
           loop(body, bound1, acc)
-        case Literal(lit) => SortedSet.empty
+        case Literal(lit) => acc
```
johnynek, line 194: this has been a lurking bug.

core/src/main/scala/org/bykn/bosatsu/Declaration.scala
```diff
@@ -163,22 +163,25 @@ sealed abstract class Declaration {
         case Annotation(term, _) => loop(term, bound, acc)
         case Apply(fn, args, _) =>
           (fn :: args).foldLeft(acc) { (acc0, d) => loop(d, bound, acc0) }
-        case ApplyOp(left, _, right) =>
+        case ao@ApplyOp(left, _, right) =>
           val acc0 = loop(left, bound, acc)
-          loop(right, bound, acc0)
+          val acc1 = loop(ao.opVar, bound, acc0)
```
johnynek, line 168: this bug was lurking.

core/src/main/scala/org/bykn/bosatsu/Declaration.scala
```diff
@@ -245,9 +270,9 @@ sealed abstract class Declaration {
         case Annotation(term, _) => loop(term, acc)
         case Apply(fn, args, _) =>
           (fn :: args).foldLeft(acc) { (acc0, d) => loop(d, acc0) }
-        case ApplyOp(left, _, right) =>
+        case ApplyOp(left, op, right) =>
           val acc0 = loop(left, acc)
-          loop(right, acc0)
+          loop(right, acc0 + op)
```
johnynek, line 275: lurking bug

core/src/main/scala/org/bykn/bosatsu/Declaration.scala
```diff
@@ -268,7 +293,7 @@ sealed abstract class Declaration {
         case Lambda(args, body) =>
           val acc1 = acc ++ args.toList.flatMap(_.names)
           loop(body, acc1)
-        case Literal(lit) => SortedSet.empty
+        case Literal(lit) => acc
```
johnynek, line 296: lurking bug.

Issue #486, opened 2020-07-31
 (a - b) n c = (a n c) - (b n c) flake


johnynek:
could be related to #460 

```
[info] - (a - b) n c = (a n c) - (b n c) *** FAILED *** (32 milliseconds)

[info]   TestFailedException was thrown during property evaluation.

[...]

Issue #475, opened 2020-07-27
x - top = 0 flake


johnynek:
```
[info] - x - top = 0 *** FAILED *** (66 milliseconds)

[info]   TestFailedException was thrown during property evaluation.

[...]

PR #492, opened 2020-08-04
Improve SeqPattern testing

### LINKED ISSUE #480: Totality check bug
With the following code:
```
package Foo

def any(lst):
    match lst:
        [*_, True, *_]: True
        [] | [False, *_]: False


f = any([False, False])

```
we get:
```
oscar@oscar-XPS-13-7390:~/oss/bosatsu$ bosatsu eval --main_file foo.bosatsu
in file: foo.bosatsu, package Foo
2|
3|def any(lst):
4|    match lst:
      ^

to:
4|    match lst:
5|        [*_, True, *_]: True
6|        [] | [False, *_]: False
                                ^

non-total match, missing: [True, *_]
```

but `[*_, True, *_]` should be `[True, *_] | [_, *_,True, *_]` which clearly contains `[True, *_]`.

As a hint, rewriting this to:
```
package Foo

def any(lst):
    match lst:
        []: False
        [*_, True, *_]: True
        [False, *_]: False


f = any([False, False])
```
passes. This is an example of a perturbation test: if we take a total match, combining any two branches with a union should keep totality. I should write that law.

### LINKED ISSUE #491: TotalityTest a n b = 0 then a - b = a flake
In TotalityTest
```
[info] - if a n b = 0 then a - b = a (7 milliseconds)

[info] - x - y = z, then x - y - z = 0 (18 milliseconds)

[info] - subset consistency: a n b == a <=> a - b = 0 *** FAILED *** (20 milliseconds)

[info]   TestFailedException was thrown during property evaluation.

[info]     Message: intSub was false

[info]     Location: (SetOpsLaws.scala:140)

[info]     Occurred when passed generated values (

[info]       arg0 = PositionalStruct((PackageName(NonEmptyList(Rmt7bxrG, WpeuFhxCttl, Dzcxddk, Fc6hpfbrfVO)),Constructor(Oe0cgBh5hkr)),List(ListPat(List()), ListPat(List()), Named(Name(bnuwm),Var(Name(dkc9haevfxJ))), Var(Operator(%)), Named(Name(zbkjafaZvvg),Var(Name(sby))), PositionalStruct((PackageName(NonEmptyList(Baub, WpBjarYbqyp, SmHify, Liey, KrivC)),Constructor(Fzhbyxy)),List(PositionalStruct((PackageName(NonEmptyList(E, Bsq9yxbk14)),Constructor(GcsasGis3)),List(StrPat(NonEmptyList(LitStr(yg0s7u), NamedStr(Name(byph6xgFSf)))), Var(Name(bpk7wzjtuzi)))), WildCard, Named(Name(ekwv),StrPat(NonEmptyList(NamedStr(Name(efufaouhs))))))))),

[info]       arg1 = ListPat(List(NamedList(Name(kc)))),

[info]       arg2 = org.bykn.bosatsu.TotalityTest$$anon$1@137e

[info]     )
```

johnynek:
This is an attempt to fix #491 #486 #480 #475.

The problem is that difference is an upper bound for SeqPattern with wildcards at both ends.

I think this is not a soundness problem because it rejects some patterns that are total, but it can't see that they are. It would be nice to minimize those however.

This code fixes a couple of real bugs, comments out one law that is only approximately true, and adds a heuristic for improving the checks for missing branches (which can be pessimistic, but should never report that a match is total when it isn't).

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/pattern/SeqPart.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/pattern/SeqPart.scala
@@ -2,6 +2,7 @@ package org.bykn.bosatsu.pattern
 
 sealed trait SeqPart[+Elem] {
   def notWild: Boolean = false
+  def isWild: Boolean = !notWild
 }
 object SeqPart {
   sealed trait SeqPart1[+Elem] extends SeqPart[Elem] {
--- a/core/src/main/scala/org/bykn/bosatsu/pattern/SeqPart.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/pattern/SeqPart.scala
@@ -76,7 +76,7 @@ object SeqPart {
           case AnyElem => true
           case Lit(c2) =>
             p1 match {
-              case Lit(c1) => setOpsA.subset(c2, c1)
+              case Lit(c1) => setOpsA.subset(c1, c2)
               case AnyElem => setOpsA.isTop(c2)
             }
         }
--- a/core/src/main/scala/org/bykn/bosatsu/pattern/SeqPattern.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/pattern/SeqPattern.scala
@@ -21,13 +21,6 @@ sealed trait SeqPattern[+A] {
     }
 
   def isEmpty: Boolean = this == Empty
-  def nonEmpty: Boolean = !isEmpty
-
-  def isLit: Boolean =
-    this match {
-      case Cat(Lit(_), Empty) => true
-      case _ => false
-    }
 
   /**
    * Concat that SeqPattern on the right
@@ -237,7 +230,12 @@ object SeqPattern {
        * return true if p1 <= p2, can give false negatives
        */
       def subset(p1: SeqPattern[A], p2: SeqPattern[A]): Boolean =
-        p2.matchesAny || subsetList(p1.toList, p2.toList)
+        p2.matchesAny || {
+          // if p2 doesn't matchEmpty but p1 does, we are done
+          // check that case quickly
+          if (!p2.matchesEmpty && p1.matchesEmpty) false
+          else subsetList(p1.toList, p2.toList)
+        }
 
       @inline
       final def isAny(p: SeqPart1[A]): Boolean =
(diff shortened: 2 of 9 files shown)


core/src/main/scala/org/bykn/bosatsu/pattern/SetOps.scala
```diff
@@ -64,9 +66,56 @@ trait SetOps[A] {
    * }
    */
   def missingBranches(top: List[A], branches: List[A]): List[A] = {
-    val missing = branches.foldLeft(top) { (missing, nextBranch) =>
-      differenceAll(missing, nextBranch :: Nil)
-    }
+    // we can subtract in any order
```
johnynek, line 73: this is a new greedy heuristic that helps for one test.

core/src/main/scala/org/bykn/bosatsu/pattern/SetOps.scala
```diff
@@ -123,6 +172,29 @@ object SetOps {
       def subset(a: A, b: A): Boolean = ordA.equiv(a, b)
     }
 
+  def fromFinite[A](items: Iterable[A]): SetOps[Set[A]] =
```
johnynek, line 181: this was important to better model the Pattern matching we have with enums. It immediately exposed a bug.

core/src/main/scala/org/bykn/bosatsu/pattern/SeqPart.scala
```diff
@@ -76,7 +76,7 @@ object SeqPart {
           case AnyElem => true
           case Lit(c2) =>
             p1 match {
-              case Lit(c1) => setOpsA.subset(c2, c1)
+              case Lit(c1) => setOpsA.subset(c1, c2)
```
johnynek, line 80: this was a bug. I'm surprised we never hit an issue with it before. The `SeqPattern[Set[Boolean]]` found the bug.

johnynek:
close #491 #486 #480 #475

core/src/main/scala/org/bykn/bosatsu/pattern/SeqPattern.scala
```diff
@@ -9,8 +9,7 @@ sealed trait SeqPattern[+A] {
   def matchesAny: Boolean =
     this match {
       case Empty => false
-      case Cat(Wildcard, Empty) => true
-      case Cat(Wildcard, t) => t.matchesAny
+      case Cat(Wildcard, t) => t.matchesEmpty
```
johnynek, line 12: I don't think this was a bug, but it is a bit clearer this way.

core/src/main/scala/org/bykn/bosatsu/pattern/SeqPattern.scala
```diff
@@ -369,63 +367,54 @@ object SeqPattern {
...
+            val dtail = difference(t1, p2).map(Cat(h1, _))
+            val u = differenceAll(dtail, t2 :: Nil)
 
             unifyUnion(u)
           case (Cat(Wildcard, t1), Cat(h2: SeqPart1[A], t2)) =>
             // *:t1 - (h2:t2) = t1 + _:p1 - h2:t2
             //   = (t1 - p2) + (_:p1 - h2:t2)
             val d12 = {
-              val dtail = difference(p1, t2)
-              if (dtail == (p1 :: Nil)) Cat(AnyElem, p1) :: Nil
```
johnynek, line 421: this is a bug, we can have an upper bound, but not be disjoint. `a - b = a` does not imply` a n b = 0`. We have to check that directly.

core/src/main/scala/org/bykn/bosatsu/pattern/SeqPattern.scala
```diff
@@ -369,63 +367,54 @@ object SeqPattern {
...
+              //
+              //or:
+              //(_ - h2):(p1 n t2) + _:(p1 - t2)
+              if (disjoint(p1, t2)) {
+                Cat(AnyElem, p1) :: Nil
+              }
               else {
-                val d1 =
-                  for {
-                    h <- part1SetOps.intersection(AnyElem, h2)
```
johnynek, line 425: there is no need to do `AnyElem n h2`. It has to be `h2`.

core/src/main/scala/org/bykn/bosatsu/pattern/SetOps.scala
```diff
@@ -64,9 +70,59 @@ trait SetOps[A] {
...
+            }
+            val smallest = trials.iterator.map(_._1).min
+            // choose the diff that ends up the smallest
+            // the most times
+            val best = trials
+              .collect { case (ulen, p) if ulen == smallest => p }
+              .flatten
+              .groupBy(identity)
+              .map { case (k, v) => (k, v.size) }
+              .minBy(_._2)
```
johnynek: should be maxBy

core/src/main/scala/org/bykn/bosatsu/pattern/SeqPattern.scala
```diff
@@ -369,63 +367,54 @@ object SeqPattern {
...
-              val dtail = difference(t1, p2)
-              if (dtail == (t1 :: Nil)) p1 :: Nil
-              else {
-                val d1 =
-                  for {
-                    h <- part1SetOps.intersection(h1, AnyElem)
-                    t <- difference(t1, p2)
-                  } yield Cat(h, t)
-
-                val d2 = part1SetOps.difference(h1, AnyElem).map(Cat(_, t1))
```
johnynek, line 408: `h1 - Any = 0`

core/src/main/scala/org/bykn/bosatsu/pattern/SeqPattern.scala
```diff
@@ -369,63 +367,54 @@ object SeqPattern {
...
+          case (_, _) if disjoint(p1, p2) => p1 :: Nil
           case (Cat(h1: SeqPart1[A], t1), Cat(Wildcard, t2)) =>
-            // h1:t1 - (*:t2) = h1:t1 - t2 - _:p2
-            val d12 = {
-              val dtail = difference(t1, p2)
-              if (dtail == (t1 :: Nil)) p1 :: Nil
-              else {
-                val d1 =
-                  for {
-                    h <- part1SetOps.intersection(h1, AnyElem)
```
johnynek, line 404: `h1 n Any = h1`

core/src/main/scala/org/bykn/bosatsu/pattern/SeqPattern.scala
```diff
@@ -369,63 +367,54 @@ object SeqPattern {
...
               // *:t1 = t1 + _:p1
               // _:p1 - [] = _:p1
               unifyUnion(Cat(AnyElem, p1) :: difference(t1, Empty))
             }
+          case (_, _) if disjoint(p1, p2) => p1 :: Nil
           case (Cat(h1: SeqPart1[A], t1), Cat(Wildcard, t2)) =>
-            // h1:t1 - (*:t2) = h1:t1 - t2 - _:p2
-            val d12 = {
-              val dtail = difference(t1, p2)
-              if (dtail == (t1 :: Nil)) p1 :: Nil
```
johnynek, line 400: you can't use difference this way, it is an upper bound, you have to directly use disjointness.

PR #494, opened 2020-08-04
Snoble/let free expressions

snoble:
Rename NormalExpressions to LetFreeExpressions (and related names). NormalExpressions was too generic of a name and can create a lot of confusion. LetFreeExpressions refers to the fact the `Lets` are all removed. The Lambda are also changed to use De Bruijn notation so that's an alternative source of naming.


snoble:
this is the first step towards https://github.com/johnynek/bosatsu/pull/488

I'll tackle code coverage failure that's showing up

johnynek:
Note codecov is not happy about your level of coverage. :)

johnynek:
This is clearer. Thanks.

Issue #473, opened 2020-07-26
a match-less IR


johnynek:
I was thinking for compiling bosatsu, it would be handy to have an intermediate representation without pattern matching and with a loop construct for tail recursion.

Some ideas:
1. replace constructor functions with calls to something `build_enum(variant, a0, a1, ...)` and `build_struct(a0, a1, ...)` Maybe we could treat structs like enums where they have variant == 0, but pattern matches would ignore them, but that seems like a waste of space given we statically know if something is an enum or struct and it would be nice to leverage that info.
2. add instructions for `get_enum_variant`, `get_enum_element` and `get_struct_element`.
3. compile pattern matches into if/else trees which are only comparing variant identifiers. So, there would be no match expression at this layer.
4. detect tail recursive loops and rewrite those at this layer so these can be directly written as while loops hoisting any constants out of the loop.
5. inline small private functions or any private functions that are only used once.

johnynek:
A partial solution to #388 or at least #331 probably needs to be considered.

In which case, a native nat type/match support should be implemented in the IR.

enum-like types would be useful for bool / fixed width intergers to be optimized.

johnynek:
Compiling list and string patterns will also be a bit non-trivial.

List is a bit clearer: we can rewrite the patterns into bosatsu code that only matches head and tail of lists.

For string patterns, we need something like substring, indexOf and rightIndexOf or something.

PR #496, opened 2020-08-05
Creates LetFreeExprFn for LetFreeEvaluation

snoble:
I'm starting with the most invasive part of https://github.com/johnynek/bosatsu/pull/488. This creates a Value class that an Evaluator can either treat as a function (`asFn`) or it can pass it a let-free expression for a value so the external function can determine its own strategy for how to produce the output (which may not require full evaluation of the argument Value, eg it may use a cached return value).

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Externals.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Externals.scala
@@ -44,7 +44,7 @@ object FfiCall {
 
     private[this] def evalExprFn(t: rankn.Type): ExprFnValue = ExprFnValue({ (e1, cache, eval) => Value.ExternalValue(wrapper(e1, t, cache, eval)) })
 
-    def call(t: rankn.Type): Value = evalExprFn(t)
+    def call(t: rankn.Type): Value = new Value.FnValue(evalExprFn(t))
   }
 
   def getJavaType(t: rankn.Type): List[Class[_]] = {
--- a/core/src/main/scala/org/bykn/bosatsu/Externals.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Externals.scala
@@ -39,6 +39,14 @@ object FfiCall {
     def call(t: rankn.Type): Value = evalFn
   }
 
+  final case class ExprFn(wrapper: (LetFreeEvaluation.LetFreeValue, rankn.Type, LetFreeEvaluation.Cache, LetFreeEvaluation.ToLFV) => Any) extends FfiCall {
+    import Value.LetFreeExprFnValue
+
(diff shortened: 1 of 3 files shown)


snoble:
I'll add some testing.

snoble:
Note that the main reason this invasiveness exists is because it allows for a lot of reuse of the tooling used by Evaluation.

johnynek:
I'm planning on totally rewriting Evaluation to use #473 internally so that can be a kind of a test for the ability to generate code.

I'm a bit concerned about making Evaluation have a dependency on LetFree because then we have two separate ASTs that this is trying to interact with.

Could we instead plug in a trait of a function to handle the callbacks in a generic way that does not create the dependency back, but allows LetFree to pass in the correct functions?

snoble:
> Could we instead plug in a trait of a function to handle the callbacks in a generic way that does not create the dependency back, but allows LetFree to pass in the correct functions?

I'm finally getting back to this. I think I can figure out how to do that. I'll give it a shot tomorrow. Thanks!

snoble:
I created a trait for FnValue to move the LetFree stuff out of Value. But I ended up adding some boxing because I couldn't figure out how to make FnValue a trait that extends Value and then extend FnValue elsewhere (without unsealing Value).

snoble:
The failures are issues downloading dependencies. I'll see what I can test when the codecov report shows up.

snoble:
I'm going to add test coverage to https://github.com/johnynek/bosatsu/pull/513 instead.

core/src/main/scala/org/bykn/bosatsu/Externals.scala
```diff
@@ -39,6 +39,14 @@ object FfiCall {
     def call(t: rankn.Type): Value = evalFn
   }
 
+  final case class ExprFn(wrapper: (LetFreeEvaluation.LetFreeValue, rankn.Type, LetFreeEvaluation.Cache, LetFreeEvaluation.ToLFV) => Any) extends FfiCall {
+    import LetFreeEvaluation.ExprFnValue
+
+    private[this] def evalExprFn(t: rankn.Type): ExprFnValue = ExprFnValue({ (e1, cache, eval) => Value.ExternalValue(wrapper(e1, t, cache, eval)) })
+
+    def call(t: rankn.Type): Value = new Value.FnValue(evalExprFn(t))
```
johnynek, line 47: I'm going to merge this, but I'd like to refactor to remove Value depending on LetFree (which I think is a circular dependency).

For instance, I think you just want a `case class FromFn(callFn: rankn.Type => Value) extends FfiCall { def call(t: rankn.Type) = callFn(t) }`

snoble: That is much simpler than what I had.

PR #499, opened 2020-08-06
Add DataRepr to improve factoring of code

johnynek:
this makes an ADT where we previously had some access functions on DefinedType.

The goal here is to use the same ADTs in the matchless IR #478 and then we should be able to swap out Evaluation
to be an interpreter for the Matchless IR which removes duplication.


johnynek:
96% of the diff was hit and the downward delta looks like it is in the noise (4 basis points).

PR #500, opened 2020-08-06
Improve SeqPattern.difference, test boolean formulas

johnynek:
I was being a bit dense about a key formula `A - (B | C)` which @non reminded me is `(A - B) n (A - C)`. By using this, then a slight upper bound, the formula in the worst case for difference gets a lot simpler.

I also added property checks for the logic formulas I'm using. They should be theorems from the axioms of set theory, but just to make sure I don't make future mistakes, I'll add tests (I guess if I were really cool I'd include coq proofs or something).

### DIFF
--- a/core/src/test/scala/org/bykn/bosatsu/pattern/SeqPatternTest.scala
+++ b/core/src/test/scala/org/bykn/bosatsu/pattern/SeqPatternTest.scala
@@ -568,6 +568,14 @@ class BoolSeqPatternTest extends SeqPatternLaws[Set[Boolean], Boolean, List[Bool
       law(p1, p2, bs)
     }
   }
+
+  test("test some missing branches") {
+    assert(setOps.missingBranches(Cat(Wildcard, Empty) :: Nil, Pattern.fromList(List(Wildcard, AnyElem, Wildcard)) :: Nil) ==
+      Pattern.fromList(Nil) :: Nil)
+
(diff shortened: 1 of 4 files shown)


core/src/test/scala/org/bykn/bosatsu/pattern/SeqPatternTest.scala
```diff
@@ -568,6 +568,14 @@ class BoolSeqPatternTest extends SeqPatternLaws[Set[Boolean], Boolean, List[Bool
       law(p1, p2, bs)
     }
   }
+
+  test("test some missing branches") {
+    assert(setOps.missingBranches(Cat(Wildcard, Empty) :: Nil, Pattern.fromList(List(Wildcard, AnyElem, Wildcard)) :: Nil) ==
+      Pattern.fromList(Nil) :: Nil)
+
+    assert(setOps.missingBranches(Cat(Wildcard, Empty) :: Nil, Pattern.fromList(List(Wildcard, Lit(Set(true)), Wildcard)) :: Nil) ==
+      Pattern.fromList(Nil) :: Pattern.fromList(Lit(Set(false)) :: Wildcard :: Nil) :: Nil)
```
johnynek, line 577: this test is pretty awesome. This is exactly what we couldn't do cleanly before because the bounds were too weak.

PR #501, opened 2020-08-06
use DataRepr in Matchless

johnynek:


core/src/main/scala/org/bykn/bosatsu/Matchless.scala
```diff
@@ -359,16 +368,36 @@ object Matchless {
...
+            case DataRepr.SuccNat =>
+              params match {
+                case single :: Nil =>
+                  // if we match, we recur on the inner pattern and prev of current
+                  val check = Some(EqualsNat(arg, DataRepr.SuccNat))
+                  val bind: F[(LocalAnon, Expr)] = makeAnon.map { nm => (LocalAnon(nm), PrevNat(arg)) }
+                  for {
+                    pre <- bind
+                    rest <- doesMatch(pre._1, single)
+                  } yield rest.map { case (preLets, cond, res) => (pre :: preLets, andInt(check, cond), res) }
```
johnynek, line 395: this is a little weird and makes me wonder if the whole design has an issue:

we have to compute prev unconditionally, we don't have a way to sequence doing a bit of extraction work here and then follow up below. I guess this is what the side-effecting conditions are for, but we're not using it here, so it seems like it will unconditionally run. I think we may want to rethink the composition.

johnynek:
I'm going to let the coverage slide since Matchless isn't used yet. The next PR will be putting it into use.

PR #508, opened 2020-08-13
Add a minimal Python code AST

johnynek:
This is a minimal python code ast which seems to be all we need to generate python code from bosatsu.

To test it, we use jython and make sure jython can parse everything.


johnynek:
I'm going to go ahead and merge this because code coverage will improve when we are actually generating python code, coming up next.

PR #513, opened 2020-08-14
Fix bugs in a LetFree version of the expressions and add an evaluator for them

snoble:
The LetFreeExpression is an attempting to be a stripped down version of a bosatsu expression to give a semantic representation of the output of the computation. It is not intended to represent the computation that produces the output, even though it does contain enough information to be evaluated. As such it can be used to sometimes tell if two expressions will produce the same outputs for cacheing purposes. It also makes functions inspectable which means it's possible to translate them into code for extrenal computation (eg sql statements).

As the name implies, it expression eliminates Lets by inlining the values. This creates an issue for how to represent a recurrsive Lets so we introduce a Recursion expression that is only valid if it contains a Lambda, where the applying the argument is the recursive call. All variable names are eliminated and replaced with De Bruijn indicies so that expressions get the same representation regardless of variable name choice or nesting. Where possible Match and App are normalized, even when it will produce a less efficient computation. No recurrsive functions or external functions are applied as they are considered runtime concerns.

LetFreeEvaluation exists so that we can verify that LetFreeExpressions do describe the same output that the TypedExpressions that they are converted from describe.
 
 It is also useful for prototyping applications that heavily use LetFreeExpressions.

In one of the tests is included a function that demonstrates how a user can take advantage of these expressions. `externalFilter` is able to check if the filter function is actually using the argument for its value, and if not it just returns either the whole list or the empty list.
https://github.com/johnynek/bosatsu/blob/0338e369c2fcd3a3e62862547a8db15e1cb0527a/core/src/test/scala/org/bykn/bosatsu/ExprFnTest.scala#L34-L76


johnynek:
if you merge master you might get a more reliable CI (and I also fixed a flakey bug).

snoble:
> if you merge master you might get a more reliable CI (and I also fixed a flakey bug).

Thank you!

johnynek:
So, this PR is dropping project coverage by 1.7%, which seems like a big hit to me. Do you think you can find a way to get coverage higher?

snoble:
> So, this PR is dropping project coverage by 1.7%, which seems like a big hit to me. Do you think you can find a way to get coverage higher?

Yes. Since this now has a full bosatsu evaluator I'm going to add the bosatsu tests to exercise this code.

cli/src/main/scala/org/bykn/bosatsu/PathModule.scala
```diff
@@ -118,6 +118,16 @@ object PathModule extends MainModule[IO] {
           val doc = resDoc.value + (Doc.lineOrEmpty + Doc.text(": ") + tDoc).nested(4)
           print(doc.render(100))
         }
+      case res@Output.LetFreeEvaluationResult(lfe, tpe, _, _) => for {
+        _ <- print(s"LetFree Expression: $lfe")
```
johnynek: can we remove the debug statements?

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -2,7 +2,6 @@ package org.bykn.bosatsu
 
 import cats.data.{NonEmptyList, State}
 import cats.implicits._
-import cats.Id
 import rankn._
```
johnynek, line 7: Can we add a comment to the class and explain what this is for (and AST without Lets I think), and why we would want it.

For instance, the motivation of the Matchless AST is that pattern matching is rather complex and a sublanguage beyond the basic lambda calculus. Why do we want to remove lets? I think it is part of a plan for normalizing expressions.

But since we still have patterns here, we should comment on pattern normalization, and how that plays a role.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -129,42 +180,69 @@ object LetFreeConversion {
   type LetFreePM = PackageMap.Typed[(Declaration, LetFreeExpressionTag)]
   type LetFreePac = Package.Typed[(Declaration, LetFreeExpressionTag)]
 
-  type PatternEnv = Map[Int, LetFreeExpression]
+  type PatternEnv[T] = Map[Int, T]
```
johnynek: I think using `LongMap[T]` will give significantly better performance since you won't have the `Int` boxing.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -129,42 +180,69 @@ object LetFreeConversion {
...
+      case _ => None
     }
   }
 
   def listAsStructList(lst: List[LetFreeExpression]): LetFreeExpression =
-    lst.foldRight(LetFreeExpression.Struct(0, Nil)) { case (lfe, acc) => LetFreeExpression.Struct(1, List(lfe, acc)) }
+    lst.foldRight(LetFreeExpression.Struct(0, Nil, DataFamily.Enum)) { case (lfe, acc) => LetFreeExpression.Struct(1, List(lfe, acc), DataFamily.Enum) }
+
+  case class LitValue(toAny: Any) {
+    def equivToLit(lit: Lit) = lit match {
```
johnynek: I think `def equivToLit(lit: Lit): Boolean = toAny == lit.unboxToAny` is the same.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -129,42 +180,69 @@ object LetFreeConversion {
...
+      case Lit.Integer(i) => i == toAny
+      case Lit.Str(s) => s == toAny
+    }
+  }
+
+  object LitValue {
+    def fromLit(lit: Lit) = lit match {
```
johnynek: `def fromList(lit: Lit): LitValue = LitValue(lit.unboxToAny)`

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -129,42 +180,69 @@ object LetFreeConversion {
...
+      case Lit.Integer(i) => LitValue(i)
+      case Lit.Str(s) => LitValue(s)
+    }
+  }
 
-  private def maybeBind(pat: LetFreePattern): (LetFreeExpression, PatternEnv) => PatternMatch[PatternEnv] =
+  implicit val neToLitValue: LetFreeExpression => Option[LitValue] = {
```
johnynek: can you commend on the implicit function, which winds up being an implicit conversion? I'm skeptical about using that pattern here.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -129,42 +180,69 @@ object LetFreeConversion {
...
+    case _ => None
+  }
+  implicit val neToStruct: (LetFreeExpression, DataFamily) => Option[(Int, List[LetFreeExpression])] = {
+    case (LetFreeExpression.Struct(enum, args, _), df) => Some((enum, args))
+    case _ => None
+  }
+  implicit val neToList: LetFreeExpression => Option[List[LetFreeExpression]] = structListAsList(_)
+  implicit val neFromList: List[LetFreeExpression] => LetFreeExpression = listAsStructList(_)
+
+  def maybeBind[T](pat: LetFreePattern)(implicit
```
johnynek: I would at least consider what I call the "module pattern" here. Where you would do this:
```scala
abstract class MaybeBind[A](pat: LetFreePattern) {
  def toLitValue(t: A): Option[LitValue]
  def toStruct(t: A, df: DataFamily): Opiton[(Int, List[A])]
  ...

  // concrete final (and some private) methods below.
}
```

johnynek (commented):
This is a very quick first pass.

Two other requests:

1. could you put a new title to the PR that summarizes the "what" of this PR?
2. could you add a summary to the comment that explains the "why" of the PR and also link to other relevant past PRs and issues (e.g. normalization issues and PRs are probably related).
3. the PR coverage seems to be at 66% according to the codecov, I'd like that to get to 90% or better if we could. Is that too much work do you think? In my experience, the coverage wins are worth it. I still find bugs when increasing coverage.

snoble:
> This is a very quick first pass.
> 
> Two other requests:
> 
> 1. could you put a new title to the PR that summarizes the "what" of this PR?
> 2. could you add a summary to the comment that explains the "why" of the PR and also link to other relevant past PRs and issues (e.g. normalization issues and PRs are probably related).

Absolutely. I'll write something up tomorrow. 

> 3. the PR coverage seems to be at 66% according to the codecov, I'd like that to get to 90% or better if we could. Is that too much work do you think? In my experience, the coverage wins are worth it. I still find bugs when increasing coverage.

Yes, I started on more coverage and already came across more bugs. I'll keep going.

snoble:
I got the coverage (even though it took me a while) and fixed a lot of bugs along the way. @johnynek let me know what you would like.

cli/src/test/scala/org/bykn/bosatsu/LetFreeEvaluationTest.scala
```diff
@@ -0,0 +1,216 @@
...
+import LetFreeEvaluation.{ComputedValue, LazyValue, ExtEnv, Cache}
+import Value.ExternalValue
+import org.scalatest.Assertion
+
+class LetFreeEvaluationTest extends AnyFunSuite {
+  import PathModule.MainCommand.LetFreeEvaluate
+  import PathModule.MainCommand.PathGen
+  import PathModule.MainCommand.MainIdentifier
+  import PathModule.MainCommand.PackageResolver
+  import PathModule.Output
```
johnynek: nit: can we move these imports to outside the class, also can we group them in scala style using`{Foo, Bar, ...` vs listing them in java style?

cli/src/test/scala/org/bykn/bosatsu/LetFreeEvaluationTest.scala
```diff
@@ -0,0 +1,216 @@
...
+
+  def letFreeTest(
+      fileNames: List[String],
+      packageName: String,
+      altAsserts: List[Output.LetFreeEvaluationResult => Assertion] = Nil
+  ) =
+    NonEmptyList.fromList(packageName.split("/").toList) match {
```
johnynek: what about `PackageName.parse(packageName).get`? Here you are hard coding the parsing again which could potentially change in the future (dots vs /).

cli/src/test/scala/org/bykn/bosatsu/LetFreeEvaluationTest.scala
```diff
@@ -0,0 +1,216 @@
...
+      letFreeTest(List("EmptyPackage"), "Bosatsu/EmptyPackage")
+      fail()
+    } catch {
+      case err: java.lang.Exception => {
+        assert(err.getMessage() == "found no main expression")
+      }
+    }
+  }
+
+  test("euler1 let free evaluate") {
```
johnynek: what about just testing everything in the test workspace? we do that in one of the tests in PathModule so we don't need to update tests each time we add new bosatsu code.

core/src/test/scala/org/bykn/bosatsu/TestUtils.scala
```diff
@@ -166,47 +167,56 @@ object TestUtils {
...
+      }
+    }
   }
 
-  def normalizeTest[A](packages: List[String], mainPackS: String, expectedMode: NormalTestMode[A]) = {
+  def normalizeTest(packages: List[String], mainPackS: String, expectedMode: NormalTestMode) = {
     def inferredHandler(infPackMap: PackageMap.Inferred, mainPack: PackageName): Assertion = {
-      val normPackMap = LetFreePackageMap(infPackMap).hashKey(ne => (ne, ne.serialize))
+      val normPackMap = LetFreePackageMap(infPackMap).letFreePackageMap
       (for {
```
johnynek, line 197: nit: instead of the big parens could we capture the value, and then operate on the value. I know this is let-free but does the implementation also have to be letfree? ;)

test_workspace/ListPat.bosatsu
```diff
@@ -16,5 +16,5 @@ tests = TestSuite("ListPat tests", [
   Assertion(any([False, True]) matches True, "any([False, True])"),
   Assertion(any([True, False]) matches True, "any([True, False])"),
   Assertion(any([True, True]) matches True, "any([True, True])"),
-  Assertion(one_count([0, 0, 1, 0, 1], 0) matches 2, "one_count 1"),
+  Assertion(one_count([0, 0, 1, 0, 1], 0) matches 2, "one_count 1")
```
johnynek: why this chage, btw? seems like trailing comma is a good style, IMO, makes it easier to add the next line.

test_workspace/ExtendedListPattern/ManyGlobs.bosatsu
```diff
@@ -0,0 +1,20 @@
+package ManyGlobs
+
+enum Result:
+  Success(n), Fail(err)
+def fn(lst):
```
johnynek, line 6: nit: newline after the Result definition

core/src/test/scala/org/bykn/bosatsu/TestUtils.scala
```diff
@@ -166,47 +167,56 @@ object TestUtils {
...
   object NormalTestMode {
-    case class TagMode(expected: LetFreeConversion.LetFreeExpressionTag, serialized: Option[String] = None) extends NormalTestMode[LetFreeConversion.LetFreeExpressionTag]
-    case class ExpressionMode(expected: LetFreeExpression, serialized: Option[String] = None) extends NormalTestMode[LetFreeExpression]
-    case class ChildrenMode(expected: Set[LetFreeExpression]) extends NormalTestMode[Set[LetFreeExpression]]
+    case class TagMode(expected: LetFreeConversion.LetFreeExpressionTag) extends NormalTestMode
+    case class ExpressionMode(expected: LetFreeExpression, extraAssertions: List[LetFreeExpression => Assertion]) extends NormalTestMode
+    case class VarSetMode(lambdaCount: Int, vars: Set[Int]) extends NormalTestMode
+  }
+
+  def unwrapLambda(expression: LetFreeExpression, lambdaCount: Int): Option[LetFreeExpression] = {
```
johnynek: nit: outer { not needed.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -2,13 +2,32 @@ package org.bykn.bosatsu
...
+/*
+ * The LetFreeExpression is an attempting to be a stripped down version of a bosatsu expression
+ * to give a semantic representation of the output of the computation. It is not intended to represent
+ * the computation that produces the output, even though it does contain enough information to be
+ * evaluated. As such it can be used to sometimes tell if two expressions will produce the same 
+ * outputs for cacheing purposes. It also makes functions inspectable which means it's possible
+ * to translate them into code for extrenal computation (eg sql statements).
+ * 
+ * As the name implies, it expression eliminates Lets by inlining the values. This creates an issue
+ * for how to represent a recurrsive Lets so we introduce a Recursion expression that is only valid
```
johnynek: typo in recursive

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -24,21 +43,37 @@ sealed abstract class LetFreeExpression {
     def escapeString(unescaped: String) = StringUtil.escape('\'', unescaped)
     this match {
```
johnynek, line 45: can we add a space between the def and this?

johnynek: nit, I usually call methods like this asString or something, maybe something other than serialize which tends to invoke a notion that it could be deserialized.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -24,21 +43,37 @@ sealed abstract class LetFreeExpression {
...
+    case LetFreeExpression.App(fn, arg) => fn.varSet ++ arg.varSet
+    case LetFreeExpression.ExternalVar(_, _, _) => Set()
+    case LetFreeExpression.Match(arg, branches) => branches.map {
+      branch => {
+        val varCount = LetFreePattern.varCount(0, List(branch._1))
+        branch._2.varSet.map(_ - varCount).filter(_ >= 0)
+      }
+    }.foldLeft(arg.varSet){ case (s1, s2) => s1 ++ s2 }
+    case LetFreeExpression.Struct(enum, args, _) => args.foldLeft(Set[Int]()) { case (s, arg) => s ++ arg.varSet }
+    case LetFreeExpression.Literal(_) => Set()
```
johnynek: Set.empty is a bit better.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -24,21 +43,37 @@ sealed abstract class LetFreeExpression {
...
         case Lit.Integer(bigInt) => s"Literal($bigInt)"
       }
       case LetFreeExpression.Recursion(lambda) => s"Recursion(${lambda.serialize})"
     }
   }
+
+  def varSet: Set[Int] = this match {
+    case LetFreeExpression.Lambda(expr) => expr.varSet.collect { case n if n > 0 => n - 1}
+    case LetFreeExpression.App(fn, arg) => fn.varSet ++ arg.varSet
+    case LetFreeExpression.ExternalVar(_, _, _) => Set()
```
johnynek: Set.empty (does not allocate the empty varargs)

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -24,21 +43,37 @@ sealed abstract class LetFreeExpression {
...
+    case LetFreeExpression.Match(arg, branches) => branches.map {
+      branch => {
+        val varCount = LetFreePattern.varCount(0, List(branch._1))
+        branch._2.varSet.map(_ - varCount).filter(_ >= 0)
+      }
+    }.foldLeft(arg.varSet){ case (s1, s2) => s1 ++ s2 }
```
johnynek: s1.union(s2)

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -24,21 +43,37 @@ sealed abstract class LetFreeExpression {
...
       case LetFreeExpression.Recursion(lambda) => s"Recursion(${lambda.serialize})"
     }
   }
+
+  def varSet: Set[Int] = this match {
+    case LetFreeExpression.Lambda(expr) => expr.varSet.collect { case n if n > 0 => n - 1}
+    case LetFreeExpression.App(fn, arg) => fn.varSet ++ arg.varSet
+    case LetFreeExpression.ExternalVar(_, _, _) => Set()
+    case LetFreeExpression.Match(arg, branches) => branches.map {
+      branch => {
```
johnynek: you don't need these inner {

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -24,21 +43,37 @@ sealed abstract class LetFreeExpression {
...
+      case LetFreeExpression.Struct(enum, args, _) => s"Struct($enum,${args.map(_.serialize).mkString(",")})"
       case LetFreeExpression.Literal(toLit) => toLit match {
         case Lit.Str(toStr) => s"Literal('${escapeString(toStr)}')"
         case Lit.Integer(bigInt) => s"Literal($bigInt)"
       }
       case LetFreeExpression.Recursion(lambda) => s"Recursion(${lambda.serialize})"
     }
   }
+
+  def varSet: Set[Int] = this match {
```
johnynek: can we add a comment to what this is, I guess it is the indices representing lambda arguments accessed in this expression.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -24,21 +43,37 @@ sealed abstract class LetFreeExpression {
...
+    case LetFreeExpression.Lambda(expr) => expr.varSet.collect { case n if n > 0 => n - 1}
+    case LetFreeExpression.App(fn, arg) => fn.varSet ++ arg.varSet
```
johnynek: I think union can be slightly better since in principle the Set can use more efficient merges of unions than generic Iterable insertion.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -24,21 +43,37 @@ sealed abstract class LetFreeExpression {
...
       case LetFreeExpression.Literal(toLit) => toLit match {
         case Lit.Str(toStr) => s"Literal('${escapeString(toStr)}')"
         case Lit.Integer(bigInt) => s"Literal($bigInt)"
       }
       case LetFreeExpression.Recursion(lambda) => s"Recursion(${lambda.serialize})"
     }
   }
+
+  def varSet: Set[Int] = this match {
+    case LetFreeExpression.Lambda(expr) => expr.varSet.collect { case n if n > 0 => n - 1}
```
johnynek: this is a really cool feature of these indices, and I guess the same property could be used to do alpha-renaming in the TypedExpression case.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -83,11 +118,12 @@ object LetFreeExpression {
   case class Recursion(lambda: LetFreeExpression) extends LetFreeExpression {
     def maxLambdaVar = lambda.maxLambdaVar
   }
+
 }
 
 sealed abstract class LetFreePattern {
   def escapeString(unescaped: String) = StringUtil.escape('\'', unescaped)
-  def serialize: String = {
+  def serialize: String =
```
johnynek: again, I think a name like asString is more consistent.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -103,13 +139,44 @@ sealed abstract class LetFreePattern {
...
+        varCount(result._1, result._2)
+      }
+      case LetFreePattern.PositionalStruct(name, params, df) => varCount(name.getOrElse(floor).max(floor), params ++ rest)
+      case LetFreePattern.Union(uHead, _) => varCount(floor, uHead :: rest)
+      case LetFreePattern.StrPat(parts) => {
+        val newFloor = (NonEmptyList.of(floor) ++ (parts.collect { case LetFreePattern.StrPart.NamedStr(name) => name+1})).maximum
+        varCount(newFloor, rest)
+      }
+    }
+    case _ => floor
```
johnynek: can we write `case Nil` here just to be clear that we are just doing the normal kind of recursion.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -103,13 +139,44 @@ sealed abstract class LetFreePattern {
...
+        }.toList.mkString(",")
+        s"StrPat($inside)"
+      }
     }
-  }
 }
 
 object LetFreePattern {
+  @tailrec
+  def varCount(floor: Int, patterns: List[LetFreePattern]): Int = patterns match {
```
johnynek: can you comment to give some intuition about how floor works? To a reader, they really need to decompile this code to get an idea what is going on. Also, can it be private?

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -103,13 +139,44 @@ sealed abstract class LetFreePattern {
...
+    case head :: rest => head match {
+      case LetFreePattern.WildCard => varCount(floor, rest)
+      case LetFreePattern.Literal(_) => varCount(0, rest)
+      case LetFreePattern.Var(name) => varCount(floor.max(name + 1), rest)
+      case LetFreePattern.Named(name, pat) => varCount(name + 1, pat :: rest)
```
johnynek: how does this disregard floor completely? I can't imagine how that works. Can you add a comment to explain some intuition there?

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -103,13 +139,44 @@ sealed abstract class LetFreePattern {
...
+      case LetFreePattern.ListPat(parts) => {
+        val result = parts.foldLeft((floor, rest)) {
+          case ((fl, lst), Left(n)) => (fl.max(n.map(_ + 1).getOrElse(fl)), lst)
```
johnynek: can we do a pattern match on n rather than this expression and split it on a couple of lines. I think enumerating the two cases will make it clear: either n is empty or not, and then we take the max in the not empty case. By combining, in one case (empty) you are taking f1 max f1, which isn't wrong, but obscures to the reader IMO.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -103,13 +139,44 @@ sealed abstract class LetFreePattern {
...
+          case ((fl, lst), Right(pat)) => (fl, pat :: lst)
+        }
+        varCount(result._1, result._2)
+      }
+      case LetFreePattern.PositionalStruct(name, params, df) => varCount(name.getOrElse(floor).max(floor), params ++ rest)
```
johnynek: again can we match on the name, I thing that is clearer what is going on.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -103,13 +139,44 @@ sealed abstract class LetFreePattern {
...
+      case LetFreePattern.Union(uHead, _) => varCount(floor, uHead :: rest)
+      case LetFreePattern.StrPat(parts) => {
```
johnynek: no need for `{` in the block.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -103,13 +139,44 @@ sealed abstract class LetFreePattern {
...
 
 object LetFreePattern {
+  @tailrec
+  def varCount(floor: Int, patterns: List[LetFreePattern]): Int = patterns match {
+    case head :: rest => head match {
+      case LetFreePattern.WildCard => varCount(floor, rest)
+      case LetFreePattern.Literal(_) => varCount(0, rest)
+      case LetFreePattern.Var(name) => varCount(floor.max(name + 1), rest)
+      case LetFreePattern.Named(name, pat) => varCount(name + 1, pat :: rest)
+      case LetFreePattern.ListPat(parts) => {
```
johnynek: no need for `{` in the block.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -130,36 +225,59 @@ object LetFreeConversion {
...
+  def structListAsList(lfe: LetFreeExpression): Option[List[LetFreeExpression]] = {
     lfe match {
-      case LetFreeExpression.Struct(0, _) => Matches(Nil)
-      case LetFreeExpression.Struct(1, List(value, tail)) => structListAsList(tail) match {
-        case Matches(lst) => Matches(value :: lst)
-        case notMatch => notMatch
-      }
-      case _ => NotProvable
+      case LetFreeExpression.Struct(0, _, _) => Some(Nil)
+      case LetFreeExpression.Struct(1, List(value, tail), _) => structListAsList(tail).map(value :: _)
```
johnynek: I think `value :: tail :: NIl` is somewhat more efficient actually, because scala can't assume much about pattern matches, I don't think it can optimize these to be the same thing.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -130,36 +225,59 @@ object LetFreeConversion {
...
   case object NoMatch extends PatternMatch[Nothing]
   case object NotProvable extends PatternMatch[Nothing]
 
-  val noop: (LetFreeExpression, PatternEnv) => PatternMatch[PatternEnv] = { (_, env) => Matches(env) }
-  val neverMatch: (LetFreeExpression, PatternEnv) => PatternMatch[Nothing] = { (_, _) => NoMatch }
+  def noop[T]: (T, PatternEnv[T]) => PatternMatch[PatternEnv[T]] = { (_, env) => Matches(env) }
+  def neverMatch[T]: (T, PatternEnv[T]) => PatternMatch[Nothing] = { (_, _) => NoMatch }
 
-  def structListAsList(lfe: LetFreeExpression): PatternMatch[List[LetFreeExpression]] = {
+  def structListAsList(lfe: LetFreeExpression): Option[List[LetFreeExpression]] = {
```
johnynek: can you add a comment as to what this is doing? It's easy for me to forget what the 0 and 1 mean below.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
johnynek: oh wait, this is literally decoding a bosatsu list for the values, right? Since List is so overloaded, can you add that to the comment?

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -173,89 +291,123 @@ object LetFreeConversion {
...
-            { (arg, acc) =>
-              arg match {
-                case LetFreeExpression.Struct(0, List()) => Matches(acc)
-                case LetFreeExpression.Struct(_, _) => NoMatch
-                case _ => NotProvable
-              }
+      case lp@LetFreePattern.ListPat(_) => {
+        sealed abstract class StructListResult
+        case object Empty extends StructListResult
+        case class Cons(h: T, t: StructList) extends StructListResult
```
johnynek, line 325: this is just a `List[T]` right? can you comment on why it is introduced? Also, I would put a high bar on a local ADT definition like this (that does not live beyond some local scope).

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -173,89 +291,123 @@ object LetFreeConversion {
...
+        case class StructList(asT: T) {
+          lazy val parts = {
+            val strct = toStruct(asT, DataFamily.Enum)
+            strct.map { 
+              case (1, List(h, t)) => Cons(h, StructList(t))
```
johnynek: `h :: t :: Nil`

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -173,89 +291,123 @@ object LetFreeConversion {
...
+              case (0, Nil) => Empty
             }
```
johnynek, line 333: this isn't a total match. Can you put a `case _ => sys.error...` to explain you hit something expected to be unreachable instead leaving the MatchError, which is what will happen now?

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -173,89 +291,123 @@ object LetFreeConversion {
...
+            case NoMatch => NoMatch
+            case NotProvable => NotProvable
+          }
+        }
+      }
+
       case LetFreePattern.Union(h, t) =>
         // we can just loop expanding these out:
-        def loop(ps: List[LetFreePattern]): (LetFreeExpression, PatternEnv) => PatternMatch[PatternEnv] =
+        def loop(ps: List[LetFreePattern]): (T, PatternEnv[T]) => PatternMatch[PatternEnv[T]] =
```
johnynek, line 433: I wonder if you could make this change: making the type parameter generic, a separate change and sent that by itself to make this change easier to digest (since the current PR is quite large). Would that work?

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -316,18 +468,50 @@ object LetFreeConversion {
...
             result
         }
+        case LetFreePattern.StrPat(parts) =>
+          val listParts: List[LetFreePattern.ListPart] = parts.toList.flatMap {
+            case LetFreePattern.StrPart.WildStr => List(Left(None))
+            case LetFreePattern.StrPart.NamedStr(n) => List(Left(Some(n)))
+            case LetFreePattern.StrPart.LitStr(str) => str.toList.map(c => Right(LetFreePattern.Literal(Lit.Str(c.toString))))
+          }
+          val listMaybeBind = maybeBind(LetFreePattern.ListPat(listParts))
+          (v, env) => toLitValue(v) match {
```
johnynek, line 508: to ease reading, can you add a space between the val and this lambda you are returning?

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -130,36 +225,59 @@ object LetFreeConversion {
...
+    case _ => None
+  }
+  val neToStruct: (LetFreeExpression, DataFamily) => Option[(Int, List[LetFreeExpression])] = {
+    case (LetFreeExpression.Struct(enum, args, _), df) => Some((enum, args))
+    case _ => None
+  }
+  val neToList: LetFreeExpression => Option[List[LetFreeExpression]] = structListAsList(_)
+  val neFromList: List[LetFreeExpression] => LetFreeExpression = listAsStructList(_)
+
+  abstract class MaybeBind[T](pat: LetFreePattern) {
```
johnynek, line 296: can you add a comment what this class is all about and what T represents? e.g. what do we use it for in different cases here? It seems partially generic in that it still depends on LetFreePattern.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -316,18 +468,50 @@ object LetFreeConversion {
...
+                  }
+                  Matches(env ++ strEnv)
+                }
+                case noMatch => noMatch
+              }
+            }       
+    }
   }
 
+  case class LetFreeExpressionMaybeBind(pat: LetFreePattern) extends MaybeBind[LetFreeExpression](pat) {
```
johnynek, line 527: could this be a separate PR after the genereic MaybeBind is added?

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -316,18 +468,50 @@ object LetFreeConversion {
...
+    def toStruct(t: LetFreeExpression, df: DataFamily): Option[(Int, List[LetFreeExpression])] = neToStruct(t, df)
+    def toList(t: LetFreeExpression): Option[List[LetFreeExpression]] = neToList(t)
+    def fromList(lst: List[LetFreeExpression]): LetFreeExpression = neFromList(lst)
+    def fromString(str: String): LetFreeExpression = LetFreeExpression.Literal(Lit.Str(str))
+    def maybeBind(pat: LetFreePattern): (LetFreeExpression, PatternEnv[LetFreeExpression]) => PatternMatch[PatternEnv[LetFreeExpression]] = LetFreeExpressionMaybeBind(pat).apply(_, _)
+  }
   def findMatch(m: LetFreeExpression.Match) =
     m.branches.collectFirst(Function.unlift( { case (pat, result) =>
-      maybeBind(pat).apply(m.arg, Map()) match {
+      LetFreeExpressionMaybeBind(pat).apply(m.arg, IntMap.empty) match {
```
johnynek, line 537: could the switch from Map to IntMap be a separate PR?

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -336,15 +520,15 @@ object LetFreeConversion {
...
+      case extVar @ ExternalVar(_, _, _) => extVar
       // check for a match reduction opportunity (beta except for Match instead of lambda)
       case Match(arg, branches) =>
         Match(normalOrderReduction(arg), branches.map{ case (p, s) => (p, normalOrderReduction(s))})
       case lv @ LambdaVar(_)  => lv
       // check for eta reduction
       case Lambda(expr)       =>
         Lambda(normalOrderReduction(expr))
-      case Struct(enum, args) => Struct(enum, args.map(normalOrderReduction(_)))
+      case Struct(enum, args, df) => Struct(enum, args.map(normalOrderReduction(_)), df)
```
johnynek, line 561: could adding data family be a separate PR?

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -612,12 +773,19 @@ case class LetFreePackageMap(pm: PackageMap.Inferred) {
...
     val names = pattern.names.collect { case b: Identifier.Bindable => b }
-    val lambdaVars = names ++ env._2
-    val nextEnv = (env._1 ++ lambdaVars.zipWithIndex
+
+    val lv = names.zipWithIndex
       .reverse
       .toMap
-      .mapValues(idx => LetFreeExpressionTag(LetFreeExpression.LambdaVar(idx), Set[LetFreeExpression]())),
-      lambdaVars)
+      .mapValues(idx => LetFreeExpressionTag(LetFreeExpression.LambdaVar(idx), Set[LetFreeExpression]()))
```
johnynek: Set.empty is marginally better.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -629,7 +797,7 @@ case class LetFreePackageMap(pm: PackageMap.Inferred) {
     Program[TypeEnv[Variance], TypedExpr[(Declaration, LetFreeConversion.LetFreeExpressionTag)], Any]] = {
     for {
       lets <- pack.program.lets.map {
-        case (name, recursive, expr) => letFreeConvertNameKindLet(name, recursive, expr, pack, (Map(), Nil)).map((name, recursive, _))
+        case (name, recursive, expr) => letFreeConvertNameKindLet(name, recursive, expr, pack, Map()).map((name, recursive, _))
```
johnynek: Map.empty

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -651,7 +819,7 @@ case class LetFreePackageMap(pm: PackageMap.Inferred) {
 
   private type SourceRef = Ref[Declaration]
   private type ResultingRef = Ref[(Declaration,  LetFreeExpressionTag)]
-  private type Env = (Map[Identifier, LetFreeExpressionTag], List[Identifier])
```
johnynek, line 654: could this change be independent?

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -665,11 +833,11 @@ case class LetFreePackageMap(pm: PackageMap.Inferred) {
...
           // we reset the environment in the other package
           for {
-            imported <- norm(pm.toMap(from.name), orig, t, (Map.empty, Nil))
+            imported <- norm(pm.toMap(from.name), orig, t, Map.empty)
             lfeTag = getTag(imported)._2
           } yield Left((item, (t, lfeTag)))
-        case NameKind.ExternalDef(pn, n, scheme) =>
-          val lfeTag = LetFreeExpressionTag(LetFreeExpression.ExternalVar(pn, n), Set())
+        case NameKind.ExternalDef(pn, n, defType) =>
+          val lfeTag = LetFreeExpressionTag(LetFreeExpression.ExternalVar(pn, n, defType), Set())
```
johnynek: Set.empty

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -686,11 +854,10 @@ case class LetFreePackageMap(pm: PackageMap.Inferred) {
...
           recursive match {
             case RecursionKind.Recursive =>
-              val lambdaVars = name :: env._2
-              val nextEnv = (env._1 ++ lambdaVars.zipWithIndex
-                .toMap
-                .mapValues(idx => LetFreeExpressionTag(LetFreeExpression.LambdaVar(idx), Set[LetFreeExpression]())),
-                lambdaVars)
+              val nextEnv = (env - name).mapValues { case ExpressionKeyTag(lfe, children) =>
+                LetFreeExpressionTag(LetFreeConversion.incrementLambdaVars(lfe, 0), children)
+              } + (name -> LetFreeExpressionTag(LetFreeExpression.LambdaVar(0), Set()))
```
johnynek: Set.empty

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
```diff
@@ -726,7 +893,7 @@ case class LetFreePackageMap(pm: PackageMap.Inferred) {
         if (params == 0) expr
         else loop(params - 1, LetFreeExpression.Lambda(expr))
 
-        loop(arity, LetFreeExpression.Struct(enum, ((arity - 1) to 0 by -1).map(LetFreeExpression.LambdaVar(_)).toList))
+        loop(arity, LetFreeExpression.Struct(enum, ((arity - 1) to 0 by -1).map(LetFreeExpression.LambdaVar(_)).toList, dt.dataFamily))
```
johnynek: putting .iterator before .map will make this a bit more efficient.

core/src/test/scala/org/bykn/bosatsu/ExprFnTest.scala
```diff
@@ -0,0 +1,437 @@
...
+  }
+  test("expression filter function that's a bit more complicated'") {
+    val counter = mutable.Map[Boolean, Int]()
+    letFreeEvaluateTest(
+      List("""
+package Ext/ExprListFilter
+
+operator < = \a, b -> a.cmp_Int(b) matches LT
+
+operator || = \a, b -> match (a,b):
```
johnynek: `True if a else b` is a bit more concise.

core/src/test/scala/org/bykn/bosatsu/LetFreeTest.scala
```diff
@@ -56,14 +68,87 @@ package Recur/Some
...
 
 out = foo
-"""
-      ), "Recur/Some", Recursion(Lambda(Lambda(Match(LambdaVar(0),NonEmptyList.of(
-        (PositionalStruct(Some(0), Nil),Struct(1,List(Literal(Str("a")), Struct(1,List(Literal(Str("b")), Struct(1,List(Literal(Str("c")), Struct(0,List())))))))),
-      (PositionalStruct(Some(1), List(WildCard, Var(0))),Lambda(Struct(1,List(Literal(Str("zero")), App(LambdaVar(2),LambdaVar(0))))))
-      )))))
+"""),
+      "Recur/Some",
+      Recursion(
```
johnynek: I get really worried about these tests. In the past, minor changes elsewhere in the code have changed these results, which can then only be updated by very carefully building the AST.

An alternate idea for such gold-set tests would be to assert that the toString or asString or whatever hashes to a given value. Then to update the test you only need to update the hash. When the hash fails, you can print what the original AST is and leave it in the comment as an expectation for someone to update, but the update process becomes much easier.

Also, is this a content change or just a formatting change? If it is a formatting change, can you make it a separate PR?

core/src/test/scala/org/bykn/bosatsu/LetFreeTest.scala
```diff
@@ -56,14 +68,87 @@ package Recur/Some
...
+                )
+              )
+            )
+          )
+        )
+      ),
+      List({ lfe: LetFreeExpression =>
+        assert(
+          lfe.serialize ==
+            "Recursion(Lambda(Lambda(Match(LambdaVar(0),PositionalStruct(0,),Struct(1,Literal('a'),Struct(1,Literal('b'),Struct(1,Literal('c'),Struct(0,)))),PositionalStruct(1,StrPat(LitStr(a: ),NamedStr(0)),Var(1)),Lambda(Lambda(Struct(1,LambdaVar(0),App(LambdaVar(3),LambdaVar(1))))),PositionalStruct(1,WildCard,Var(0)),Lambda(Struct(1,Literal('zero'),App(LambdaVar(2),LambdaVar(0))))))))",
```
johnynek: as I mentioned above, can we just hash this and print the hex or something? Leave the string in a comment?

core/src/test/scala/org/bykn/bosatsu/LetFreeTest.scala
```diff
@@ -77,11 +162,41 @@ def foldLeft(lst: List[a], item: b, fn: b -> a -> b) -> b:
...
 
 out = foldLeft
-"""
-      ), "Recur/FoldLeft", Recursion(Lambda(Lambda(Match(LambdaVar(0),NonEmptyList.of(
-        (PositionalStruct(Some(0),List()),Lambda(Lambda(LambdaVar(1)))),
-        (PositionalStruct(Some(1),List(Var(0), Var(1))),Lambda(Lambda(Lambda(Lambda(App(App(App(LambdaVar(5),LambdaVar(3)),App(App(LambdaVar(0),LambdaVar(1)),LambdaVar(2))),LambdaVar(0)))))))
-      )))))
+"""),
+      "Recur/FoldLeft",
+      Recursion(
```
johnynek: same comment

johnynek (commented):
Congrats on such a big piece of work! I think there is a lot in common with this work and general symbolic evaluation which would be needed to solve compile time execution, so that's interesting to see.

My main issues are:

1. this PR is very very large. It is about 7 to 8 times larger than the ideal change is, IMO. This makes it very difficult and very time consuming to review, and no doubt lowers the quality and value of my review. I would really love to see if we can bring changes off of this to merge in that reduce the size and scope.

2. I continue to be a bit concerned that we basically have two execution engines mostly, it seems to me, so that each of us can work in our own space. This isn't a blocker, but the goals you want, are goals I want in general, so in the best case I will wind up reimplementing most of this, which is a shame. The goal of normalization (including alpha normalization) and support for expression level caching (which I think is a goal if your worksheet system) is something I want in general. And as I mentioned, the ability to symbolically evaluate as much as possible is needed for the compile time macro (the meta-ocaml inspired stuff).

Anyway, I'd love to just aggressively see how many PRs can be pulled out of this. The smaller the better. One liners, even, no problem.... :)

snoble:
> Congrats on such a big piece of work! I think there is a lot in common with this work and general symbolic evaluation which would be needed to solve compile time execution, so that's interesting to see.
> 
> My main issues are:
> 
> 1. this PR is very very large. It is about 7 to 8 times larger than the ideal change is, IMO. This makes it very difficult and very time consuming to review, and no doubt lowers the quality and value of my review. I would really love to see if we can bring changes off of this to merge in that reduce the size and scope.
> 2. I continue to be a bit concerned that we basically have two execution engines mostly, it seems to me, so that each of us can work in our own space. This isn't a blocker, but the goals you want, are goals I want in general, so in the best case I will wind up reimplementing most of this, which is a shame. The goal of normalization (including alpha normalization) and support for expression level caching (which I think is a goal if your worksheet system) is something I want in general. And as I mentioned, the ability to symbolically evaluate as much as possible is needed for the compile time macro (the meta-ocaml inspired stuff).
> 
> Anyway, I'd love to just aggressively see how many PRs can be pulled out of this. The smaller the better. One liners, even, no problem.... :)

Just want to reply really quicky to say I totally agree and clarify my use case a bit (in case there's a better solution).

The main thing I want from this for my notebooks use case is to be able to build language extensions that can inspect functions in particular. This is helpful for caching ("this `map` has been done before and is cached and we can verify that because we can inspect the function") but also for being able to translate part of the work of functions to other systems ("this function being used to filter this db result set can just be translated into sql and added to the db query").

I would love to be able to have executor that builds off of `Matchless` and has the expressions as tags on the values that get passed to the external functions. The let free execution engine was super useful for finding a bunch of bugs in the let free expression, and it was convenient to add on this use case. But if I can get this use case without it that would be way better.

I'll go through the comments and make some smaller PRs. I am definitely open to other paths to this use case that removes redundancies, even if it means throwing out big chunks of the code here.

cli/src/test/scala/org/bykn/bosatsu/LetFreeEvaluationTest.scala
snoble: I'll try to copy that.

test_workspace/ListPat.bosatsu
snoble: this was probably me commenting and uncommenting assertions to try to narrow in on the bug and I deleted too much one time.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
snoble: It's a bug! I expect it's never found because the index of Named was always high? Thank you for finding it.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
snoble: It's not quite a `List[T]`. `T` is something that can possibly be converted to a head and a tail; eg a `Value` or a `LetFreeExpression`. But in the case of a `LetFreeExpression`, it might not need runtime information to do the conversion and this might be running at compile time. So this structure lets me check if a pattern can be ruled in or out, even if we can't evaluate the tail. There's probably an existing way to do this in cats that I don't know about.

does that make sense?

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
snoble: I think so. I could just have a PR with no LetFreeEvaluation and then this would only need LetFreeExpression.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
snoble: Totally. Of the code in here I think MaybeBind is probably the most interesting.

core/src/main/scala/org/bykn/bosatsu/LetFree.scala
snoble: It would just mean it was buggy between the PRs... but that seems fine.

core/src/test/scala/org/bykn/bosatsu/LetFreeTest.scala
snoble: They aren't particularly great tests because the expressions are so big to be totally opaque. They are usually here to test some edge of normalization.

Let me see if I can make something better.

I believe in this case there was a bug and the expression did change in some subtle way.

PR #517, opened 2020-08-17
Refactor FfiCall a bit

johnynek:
please take a look @snoble and let me know if you see any problems with this approach.


core/src/main/scala/org/bykn/bosatsu/LetFreeEvaluation.scala
```diff
@@ -32,4 +32,12 @@ object LetFreeEvaluation {
 
   type Cache = Option[CMap[String, (Future[Value], Type)]]
   type ToLFV = Option[LetFreeValue => Future[Value]]
+
+
+  def exprFn(wrapper: (LetFreeEvaluation.LetFreeValue, rankn.Type, LetFreeEvaluation.Cache, LetFreeEvaluation.ToLFV) => Any): FfiCall = {
```
johnynek, line 37: we move the function from Ffi here so we keep value agnostic of as much as possible.

core/src/main/scala/org/bykn/bosatsu/Externals.scala
```diff
@@ -1,69 +1,5 @@
...
-          FnValue { e3 =>
-            fn(e1, e2, e3)
-          }
-        }
-      }
-
-    def call(t: rankn.Type): Value = evalFn
-  }
-
-  final case class ExprFn(wrapper: (LetFreeEvaluation.LetFreeValue, rankn.Type, LetFreeEvaluation.Cache, LetFreeEvaluation.ToLFV) => Any) extends FfiCall {
```
johnynek, line 42: this is moved to a function in LetFree

core/src/main/scala/org/bykn/bosatsu/Value.scala
```diff
@@ -120,18 +120,20 @@ object Value {
...
 
-  trait FnValueArg {
-    def toFn: Value => Value
+  class FnValue(fnValueArg: FnValue.Arg) extends Value {
+    val arg: FnValue.Arg = fnValueArg
   }
 
-  case class SimpleFnValue(toFn: Value => Value) extends FnValueArg
+  object FnValue {
+    trait Arg {
```
johnynek, line 128: minor nit: I prefer to use objects as namespaces than use longer names.

core/src/main/scala/org/bykn/bosatsu/FfiCall.scala
```diff
@@ -0,0 +1,62 @@
+package org.bykn.bosatsu
+
+sealed abstract class FfiCall {
```
johnynek, line 3: move this to a separate file since I keep not finding it when I look for it.

johnynek:
see #496 for related change.

core/src/main/scala/org/bykn/bosatsu/LetFreeEvaluation.scala
snoble: This is much nicer here

snoble:
This feels much nicer and can probably accomodate future experimental evaluators better. Thank you!

johnynek:
going to merge even though the patch isn't well covered (because it is staged to be covered by #513

PR #520, opened 2020-08-18
Fix an issue with top-level recursives

johnynek:
increase the test coverage of compiling to python.


johnynek:
codecov can be pretty frustrating with scalacheck. Randomly we get slightly different coverage on each run, but codecov will mark it red unless all are green.

PR #531, opened 2020-08-21
Implement Predef Dict via AvlTree

### LINKED ISSUE #432: implement AVL-tree and replace External Dict implementation with it.
Since Dict is an opaque external type, we can't easily recurse on it. We can convert it to a list and recurse on that, which is possibly fine, but a nice exercise would be showing that we could implement an AVL-tree in bosatsu.

It would also open up more optimizations later since we can see the structure of the nodes and use them in constant propagation and inlining.

johnynek:
close #432 

The motivation to work on this now is I am trying to minimize the amount of code I need to implement in python in order to run.

Unfortunately, this does push around work since `Dict[String, v]` is a type the JSON code knows about, so we still need to convert those dicts into Values.


core/src/main/resources/bosatsu/predef.bosatsu
```diff
@@ -127,15 +132,235 @@ def range_fold(inclusiveLower: Int, exclusiveUpper: Int, init: a, fn: a -> Int -
...
+#############
+# Support for built-in testing:
+#############
 enum Test:
   Assertion(value: Bool, message: String)
   TestSuite(name: String, tests: List[Test])
 
-external def trace(prefix: String, item: a) -> a
+#############
+# AVL trees used to implement dictionaries
```
johnynek, line 156: this code is duplicative with AvlTree.bosatsu, but debugging the predef is very painful since we parse and compile it once on initialization. Having it copied makes it easier to have tests in another file.

core/src/main/scala/org/bykn/bosatsu/Value.scala
```diff
@@ -287,19 +287,94 @@ object Value {
...
+              sys.error(s"ill-typed in String Dict order: $tup1, $tup2")
+              // $COVERAGE-ON$
+            }
+        }
+      }
+
     def fromStringKeys(kvs: List[(String, Value)]): Value = {
-      implicit val ord: Ordering[Value] =
-        implicitly[Ordering[String]].on { case Str(str) => str }
+      val allItems: Array[(String, Value)] = kvs.toArray
```
johnynek: these have to be sorted

PR #533, opened 2020-08-21
Make missing externals an error on transpile

johnynek:
this also implements the remaining python externals in predef.


core/src/main/scala/org/bykn/bosatsu/codegen/python/PythonGen.scala
```diff
@@ -240,25 +240,13 @@ object PythonGen {
...
             case IfElse(crest, er) =>
               // preserve IfElse chains
               IfElse(head :: crest, er)
-            case nestX: Expression =>
-              t match {
-                case tx: Expression =>
-                  Ternary(tx, cx, nestX)
-                case _ =>
-                  IfElse(NonEmptyList(head, Nil), nestX)
-              }
```
johnynek, line 261: this is a bad merge it looks like.

Issue #535, opened 2020-08-22
make unittest tests out of Bosatsu tests in python output


johnynek:
unittest is built in to python, and we can run the tests with `python -m unittest module_name`

So, the idea would be to take the final test value (like the bosatsu test command does) and builds `unittest.TestCase` out of it.

If we have a TestSuite, we can take the strings and turn them into method names of the TestCase class and assertTrue all the items in the list (or expand out inner lists and recurse).

To do this, we need to not lose track of the types in Matchless.Expr, that said, since we only need to last Test value, if it exists, we can do that from the TypedExpr without having to preserve types in Matchless.Expr.

PR #541, opened 2020-08-25
Add test generation for python, and run them

johnynek:
this automatically generates `unittest` tests from the Bosatsu tests and runs them.

This allows us to test all the python code from the existing bosatsu tests.


johnynek:
Close #535

PR #542, opened 2020-08-26
Remove bazel build/rules 527

### LINKED ISSUE #527: remove bazel build
currently we have both bazel rules and a bazel build of the code.

Having a bazel build of the code makes build updates twice as hard. To be clear, I want to retain the bazel rules for bosatsu (and expand them to support python and other code generation). I don't want to maintain two builds of the main code.

Plan:

1. remove the bazel build of the code.
2. change the bazel bosatsu rules to use *either* a fat jar produced by sbt (or potentially coursier) or a graalvm native image.

johnynek:
close #527


johnynek:
the test that timed out wasn't related to the bazel tests (and sometimes happens). The travis CI passed which is the one that knew about bazel.

The test coverage change is just noise due to scalacheck.

PR #543, opened 2020-08-26
Test the native image build

johnynek:


johnynek:
the bazel build was still being tested here, but it was removed from master, which is why the test failed.

Going to merge.

PR #556, opened 2020-08-27
Reduce size of scalajs scalachecks (see #548)

### LINKED ISSUE #548: reduce the size of js tests
they take too long.

Use a isScalaJs flag to use a lot fewer tests cases for scalacheck should work.

johnynek:
close #548


johnynek:
this uses crossproject to reduce the size of some of the longer scalachecks.

Since the logic is the same between JVM and JS, one can argue that we shouldn't need to test as extensively on scalajs (of course, that assumes scalajs works exactly as scala-jvm which isn't really true).

Issue #471, opened 2020-07-04
making package name and imports more pythonic


johnynek:
One significant deviation of bosatsu from python is that we call "modules" by the name "package" and we normally write the package name into the file.

Next, we deviate from the import syntax of python somewhat needlessly:
```
import Foo/Bar [ fn1, fn2 ]
```

In python this would be:
```python
from foo.bar import fn1, fn2
```

I'm not sure why I bothered to have this deviation, but it seems like maybe calling "package" "module" and using python import syntax is a good idea.

About the module name being in the file, the motivation for this was to make only the content needed to compile, but I can't think of a real issue with that. you can always attach a virtual path/module name to the content before you compile. So, I could remove the explicit package syntax and like python just use the path relative to a root compilation path (I already have this implemented as #375 #10).

Finally, types and constructors are capitalized and in a different name space from lower case value (which are bindable names). For packages it doesn't really matter since we require all external names to be declared in bosatsu (something I like because it makes tooling easier as it makes the dependency graph more explicit). So we could use lower case names as python does for module names. This would remove a difference with python.

After that, the main differences with python becomes:
1. a shorter lambda syntax (but we could also support python's syntax in addition).
2. if is always an expression (ternary or not).
3. match exists and is an expression (python may get match working almost identical to this match: https://www.python.org/dev/peps/pep-0622/ but it is not an expression).
4. no while in bosatsu
5. return isn't needed as the last value is the return value (although we could add a return statement that is just identity to make the code read more like python).
6. the struct/enum syntax deviates from dataclass, but I think this pays its way, and I'm less interested in changing.

@snoble do you have any views here?

snoble:
I've been wondering about 2. a lot. Is it reasonable to make non-ternary `if`s not an expression? ie any assignments that happens on both branches of the if statement becomes a return value. eg

```
if b > a:
  x = 1
  y = 2
  z = 0
else:
  x = 3
  y = 4
```
is equivalent to
```
(x,y) = match (b > a):
  True:
      x = 1
      y = 2
      z = 0
      (x,y)
  False:
      x = 3
      y = 4
      (x,y)
```

it's hard for me to imagine how a `while` would work without being extremely constrained. probably should just have a required `return`. similarly seems right to just use python's import syntax.

johnynek:
This is totally doable and similar to the idea for #99 basically: these kinds of constructs become let bindings.

The linter should error in the above on z, since it would become an unused let binding.

[...]

PR #565, opened 2020-08-28
Support python-like import/from syntax for imports and prefer it

johnynek:
change the import syntax to:

```
from package import foo
```

The export syntax becomes just 
```
export foo, bar
```
If you need multiline, you can use parens:
```
from package import (
    foo,
    bar,
    baz as quux,
)
```

Similarly with exports.


johnynek:
part of #471

Issue #573, opened 2020-09-25
recursion checking is failing unexpectedly


snoble:
the code
```
package VarSet/Recursion

enum Thing:
  Thing1, Thing2(a: Int, t: Thing)

def bar(y, _: String, x):
  recur x:
    Thing1: y
    Thing2(i, t): bar(i, "boom", t)

out = bar
```
gives a surpring error
```
error:
error: in file: /file, package VarSet/Recursion, recursion in bar is not on a name (expect a name which is exactly a arg to the def)
7|  recur x:
8|    Thing1: y
9|    Thing2(i, t): bar(i, "boom", t)
                           ^^^^^^
```

johnynek:
huh, that's weird. It seems the `_` is causing some issue, changing a bit to below works.
```
package VarSet/Recursion

enum Thing:
  Thing1, Thing2(a: Int, t: Thing)

def bar(y, z: String, x):
  recur x:
    Thing1: z
    Thing2(_, t): bar(y, "boom", t)

out = bar
```

PR #575, opened 2020-10-01
Fix issue 573, incorrect recursion checking with pattern arguments to defs

johnynek:
close issue #573 

The issue was that we were just totally broken for pattern arguments to defs.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
@@ -40,10 +40,10 @@ object DefRecursionCheck {
     def region = decl.region
     def message = "unexpected recur: may only appear unnested inside a def"
   }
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
```diff
@@ -398,9 +398,9 @@ object DefRecursionCheck {
      */
     def checkDef[A](state: State, defstmt: DefStatement[Pattern.Parsed, (OptIndent[Declaration], A)]): Res = {
       val body = defstmt.result._1.get
-      val args = defstmt.args.flatMap(_.names)
-      val state1 = state.inDef(defstmt.name, args)
-      checkForIllegalBinds(state, defstmt.name :: args, body) {
+      val nameArgs = defstmt.args.flatMap(_.names)
```
johnynek, line 415: this was just totally wrong. This creates a list of names, but it isn't the parameters in order.

Issue #545, opened 2020-08-27
update to fastparse 2 or move to a new parsing library


johnynek:
fastparse 1 is blocking updating the scala protobuf library.

I'm a bit worried about fastparse 2 being macro based and not being able to update to scala 3.

I've threatened to write another parsing library that doesn't use macros and use that. Maybe I'll do that within this project following the the code I wrote for this talk:

https://www.youtube.com/watch?v=kTyvu0uxPzg

johnynek:
this is also blocking scalajs 1.0.0

ngbinh:
can't wait for this. I am already on scala.js 1.2.0

johnynek:
#588 updates to scalajs 1.2.0.

PR #576, opened 2020-10-02
Add parsing module

johnynek:
this is to address #545 

for a simple benchmark, this is as fast as fastparse:
```
[info] Do not assume the numbers tell you what you want them to tell.
[info] Benchmark                       Mode  Cnt  Score   Error  Units
[info] ParserBench.benchBigIntsFP     thrpt    3  1.588 ± 0.578  ops/s
[info] ParserBench.benchBigIntsLocal  thrpt    3  1.553 ± 0.489  ops/s
[success] Total time: 100 s (01:40), completed Oct 2, 2020 12:53:44 PM
```

### DIFF
--- a/parser/src/main/scala/org/bykn/bosatsu/parser/Parser.scala
+++ b/parser/src/main/scala/org/bykn/bosatsu/parser/Parser.scala
@@ -143,21 +143,8 @@ sealed abstract class Parser1[+A] extends Parser[A] {
 
 object Parser extends ParserInstances {
   sealed abstract class Expectation {
-    import Expectation._
-
     def offset: Int
     def toError: Error = Error.MissedExpectation(this)
-
-    def translateOffset(inc: Int): Expectation = {
-      this match {
-        case Str(o, str) => Str(o + inc, str)
-        case InRange(o, l, c) => InRange(o + inc, l, c)
-        case Failure(o) => Failure(o + inc)
-        case StartOfString(o) => StartOfString(o + inc)
-        case EndOfString(o, length) => EndOfString(o + inc, length)
-        case Length(o, expected, actual) => Length(o + inc, expected, actual)
-      }
-    }
   }
 
   object Expectation {
@@ -174,12 +161,6 @@ object Parser extends ParserInstances {
     def expected: NonEmptyList[Expectation]
     def offsets: NonEmptyList[Int] =
       expected.map(_.offset).distinct
-
-    def translateOffset(inc: Int): Error =
-      this match {
-        case Error.MissedExpectation(x) => Error.MissedExpectation(x.translateOffset(inc))
-        case Error.Combined(es) => Error.Combined(es.map(_.translateOffset(inc)))
-      }
   }
   object Error {
     case class MissedExpectation(expectation: Expectation) extends Error {
@@ -221,13 +202,15 @@ object Parser extends ParserInstances {
     Impl.Expect(str)
 
   def oneOf1[A](parsers: List[Parser1[A]]): Parser1[A] = {
+    // TODO: we can implement a union on CharIn which will be more
+    // efficient
     @annotation.tailrec
     def flatten(ls: List[Parser1[A]], acc: List[Parser1[A]]): Parser1[A] =
       ls match {
         case Nil =>
           acc match {
             case h :: Nil => h
-            case other => Impl.OneOf1(other.reverse)
+            case other => Impl.OneOf1(other.reverse.distinct)
           }
         case Impl.OneOf1(ps) :: rest =>
           flatten(ps ::: rest, acc)
@@ -245,9 +228,11 @@ object Parser extends ParserInstances {
         case Nil =>
           acc match {
             case h :: Nil => h
-            case other => Impl.OneOf(other.reverse)
+            case other => Impl.OneOf(other.reverse.distinct)
           }
         case Impl.OneOf(ps) :: rest =>
+          flatten(ps ::: rest, acc)
+        case Impl.OneOf1(ps) :: rest =>
           flatten(ps ::: rest, acc)
         case notOneOf :: rest =>
           flatten(rest, notOneOf :: acc)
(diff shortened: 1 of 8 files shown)


parser/src/main/scala/org/bykn/bosatsu/parser/Parser.scala
```diff
@@ -0,0 +1,938 @@
...
+   */
+  def expect(str: String): Parser1[Unit] =
+    Impl.Expect(str)
+
+  def oneOf1[A](parsers: List[Parser1[A]]): Parser1[A] = {
+    @annotation.tailrec
+    def flatten(ls: List[Parser1[A]], acc: List[Parser1[A]]): Parser1[A] =
+      ls match {
+        case Nil =>
+          acc match {
```
johnynek, line 211: We could call distinct here I think, which would help when composing possibly overlapping unions.

parser/src/main/scala/org/bykn/bosatsu/parser/Parser.scala
```diff
@@ -0,0 +1,938 @@
...
+
+    flatten(parsers, Nil)
+  }
+
+  def oneOf[A](ps: List[Parser[A]]): Parser[A] = {
+    @annotation.tailrec
+    def flatten(ls: List[Parser[A]], acc: List[Parser[A]]): Parser[A] =
+      ls match {
+        case Nil =>
+          acc match {
```
johnynek, line 229: Distinct here.

parser/src/main/scala/org/bykn/bosatsu/parser/Parser.scala
```diff
@@ -0,0 +1,938 @@
...
+   */
+  def length(len: Int): Parser1[String] =
+    Impl.Length(len)
+
+  /**
+   * Repeat this parser 0 or more times
+   * note: this can wind up parsing nothing
+   */
+  def rep[A](p1: Parser1[A]): Parser[List[A]] =
+    Impl.Rep(p1)
```
johnynek, line 263: We could check that Rep isn't directly nesting. rep of rep won't parse any more, the inner one will scoop everything up.

This is especially useful potentially after a void.

parser/src/main/scala/org/bykn/bosatsu/parser/Parser.scala
```diff
@@ -0,0 +1,938 @@
...
+  def charIn(c0: Char, cs: Char*): Parser1[Char] =
+    charIn1(c0 :: cs.toList)
+
+  def charIn1(cs: Iterable[Char]): Parser1[Char] = {
+    val ary = cs.toArray
+    java.util.Arrays.sort(ary)
+    Impl.CharIn(ary(0).toInt, BitSetUtil.bitSetFor(ary), Impl.rangesFor(ary))
+  }
+
+  def charIn(cs: Iterable[Char]): Parser[Char] =
```
johnynek: Add a comment that if cs is empty this parser always fails.

That should still be a Parser1 though.

parser/src/main/scala/org/bykn/bosatsu/parser/Parser.scala
```diff
@@ -0,0 +1,938 @@
...
+    if (cs.isEmpty) fail
+    else charIn1(cs)
+
+  def charWhere(fn: Char => Boolean): Parser[Char] = {
```
johnynek: This should return Parser1.

parser/src/main/scala/org/bykn/bosatsu/parser/Parser.scala
```diff
@@ -0,0 +1,938 @@
...
+    // we use defer to avoid enumerating this function unless needed
+    defer(charIn(Impl.allChars.filter(fn)))
+  }
+
+  def charWhere1(fn: Char => Boolean): Parser1[Char] =
+    // we use defer to avoid enumerating this function unless needed
+    defer1(charIn1(Impl.allChars.filter(fn)))
+
+  def void[A](pa: Parser[A]): Parser[Unit] =
```
johnynek, line 335: Add a comment that void is a considerable optimization as is string on a large parse since it removes the internal allocations to build up structure.

parser/src/main/scala/org/bykn/bosatsu/parser/Parser.scala
```diff
@@ -0,0 +1,938 @@
...
+      case notVoid => Impl.StringP(Impl.unmap(pa))
+    }
+
+  def string1[A](pa: Parser1[A]): Parser1[String] =
+    pa match {
+      case str@Impl.StringP1(_) => str
+      case notVoid => Impl.StringP1(Impl.unmap1(pa))
+    }
+
+  val index: Parser[Int] =
```
johnynek, line 362: Consider using Long here.

parser/src/main/scala/org/bykn/bosatsu/parser/Parser.scala
```diff
@@ -0,0 +1,938 @@
...
+          case later => map2(pa, defer(later.value))(fn)
+        })
+
+      def tailRecM[A, B](init: A)(fn: A => Parser1[Either[A, B]]): Parser1[B] =
+        tailRecM1(init)(fn)
+
+      def combineK[A](pa: Parser1[A], pb: Parser1[A]): Parser1[A] =
+        Parser.oneOf1(pa :: pb :: Nil)
+
+      override def void[A](pa: Parser1[A]): Parser1[Unit] =
```
johnynek, line 427: We should override as and <* and *> to leverage void. So as(a) = void.map(_ => a)

johnynek (commented):
Some notes I saw.

johnynek:
1. Need to add anyChar (and maybe review fastparse for other parsers)
2. Law p.orElse(p) == p
3. If p ~ end parses s then p ~ p ~ end parses s + s (or more generally with p1 and s1 with p2 and s2)
4. p1.backtrack.orElse(p2) succeeds if either p1 or p2 do.

johnynek:
going to merge even though some tests timed out. I have seen this rarely locally. It seems like a dead-lock. I'm not sure what it is but I think it could be related to jython.

PR #586, opened 2020-10-12
remove fastparse and use parser module

### LINKED ISSUE #545: update to fastparse 2 or move to a new parsing library

johnynek:
close #545 

Move to the new parsing library. The benefit is that it seems marginally faster, and we remove some of the backtracking we had previously (since the new library has backtracking as opt-in), which allows us to point more clearly to parse errors.


snoble:
This is amazing to get rid of such a big dependency and implement in a way that works the way that's most useful to bosatsu. Congratulations!

Issue #594, opened 2020-10-15
fix the browser compiler


johnynek:
since updating to use munit, I think I turned on common js modules, which means that the browser is no longer working.

I get `Uncaught ReferenceError: exports is not defined` in the console, which seems to be symptomatic of trying to directly run the commonjs module.

without turning on commonjs modules, I couldn't run munit.

It seems I have to understand more about using commonjs modules in scalajs now... evidently this is related:

https://scalacenter.github.io/scalajs-bundler/

cc @snoble

johnynek:
maybe related to #574

PR #596, opened 2020-10-16
Remove module configuration that broke browser imports

snoble:
just add an exports object to index.html so module like code works. webpack would be a more complete solution but seems like overkill.

fixes https://github.com/johnynek/bosatsu/issues/594


snoble:
which of course breaks the tests...

snoble:
also added in the referenced sourcemap to remove a browser warning.

johnynek:
Thanks so much man!

PR #617, opened 2020-11-18
Remove LetFree

snoble:
Many of the normalizations of LetFree are now applied directly to
TypedExpr. The remaining normalizations can lead to massive expressions.


johnynek:
Poor one out for LetFree!

I think one thing you had that would be interesting to get in, and I think you figured out how to do it is alpha renaming. LetFree effectively did alpha renaming by using de bruijn indices. It would be good to do that in TypedExpr so that if you use different variable names on non exported values the results are the same.

Or maybe even alpha rename everything and for exported values set up a final binding for each exported name.

Anyway, thanks for all your work on this!

PR #618, opened 2020-11-19
update scala-setup github action to deal with compatibility change

snoble:


johnynek (approved):
Thanks for doing this!

snoble:
v6 was making all of the checks fail because of this change https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

PR #625, opened 2020-11-24
Update scalapb to the latest released version

### LINKED ISSUE #589: update scalapb
https://github.com/scalapb/ScalaPB

the latest version #550 doesn't just work. It seems the generated code has changed so it may require some significant changes, or it may be that there is some option I can set to make the change more modest.

johnynek:
close #589


johnynek:
the change seems to be that scalapb now generates a field that carries unknown data so if the schema changes later and more data is sent you can potentially inspect it.

This can be useful for some data engineering tasks, but I don't think it is something that is relevant here, so I just ignore any extra fields.

PR #649, opened 2021-01-12
Upgrade to cats-parse 0.3.0

johnynek:
replaces #648 

the scalafix got most of the way, but there were a few imports that were unneeded

cc @martijnhoekstra


martijnhoekstra:
That scalafix is looking pretty good for this repo. This is about what we can expect for others too, I think.

PR #657, opened 2021-01-17
Improving comments in Declaration

johnynek:
Currently, Comment is not a NonBinding Declaration. As a result of that, any place that prohibits binding expressions or defs also prohibits comments.

This is really annoying.

The goal we have had so far, is be able to reconstruct the code from the parsed representation, or at least a formatted version of it.

The challenge is that means we have to store the comment in the AST somehow to do that.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/CommentStatement.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/CommentStatement.scala
@@ -25,18 +25,18 @@ object CommentStatement {
    */
   def parser[T](onP: String => P0[T]): Parser.Indy[CommentStatement[T]] =
     Parser.Indy { indent =>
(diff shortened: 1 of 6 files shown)


core/src/main/scala/org/bykn/bosatsu/CommentStatement.scala
```diff
@@ -25,18 +25,18 @@ object CommentStatement {
    */
   def parser[T](onP: String => P0[T]): Parser.Indy[CommentStatement[T]] =
     Parser.Indy { indent =>
-      val sep = Parser.newline ~ P.string0(indent)
+      val sep = Parser.newline ~ Parser.maybeSpace
```
johnynek, line 28: we can weeken this to allow any number of space or tab between comment lines. Why not.

PR #660, opened 2021-01-20
Fix an issue with imports and enums

johnynek:
There was a bug where if you use and enum, but only import the names of the constructors, you would get a confusing error message. This was due to a bad check in the TotalityChecker.

The check itself was redundant, and removing it seems to be a fix.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
@@ -92,12 +92,7 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
       case PositionalStruct(name, args) =>
         // This is total if the struct has a single constructor AND each of the patterns is total
         val argCheck = args.parTraverse_(validatePattern)
-        val nameCheck = inEnv.definedTypeFor(name) match {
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
```diff
@@ -92,12 +92,7 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
       case PositionalStruct(name, args) =>
         // This is total if the struct has a single constructor AND each of the patterns is total
         val argCheck = args.parTraverse_(validatePattern)
-        val nameCheck = inEnv.definedTypeFor(name) match {
-          case None =>
-            Left(NonEmptyList.of(UnknownConstructor(name, p, inEnv)))
-          case Some(_) =>
-            checkArity(name, args.size, p)
-        }
+        val nameCheck = checkArity(name, args.size, p)
```
johnynek, line 95: this check is already implemented correctly in checkArity, so there is no need to check it first.

The bug is that we don't need the definedType, which might only be present if we import the type (maybe it should be there if only the constructor is imported). Instead we can just see if we have the constructor imported or not.

PR #661, opened 2021-01-21
Follow up to 660 on imported enum totality checks

johnynek:
This is a continuation of #660 

That fix was only partial. I think this is a more complete fix.


core/src/main/scala/org/bykn/bosatsu/rankn/TypeEnv.scala
```diff
@@ -55,7 +55,8 @@ class TypeEnv[+A] private (
       val nec = constructors.updated((pack, cf.name), (cf.args, dt, cf.fnType))
       // add this constructor to the values
       val v1 = values.updated((pack, cf.name), cf.fnType)
-      new TypeEnv(values = v1, constructors = nec, definedTypes = definedTypes)
```
johnynek, line 58: note definedTypes was not updated previously, despite we had a new `dt`.

Issue #558, opened 2020-08-27
update to 2.12.12 (or later) when native-image issues can be solved


johnynek:
see: https://github.com/scala/bug/issues/12116

there is a workaround here:
https://github.com/scala/bug/issues/11634#issuecomment-670386111

I'd rather see native-image just work of course.

PR #737, opened 2021-05-29
Try to see if 2.12.14 just works with graalvm

johnynek:


johnynek:
after updating to graalvm to 21.1.0 in #736 I want to see if 2.12.14 just works (2.12.12 used method handles that tripped up native-image).

johnynek:
okay, it doesn't "just work".

I guess I should use the sbt-native-image plugin which has a work around:

https://github.com/scalameta/sbt-native-image

johnynek:
closes #558

PR #784, opened 2021-08-28
Try to get codecov configured better

johnynek:


johnynek:
this was totally useless and I already had the right file name created in #733

PR #785, opened 2021-08-28
Add a patch setting to codecov.yml

johnynek:


johnynek:
follow up to #784 and #733

PR #791, opened 2021-09-11
error on private type leaks (issue #782)

### LINKED ISSUE #782: private types escaping via exported values
Currently there is no check that all the types necessary to describe a value are exported. So you can have:

```
struct Foo(i: Int)

some_foo = Foo(42)
```
and export `some_foo` without exporting `Foo`. It should be an error to export a value that references a type that is not exported, and the error message should explicitly say types need to be exported.

johnynek:
close #782

### DIFF
--- a/test_workspace/bazel_deps_api.bosatsu
+++ b/test_workspace/bazel_deps_api.bosatsu
@@ -3,7 +3,7 @@ package BazelDepsApi
 from DictTools import merge as merge_Dict
 
 export (
-  Artifact(), Config(), Resolver(), Replacement(), maven_central,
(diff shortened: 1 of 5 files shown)


test_workspace/bazel_deps_api.bosatsu
```diff
@@ -3,7 +3,7 @@ package BazelDepsApi
 from DictTools import merge as merge_Dict
 
 export (
-  Artifact(), Config(), Resolver(), Replacement(), maven_central,
+  Artifact(), Config(), Resolver(), Replacement(), Options(), maven_central,
```
johnynek, line 6: seems like there was only one bug!

PR #808, opened 2021-11-11
Show expected characters on parse failure

johnynek:


johnynek:
I tested these messages by making simple edits and watching the messages...

It would be nice to get better tests, but I think this is an improvement and I'm will to merge even with lower test coverage.

PR #810, opened 2021-11-13
Upgrade cats-parse to 0.3.5

johnynek:
This makes type-check on the test_workspace directory about 10% faster.


johnynek:
replaces #809

PR #826, opened 2021-12-26
Remove zero arg defs

johnynek:
The idea here is in general there should only be one way to do things.

With blocks like:
```
x = (
  y = 1
  f(y)
)
```

We don't really need `def x:`.

This also has the nice side effect that `def` *always* defines a function now and has a more regular syntax: there is always a `(` after def.

This requires adding `external foo: Int` as a valid syntax for values, but that is more regular anyway, so I think that's fine.


johnynek:
@snoble what do you think of this change?

PR #830, opened 2021-12-29
Turn on -Wunused

johnynek:


cli/src/test/scala/org/bykn/bosatsu/codegen/python/PythonGenTest.scala
```diff
@@ -64,14 +64,14 @@ class PythonGenTest extends AnyFunSuite {
...
-        assert(false, s"expected a Test to have 0 or 1 in first tuple entry: $tup")
+        assert(false, s"expected a Test to have 0 or 1 in first tuple entry: $tup, $other")
         ()
     }
   }
 
   def compileFile(path: String, rest: String*): PackageMap.Typed[Any] = {
     def toS(s: String): String =
-      new String(Files.readAllBytes(Paths.get(path)), "UTF-8")
+      new String(Files.readAllBytes(Paths.get(s)), "UTF-8")
```
johnynek, line 74: this was a real bug

core/src/main/scala/org/bykn/bosatsu/Pattern.scala
```diff
@@ -274,7 +274,7 @@ object Pattern {
       Document.instance {
         case WildStr => wildDoc
         case NamedStr(b) => prefix + Document[Bindable].document(b) + Doc.char('}')
-        case LitStr(s) => Doc.text(s)
+        case LitStr(s) => Doc.text(StringUtil.escape(q, s))
```
johnynek, line 277: this seems like a real bug.

core/src/main/scala/org/bykn/bosatsu/codegen/python/PythonGen.scala
```diff
@@ -569,7 +569,7 @@ object PythonGen {
     // if we have a top level let rec with the same name, handle it more cleanly
     val nmVeEnv =
       me match {
-        case Let(Right((n1, r)), inner, Local(n2)) if ((n1 === name) && (n2 === name)) =>
+        case Let(Right((n1, RecursionKind.NonRecursive)), inner, Local(n2)) if ((n1 === name) && (n2 === name)) =>
```
johnynek, line 572: this is a change that seems like it matters... but there is a fallback in any case.

PR #835, opened 2021-12-30
Add Evaluators to transpiler

johnynek:
the idea here is to specify different evaluators for pure values so we never have to admit impure or illegal functions into the bosatsu code.

The motivating example is IO. We could put literally no way to run IO directly in Bosatsu, so once you get an IO, you are stuck in IO land.

But when you compile, you can generate an IO evaluator for your program, and if you run that program you could then do impure things, but evaluating pure Bosatsu code is never impure (even though a bosatsu program may be a builder of another impure program).


johnynek:
there are a few issues here:

1. if you specify a type that is unknown or ill-formed, it is just ignored. That is bad.
2. the type match has to be exact now (there is no `forall a. IO[a]` or something).
3. if there is no match, you just silently get no evaluator. Maybe we should require a match, and just allow a no-op evaluator for Unit or something.

Issue #121, opened 2018-11-29
Kind inference


johnynek:
Currently, we don't know the kinds of types, and we don't check the kinds. We kind of implicit do so as part of the type checking process, but it might be good for us to explicitly check for instance that if we have `f[g]` that `f` has kind `k0 -> k1` and `g` has kind `k0`.

I think we can compute the kinds pretty directly for each of the type parameters of any structs and enums declared. I guess for external structs we can't see, we would need to annotate the kind.

Alternatively, maybe what we have now is effectively kind polymorphism:
https://downloads.haskell.org/~ghc/7.8.1/docs/html/users_guide/kind-polymorphism.html

so, maybe we should have concrete bugs or issues rather than working on more problems.

PR #926, opened 2023-02-06
Simplify ConstructorFn

johnynek:
ConstructorFn has some codesmell that so far was ignored. For instance, it carries its full type, but that type is actually ignored by serialization.

Secondly, we require to know the full type of the constructor for all `TypeEnv[A]` but we actually only query that type once we have fully inferred types (currently that means A =:= Variance in practice).

In the work to check (or infer) Kinds, we wouldn't know the full Kind of generic type parameters at SourceConverter time, and only later at Infer time. This currently presents a problem, but it is only a problem due to how the code is factored, not an actual problem with the algorithms.

relates to #121

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/rankn/ConstructorFn.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/rankn/ConstructorFn.scala
@@ -1,11 +1,11 @@
 package org.bykn.bosatsu.rankn
 
-import org.bykn.bosatsu.{TypeName, PackageName}
 import org.bykn.bosatsu.Identifier.{Bindable, Constructor}
 
 final case class ConstructorFn(
-  name: Constructor,
-  args: List[(Bindable, Type)]) {
+    name: Constructor,
+    args: List[(Bindable, Type)]
+) {
 
   def isZeroArg: Boolean = args == Nil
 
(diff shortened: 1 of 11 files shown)


core/src/main/scala/org/bykn/bosatsu/rankn/TypeEnv.scala
```diff
@@ -37,26 +35,50 @@ class TypeEnv[+A] private (
...
   def getType(p: PackageName, t: TypeName): Option[DefinedType[A]] =
     definedTypes.get((p, t))
 
-  def getValue(p: PackageName, n: Identifier): Option[Type] =
+  def getExternalValue(p: PackageName, n: Identifier): Option[Type] =
     values.get((p, n))
 
-  def localValuesOf(p: PackageName): SortedMap[Identifier, Type] =
-    (SortedMap.newBuilder[Identifier, Type] ++= values.iterator.collect { case ((pn, n), v) if pn == p => (n, v) }).result()
+  private def getConsType[A1](dt: DefinedType[A1], cf: ConstructorFn)(implicit ev: A1 <:< Variance): Type =
```
johnynek: we don't really use `ev` currently, but it is here to enforce that we only access the constructor type *after* we have inferred the types (or variance) of the defined types.

core/src/main/scala/org/bykn/bosatsu/rankn/ConstructorFn.scala
```diff
@@ -19,26 +18,21 @@ final case class ConstructorFn(
     }
 
   def arity: Int = args.length
-}
 
-object ConstructorFn {
-
-  def build(packageName: PackageName, typeName: TypeName, tparams: List[Type.Var.Bound], name: Constructor, fnParams: List[(Bindable, Type)]): ConstructorFn = {
+  def fnType(packageName: PackageName, typeName: TypeName, dtTypeParams: List[Type.Var.Bound]): Type = {
```
johnynek: move this to inside `DefinedType.fnTypeOf`.

PR #939, opened 2023-02-22
Simplify Expr by removing annotated Lambda

johnynek:
This just unifies Lambda and AnnotatedLambda by making the type Option. This nets less code.

I'm about to make another change to Expr so I thought I'd do this first.


johnynek:
taking my chances the timed out test would go green.

PR #940, opened 2023-02-22
Add syntax for def type annotation

### LINKED ISSUE #938: syntax for universal quantification of def
currently, any unbound Bindable names are universally quantified in defs, but with kind inference #121 we are only inferring kinds for types, not the kinds for universally quantified types. It may be possible to infer the kinds for universally quantified types as well, but it's tricky since you have to conflate it with type inference (it seems), which is not true of inferring kinds of declared types.

Two ideas for syntax:

```
# follow types using [ ] for type arguments and parameters.

def foo[a](some_arg: a) -> Option[a]
def foo[c: * -> *](some_c: c[Int]) -> c[Bool]

# use forall:

forall a. def foo(some_arg: a) -> Option[a]
forall c: * -> *. def foo(some_c: c[Int]) -> c[Bool]

```

The second has the analog of following the type syntax, while the first mirrors the `struct/enum` syntax. I'd like to have a very regular language that makes sense, but it should also be compact and not too ugly. Looking at these two suggestions, I find the first more readable.

johnynek:
close #938 

This turns out to be needed to make kinds work with many of our current examples (which do have types mentioned which are not of kind `*`).

The change is pretty mechanical and minor.

We cannot use any kind of strict checking that all free vars are mentioned like we do with structs/enums because unlike structs and enums `defs` are recursive. We assume any param not mentioned is free in the higher level until we reach the top and then it is a top level free param.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/DefStatement.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/DefStatement.scala
@@ -1,76 +1,67 @@
 package org.bykn.bosatsu
 
-import Parser.{ Combinators, maybeSpace }
-import cats.Applicative
+import Parser.{Combinators, maybeSpace}
 import cats.data.NonEmptyList
-import cats.implicits._
 import cats.parse.{Parser => P}
-import org.typelevel.paiges.{ Doc, Document }
+import org.typelevel.paiges.{Doc, Document}
 
-import Identifier.{Bindable, Constructor}
+import Identifier.Bindable
+
+import cats.syntax.all._
 
 case class DefStatement[A, B](
-  name: Bindable,
-  typeArgs: Option[NonEmptyList[TypeRef.TypeVar]],
-  args: NonEmptyList[A],
-  retType: Option[TypeRef], result: B) {
-
-  /**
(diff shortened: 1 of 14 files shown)


core/src/main/scala/org/bykn/bosatsu/DefStatement.scala
```diff
@@ -26,7 +27,13 @@ case class DefStatement[A, B](
       }
 
     (args.traverse(argFn), bodyExp, tag).mapN { (as, b, t) =>
-      Expr.buildPatternLambda(as, b, t)
+      val lambda = Expr.buildPatternLambda(as, b, t)
+      typeArgs match {
+        case None => lambda
+        case Some(args) =>
+          val bs = args.map { case TypeRef.TypeVar(b) => rankn.Type.Var.Bound(b) }
+          Expr.Generic(bs, lambda)
```
johnynek: We need to check in SourceConverter, which may be the only caller of this method, that all the declared type parameters are actually free.

We shouldn't allow totally free parameters. (In general, declaring forall x. y should require that x is free in y, otherwise it is likely a typo).

PR #943, opened 2023-02-25
Add kind system with kind inference for structs

### LINKED ISSUE #121: Kind inference

johnynek:
close #121 

This is a very large PR. It's unfortunate to make large PRs, but this it's a bit hard to finish the last bit without making a lot of changes.

I will take some time to polish and pull out smaller bits of the PR, but it seems to be working.

Left to later PRs:

1. I don't think I'm unifying kinds early enough and some ill-kinded methods are only caught at use time, rather than definition time. We want to error at the definition.
2. The Kind and Shape errors are just dumped out. They are totally inscrutable. They need sane messages that show the regions of the shape errors.
3. In the distant future it may be possible to infer kinds on methods not just data types. This would require changing the inference algorithm because we would need shape and kind variables, not just type variables. But since higher kinds are actually fairly rare, requiring annotation of them doesn't seem like a disaster for now (in fact, it may be good and discourage overly abstract code).
4. this code now triggers errors if you use a function that returns type X but don't mention type X in your code explicitly, because we still need to see the type. Previously since we knew the typename, and know no two types are identical and all are identical to themselves, this wasn't an issue. This changes the ergonomics some. It's weird to import a type just because a function returns it, IMO. If you don't mention a type, but we can infer your use, you shouldn't have to import it. But if you do mention it, IMO, you should import it so it can be clear to a reader where all names come from.


core/src/main/resources/bosatsu/predef.bosatsu
```diff
@@ -52,6 +52,8 @@ struct TupleCons(first, second)
 enum Bool:
   False, True
 
+external struct Fn[a: -*, b: +*]
```
johnynek, line 55: With variance annotations we can finally correctly define Fn in predef.

core/src/main/scala/org/bykn/bosatsu/Kind.scala
```diff
@@ -67,9 +72,18 @@ object Kind {
   // Function[-A, +B] => -* -> +* -> *
   case class Cons(arg: Arg, result: Kind) extends Kind
 
+  val invariantTypeArg: Arg = Arg(Variance.in, Type)
+
   def apply(args: Arg*): Kind =
     args.foldRight(Type: Kind)(_.returns(_))
 
+  // The kind of Bosatsu.Predef::Fn
+  val FnKind: Kind = Kind(Type.contra, Type.co)
```
johnynek: Remove these. This was moved to rankn.Type.

core/src/main/scala/org/bykn/bosatsu/Kind.scala
```diff
@@ -95,45 +109,138 @@ object Kind {
...
+      (phantom, true) ->
+        List(phantom),
+      (phantom, false) ->
+        List(phantom, contra, co, in)
+    )
+  }
+  private def vars(v: Variance, sub: Boolean): List[Variance] =
+    varMap((v, sub))
+
+  private val varMapSize: Map[(Variance, Boolean), Long] = {
```
johnynek, line 129: Two maps would probably be better, one for true and one for false.

core/src/main/scala/org/bykn/bosatsu/Kind.scala
```diff
@@ -95,45 +109,138 @@ object Kind {
       case _ => false
     }
 
-  // allSubKinds(k).forall(leftSubsumesRight(k, _))
-  def allSubKinds(k: Kind): LazyList[Kind] =
+  private val varMap: Map[(Variance, Boolean), List[Variance]] = {
```
johnynek, line 105: Probably two different maps are better.

core/src/main/scala/org/bykn/bosatsu/Package.scala
```diff
@@ -189,38 +189,55 @@ object Package {
...
     stmts: List[Statement]):
       Ior[NonEmptyList[PackageError],
-      Program[TypeEnv[Variance], TypedExpr[Declaration], List[Statement]]] =
+      Program[TypeEnv[Kind.Arg], TypedExpr[Declaration], List[Statement]]] =
         inferBodyUnopt(p, imps, stmts).map {
           case (fullTypeEnv, prog) =>
             TypedExprNormalization.normalizeProgram(p, fullTypeEnv, prog)
         }
 
+  private def kindMap(
```
johnynek: Feels like maybe this should be somewhere else so we can maybe test it.

core/src/main/scala/org/bykn/bosatsu/Package.scala
```diff
@@ -189,38 +189,55 @@ object Package {
...
-          VarianceFormula.solve(importedTypeEnv, parsedTypeEnv.allDefinedTypes) match {
-            case Right(infDTs) =>
-              Ior.right(ParsedTypeEnv(infDTs, parsedTypeEnv.externalDefs))
-            case Left(err) =>
-              Ior.left(NonEmptyList.one(PackageError.VarianceInferenceFailure(p, err)))
-          }
+        val inferVarianceParsed: Ior[NonEmptyList[PackageError], ParsedTypeEnv[Kind.Arg]] =
+          KindFormula.solveShapesAndKinds(importedTypeEnv, parsedTypeEnv.allDefinedTypes.reverse)
+            .bimap({ necError =>
+              necError.map(PackageError.KindInferenceError(p, _)).toNonEmptyList
```
johnynek, line 220: This needs to have a region of the type that failed.

core/src/main/scala/org/bykn/bosatsu/TypeRecursionCheck.scala
```diff
@@ -20,14 +20,14 @@ object TypeRecursionCheck {
    * this is okay, and cannot be used to break totality
    */
   def checkLegitRecursion(
```
johnynek: Try to copy the tests to KindInferenceTests and remove this. The kind inference should be able to verify this because it has constraints for this.

core/src/main/scala/org/bykn/bosatsu/Variance.scala
```diff
@@ -58,4 +59,25 @@ object Variance {
       override def combine(a: Variance, b: Variance): Variance =
         a + b
     }
+
+  implicit val orderVariance: Order[Variance] =
```
johnynek, line 63: Test with order laws.

core/src/main/scala/org/bykn/bosatsu/rankn/Type.scala
```diff
@@ -216,16 +218,28 @@ object Type {
...
+  val UnitType: Type.TyConst = TyConst(Type.Const.predef("Unit"))
+
+  val builtInKinds: Map[Type.Const.Defined, Kind] =
+    List(
+      BoolType -> Kind.Type,
+      DictType -> Kind.DictKind,
+      FnType -> Kind.FnKind,
+      IntType -> Kind.Type,
+      ListType -> Kind.ListKind,
+      StrType -> Kind.Type
```
johnynek: Add Unit and TupleCons

core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
```diff
@@ -47,7 +47,8 @@ final class SourceConverter(
   private val importedNames: Map[Identifier, (PackageName, Identifier)] =
     imports.iterator.flatMap(_.resolveToGlobal).toMap
 
-  val importedTypeEnv = Referant.importedTypeEnv(imports)(identity)
+  val importedTypeEnv: TypeEnv[Kind.Arg] =
+    Referant.importedTypeEnv(imports)(identity)
```
johnynek, line 51: Forgot about this method. Seems this is implemented twice. See kindMap.

core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
```diff
@@ -1180,12 +1191,13 @@ final class SourceConverter(
...
             val freeBound = freeVars.map {
               case b@rankn.Type.Var.Bound(_) => b
-              case s@rankn.Type.Var.Skolem(_, _) =>
+              case s@rankn.Type.Var.Skolem(_, _, _) =>
                 // $COVERAGE-OFF$ this should be unreachable
                 sys.error(s"invariant violation: parsed a skolem var: $s")
                 // $COVERAGE-ON$
             }
-            val maybeForAll = rankn.Type.forAll(freeBound, tpe)
+            // TODO: Kind support parsing kinds
```
johnynek, line 1199: This was done.

PR #949, opened 2023-02-26
Make sure all types are well kinded

### LINKED ISSUE #944: Add more negative tests for shape and kind inference and checking
We don't have many (any?) ill kinded function tests. We need to make sure the functions fail to compile.

johnynek:
closes #944

In type checking, we were not checking that `TyApply` had compatible kinds.

This rules out (most? all?) ill-kinded types at definition time rather than at function application time, which we were seeing before.


johnynek:
also, should have been in a separate change, but I moved PackageError into its own file.

johnynek:
PackageError is pretty uncovered by unit tests so I think that is what brings down the code coverage of this diff.

PR #952, opened 2023-02-27
Improve inference of types containing forall and matching.

### LINKED ISSUE #264: function doesn't stay polymorphic enough when it is declared to be
I'll take another pass and simplify this later but I thought I'd post this in case the solution is obvious

This code:
```
package RecordSet

enum RowEntry[w]:
  REBool(value: w[Bool])
  REInt(value: w[Int])
  REString(value: w[String])

struct RecordField[t](name: String, to_entry: forall w. w[t] -> RowEntry[w])
struct RecordValue[t](value: t)
struct RecordGetter[shape, t](f: forall w. shape[w] -> w[t])

struct RecordSet[shape](fields: shape[RecordField], rows: List[shape[RecordValue]], getters: shape[RecordGetter[shape]])

def restructure(RecordSet(fields, rows, getters), f: shape1[RecordGetter[shape1]] -> (forall w. shape1[w] -> shape2[w], shape2[RecordGetter[shape2]])) -> RecordSet[shape2]:
  (reshaper, new_getters) = f(getters)
  RecordSet(reshaper(fields), rows.map_List(reshaper), new_getters)
```
results in:
```
-----------------------------------------------------------------------------
in file: test_workspace/recordset.bosatsu, package RecordSet, type RecordField does not unify with type RecordValue
14|def restructure(RecordSet(fields, rows, getters), f: shape1[RecordGetter[shape1]] -> (forall w. shape1[w] -> shape2[w], shape2[RecordGetter[shape2]])) -> RecordSet[shape2]:
15|  (reshaper, new_getters) = f(getters)
16|  RecordSet(reshaper(fields), rows.map_List(reshaper), new_getters)
                                               ^^^^^^^^
```

even though `reshaper` should have type `forall w. shape1[w] -> shape2[w]`

johnynek:
close #264 and #267

I think this and the other variance related changes address all of the issues in #264  we can reopen later if not.

The pushdownCovariant idea can be taken further I think. The current algorithm is only catching easy, not all cases. And we can possibly use it in more than just inferring types on matches.

cc @snoble

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
@@ -1065,6 +1065,9 @@ object Infer {
                 _ <- unifyKind(k.kind, Type.TyVar(v0), rk, right, reg, sigmaRegion)
                 rest <- loop(vs, Kind.Cons(k, leftKind), left)
               } yield rest.updated(v0, right)
+            case (_, fa@Type.ForAll(_, _)) =>
+              // we have to instantiate a rho type
+              instantiate(fa).flatMap(loop(revArgs, leftKind, _))
             case ((v0, k) :: rest, _) =>
               // (k -> leftKind)(k)
               for {
@@ -1089,6 +1092,10 @@ object Infer {
               pushDownCovariant(args, params.toList reverse_::: revForAlls, over)
             case ((_, Kind.Arg(Variance.Covariant, _)) :: rest,
               Type.TyApply(left, rightArg @ Type.TyVar(arg: Type.Var.Bound))) =>
+                // TODO it seems like actually we want to partition:
+                // all the frees in the right side that don't appear in the
+                // left can be pushed down only on the right, where
+                // we can recurse
                 val leftFree = Type.freeBoundTyVars(left :: Nil)
                 revForAlls.collectFirst { case (leftA, k) if leftA == arg && !leftFree.contains(arg) => k } match {
                   case Some(k) =>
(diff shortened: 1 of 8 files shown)


johnynek:
an example in the tests come from:

https://github.com/johnynek/bosatsu/issues/267#issuecomment-498856002

core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
```diff
@@ -870,25 +901,87 @@ object Infer {
                 tbranches <- branches.traverse { case (p, r) =>
                   inferBranch(p, Expected.Check((tsigma.getType, region(term))), r)
                 }
-                resT = tbranches.map { case (_, te) => (te.getType, region(te)) }
-                _ <- resT.flatMap { t0 => resT.map((t0, _)) }.traverse_ {
-                  case (t0, t1) if t0 eq t1 => Infer.unit
-                  // TODO
```
johnynek, line 876: this has been fixed I think.

core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -1993,6 +1997,17 @@ w = ListWrapper([], True)
 
 ListWrapper([*_], r) = w
 
+tests = Assertion(r, "match with total list pattern")
+"""), "A", 1)
+
+    runBosatsuTest(List("""package A
+
+struct ListWrapper2(items: List[a], others: List[b], b: Bool)
```
johnynek, line 2005: add another example with `struct ListWrapperTup(items: List[(a, b)], b: Bool)` and make sure we push down on match in there as well.

Issue #945, opened 2023-02-26
Give decent error messages for kind and shape errors


johnynek:
Currently it's just toString on a list of failed constraints.

PR #956, opened 2023-03-01
Improve Kind check Error classes

johnynek:
follow up to #952


johnynek:
partially address #945

PR #961, opened 2023-03-04
remove the leading `\` from lambdas

### LINKED ISSUE #659: consider removing `\` from being required for a lambda
Instead of a lambda being marked by `\` we would use `->` as a postfix operator to imply a lambda.

This may be tricky for multi-arg functions: `\a, b, c -> fn(c, b, a)` would be `a, b, c -> fn(c, b, a)` which would be ambiguous if you wrap in parens: `(a, b, c -> fn(c, b, a),)` is that a tuple1 with a three param lambda, or a tuple3 with the last item being a lambda?

johnynek:
close #659 

cc @avibryant 

Before merging the PR I'll go ahead and try to migrate all internal syntax, but the generator testing is only generating bare (no `\`) style, so I'm pretty sure it's right.


avibryant (commented):
Looks great!

PR #972, opened 2023-03-20
Improve TypedExpr.coerce

johnynek:


core/src/main/scala/org/bykn/bosatsu/TypedExpr.scala
```diff
@@ -703,52 +705,160 @@ object TypedExpr {
...
+            else None
+          }
+      case _ => None
+    }
+
+  private def allPatternTypes[N](p: Pattern[N, Type]): SortedSet[Type] =
+    p.traverseType { t => Writer[SortedSet[Type], Type](SortedSet(t), t) }.run._1
+
+  private def pushGeneric[A](g: Generic[A]): Option[TypedExpr[A]] =
+    g match {
```
johnynek: Match on g.in

core/src/main/scala/org/bykn/bosatsu/TypedExpr.scala
```diff
@@ -916,29 +1026,42 @@ object TypedExpr {
...
+          case gen@Generic(_, _) =>
+            pushGeneric(gen) match {
+              case Some(e1) => self(e1)
+              case None =>
+                instantiateTo(gen, fntpe, kinds) match {
+                  case Some(res) => res
+                  case None => Annotation(gen, fntpe, gen.tag)
+                }
+              }
           case Local(n, _, tag) =>
```
johnynek: Change this to use Annotation like coerce does.

PR #1022, opened 2023-08-19
Support Existential Recursion

johnynek:
the encoding of existential types `forall a. (forall b. (P[b] -> a) -> a)` inside a recursive data type currently isn't supported.

It should be safe for covariant paths to the same time inside the `P[b]` in the above.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
@@ -28,6 +28,8 @@ object DefRecursionCheck {
     def region: Region
     def message: String
   }
+  // TODO: this includes the case where don't apply enough args to get to the recursion
+  // but that is a somewhat common error, so we should split that case to give a better message
   case class InvalidRecursion(name: Bindable, illegalPosition: Region) extends RecursionError {
     def region = illegalPosition
     def message = s"invalid recursion on ${name.sourceCodeRepr}"
@@ -99,7 +101,7 @@ object DefRecursionCheck {
           case TopLevel => Nil
           case InDef(outer, n, _, _) => n :: outer.outerDefNames
           case InDefRecurred(id, _, _, _) => id.outerDefNames
-          case InRecurBranch(ir, _) => ir.outerDefNames
+          case InRecurBranch(ir, _, _) => ir.outerDefNames
         }
 
       final def defNamesContain(n: Bindable): Boolean =
@@ -107,7 +109,7 @@ object DefRecursionCheck {
           case TopLevel => false
           case InDef(outer, dn, _, _) => (dn == n) || outer.defNamesContain(n)
           case InDefRecurred(id, _, _, _) => id.defNamesContain(n)
-          case InRecurBranch(ir, _) => ir.defNamesContain(n)
+          case InRecurBranch(ir, _, _) => ir.defNamesContain(n)
         }
 
       def inDef(fnname: Bindable, args: NonEmptyList[Pattern.Parsed]): InDef =
@@ -118,7 +120,7 @@ object DefRecursionCheck {
         this match {
           case InDef(_, defname, _, _) => defname
           case InDefRecurred(ir, _, _, _) => ir.defname
-          case InRecurBranch(InDefRecurred(ir, _, _, _), _) => ir.defname
+          case InRecurBranch(InDefRecurred(ir, _, _, _), _, _) => ir.defname
         }
     }
     case object TopLevel extends State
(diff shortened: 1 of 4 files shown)


core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -852,6 +852,52 @@ main = sum(Succ(Succ(Succ(Zero))))
...
+
+def map[a, b](box: Box[a], fn: a -> b) -> Box[b]:
+  Next(cont -> cont((b, fn)))
+
+b = Item(1)
+
+def loop[a](box: Box[a]) -> a:
+  recur box:
+    case Item(a): a
+    case Next(cont): cont((box, fn) -> fn(loop(box)))
```
johnynek: we should be able to see that since `cont` is a substructure of the `recur` variable, and that the `box` parameter is in a covariant position of `cont`, that calling `loop` on it is safe.

johnynek:
the test here needs more than just better recursion checking, it also needs polymorphic recursion, which I'm not sure I would even want to add (monomorphic recursion means we could monomorphize in principle before emitting code, which could be very nice if we ever wanted to add an optimizing back end).

It would be nice to find a test that separates the concerns.

core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
```diff
@@ -244,10 +294,26 @@ object DefRecursionCheck {
...
                 failSt(InvalidRecursion(nm, region))
               case Some(arg) =>
-                toSt(strictSubstructure(irb.defname, branch, arg)) *>
+                toSt(strictSubstructure(irb.defname, branch, arg, optApplyState)) *>
                   setSt(irb.incRecCount) // we have recurred again
             }
           }
+          else if (branch.names.contains(nm)) {
+            // we are calling a function referenced by the recur variable
+            val newIrb = irb.copy(applyState = Some(ApplyState(None, nm, Variance.co, Map.empty)))
```
johnynek: seems like this should not just ignore optApplyState

core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
```diff
@@ -165,11 +184,22 @@ object DefRecursionCheck {
...
+    private def strictSubstructure(fnname: Bindable, pat: Pattern.Parsed, decl: Declaration, optApplyState: Option[ApplyState]): Res =
       decl match {
-        case v@Declaration.Var(nm) =>
+        case v@Declaration.Var(nm: Bindable) =>
           if (pat.substructures.contains(nm)) unitValid
-          else Validated.invalidNel(RecursionNotSubstructural(fnname, pat, v))
+          else {
+            optApplyState match {
+              case None => Validated.invalidNel(RecursionNotSubstructural(fnname, pat, v))
+              case Some(ApplyState(_, _, apV, binds)) =>
```
johnynek, line 207: seems strange that we are ignoring the function name here... can we remove it entirely? What about the parent?

core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
```diff
@@ -244,10 +294,26 @@ object DefRecursionCheck {
...
           }
+          else if (branch.names.contains(nm)) {
+            // we are calling a function referenced by the recur variable
+            val newIrb = irb.copy(applyState = Some(ApplyState(None, nm, Variance.co, Map.empty)))
+            for {
+              _ <- setSt(newIrb)
+              _ <- args.traverse_(checkDecl)
+              _ <- getSt.flatMap {
+                case irb1@InRecurBranch(_, _, _) =>
+                  setSt(irb1.copy(applyState = irb.applyState))
```
johnynek, line 325: Add a comment that we are maintaining the recursion count.

core/src/main/scala/org/bykn/bosatsu/DefRecursionCheck.scala
johnynek: I think we want:

```scala

def call(opt: Option[ApplyState], nm: Bindable): ActionState 
```
To correctly build the next state.

johnynek:
Thinking on this algorithm more:

The variance/kind checker already bans recursive ADTs where the recursion isn't covariant. So, I think this checker only needs to check that we are calling the function on something reachable only an arg in the recur position.

I think the algorithm could work like this: we track all the names in the recur pattern for the branch. We allow a recursive call in the correct index of a parameter reachable from one of those names. We say it is reachable if:
1. it is one of those names
2. it is an argument of a lambda passed to a reachable name.

thus once we are in a branch, we have to track all the shadowing bindings which could confuse us into thinking we are using a name from a match (or reachable) when in fact we aren't.

johnynek:
we can pick this back up after #1041

Issue #1026, opened 2023-08-27
removing automatic currying


johnynek:
The current representation of functions is that there is only `a -> b`. We encode `def foo(a: a, b: b) -> c:` as `a -> b -> c`. Like Haskell.

This creates two challenges:

1. it deoptimizes functions which naturally often have multiple arguments. In those cases we wind up building closures over the first N-1 args only to do the real work at the very end.
2. it creates a situation where we can only type check and see an error when we apply too many arguments to a function. When we don't apply enough it still type checks at the site in many cases, but then gives a confusing error later.

So, instead of copying Haskell, we would be more like scala 2 and have fixed sized Functions, say up to size 8 or something. ([Haskell 98](https://www.haskell.org/onlinereport/basic.html#:~:text=6.1.,-4%20Tuples&text=There%20is%20no%20upper%20bound,Bounded%2C%20Read%2C%20and%20Show.) requires tuples up to size 15, and standard functions up to size 7). It sometimes said humans can remember [7 plus or minus 2 things at one time](https://thebrain.mcgill.ca/flash/capsules/experience_jaune03.html#:~:text=Many%20psychology%20experiments%20have%20shown,2%2C%20depending%20on%20the%20individual.). Choosing 8 is more than 7, and a power of two.

So, then we would require applying all the arguments of a function at once, or we would fail the type check because we would try unifying with the wrong arity of function.

There are three syntactic forms of function application:
1. normal application `foo(a, b)`
2. dot-syntax: `a.foo(b)`
3. left apply:
```
x <- list.flat_map
[x, x]
#desugared into list.flat_map(x -> [x, x])
```

Normal application wouldn't change, except that it wouldn't be auto-currying. You would need to manually curry if you want that with e.g. `x -> foo(y, x)`

But for the dot apply, we need to change things. Plausibly, `a.foo(b)` could desugar to `foo(a)(b)` or `foo(a, b)`. Since the latter is more in-keeping with the ideas here, we have to deal with the fact that `a.foo` can't be a valid value, since `(a.foo)(b)` would then be different from `(a.foo(b))` if we use the second desugaring. Therefore, we need to ban `a.foo` and require `a.foo()(b)` for the case which is the same as `foo(a)(b)`. Again, very arguably, this is more in keeping with the bosatsu way: if we have a function application it always has a `()`.

[...]

PR #1027, opened 2023-08-28
Require parens on dot apply

johnynek:
this is the main syntax change in #1026. We can do this before changing the type-system to have FunctionN types.

### DIFF
--- a/core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
+++ b/core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
@@ -5,7 +5,7 @@ import Value._
 import LocationMap.Colorize
 import org.scalatest.funsuite.AnyFunSuite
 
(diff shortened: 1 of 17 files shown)


core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -2458,7 +2458,7 @@ test = Assertion(substitute.eq_Int(42), "basis substitution")
   runBosatsuTest(List("""
 package A
 
-three = 2.(x -> add(x, 1))
+three = 2.(x -> add(x, 1))()
```
johnynek, line 2461: this is kind of weird... seems like it would be better to have:
```scala
def pipe(x, f): f(x)

2.pipe(x -> add(x, 1))
# or
operator |> = pipe
2 |> (x -> add(x, 1))
```

PR #1029, opened 2023-09-04
so long to automatic currying

### LINKED ISSUE #1026: removing automatic currying

johnynek:
closes #1026

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/MatchlessToValue.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/MatchlessToValue.scala
@@ -47,14 +47,14 @@ object MatchlessToValue {
         else
           // arity > 1
           FnValue { args =>
-            val prod = ProductValue.fromList(args.toList.take(arity))
+            val prod = ProductValue.fromList(args.toList)
             SumValue(variant, prod)
           }
       case MakeStruct(arity) =>
         if (arity == 0) UnitValue
         else if (arity == 1) FnValue.identity
         else FnValue { args =>
-          ProductValue.fromList(args.toList.take(arity))
+          ProductValue.fromList(args.toList)
         }
       case ZeroNat => zeroNat
       case SuccNat => succNat
@@ -388,8 +388,6 @@ object MatchlessToValue {
                 // hopefully optimization/normalization has lifted anything
                 // that doesn't depend on argV above this lambda
                 FnValue { argV =>
-                  // TODO remove when tests are passing
-                  assert(argV.length == args.length)
                   val scope2 = scope1.letAll(args.zip(argV))
                   resFn(scope2)
                 }
--- a/core/src/main/scala/org/bykn/bosatsu/MatchlessToValue.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/MatchlessToValue.scala
@@ -1,6 +1,7 @@
 package org.bykn.bosatsu
 
-import cats.{Eval, Functor, Applicative}
+import cats.{Eval, Foldable, Functor, Applicative}
+import cats.data.NonEmptyList
 import cats.evidence.Is
 import java.math.BigInteger
 import scala.collection.immutable.LongMap
@@ -33,7 +34,7 @@ object MatchlessToValue {
       val bi = v.asExternal.toAny.asInstanceOf[BigInteger]
       ExternalValue(bi.add(BigInteger.ONE))
     }
-    FnValue(inc(_))
+    FnValue { case NonEmptyList(a, _) => inc(a) }
   }
 
   def makeCons(c: ConsExpr): Value =
@@ -41,17 +42,20 @@ object MatchlessToValue {
       case MakeEnum(variant, arity, _) =>
         if (arity == 0) SumValue(variant, UnitValue)
         else if (arity == 1) {
-          FnValue { v => SumValue(variant, ConsValue(v, UnitValue)) }
+          FnValue { case NonEmptyList(v, _) => SumValue(variant, ConsValue(v, UnitValue)) }
         }
         else
-          FnValue.curry(arity) { args =>
-            val prod = ProductValue.fromList(args)
+          // arity > 1
+          FnValue { args =>
+            val prod = ProductValue.fromList(args.toList.take(arity))
             SumValue(variant, prod)
           }
       case MakeStruct(arity) =>
         if (arity == 0) UnitValue
         else if (arity == 1) FnValue.identity
-        else FnValue.curry(arity)(ProductValue.fromList(_))
+        else FnValue { args =>
+          ProductValue.fromList(args.toList.take(arity))
+        }
       case ZeroNat => zeroNat
       case SuccNat => succNat
     }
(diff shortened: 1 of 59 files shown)


core/src/main/scala/org/bykn/bosatsu/Package.scala
```diff
@@ -332,12 +332,7 @@ object Package {
     def errorFor(t: (Type.Const, Exp, Type)): List[PackageError] =
       exportedTE.toDefinedType(t._1) match {
         case None =>
-          if (Type.TyConst(t._1) != Type.FnType)
```
johnynek, line 335: maybe we should put this back and remove the external struct stuff... it's better to programmatically define some of these types (Fn and Tuple if we stop the cons-cell encoding of Tuple to match Fn)

core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
```diff
@@ -407,7 +407,7 @@ final class SourceConverter(
         def mkN(n: String): Expr[Declaration] =
           Expr.Global(pn, Identifier.Name(n), l)
         val empty: Expr[Declaration] =
-          Expr.App(mkN("empty_Dict"), mkN("string_Order"), l)
+          Expr.App(mkN("empty_Dict"), NonEmptyList(mkN("string_Order"), Nil), l)
```
johnynek: NonEmptyList.one

core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
```diff
@@ -578,6 +576,8 @@ final class SourceConverter(
             else success(neBound.toList ::: missingFromDecl.map((_, None)))
         }
 
+    // TODO we have to make sure we don't have more than 8 arguments to a struct
```
johnynek, line 584: make sure to use `FnType.isValid` or whatever so we can set the function maximum size in one place.

core/src/main/scala/org/bykn/bosatsu/TypeRef.scala
```diff
@@ -58,7 +58,14 @@ object TypeRef {
     def toBoundVar: Type.Var.Bound = Type.Var.Bound(asString)
   }
   case class TypeName(name: Name) extends TypeRef
-  case class TypeArrow(from: TypeRef, to: TypeRef) extends TypeRef
+  case class TypeArrow(from: NonEmptyList[TypeRef], to: TypeRef) extends TypeRef
+
+  object TypeArrow {
+    // the common case of Fn1
```
johnynek, line 64: remove this and move it to the tests or give it a name like `TypeArrow.oneArg` or something so it isn't confused with the default apply.

core/src/main/scala/org/bykn/bosatsu/TypeRef.scala
```diff
@@ -77,7 +84,7 @@ object TypeRef {
           case (TypeName(_), TypeVar(_)) => 1
           case (TypeName(_), _) => -1
           case (TypeArrow(a0, b0), TypeArrow(a1, b1)) =>
-            val c = compare(a0, a1)
+            val c = list.compare(a0.toList, a1.toList)
```
johnynek: can probably avoid an allocation here by using `Order[NonEmptyList` from cats.

core/src/main/scala/org/bykn/bosatsu/TypedExpr.scala
```diff
@@ -233,31 +240,46 @@ object TypedExpr {
...
+
+    val ifNoCallSemigroup: Semigroup[SelfCallKind] =
+      Semigroup.instance(_.ifNoCallThen(_))
   }
 
-  private def callKind[A](n: Bindable, arity: Int, te: TypedExpr[A]): SelfCallKind =
+  /**
+   * assuming expr is bound to nm, what kind of self call does it contain?
+   */
+  def selfCallKind[A](n: Bindable, te: TypedExpr[A]): SelfCallKind =
```
johnynek: this should be the `def apply` of `SelfCallKind` and possibly moved to a new file if it isn't circular

core/src/main/scala/org/bykn/bosatsu/MatchlessToValue.scala
```diff
@@ -361,14 +368,16 @@ object MatchlessToValue {
...
+          case Lambda(caps, args, res) =>
             val resFn = loop(res)
 
             if (caps.isEmpty) {
               // we can allocate once if there is no closure
               val scope1 = Scope.empty()
               val fn = FnValue { argV =>
-                val scope2 = scope1.let(arg, Eval.now(argV))
+                // TODO remove when tests are passing
+                assert(argV.length == args.length)
```
johnynek, line 379: remove this

core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
johnynek: remove this TODO

PR #1030, opened 2023-09-08
Improve inference for recursive functions

johnynek:


core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -1567,6 +1567,27 @@ main = fn(0, 1, 2)
       ()
     }
 
+    // we should have the region set inside
```
johnynek, line 1570: this failed previously. We pointed to the entire body since fn was totally free at the start of infer, then set to be a function 2 and you get an error like you can't unify the body with Fn1.

PR #1037, opened 2023-09-09
Allow more than one parameter group on def

johnynek:


johnynek:
related to #1026 -- with automatic currying, this syntax is superfluous, but with explicit function sizes, then it is useful when you have cases that will almost always be curried (like building comparison or equality functions from other functions).

PR #1038, opened 2023-09-09
Add annotation on lambdas to improve type error locations

johnynek:


core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -2891,6 +2891,27 @@ Region(396,450)""")
...
+  recur n:
+    case Z: acc
+    case S(n): toInt(n, "foo")
+
+"""
+   evalFail(List(testCode)) { case kie@PackageError.TypeErrorIn(_, _) =>
+      val message = kie.message(Map.empty, Colorize.None)
+      assert(message.contains("Region(122,127)"))
+      val badRegion = testCode.substring(122, 127)
+      assert(badRegion == "\"foo\"")
```
johnynek, line 2910: without this change the bad region is the entire body of `toInt` which is very nondescript.

PR #1050, opened 2023-09-20
Hacking on TypedExprNormalization

johnynek:


core/src/main/scala/org/bykn/bosatsu/TypedExpr.scala
```diff
@@ -606,46 +620,139 @@ object TypedExpr {
...
-                s2 <- solve(arg, arg2, s1, solveSet)
-              } yield s2
-            case (TyConst(_) | TyMeta(_) | TyVar(_), _) =>
-              if (left == right) {
-                // can't recurse further into left
-                Some(state)
-              }
-              else None
-            case (TyApply(_, _), _) => None
+  def instantiateTo[A](gen: Generic[A], instTpe: Type.Rho, kinds: Type => Option[Kind]): TypedExpr[A] = {
```
johnynek, line 676: this change seems good, and should probably go in.

core/src/main/scala/org/bykn/bosatsu/TypedExprNormalization.scala
```diff
@@ -374,22 +389,39 @@ object TypedExprNormalization {
...
+            typeEnv.getType(const).map(_.kindOf)
+          case _ => None
+        }
+
       object ResolveToLambda {
-        def unapply(te: TypedExpr[A]): Option[(NonEmptyList[(Bindable, Type)], TypedExpr[A], A)] =
+        def unapply(te: TypedExpr[A]): Option[(List[(Type.Var.Bound, Kind)], NonEmptyList[(Bindable, Type)], TypedExpr[A], A)] =
           te match {
-            case AnnotatedLambda(args, expr, ltag) => Some((args, expr, ltag))
+            case Annotation(ResolveToLambda((h :: t), args, ex, tag), rho: Type.Rho) =>
```
johnynek, line 465: these changes make function inlining much more powerful, which turns out to blow a lot of things up.

With unconditional function inlining, we can definitely make code exponentially large.

It either caused some issues with infinite loops, or maybe just an exponential blow up that was blowing the stack.

I tried to hack around it by limiting symbolic evaluation to when it can make expressions smaller, but that broke a bunch of other tests.

core/src/test/scala/org/bykn/bosatsu/TypedExprTest.scala
```diff
@@ -522,7 +522,8 @@ res = y -> (x -> f(x, z))(y)
   
       checkLast("""
 f = (_, y) -> y
-res = y -> f(y, 1)
+#res = y -> f(y, 1)
+res = _ -> 1
```
johnynek, line 539: I actually got this to almost pass, but it needs alpha renaming (the inlined function gets a different name that we have here.

core/src/main/scala/org/bykn/bosatsu/TypedExpr.scala
```diff
@@ -606,46 +620,130 @@ object TypedExpr {
...
   type Coerce = FunctionK[TypedExpr, TypedExpr]
 
+  private def pushDownCovariant(tpe: Type, kinds: Type => Option[Kind]): Type = {
+    import Type._
+    tpe match {
+      case ForAll(targs, in) =>
+        val (cons, cargs) = Type.unapplyAll(in)
+        kinds(cons) match {
+          case None =>
+            sys.error(s"unknown kind of $cons in $tpe")
```
johnynek: `cons` could be a type variable from an outer `Generic`. This can't be safe.

core/src/test/scala/org/bykn/bosatsu/TypedExprTest.scala
johnynek: got this to pass.

PR #1079, opened 2023-12-09
Explicitly build closure captures

johnynek:
In order to implement a backend on a platform without built-in closures (such as a C backend), it is better to explicitly build the context over which the lambdas close over, and also to insure that it only reaches into the current scope (not across several scopes).

This PR makes that change.

This doesn't really help python code generation or the interpreter (both of which use the Matchless IR). But it does open the door to easier implementation of a C or Java backend.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Matchless.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Matchless.scala
@@ -75,8 +75,8 @@ object Matchless {
     }
   }
 
-  // Self-name is set for recursive (but not tail recursive) methods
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/bykn/bosatsu/Matchless.scala
```diff
@@ -244,8 +243,47 @@ object Matchless {
           }
       }
 
-    def loopLetVal(name: Bindable, e: TypedExpr[A], rec: RecursionKind): F[Expr] = {
-      lazy val e0 = loop(e)
+    case class LambdaState(name: Option[Bindable], slots: Map[Bindable, Expr]) {
+      def apply(b: Bindable): Expr =
+        name match {
+          case Some(b1) if b1 === b => Local(b)
```
johnynek: I think this isn't needed now since we should never capture b1

Issue #650, opened 2021-01-15
add support for quick look in type inference


johnynek:
see: https://www.microsoft.com/en-us/research/publication/a-quick-look-at-impredicativity/

this can infer a few more types with a relatively minor change to type inference.

PR #1080, opened 2023-12-09
WIP: quicklook-like inference

johnynek:
relates to #650 

This is not exactly the algorithm, I don't think, but it is at least very related.

I should add some of the tests from the paper.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
@@ -2,7 +2,7 @@ package org.bykn.bosatsu.rankn
 
 import cats.{Functor, Monad}
 import cats.arrow.FunctionK
-import cats.data.{Chain, NonEmptyChain, NonEmptyList}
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
```diff
@@ -1231,10 +1381,65 @@ object Infer {
...
           }
 
-          (checkSigma(inner, tpe), instSigma(tpe, expect, region(tag)))
-            .parFlatMapN { (typedTerm, coerce) =>
-              zonkTypedExpr(typedTerm).map(coerce(_))
-            }
+          @inline def default = checkAnnotated(inner, tpe, region(tag), expect)
+
+          tpe match {
+            case rho: Type.Rho =>
```
johnynek, line 1415: seems like this should only check `Type.Rho` when expect is Infer. Otherwise we can use `subsCheck`.

Secondly, it seems like this whole algorithm should actually be inside `checkSigma` which is logically what it is doing, but checkSigma is called in some other places.

PR #1101, opened 2024-01-14
Error on unused imports in Bosatsu packages

### LINKED ISSUE #1098: unused imports are not an error
Adding an unused import should be an error.

This makes reading easier and also allows for easy managing of dependencies.

johnynek:
closes #1098

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/PackageCustoms.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/PackageCustoms.scala
@@ -107,29 +107,32 @@ object PackageCustoms {
       val unusedTypeMap = unusedTypes.groupByNes(_._1)
       val unusedPacks = unusedValMap.keySet | unusedTypeMap.keySet
 
-      Chain.fromIterableOnce(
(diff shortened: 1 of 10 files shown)


core/src/main/scala/org/bykn/bosatsu/PackageCustoms.scala
```diff
@@ -0,0 +1,171 @@
...
+      val unusedTypes = impTypes.filterNot { case (pn, t) =>
+        // deal with the ambiguity of capital names
+        usedTypes(t) || usedValues((pn, t.toDefined.name.ident))
+      }
+
+      val unusedValMap = unusedValues.groupByNes(_._1)
+      val unusedTypeMap = unusedTypes.groupByNes(_._1)
+      val unusedPacks = unusedValMap.keySet | unusedTypeMap.keySet
+
+      Chain.fromIterableOnce(
```
johnynek: all these errors are for a single package, we should go ahead and group them rather than forget that we have a set of errors of the same type from the same package.

Issue #322, opened 2019-08-10
Add Generative testing to Bosatsu


johnynek:
It would be nice to have a Generative/Property testing library.

PR #1102, opened 2024-01-15
Add basic property testing

johnynek:
this gives an example of a basic property testing library for bosatsu.

Hopefully this makes it easier to write more powerful tests directly in bosatsu.


johnynek:
closes #322

PR #1116, opened 2024-01-27
leverage UnionRelatable

johnynek:
The idea here is to use the UnionRelModule to improve relation computation for Patterns.

Computing the relationships of patterns in the presence of nested unions and non-normalized representations is pretty challenging.

The UnionRelModule allows you to have a recursive structure that is well tested. So you only have to solve simpler problems and be sure it's correct.

This and the added test coverage wound up fixing some lurking bugs in the original set relationship calculations on Patterns, but since TotalityCheck is only a compilation lint, the bugs could only have manifested by a runtime error of a non-total pattern match (which I never saw so far).

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
@@ -609,10 +609,10 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
             u.exists(isTopCheap(_))
         }
 
-      def subset(a0: Pattern[Cons, Type], b0: Pattern[Cons, Type]): Boolean =
+      override def subset(a0: Pattern[Cons, Type], b0: Pattern[Cons, Type]): Boolean =
         isTopCheap(b0) || relate(a0, b0).isSubtype
 
-      override def relate(a0: Pattern[Cons,Type], b0: Pattern[Cons,Type]): Rel = {
+      def relate(a0: Pattern[Cons,Type], b0: Pattern[Cons,Type]): Rel = {
         def loop(a: Pattern[Cons, Type], b: Pattern[Cons, Type]): Rel =
           (a, b) match {
             case _ if a == b => Rel.Same
@@ -749,6 +749,6 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
   val eqPat: Eq[Pattern[Cons, Type]] =
     new Eq[Pattern[Cons, Type]] {
       def eqv(l: Pattern[Cons, Type], r: Pattern[Cons, Type]) =
-        normalizePattern(l) == normalizePattern(r)
+        patternSetOps.equiv(l, r)
     }
 }
--- a/core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
@@ -12,7 +12,6 @@ import rankn.{Type, TypeEnv}
 import Pattern._
 
 import Identifier.{Bindable, Constructor}
-import org.bykn.bosatsu.set.Rel.Sub
 
 object TotalityCheck {
   type Cons = (PackageName, Constructor)
@@ -396,6 +395,8 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
         right: Pattern[Cons, Type]): List[Pattern[Cons, Type]] =
           (left, right) match {
             case (Var(va), Var(vb)) => Var(Ordering[Bindable].min(va, vb)) :: Nil
+            case (Var(_), v) => v :: Nil
+            case (v, Var(_)) => v :: Nil
             case (Named(va, pa), Named(vb, pb)) if va == vb =>
               intersection(pa, pb).map(Named(va, _))
             case (Named(_, pa), r) => intersection(pa, r)
@@ -403,8 +404,6 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
             case (WildCard, v) => v :: Nil
             case (v, WildCard) => v :: Nil
             case (_, _) if left == right => left :: Nil
-            case (Var(_), v) => v :: Nil
-            case (v, Var(_)) => v :: Nil
             case (Annotation(p, _), t) => intersection(p, t)
             case (t, Annotation(p, _)) => intersection(t, p)
             case (Literal(a), Literal(b)) =>
@@ -427,13 +426,8 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
                   if (isTop(rp)) left :: Nil
                   else Nil
               }
-            case (lp@ListPat(_), PositionalStruct(n, as)) =>
-              structToList(n, as) match {
-                case Some(rp) => intersection(lp, rp)
-                case None =>
-                  if (isTop(lp)) right :: Nil
-                  else Nil
-              }
+            case (lp@ListPat(_), pos@PositionalStruct(_, _)) =>
+              intersection(pos, lp)
             case (PositionalStruct(ln, lps), PositionalStruct(rn, rps)) =>
               if (ln == rn) {
                 val la = lps.size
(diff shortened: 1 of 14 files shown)


core/src/test/scala/org/bykn/bosatsu/TotalityTest.scala
```diff
@@ -464,4 +464,28 @@ enum Either: Left(l), Right(r)
...
+    assertEquals(rel, Rel.Super) 
+  }
+
+  test("union commutes") {
+    val tc = TotalityCheck(predefTE)
+
+    {
+      val p1 :: p2 :: _ = patterns("""[Some(1 | 2), Some(1) | Some(2)]""")
+      val rel = tc.patternSetOps.relate(p1, p2)
+      assertEquals(rel, Rel.Same)
```
johnynek, line 506: this test fails on master:
```
==> X org.bykn.bosatsu.TotalityTest.union commutes  0.025s munit.ComparisonFailException: /Users/oscar/code/oss/bosatsu/core/src/test/scala/org/bykn/bosatsu/TotalityTest.scala:474
473:      val rel = tc.patternSetOps.relate(p1, p2)
474:      assertEquals(rel, Rel.Same) 
475:    }
values are not the same
=> Obtained
Intersects()
=> Diff (- obtained, + expected)
-Intersects()
+Same()
```

core/src/test/scala/org/bykn/bosatsu/TotalityTest.scala
```diff
@@ -464,4 +464,28 @@ enum Either: Left(l), Right(r)
...
+    {
+      val p1 :: p2 :: _ = patterns("""[Some(1 | 2), Some(1) | Some(2)]""")
+      val rel = tc.patternSetOps.relate(p1, p2)
+      assertEquals(rel, Rel.Same) 
+    }
+
+    {
+      val p1 :: p2 :: _ = patterns("""[Some(1 | 2 | 3), Some(1) | Some(2)]""")
+      val rel = tc.patternSetOps.relate(p1, p2)
+      assertEquals(rel, Rel.Super)
```
johnynek, line 512: this test fails on master:
```
==> X org.bykn.bosatsu.TotalityTest.union commutes  0.029s munit.ComparisonFailException: /Users/oscar/code/oss/bosatsu/core/src/test/scala/org/bykn/bosatsu/TotalityTest.scala:482
481:      val rel = tc.patternSetOps.relate(p1, p2)
482:      assertEquals(rel, Rel.Super) 
483:    }
values are not the same
=> Obtained
Intersects()
=> Diff (- obtained, + expected)
-Intersects()
+Super()
```

core/src/test/scala/org/bykn/bosatsu/set/SetOpsLaws.scala
```diff
@@ -272,6 +274,53 @@ abstract class SetOpsLaws[A] extends munit.ScalaCheckSuite {
...
     }
   }
+
+  test("relate consistency") {
+    forAll(genItem, genItem, eqUnion) { (a, b, eqv) =>
+      val relAb = setOps.relate(a, b)
+      assertEquals(setOps.relate(b, a).invert, relAb)
+      relAb match {
+        case Rel.Same =>
+          val intr = setOps.intersection(a, b)
```
johnynek, line 293: We should also check eqv(a, b)

core/src/main/scala/org/bykn/bosatsu/pattern/SeqPattern.scala
```diff
@@ -307,6 +311,57 @@ object SeqPattern {
             subsetList(t1, p2)
         }
 
+      private val viaIntersection: Relatable[SeqPattern[A]] =
+        Relatable.fromSubsetIntersects(
+          self.subset(_, _),
+          (l, r) => self.intersection(l, r).nonEmpty
+        )
+
+      private def relateList(p1: List[SeqPart[A]], p2: List[SeqPart[A]]): Rel =
```
johnynek, line 320: No need to convert to and from List. Just use the SeqPattern representation.

core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
```diff
@@ -287,126 +449,127 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
+                  case None => left :: Nil
+                }
+              case (_, PositionalStruct(nm, _)) if isTop(left) =>
+                inEnv.definedTypeFor(nm) match {
+                  case None =>
+                    // just assume this is infinitely bigger than unknown
+                    // types
+                    left :: Nil
+                  case Some(dt) =>
+                    dt.constructors.flatMap {
```
johnynek: Factor this out. topFor(dt)

core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
```diff
@@ -287,126 +449,127 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
-                case _ =>
-                  // we can't solve this
-                  left :: Nil
-              }
-            }
-            else left :: Nil
-        }
+          }
 
       def isTop(p: Pattern[Cons, Type]): Boolean =
```
johnynek, line 582: Can we make this as simple as checking `relate(Wildcard, p) == Rel.Same`

core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
```diff
@@ -454,24 +617,49 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
+                else Rel.Super
+              }
+              else if (tb) {
+                Rel.Sub
+              }
+              else Rel.Disjoint
+          }
 
           loop(normalizePattern(a0), normalizePattern(b0))
         }
```
johnynek, line 695: Below normalizeUnion doesn't seem to be checking if it has all the needed branches of a defined type to become Wildcard.

core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
```diff
@@ -454,24 +617,49 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
-              case _ => false
-            }
+      override def subset(a0: Pattern[Cons, Type], b0: Pattern[Cons, Type]): Boolean =
+        isTopCheap(b0) || relate(a0, b0).isSubtype
+
+      def relate(a0: Pattern[Cons,Type], b0: Pattern[Cons,Type]): Rel = {
+        def loop(a: Pattern[Cons, Type], b: Pattern[Cons, Type]): Rel =
+          (a, b) match {
+            case _ if a == b => Rel.Same
+            case (_, Union(h, t)) =>
```
johnynek: We should split this Union is half, not take the head item. This ensures the depth is logarithmic.

core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
```diff
@@ -454,24 +617,49 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
+              listPatternSetOps.relate(l1, l2)
+            case (Pattern.PositionalStruct(ln, lp), Pattern.PositionalStruct(rn, rp)) =>
+              if ((ln == rn) && (lp.size == rp.size)) {
+                (lp.zip(rp).foldLeft(Rel.Same: Rel) { case (acc, (l, r)) =>
+                  acc.lazyCombine(loop(l, r))
+                })
+              }
+              else Rel.Disjoint
+            case _ =>
+              val ta = isTop(a)
```
johnynek: Can we avoid using isTop here? If so, isTop could become as simple as relate.

Since the pattern is normalized, isTop may be the same as Wildcard here.

core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
```diff
@@ -531,16 +719,19 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
         patternSetOps.unifyUnion(flattened.toList) match {
+          case h :: t => Pattern.union(h, t)
           case Nil =>
             // $COVERAGE-OFF$
             sys.error("unreachable: union can't remove items")
             // $COVERAGE-ON$
-          case h :: t => Pattern.union(h, t)
         }
 
       case _ if patternSetOps.isTop(p) => WildCard
```
johnynek: It would be nice to be able to remove this.

core/src/main/scala/org/bykn/bosatsu/TotalityCheck.scala
johnynek: no, we can't. `relate` with `Wildcard` always results in the `solve` branch happening, which calls unifyUnion, which needs isTop.

PR #1117, opened 2024-01-31
add a show command

johnynek:
This allows you to print out compiled bosatsu code. This is useful for inspecting what the compiler has produced already shows a few issues I should improve.

### DIFF
--- a/cli/src/main/scala/org/bykn/bosatsu/PathModule.scala
+++ b/cli/src/main/scala/org/bykn/bosatsu/PathModule.scala
@@ -191,7 +191,13 @@ object PathModule extends MainModule[IO] {
           output match {
             case None =>
               docs.traverse_ { doc =>
-                IO.println(doc.render(80))
+                IO.blocking {
+                  doc.renderStreamTrim(100)
+                    .iterator
+                    .foreach(System.out.print)
+
+                  System.out.println("")
+                }
               }
               .as(ExitCode.Success)
             case Some(p) =>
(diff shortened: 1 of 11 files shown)


core/src/main/scala/org/bykn/bosatsu/MainModule.scala
```diff
@@ -1146,6 +1149,21 @@ abstract class MainModule[IO[_]](implicit
...
+        inputs: Inputs.Runtime,
+        output: Option[Path],
+        errColor: Colorize
+    ) extends MainCommand("show") {
+
+      type Result = Output.ShowOutput
+
+      def run = withEC { implicit ec =>
+        for {
+          (packs, _) <- inputs.packMap(this, Nil, errColor)
```
johnynek: I think packMap is added predef, which we shouldn't include by default, I don't think. Maybe a flag?

core/src/main/scala/org/bykn/bosatsu/MainModule.scala
```diff
@@ -1146,6 +1149,21 @@ abstract class MainModule[IO[_]](implicit
       }
     }
 
+    case class Show(
+        inputs: Inputs.Runtime,
```
johnynek: it would be nice to support showing interfaces as well, but there is no defined Inputs that takes interfaces and packages only, so we would need to define a new input.

cli/src/main/scala/org/bykn/bosatsu/PathModule.scala
```diff
@@ -153,6 +153,58 @@ object PathModule extends MainModule[IO] {
...
+                }).grouped.nested(4)
+
+            val tpes = Doc.text("types: ") + Doc.intercalate(Doc.comma + Doc.line,
+              pack.program.types.definedTypes.toList.map { case (_, t) =>
+                Doc.text(t.name.ident.sourceCodeRepr)
+              }).grouped.nested(4)
+
+            val exprs = Doc.intercalate(Doc.hardLine + Doc.hardLine,
+              pack.program.lets.map { case (n, _, te) =>
+                // TODO: we really need to use Doc in repr
```
johnynek: this was done in this Pr.

johnynek:
Part of #394

PR #1124, opened 2024-02-04
Inline applying lambda in Python code generation

johnynek:
Here is an example:
```
 def log2(i):
-    ___t64 = 0 < i
-    ___t65 = 0
-    ___t66 = i
-    ___t67 = 0
-    while ___t64:
-        ___t65 = (lambda n, cnt: (n // 2, cnt + 1))(___t66, ___t67)
-        ___t68 = ___t65[0]
-        ___t67 = ___t65[1]
-        ___t64 = (0 < ___t68) and (___t68 < ___t66)
-        ___t66 = ___t68
-    return ___t67
+    ___t16 = 0 < i
+    ___t17 = 0
+    ___t18 = i
+    ___t19 = 0
+    while ___t16:
+        ___t20 = ___t18 // 2
+        ___t19 = ___t19 + 1
+        ___t16 = (0 < ___t20) and (___t20 < ___t18)
+        ___t18 = ___t20
+    return ___t19
```

In the `int_loop` built-in function, we often wind up with `(lambda x, y: ...)(x_arg, y_arg)` patterns, which could be inlined. Also, in this example, we know we are returning a tuple2. So, we can try to inline the lambda, and if that works, we can often avoid allocating the tuple as well.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/codegen/python/Code.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/codegen/python/Code.scala
@@ -431,9 +431,11 @@ object Code {
   case class SelectItem(arg: Expression, position: Expression) extends Expression {
     def simplify: Expression =
       (arg.simplify, position.simplify) match {
-        case (MakeTuple(items), PyInt(bi)) if items.lengthCompare(bi.intValue()) > 0 =>
+        case (MakeTuple(items), PyInt(bi))
+          if (items.lengthCompare(bi.intValue()) > 0) && (bi.intValue() >= 0) =>
           items(bi.intValue())
-        case (MakeList(items), PyInt(bi)) if items.lengthCompare(bi.intValue()) > 0 =>
+        case (MakeList(items), PyInt(bi))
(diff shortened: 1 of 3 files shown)


core/src/main/scala/org/bykn/bosatsu/codegen/python/Code.scala
```diff
@@ -468,7 +470,27 @@ object Code {
...
   }
   case class Apply(fn: Expression, args: List[Expression]) extends Expression {
-    def simplify: Expression = Apply(fn.simplify, args.map(_.simplify))
+    def simplify: Expression = {
+      fn.simplify match {
+        case Lambda(largs, result) if largs.length == args.length => 
+          // if this is a lambda, but the args don't match, let
+          // the python error
+          val subMap = largs.iterator.zip(args).toMap
+          val subs = substitute(subMap, result)
```
johnynek, line 517: have to be careful here... what about a case like `(lambda x: lambda y: x + y)(y)` I don't think we are correctly handling this. To handle it correctly, we need to do something like return `lambda y1: y + y1`.

Add a test for this before merging.

PR #1156, opened 2024-02-25
scalafmt 3.8.0

johnynek:
replace #1153


johnynek:
this has to be merged as a merge commit or the the ignore commit number won't work.

PR #1162, opened 2024-03-03
ToyIO demonstration of a basic IO implementation for runtimes

johnynek:
The idea here is to encode general recursion into a data structure, then have the recursive loop be implemented in the runtime or host language.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -183,8 +183,8 @@ lazy val core =
     // periodically we use acyclic to ban cyclic dependencies and make compilation faster
     ,
     autoCompilerPlugins := true,
-    addCompilerPlugin("com.lihaoyi" % "acyclic_2.13.12" % "0.3.11"),
-    scalacOptions += "-P:acyclic:force"
+    //addCompilerPlugin("com.lihaoyi" % "acyclic_2.13.12" % "0.3.11"),
+    //scalacOptions += "-P:acyclic:force"
   ).dependsOn(base)
     .jsSettings(commonJsSettings)
 
--- a/core/src/main/scala/org/bykn/bosatsu/ToyIO.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/ToyIO.scala
@@ -10,7 +10,17 @@ object ToyIO {
   /**
    * fix(f) = f(fix(f))
    */
-  case class ApplyFix[E, A, B](arg: A, fixed: (A => ToyIO[E, B]) => (A => ToyIO[E, B])) extends ToyIO[E, B]
+  case class ApplyFix[E, A, B](arg: A, fixed: (A => ToyIO[E, B]) => (A => ToyIO[E, B])) extends ToyIO[E, B] {
+    def step: ToyIO[E, B] =
+      // it is temping to simplify fixed(fix(fixed)) == fix(fixed)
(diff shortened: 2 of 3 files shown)


core/src/main/scala/org/bykn/bosatsu/ToyIO.scala
```diff
@@ -0,0 +1,95 @@
...
+
+object ToyIO {
+  case class Pure[A](get: A) extends ToyIO[Nothing, A]
+  case class Err[E](get: E) extends ToyIO[E, Nothing]
+  case class FlatMap[E, A, B](init: ToyIO[E, A], fn: A => ToyIO[E, B]) extends ToyIO[E, B]
+  case class RecoverWith[E, E1, A](init: ToyIO[E, A], fn: E => ToyIO[E1, A]) extends ToyIO[E1, A]
+  /**
+   * fix(f) = f(fix(f))
+   */
+  case class ApplyFix[E, A, B](arg: A, fixed: (A => ToyIO[E, B]) => (A => ToyIO[E, B])) extends ToyIO[E, B] {
```
johnynek: this constructor can't be written in Bosatsu because `ToyIO[E, B]` is in an invariant position (contravariant and covariant in the fixed function, and we can only have co-variant recursion in bosatsu. But we could define this constructor outside of bosatsu and return it via an `external def`.

core/src/main/scala/org/bykn/bosatsu/ToyIO.scala
```diff
@@ -0,0 +1,95 @@
...
+  sealed trait Stack[E, A, E1, A1]
+
+  case class Done[E1, A1, E, A](ev: E1 =:= E, av: A1 =:= A) extends Stack[E1, A1, E, A]
+  case class FMStep[E, E1, A, B, B1](fn: A => ToyIO[E, B], stack: Stack[E, B, E1, B1]) extends Stack[E, A, E1, B1]
+  case class RecStep[E, E1, E2, A, B](fn: E => ToyIO[E1, A], stack: Stack[E1, A, E2, B]) extends Stack[E, A, E2, B]
+
+  def run[E, A](io: ToyIO[E, A]): Either[E, A] = {
+
+    @annotation.tailrec
+    def loop[E1, A1](arg: ToyIO[E1, A1], stack: Stack[E1, A1, E, A]): Either[E, A] =
```
johnynek: this loop is the only place with explicit recursion. This loop would be implemented in the runtime or host language using a loop to avoid blowing the stack.

build.sbt
```diff
@@ -183,8 +183,8 @@ lazy val core =
     // periodically we use acyclic to ban cyclic dependencies and make compilation faster
     ,
     autoCompilerPlugins := true,
-    addCompilerPlugin("com.lihaoyi" % "acyclic_2.13.12" % "0.3.11"),
-    scalacOptions += "-P:acyclic:force"
+    //addCompilerPlugin("com.lihaoyi" % "acyclic_2.13.12" % "0.3.11"),
```
johnynek, line 186: Commented this because it blows up the repl

johnynek:
This approach has been merged in the Prog data type and the runtimes in C, python and Evaluation use this approach.

PR #1163, opened 2024-03-06
Error on unused lets

johnynek:


core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
```diff
@@ -1218,7 +1218,7 @@ final class SourceConverter(
 
   private def anonNameStrings(): Iterator[String] =
     rankn.Type.allBinders.iterator
-      .map(_.name)
+      .map { bn => "_" + bn.name }
```
johnynek: this is a weird hidden hack: we are making names that are not parseable to make synthetic names.

Probably a better solution is to just make a new Bindable type called Synthetic, that has no Parser, so it is never the result of a parsing rather than backdoor it here.

core/src/main/scala/org/bykn/bosatsu/PackageCustoms.scala
```diff
@@ -224,4 +233,66 @@ object PackageCustoms {
...
+                case Left(_) => 0
+                case Right(_) => -1
+              }
+          }
+      }
+
+    val exports: Node = Left(pack.exports)
+    val roots: List[Node] =
+      (exports ::
+        pack.program.lets.collect { case (b, _, _) if b.asString.startsWith("_") => Right(b) } :::
```
johnynek: this should really be collecting all synthentic names and ignoring any synthetic name.

This probably should not be a root, just filtered from the errors. synthetic names are used to rewrite:
```
# top level
(a, b) = foo
```
into:
```
_a = foo
a = match _a:
  case (a, _): a
b = match _a:
  case (_, b): b
```
so, the only case where `_a` would be unused is for constructs like:
```
_ = foo(1)
```
where we intend this idiom to exist only to check that `foo(1)` type-checks, but can then be removed from runtime.

core/src/main/scala/org/bykn/bosatsu/MainModule.scala
```diff
@@ -1018,16 +1018,14 @@ abstract class MainModule[IO[_]](implicit
...
-            mainPackageNameValue <- mainPackage.getMain(names)
-            (mainPackageName, value) = mainPackageNameValue
+            (packs, names) <- inputs.packMap(this, List(mainPackage), errColor)
+            (mainPackageName, value) <- mainPackage.getMain(names)
             out <-
               if (packs.toMap.contains(mainPackageName)) {
                 val ev = Evaluation(packs, Predef.jvmExternals)
 
                 val res = value match {
-                  case None        => ev.evaluateLast(mainPackageName)
```
johnynek, line 1030: we should just remove evaluateLast and rewrite to `evaluateMain`. `Last` is how, `Main` is what.

core/src/main/scala/org/bykn/bosatsu/PackageCustoms.scala
```diff
@@ -224,4 +233,65 @@ object PackageCustoms {
...
+              y match {
+                case Left(_) => 0
+                case Right(_) => -1
+              }
+          }
+      }
+
+    val exports: Node = Left(pack.exports)
+    val roots: List[Node] =
+      (exports ::
```
johnynek, line 234: Nit: put exports last.

core/src/main/scala/org/bykn/bosatsu/PackageCustoms.scala
```diff
@@ -59,6 +62,39 @@ object PackageCustoms {
         pack.copy(imports = i)
     }
 
+  private type VSet = Set[(PackageName, Identifier)]
+  private type VState[X] = State[VSet, X]
+
+  private def usedGlobals[A](te: TypedExpr[A]): VState[TypedExpr[A]] =
```
johnynek: This should be moved to TypedExpr.

core/src/main/scala/org/bykn/bosatsu/PackageCustoms.scala
```diff
@@ -59,6 +62,39 @@ object PackageCustoms {
         pack.copy(imports = i)
     }
 
+  private type VSet = Set[(PackageName, Identifier)]
```
johnynek, line 63: Identifier should be changed to Bindable for unused names.

johnynek: We don't check that types are used. There should be a similar check for a type that is never used.

core/src/main/scala/org/bykn/bosatsu/PackageCustoms.scala
```diff
@@ -224,4 +233,65 @@ object PackageCustoms {
...
+      }
+    val canReach: SortedSet[Node] = Dag.transitiveSet(roots)(depsOf _)
+
+    val unused = pack.program.lets.filter {
+      case (bn, _, _) => !(bn.isSynthetic || canReach.contains(Right(bn)))
+    }
+
+    NonEmptyList.fromList(unused) match {
+      case None => Validated.unit
+      case Some(value) =>
```
johnynek, line 266: We should add a test to exercise the error message from this.

core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -3054,8 +3061,7 @@ test = Assertion(True, "")
 package A
 
 # this is basically a typecheck only
-x = (1, "1")
-(_, _) = x
```
johnynek, line 3058: This shouldn't be allowed. In a pure language a total pattern match with no binding doesn't do anything.

The source converter should give an error in this case.

core/src/main/scala/org/bykn/bosatsu/PackageCustoms.scala
johnynek: This isn't needed. I'm being ticky here and `exports` is a single node, not a list.

core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
```diff
@@ -1593,6 +1595,12 @@ object SourceConverter {
   def addError[A](r: Result[A], err: Error): Result[A] =
     parallelIor.<*(r)(failure(err))
 
+  def maybeError[A](r: Result[A], opt: Option[Error]): Result[A] =
```
johnynek: This wasn't used

core/src/test/scala/org/bykn/bosatsu/EvaluationTest.scala
```diff
@@ -3338,7 +3373,7 @@ def foo1(fn) -> Int:
 def foo2(fn: List[forall a. a -> a]) -> Int:
   fn.foldLeft(0, (x, _) -> x.add(1))
 
-count = foo1(single(id))
+count0 = foo1(single(id))
```
johnynek, line 3376: This is good. We now see that this original count was useless.

Issue #1164, opened 2024-03-06
Some types seem somewhat unusable


johnynek:
see for instance: https://github.com/johnynek/bosatsu/pull/1163/files#diff-25a462d709b48d25e8d8d95909ecad8339382ba129c10995a5e7a45e62b7f2b1R76

I had values like:
```
refl_Int: Sub[forall a. a, Int] = refl_sub
refl_any: Sub[forall a. a, exists a. a] = refl_sub
refl_any1: Sub[exists a. a, exists a. a] = refl_sub
```
but then using them like:
```
ignore = (refl_Int, refl_any)
```
wouldn't typecheck. Then I defined a function like:
```
def hide(a) -> exists b. b: a

ignore = [hide(refl_Int), hide(refl_any)]
```
and that fails too.

I am starting to wonder if the extensions I've made to the type system basically mean some values are unusable.  It could also just be some lurking bugs with existential types.

But a message like `type ?211 does not subsume type forall a: *. a` seems like we need to set a metavar to `forall a. a` but that's not possible because metavars must be rho types.

It seems like it should be trivial that if `hide: forall a. (a -> exists. b. b)` and we call `hide(x)` and we know the type of `x`, then we should be able to instantiate `a := typeof(x)`.

johnynek:
a related issue is that:
```
x = (a, b, c, ...)
```
should always typecheck, if everything before that type checks. I was surprised I couldn't even put these values in a tuple and have it typecheck. This is similar to:
```
struct Box(a)
x = Box(y)
```
should always compile and produce a good type. I imagine this is also failing for either existential or universal quantification or both (maybe when existentials are in covariant position or universals are in contravariant positions?)

johnynek:
If we look at line 577 of rankn/Type.scala there is a todo about handling shadowing of ForAll types. It could be that if we handle this using the algorithm of #1156 then this issue would also be solved. My guess is based on the idea that types are often automatically assigned starting with a, b, c... so collisions are somewhat frequent in practice. So this may be blocking substitution from working just because we don't unshadow the types at that stage.

We could check this theory cheaply by throwing an exception in that branch and seeing how often it fails.

PR #1166, opened 2024-03-08
improve 1164

johnynek:
This improves type inference on function application in the common case that the fn and argument types are simple.

This seems to fix the issue with #1164 for now, but does not address the unshadowing approach I thought would be useful (that turns out to never actually be a codepath hit for reasons I don't understand).


core/src/test/scala/org/bykn/bosatsu/rankn/RankNInferTest.scala
```diff
@@ -1835,4 +1846,19 @@ f3: Foo[forall a. a] = f1
...
+  test("test Liskov example") {
+    parseProgram(
+      """
+struct Sub[a: -*, b: +*](sub: forall f: +* -> *. f[a] -> f[b])
+struct Tup(a, b)
+
+refl_sub: forall a. Sub[a, a] = Sub(x -> x) 
+refl_any: Sub[forall a. a, exists a. a] = refl_sub
+#ignore: Tup[forall a. Sub[a, a], Sub[forall a. a, exists a. a]] = Tup(refl_sub, refl_any)
+ignore = Tup(refl_sub, refl_any)
```
johnynek: add more tests here, looks like we haven't solved it yet.

core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
```diff
@@ -1191,8 +1189,26 @@ object Infer {
...
+                      // we apply the annotation here and let Normalization
+                      // instantiate. We could explicitly have
+                      // instantiation TypedExpr where you pass the variables to set
+                      TypedExpr.Annotation(te, declared)
+                  }
+                }
+          }
+        case _ =>
+          None
+      })
```
johnynek, line 1215: We should be able to match on any Quantification and instantiate it correctly. The existential and universal require instantiating two different directions to only one or the other is easy.

core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
```diff
@@ -1462,6 +1479,130 @@ object Infer {
       }
     }
 
+    // noshadow must include any free vars of args
+    def liftQuantification[A](
```
johnynek: Should add direct tests of this in simple cases and possibly property checks. For instance, if there is a quantification on any arg, it should be lifted up. Or rather the result should not have quantification on any argument.

PR #1168, opened 2024-03-15
Support comments in headers

johnynek:
this allows comments in headers, but discards them: it does not retain them in the parsed AST.

This is convenient when we are really strict about imports and exports.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Package.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Package.scala
@@ -190,10 +190,12 @@ object Package {
         Doc.intercalate(Doc.empty, p :: i :: e :: b)
     }
 
+
   def headerParser(defaultPack: Option[PackageName]): P0[Header] = {
     val spaceComment: P0[Unit] =
-      (Parser.spaces | (P.char('#') <* P.charsWhile0(_ != '\n'))).?.void
-    val eol = Parser.commentToEOL.void | Parser.toEOL
+      (Parser.spaces.? ~ CommentStatement.commentPart.?).void
+
+    val eol = spaceComment <* Parser.termination
     val parsePack = Padding
       .parser(
         (P.string("package")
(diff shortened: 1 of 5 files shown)


core/src/main/scala/org/bykn/bosatsu/Parser.scala
```diff
@@ -448,7 +448,11 @@ object Parser {
     (P.char('(') ~ ws) *> pa <* (ws ~ P.char(')'))
 
   val newline: P[Unit] = P.char('\n')
-  val toEOL: P0[Unit] = maybeSpace *> newline.orElse(P.end)
+  // The comment excluding the leading # up to the EOL or EOF
+  val termination: P0[Unit] = newline.orElse(P.end)
+  val commentToEOL: P[String] =
+    P.char('#') *> P.charsWhile0(_ != '\n') <* termination
```
johnynek: This only parses a single comment to EOL. We could have several. This isn't a problem for the code as it is now due to the way it's used but it's dangerous to leave it around since it isn't clear if it parses one or many lines.

Also it doesn't allow space then a comment which makes it rarely useful.

core/src/main/scala/org/bykn/bosatsu/Package.scala
```diff
@@ -191,11 +191,14 @@ object Package {
     }
 
   def headerParser(defaultPack: Option[PackageName]): P0[Header] = {
-    // TODO: support comments before the Statement
+    val spaceComment: P0[Unit] =
+      (Parser.spaces | (P.char('#') <* P.charsWhile0(_ != '\n'))).?.void
+    val eol = Parser.commentToEOL.void | Parser.toEOL
```
johnynek: This should be spaceComment *> termination.

core/src/main/scala/org/bykn/bosatsu/Package.scala
```diff
@@ -191,11 +191,14 @@ object Package {
     }
 
   def headerParser(defaultPack: Option[PackageName]): P0[Header] = {
-    // TODO: support comments before the Statement
+    val spaceComment: P0[Unit] =
+      (Parser.spaces | (P.char('#') <* P.charsWhile0(_ != '\n'))).?.void
```
johnynek: Parser should be moved to a private val above.

johnynek (commented):
Should add tests of header parsing.

PR #1169, opened 2024-03-15
Add a Bosatsu/Queue property check

johnynek:
This is trying to exercise the property check API a bit, and of course the language at the same time.

### DIFF
--- a/test_workspace/Queue.bosatsu
+++ b/test_workspace/Queue.bosatsu
@@ -1,6 +1,11 @@
 package Queue
 
 from Bosatsu/List import eq_List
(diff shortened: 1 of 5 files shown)


test_workspace/Queue.bosatsu
```diff
@@ -78,6 +83,32 @@ eq_li = eq_List(eq_Int)
 
 q12 = empty.push(1).push(2)
 
+def samp(r)(fn): flat_map_Rand(r, fn)
```
johnynek, line 86: this isn't great. Either we should be really good at optimizing this to remove the uncurrying or I think bias to the uncurried desugaring in the first place.

PR #1171, opened 2024-03-16
add complete queue property checks

johnynek:
trying to exercise property checks by adding more checks for Queue. Definitely it can be tricky to implement functions in Bosatsu due to the totality requirements. For instance, implementing Rand::one_of was a bit tricky. You have to find the right recursion strategy. I was trying to make sampling O(log N), so I needed to make a tree. To do that, I used BinNat to be able to recur on half the size of the items, rather than the naive approach.

I also found a nice trick.

Define:
```
external def todo(ignore: x) -> forall a. a
```
then we can consume args and magically produce things by doing `todo((foo, bar))` and run `type-check`. Then it fails at runtime since we don't actually implement todo. We could use this strategy except possibly make that function only exist in predef when running type-check (which doesn't output anything).

### DIFF
--- a/test_workspace/Nat.bosatsu
+++ b/test_workspace/Nat.bosatsu
@@ -1,7 +1,7 @@
 package Bosatsu/Nat
 
 from Bosatsu/Predef import add as operator +, times as operator *
-export Nat(), times2, add, mult, exp, to_Int, to_Nat
+export Nat(), times2, add, mult, exp, to_Int, to_Nat, is_even, div2
 
 # This is the traditional encoding of natural numbers
 # it is useful when you are iterating on all values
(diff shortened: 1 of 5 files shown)


test_workspace/Nat.bosatsu
```diff
@@ -33,6 +33,23 @@ def mult(n1: Nat, n2: Nat) -> Nat:
...
+            case Succ(n): loop(n, False if res else True)
+    loop(n, True)
+
+def div2(n: Nat) -> Nat:
+    def loop(n: Nat, acc: Nat):
+        recur n:
+            case Zero: acc
+            case Succ(n):
+                # (n + 1) / 2 = n/2 if n is even, else n/2 + 1
+                if is_even(n): loop(n, acc)
```
johnynek: we know that every other item is even. This algorithm is quadratic, but we can make it O(N).

test_workspace/Queue.bosatsu
```diff
@@ -48,7 +50,7 @@ def fold_Queue(Queue(f, b): Queue[a], init: b, fold_fn: (b, a) -> b) -> b:
   b.reverse().foldLeft(front, fold_fn)
 
 def reverse_Queue(Queue(f, b): Queue[a]) -> Queue[a]:
-  Queue(b.reverse(), f.reverse())
```
johnynek, line 51: this bug was detected by property checks! Yay!

test_workspace/Queue.bosatsu
```diff
@@ -85,14 +87,46 @@ q12 = empty.push(1).push(2)
...
+    recur depth:
+        case Zero: queue_from_list
+        case Succ(n):
+          smaller = rand_Queue_depth(n)
+          pop_rand = smaller.map_Rand(pop)
+          push_rand = prod_Rand(rand_int, smaller).map_Rand(((h, q)) -> push(q, h))
+          rev_rand = smaller.map_Rand(reverse_Queue)
+          one_of(pop_rand, [push_rand, rev_rand])
+
+rand_Queue_Int: Rand[Queue[Int]] = rand_Queue_depth(int_to_Nat(50))
```
johnynek, line 110: this is building a pretty giant tree perhaps... definitely the checks take a while to run.

PR #1176, opened 2024-03-18
Add property checks for BinNat

johnynek:
add the same tests for BinNat as Nat.

### DIFF
--- a/test_workspace/BinNat.bosatsu
+++ b/test_workspace/BinNat.bosatsu
@@ -3,7 +3,7 @@ package Bosatsu/BinNat
 from Bosatsu/Nat import Nat, Zero as NatZero, Succ as NatSucc, times2 as times2_Nat
 
 export (BinNat(), toInt, toNat, toBinNat, next, add_BinNat, times2, div2,
-  prev, times_BinNat, exp, cmp_BinNat, is_even, sub_BinNat)
+  prev, times_BinNat, exp, cmp_BinNat, is_even, sub_BinNat, eq_BinNat)
 # a natural number with three variants:
 # Zero = 0
 # Odd(n) = 2n + 1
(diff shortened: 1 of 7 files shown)


test_workspace/BinNat.bosatsu
```diff
@@ -97,11 +130,45 @@ def times2(b: BinNat) -> BinNat:
...
+            case _: Odd(diff)
+        Even(right):
+          # (2n + 2) - (2m + 2) = 2(n - m)
+          times2(sub_BinNat(left, right))
+
 def div2(b: BinNat) -> BinNat:
     match b:
         case Zero: Zero
-        case Odd(n): n
-        case Even(n): prev(n)
```
johnynek, line 104: this was a bug! yay for property checks!

test_workspace/BinNat.bosatsu
```diff
@@ -97,11 +130,45 @@ def times2(b: BinNat) -> BinNat:
...
+      match right:
+        Zero: even
+        Odd(right):
+          # (2n + 2) - (2m + 1)
+          # if n >= m: 2(n - m) + 1
+          # if n < m: 0
+          diff = sub_BinNat(left, right)
+          match diff:
+            case Zero:
+              match cmp_BinNat(left, right):
```
johnynek: this seems bad to do internally, seems like we should be able to do this once outside the loop... should think some more.

test_workspace/BinNat.bosatsu
```diff
@@ -97,11 +142,51 @@ def times2(b: BinNat) -> BinNat:
...
+            # (2n + 1) - (2m + 2) = 2(n - m) - 1
+            # note if (2n + 1) > (2m + 2), then n > m
+            times2(loop(left, right)).prev()
+      case Even(left) as even:
+        match right:
+          case Zero: even
+          case Odd(right):
+            # (2n + 2) - (2m + 1)
+            # if n >= m: 2(n - m) + 1
+            # if (2n + 2) > (2m + 1), then n > m
```
johnynek: this is wrong... if (2n + 2) > (2m + 1) then n >= m. So, that means the invariant must actually be left >= right. So that means we only know that (2n + 2) >= (2m + 1), but that's okay, because we know an odd number can't equal an even number, so we recover n >= m.

test_workspace/BinNat.bosatsu
```diff
@@ -97,11 +142,51 @@ def times2(b: BinNat) -> BinNat:
...
+            Odd(loop(left, right))
+          case Even(right):
+            # (2n + 2) - (2m + 2) = 2(n - m)
+            times2(loop(left, right))
+      case Zero:
+        # this is last because it is actually unreachable
```
johnynek: this isn't true. The invariant is left >= right

test_workspace/BinNat.bosatsu
```diff
@@ -97,11 +142,51 @@ def times2(b: BinNat) -> BinNat:
       #2(2(n + 1)) = 2((2n + 1) + 1)
       Even(Odd(n))
 
+def sub_BinNat(left: BinNat, right: BinNat) -> BinNat:
+  # invariant: left > right
```
johnynek: see below: invariant: left >= right, so the comment below is wrong, we can return Zero.

PR #1183, opened 2024-03-23
Fix bug in python if/else chains

### LINKED ISSUE #1181: python code generation has a bug
see the diff here:

https://github.com/johnynek/bosatsu/pull/1176/commits/d32020f178de4e87fec06256c213aa4d6546d82f

this shouldn't change anything: the order we match cases may change efficiency, but it shouldn't change the result in cases like this.

But inspecting the generated python code it is definitely wrong. The bug seems to be with python generation and not the Matchless AST since that is used for the interpretter tests which pass in both cases.

The cmp_BinNat expression is simple enough, but clearly miscompiled when the order isn't starting with Zero.

johnynek:
close #1181 

This bug has been lurking in python code generation for some time, but only with generative testing did we push hard enough to find it.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Matchless.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Matchless.scala
@@ -173,7 +173,20 @@ object Matchless {
         l.nonEmpty || hasSideEffect(b)
     }
 
-  case class If(cond: BoolExpr, thenExpr: Expr, elseExpr: Expr) extends Expr
+  case class If(cond: BoolExpr, thenExpr: Expr, elseExpr: Expr) extends Expr {
+    def flatten: (NonEmptyList[(BoolExpr, Expr)], Expr) = {
+      def combine(expr: Expr): (List[(BoolExpr, Expr)], Expr) =
+        expr match {
+          case If(c1, t1, e1) =>
+            val (ifs, e2) = combine(e1)
+            (((c1, t1)) :: ifs, e2)
+          case last => (Nil, last)
+        }
+
+      val (rest, last) = combine(elseExpr)
+      (NonEmptyList((cond, thenExpr), rest), last)
+    }
+  }
   case class Always(cond: BoolExpr, thenExpr: Expr) extends Expr
   def always(cond: BoolExpr, thenExpr: Expr): Expr =
     if (hasSideEffect(cond)) Always(cond, thenExpr)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/org/bykn/bosatsu/Matchless.scala
```diff
@@ -173,7 +173,20 @@ object Matchless {
...
     }
 
-  case class If(cond: BoolExpr, thenExpr: Expr, elseExpr: Expr) extends Expr
+  case class If(cond: BoolExpr, thenExpr: Expr, elseExpr: Expr) extends Expr {
+    def flatten: (NonEmptyList[(BoolExpr, Expr)], Expr) = {
+      def combine(expr: Expr): (List[(BoolExpr, Expr)], Expr) =
+        expr match {
+          case If(c1, t1, e1) =>
+            val (ifs, e2) = combine(e1)
+            (((c1, t1)) :: ifs, e2)
```
johnynek, line 182: previously this code was in PythonGen, but it really only depends on Matchless.Expr. The order was being reversed by appending to the tail: `ifs :+ ((c1, t1))` which for some branches changes semantics (where Matchless has leveraged prior information to do less work).

core/src/main/scala/org/bykn/bosatsu/codegen/python/PythonGen.scala
```diff
@@ -1878,17 +1878,8 @@ object PythonGen {
             // there is no need to
             loop(in, slotName)
           case Literal(lit) => Env.pure(Code.litToExpr(lit))
-          case If(cond, thenExpr, elseExpr) =>
-            def combine(expr: Expr): (List[(BoolExpr, Expr)], Expr) =
-              expr match {
-                case If(c1, t1, e1) =>
-                  val (ifs, e2) = combine(e1)
-                  (ifs :+ ((c1, t1)), e2)
```
johnynek, line 1886: note the order is wrong here: we are pushing `c1, t1` on to the end of `ifs` but that came from expanding `e1` so it should come before `e1` since it was before originally.

core/src/test/scala/org/bykn/bosatsu/MatchlessTests.scala
```diff
@@ -151,4 +156,21 @@ class MatchlessTest extends AnyFunSuite {
       assert(matchlessRes == matchRes)
     }
   }
+
+  test("If.flatten can be unflattened") {
```
johnynek, line 160: this test didn't always find the bug (since randomly finding a case is somewhat rare, but after 3 runs, one found it, so I think it would be flakey enough that we would discover the issue).

test_workspace/BinNat.bosatsu
```diff
@@ -71,6 +67,10 @@ def cmp_BinNat(a: BinNat, b: BinNat) -> Comparison:
             case GT | EQ: GT
             case LT: LT
         case Zero: GT
+    case Zero:
```
johnynek, line 70: this is more efficient since Zero is rare, but triggered the bug before this fix.

PR #1193, opened 2024-04-01
Optimize if-matches case

johnynek:
the pattern:
```
if x matches p:
  trueCase
else:
  falseCase
```

was being compiled into:
```
b = match x:
  case p: True
  case _: False

match b:
  case True: trueCase
  case False: falseCase
```
rather than relying on optimization later, it's easy to detect this pattern while still at the source AST and directly convert to:
```
match x:
  case p: trueCase
  case _: falseCase
```

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
@@ -348,6 +348,19 @@ final class SourceConverter(
             else RecursionKind.NonRecursive
           Expr.Let(boundName, lam, in, recursive = rec, decl)
         }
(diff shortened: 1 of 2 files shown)


core/src/main/scala/org/bykn/bosatsu/SourceConverter.scala
```diff
@@ -348,6 +348,19 @@ final class SourceConverter(
             else RecursionKind.NonRecursive
           Expr.Let(boundName, lam, in, recursive = rec, decl)
         }
+      case IfElse(NonEmptyList((Matches(a, p), res), Nil), elseCase) =>
```
johnynek: It's better to make sure there are no names here and error than rewrite possibly with names and then have a less precise error condition.

Issue #1233, opened 2024-10-23
existential type unification issues


johnynek:
Here is a simple case that does not typecheck:
```
in file: test_workspace/Prog.bosatsu:18:20, package Bosatsu/Prog
type error: expected type $env$2e to be the same as type $env$8e
16|
17|def main(p: Prog[exists env. env, String, Int]) -> Prog[exists env. env, String, Int]:
18|  p
```

in this case we are literally just annotating the return type, but still, we can't unify. Clearly something is wrong.

PR #1236, opened 2024-10-25
fix some existential type unification cases

johnynek:
this shows the https://github.com/johnynek/bosatsu/issues/1233 failure.

For some reason, the existential case works, I bet due to the special casing of substitution that was added.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
@@ -43,12 +43,6 @@ sealed abstract class Infer[+A] {
   final def mapEither[B](fn: A => Either[Error, B]): Infer[B] =
     Infer.Impl.MapEither(this, fn)
 
-  def debug(showFn: A => String): Infer[A] =
-    mapEither { a =>
-      println(showFn(a))  
-      Right(a)
-    }
-
   final def runVar(
       v: Map[Infer.Name, Type],
       tpes: Map[(PackageName, Constructor), Infer.Cons],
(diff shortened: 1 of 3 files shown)


johnynek:
close #1233

core/src/main/scala/org/bykn/bosatsu/rankn/Infer.scala
```diff
@@ -500,20 +502,20 @@ object Infer {
...
             // we know the kind of left is k -> x, and right has kind k
             // since left: Rho, we know loop(left, path)._3 is Rho
-            (varianceOfCons(ta, region), loop(left, path))
+            (varianceOfCons(ta, region), loop(left, allCo))
               .flatMapN { case (consVar, (sksl, el, ltpe0)) =>
                 // due to loop invariant
                 val ltpe: Type.Rho = ltpe0.asInstanceOf[Type.Rho]
-                val rightPath = consVar * path
-                loop(right, rightPath)
+                val allCoRight = allCo && (consVar == Variance.co)
```
johnynek, line 531: I don't *really* know why this works, but it is skolemizing in a smaller set of cases from before, yet still keeping functions in weak-prenex form (which isn't type checked, but is asserted in the inference and typechecking process).

So, I guess one argument is we want to skolemize the least we can to make inference work, especially with function return types, and this is sufficient to do it. It also passes all the current tests.

The paper I based the type inference on doesn't have a type system as complex as bosatsu, so I had to generalize it, and in this case I suppose my guess was wrong (which was refusing to typecheck things that seem like they should, such as the examples in this PR).

PR #1262, opened 2024-11-19
evaluate static string matches in TypedExprNormalization

johnynek:
I noticed we are generating code for static string matches that could be done in advance.

This refactors to use the existing string matching code statically.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Matchless.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Matchless.scala
@@ -2,6 +2,7 @@ package org.bykn.bosatsu
 
 import cats.{Monad, Monoid}
 import cats.data.{Chain, NonEmptyList, WriterT}
+import org.bykn.bosatsu.pattern.StrPart
 import org.bykn.bosatsu.rankn.{DataRepr, Type, RefSpace}
 
 import Identifier.{Bindable, Constructor}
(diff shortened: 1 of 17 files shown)


core/src/main/scala/org/bykn/bosatsu/Matchless.scala
```diff
@@ -810,63 +930,139 @@ object Matchless {
         LetMut(anon, rest)
       }
 
+    def staticMatch(arg: Expr, branches: NonEmptyList[
+        (Pattern[(PackageName, Constructor), Type], Expr)
+      ]
+    ): Option[Expr] =
+      arg match {
+        case Literal(Lit.Str(s)) =>
```
johnynek: We could extend this very easily to any Literal which is probably all we need to handle here.

That said, maybe this code should really be pushed into TypedExprNormalization since applying these optimizations could open other normalizations.

core/src/test/scala/org/bykn/bosatsu/pattern/StrPartTest.scala
```diff
@@ -0,0 +1,103 @@
+package org.bykn.bosatsu.pattern
+
+import org.bykn.bosatsu.Generators.{genStrPat, genValidUtf}
+import org.bykn.bosatsu.StringUtil
+import org.scalacheck.Prop.forAll
+import org.scalacheck.Gen
+
+class StrPartTest extends munit.ScalaCheckSuite {
```
johnynek, line 8: this test actually also tests that all the matching algorithms agree. It seems we have something like 4 of them:

1. StrPart
2. SeqPattern + matcher
3. NamedSeqPattern + matcher
4. java regex

johnynek: I *think* StrPart is the fastest, but I haven't carefully benchmarked it, but they should all agree, and the speed only matters for the MatchlessToValue runtime, in normalization, it doesn't matter I don't think.

PR #1266, opened 2024-11-20
Pass in MatchString if a string match must be total (which we already knew)

johnynek:
This also improves the python code generation somewhat after looking more at the generated code.

### DIFF
--- a/cli/src/test/scala/org/bykn/bosatsu/codegen/python/CodeTest.scala
+++ b/cli/src/test/scala/org/bykn/bosatsu/codegen/python/CodeTest.scala
@@ -521,7 +521,17 @@ else:
             assert(tern == f.simplify)
           }
         case whoKnows =>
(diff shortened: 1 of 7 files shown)


johnynek:
with this we get

```
def string_to_Char(s: String) -> Option[Char]:
  match s:
    case "$.{c}": Some(c)
    case _: None
```
compiling to:

```python
def string_to_Char(s):
    if s.__len__() == 1:
        return (1, s[0])
    else:
        return (0,)
```

and
```
def last_String(s: String) -> Option[Char]:
  match s:
    case "": None
    case "${_}$.{l}": Some(l)
```

compiling to:

```python
def last_String(s):
    if s == "":
        return (0,)
    else:
        return (1, s[-1])
```

Even though C backend is the priority, by improving python code-gen we learn about how to make codegen in general better.

PR #1267, opened 2024-11-21
Implement matchString in C backend

johnynek:


johnynek:
looks like we are generating matching code in the wrong order:

```c
BValue ___bsts_g_Bosatsu_l_Char_l_starts__with__foo(BValue __bsts_b_s0) {
    int __bsts_l_offset70 = 0;
    __bsts_l_offset70 = __bsts_l_offset70 + (3);
    return BSTS_AND(bsts_string_matches_at(__bsts_b_s0,
            __bsts_l_offset70,
            BSTS_NULL_TERM_STATIC_STR("foo")),
        1) ?
        alloc_enum0(1) :
        alloc_enum0(0);
}
```
note how we update the length of foo *before* we check if we match. We should update after and only if that is successful

johnynek:
oh, I think onExpr in the C code may just be wrong... it is ignoring the fact that the scope can change. That or the usage of it in andCode.

PR #1278, opened 2024-11-27
run list tests

johnynek:
this runs the tests for the Bosatsu/List module using the generated C code.

Two tests currently fail (I'm sure there are many bugs still to squash).


c_runtime/bosatsu_runtime.c
```diff
@@ -20,12 +21,14 @@ BValue* closure_data_of(Closure1Data* s) {
...
   size_t slots = s->slot_len;
-  BValue* items = closure_data_of(s);
+  BValue* items0 = closure_data_of(s);
+  BValue* items = items0;
   while (slots > 0) {
     release_value(items);
     items = items + 1;
     slots = slots - 1;
   }
+  free(items0);
```
johnynek: we don't need to do this... this memory was allocated in the same malloc as s.

PR #1289, opened 2024-12-02
Fix bug with chars packed into BValue

johnynek:


c_runtime/bosatsu_runtime.c
```diff
@@ -69,7 +69,7 @@ BValue bsts_unit_value() {
 }
 
 BValue bsts_char_from_code_point(int codepoint) {
-  return (BValue)PURE_VALUE(codepoint);
+  return (BValue)TO_PURE_VALUE(codepoint);
```
johnynek, line 72: this was the bug... very easy to make bugs without the compiler to check things.

PR #1290, opened 2024-12-02
Fix partition in the C runtime

johnynek:


c_runtime/bosatsu_ext_Bosatsu_l_Predef.c
```diff
@@ -182,7 +182,7 @@ BValue ___bsts_g_Bosatsu_l_Predef_l_partition__String(BValue a, BValue b) {
   // Some((x, y)) with x = a[0:offset], y = a[offset + b.len():]
   BValue x = bsts_string_substring(a, 0, offset);
   BValue y = bsts_string_substring_tail(a, offset + blen);
-  res = alloc_enum2(1, x, y);
+  res = alloc_enum1(1, alloc_struct2(x, y));
```
johnynek, line 185: this should have been `Some((x, y))` but the tuple was being flattened into an ADT that wouldn't match the shape.

PR #1293, opened 2024-12-04
Fix bug with shifts and recursive closures

johnynek:
There are two bugs fixed here (sorry)...

1. the shifting code wasn't using twos compliment (I am having second thoughts about using a sign + magnitude representation, but as long as it is here, for negative numbers we need to convert to twos compliment for bitwise ops).
2. if a closure is used in a nested recursive position, we were incorrectly allocating the value for it. This can happen rarely when the optimizer rewrites a function call to being nested. This seems to be due to an optimization that we push down calls as low as we need them, which can result in pushing them into nested calls. Maybe we don't want to push values down since it can also wind up multiplying work and de-optimizing calls (e.g. imagine `let x = expensive(y) in z.map(w -> (w, x))` here we don't want to push x down since we would wind up computing it on every call. That substitution should probably only be done when we can be sure it will be called 0 or 1 times, (e.g. into a match or lets).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,7 +127,7 @@ jobs:
       - name: "build assembly"
         run: "sbt \"++${{matrix.scala}}; cli/assembly\""
       - name: "generate c code"
-        run: "./bosatsuj transpile --input_dir test_workspace/ --package_root test_workspace/ --outdir c_out c --test --filter Bosatsu/List --filter IntTest --filter Bosatsu/BinNat --filter Bosatsu/Char --filter PredefTests"
+        run: "./bosatsuj transpile --input_dir test_workspace/ --package_root test_workspace/ --outdir c_out c --test"
       - name: "compile generated c code"
         run: |
           cp c_runtime/*.h c_out
--- a/core/src/main/scala/org/bykn/bosatsu/codegen/clang/ClangGen.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/codegen/clang/ClangGen.scala
@@ -1033,12 +1033,8 @@ object ClangGen {
                     pv(boxFn(nm, arity))
                   }
                   else {
-                    // we need to allocate another wrapper
-                    pv(Code.Ident(s"alloc_closure${arity}")(
-                      Code.IntLiteral(BigInt(arity)),
-                      slotsArgName,
-                      nm
-                    ))
+                    // recover the pointer to this closure from the slots argument
+                    pv(Code.Ident("bsts_closure_from_slots")(slotsArgName))
                   }
                 case None =>
                   getBinding(arg).widen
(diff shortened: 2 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -127,7 +127,7 @@ jobs:
       - name: "build assembly"
         run: "sbt \"++${{matrix.scala}}; cli/assembly\""
       - name: "generate c code"
-        run: "./bosatsuj transpile --input_dir test_workspace/ --package_root test_workspace/ --outdir c_out c --test --filter Bosatsu/List --filter IntTest --filter Bosatsu/BinNat --filter Bosatsu/Char --filter PredefTests"
+        run: "./bosatsuj transpile --input_dir test_workspace/ --package_root test_workspace/ --outdir c_out c --test"
```
johnynek, line 130: we are removing the filters... this should test all the calls.

core/src/main/scala/org/bykn/bosatsu/codegen/clang/ClangGen.scala
```diff
@@ -844,10 +844,14 @@ object ClangGen {
 
       // We have to lift functions to the top level and not
       // create any nesting
-      def innerFn(fn: FnExpr): T[Code.ValueLike] = 
+      def innerFn(fn: FnExpr): T[Code.ValueLike] = {
+        val nameSuffix = fn.recursiveName match {
+          case None => ""
+          case Some(n) => Idents.escape("_", n.asString)
```
johnynek, line 850: this makes it easier to read the generated code.

core/src/main/scala/org/bykn/bosatsu/codegen/clang/ClangGen.scala
```diff
@@ -1022,10 +1027,20 @@ object ClangGen {
...
-                  // a closure can't be a static name
-                  pv(boxFn(nm, arity))
-                case _ =>
+                case Some((nm, isClosure, arity)) =>
+                  if (!isClosure) {
+                    // a closure can't be a static name
+                    pv(boxFn(nm, arity))
+                  }
+                  else {
+                    // we need to allocate another wrapper
```
johnynek: this was a bug... if we have a direct function that is a closure, we can't just pass it by value, we have to box it in a closure wrapper to carry the environment.

core/src/main/scala/org/bykn/bosatsu/codegen/clang/ClangGen.scala
```diff
@@ -1022,10 +1027,20 @@ object ClangGen {
...
+                    pv(Code.Ident(s"alloc_closure${arity}")(
+                      Code.IntLiteral(BigInt(arity)),
+                      slotsArgName,
```
johnynek: eventually we will need to bump the references on these values.

PR #1298, opened 2024-12-05
track memory state c

johnynek:
The goal of this work is to make the code pass tests and have valgrind believe there are no memory leaks.


johnynek:
#1302 using Boehm libgc superseded this work for now.

PR #1302, opened 2024-12-09
try boehm gc

johnynek:
in #1298 I was trying to leverage the fact that all objects are immutable and form a DAG to justify reference counting.

Doing this correctly is a significant challenge, but also it makes the code much more complex. My reading of the GC literature is that it is actually very hard to make reference counting perform as well as GC and its conceptual simplicity is appealing, but a mirage. I've seen reports the tail GC time of ref counting (which still exists when large dags fall out of scope) is also very large, but since most objects don't get many references and are almost immediately collected, you are doing a lot of extra work (maybe 30-50%) not to mention making the code harder for the c compiler to optimize.

I can always come back to this if I want to continue, but it seems wise to just use boehm like virtually every other project I can find that needs some form of garbage collection.

### DIFF
--- a/c_runtime/bosatsu_runtime.h
+++ b/c_runtime/bosatsu_runtime.h
@@ -29,6 +29,8 @@ BValue get_struct_index(BValue v, int idx);
 
 // &BValue -> Tag
 ENUM_TAG get_variant(BValue v);
+// This is only safe if all enums have zero args
+ENUM_TAG get_variant_value(BValue v);
 // (&BValue, int) -> &BValue
 BValue get_enum_index(BValue v, int idx);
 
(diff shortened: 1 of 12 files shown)


c_runtime/bosatsu_runtime.h
```diff
@@ -21,11 +21,8 @@ typedef void (*FreeFn)(void*);
 // A function which constructs a BValue
 typedef BValue (*BConstruct)();
 
-// Function to determine the type of the given value pointer and clone if necessary
-// &BValue -> BValue
-BValue clone_value(BValue value);
-// BValue -> ()
-void release_value(BValue value);
+// delta may be negative or positive
+void bsts_increment_value(BValue value, int delta);
```
johnynek, line 25: Remove this

PR #1313, opened 2024-12-11
Improve TypedExpr.substitute issue 1126

### LINKED ISSUE #1126: improve TypedExpr.substitute and Declaration.substitute
The unshadow algorithm in Code.Lambda added in this commit: https://github.com/johnynek/bosatsu/pull/1124/commits/dc5389bea86bbd3910e07ed1c230c66c52355934

could be used in TypedExpr and Declaration. Both of which give up and return None if they hit a condition where it is needed.

By making that change, we could make those functions always return, which could improve optimizations.

johnynek:
close #1126

This properly handles masking by renaming so that substitution can always proceed. This opened up three optimizations in the tests, but in real code the kind of shadowing this is dealing with may not often happen, so I don't know if it matters much.

(I should really do this with a real benchmark suite).

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/TypedExprNormalization.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/TypedExprNormalization.scala
@@ -119,6 +119,16 @@ object TypedExprNormalization {
   private def setType[A](expr: TypedExpr[A], tpe: Type): TypedExpr[A] =
     if (!tpe.sameAs(expr.getType)) Annotation(expr, tpe) else expr
 
+  private def appLambda[A](f1: AnnotatedLambda[A], args: NonEmptyList[TypedExpr[A]], tpe: Type, tag: A): TypedExpr[A] = {
+    val AnnotatedLambda(lamArgs, expr, _) = f1
+    // (y -> z)(x) = let y = x in z
+    val lets = lamArgs.zip(args).map { case ((n, ltpe), arg) =>
+      (n, setType(arg, ltpe))
+    }
+    val expr2 = setType(expr, tpe)
+    TypedExpr.letAllNonRec(lets, expr2, tag)
+  }
+
   /** if the te is not in normal form, transform it into normal form
     */
   private def normalizeLetOpt[A, V](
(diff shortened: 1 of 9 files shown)


core/src/main/scala/org/bykn/bosatsu/TypedExprNormalization.scala
```diff
@@ -119,6 +119,16 @@ object TypedExprNormalization {
   private def setType[A](expr: TypedExpr[A], tpe: Type): TypedExpr[A] =
     if (!tpe.sameAs(expr.getType)) Annotation(expr, tpe) else expr
 
+  private def appLambda[A](f1: AnnotatedLambda[A], args: NonEmptyList[TypedExpr[A]], tpe: Type, tag: A): TypedExpr[A] = {
+    val AnnotatedLambda(lamArgs, expr, _) = f1
+    // (y -> z)(x) = let y = x in z
```
johnynek: This law seems suspect. What if the args have free variables that intersect with the lambda args? It isn't this simple.

core/src/main/scala/org/bykn/bosatsu/codegen/clang/ClangGen.scala
```diff
@@ -1168,35 +1169,43 @@ object ClangGen {
...
-                case Some(items @ NonEmptyList(Right(idx), _)) =>
-                  Right(idx + 1) :: items
-                case Some(items @ NonEmptyList(Left((_, idx)), _)) =>
-                  Right(idx + 1) :: items
+                case None => NonEmptyList.one(Right(cnt))
+                case Some(items) =>
+                  Right(cnt) :: items
               }  
-              result(s.copy(binds = s.binds.updated(bn, v)), ())
+              (s.copy(bindCount = cnt + 1, binds = s.binds.updated(bn, v)), ())
```
johnynek: this was a bug. We need each scope to be a distinct variable in C, but popping off the stack reused the variable names. With some of the inlining changes, we triggered the bug. The fix here is to just allocate a new ID for each binding inside of a top level name and reset on each value.

core/src/main/scala/org/bykn/bosatsu/TypedExprNormalization.scala
```diff
@@ -247,10 +262,12 @@ object TypedExprNormalization {
                   if ((e1 eq expr) && (lamArgs === lamArgs0)) None
                   else Some(AnnotatedLambda(lamArgs, e1, tag))
                 }
+                /*
               } else {
                 if ((e1 eq expr) && (lamArgs === lamArgs0)) None
                 else Some(AnnotatedLambda(lamArgs, e1, tag))
               }
+              */
```
johnynek: need to clean this up before merging

core/src/main/scala/org/bykn/bosatsu/TypedExprNormalization.scala
```diff
@@ -227,18 +239,21 @@ object TypedExprNormalization {
...
+              //if (e1 != e2) {
                 // in this case we have inlined, vs there already being
                 // a literal lambda being applied
                 // by normalizing this, it will become a let binding
-                val e3 = normalize1(None, e2, bodyScope, typeEnv).get
+                //val e3 = normalize1(None, e2, bodyScope, typeEnv).get
 
-                if (e3.size <= expr.size) {
+                //if (e3.size <= expr.size) {
+                if (true) {
```
johnynek: this is janky, need to remove

PR #1323, opened 2024-12-17
WIP: Use cats-effect in core

### LINKED ISSUE #1185: Use cats-effect in core
Cats effect works on all scala platforms now. No reason not to use it in core and we can avoid some of the tagless and also the Par wrapper for parallelism.

johnynek:
close #1185


johnynek:
better to just restart in light of #1325 and #1327

PR #1327, opened 2024-12-19
Introduce Fs2 based PlatformIO supporting scalajs

### LINKED ISSUE #393: build a nodejs version of the compiler
Not sure who would be very interested, but it is interesting to note that I think it is very easy to make a nodejs version of bosatsu since we use very pure scala and scalajs works.

This could be nice for users who really hate the JVM.

Although, shipping static native-image binaries is probably the way to go there since the performance will be much better.

johnynek:
close #393 

This allows users to run bosatsu without the jvm.


core/src/main/scala/org/bykn/bosatsu/tool/Fs2Module.scala
```diff
@@ -0,0 +1,37 @@
+package org.bykn.bosatsu.tool
+
+import cats.{effect => ce}
+import cats.effect.{IO, Resource}
+import fs2.io.file.Path
+import org.bykn.bosatsu.{Par, MainModule, Fs2PlatformIO}
+
+object Fs2Module extends MainModule[IO, Path](Fs2PlatformIO) { self =>
```
johnynek: We could move this to cliJS module to reduce the dependencies of core.

PR #1328, opened 2024-12-22
Remove optional ops from PlatformIO

johnynek:
really PlatformIO should abstract over a real filesytem and not be too vague and pretend like there maybe not even such thing as directories. Simulating that in memory isn't too much more work, so just do it.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/MemoryMain.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/MemoryMain.scala
@@ -274,7 +274,7 @@ object MemoryMain {
         def writeStdout(doc: Doc): F[Unit] =
           catsDefaultME.raiseError(new Exception(s"writeStdout($doc) is unimplemented on a read only platform"))
 
-        def print(str: String): F[Unit] =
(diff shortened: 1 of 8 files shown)


core/src/main/scala/org/bykn/bosatsu/MemoryMain.scala
```diff
@@ -1,17 +1,18 @@
...
 
 import cats.syntax.all._
+import cats.data.ValidatedNel
 
-class MemoryMain[G[_], K](
-  platform: PlatformIO[Kleisli[G, MemoryMain.State[K], *], K]) extends
-  MainModule[Kleisli[G, MemoryMain.State[K], *], K](platform) {
+class MemoryMain[G[_]](
+  platform: PlatformIO[Kleisli[G, MemoryMain.State, *], Chain[String]]) extends
+  MainModule[Kleisli[G, MemoryMain.State, *], Chain[String]](platform) {
```
johnynek, line 15: next we should change this to being StateT not Kleisli, then we can support writing as well as reading and we can also include a stdout and stderr `Chain[String]` of lines to print.

PR #1336, opened 2024-12-24
support directly building an executable from bosatsu code

johnynek:
this simplifies iteration because we can have bosatsu directly invoke the c compiler on the platform.

This means we can change something, then directly just call bosatsu and not have to do that and make, etc...


core/src/main/scala/org/bykn/bosatsu/codegen/clang/ClangTranspiler.scala
```diff
@@ -81,15 +80,73 @@ case object ClangTranspiler extends Transpiler {
...
-  def traverseArgs: Traverse[Args] = implicitly
+      val cOpt = Opts.option[P]("output", "o", "name of output c code file.")
+        .orElse {
+          Opts("output.c")
+            .mapValidated(platformIO.path(_))
+        }
+
+      def parseCCFile(confPath: P): F[CcConf] =
+        for {
+          jsonStr <- platformIO.readUtf8(confPath)
```
johnynek: Should add a parse method to platformIO and maybe use that in MainModule

PR #1338, opened 2024-12-27
Add semver Version implementation

johnynek:


core/src/main/scala/org/bykn/bosatsu/library/Version.scala
```diff
@@ -0,0 +1,195 @@
...
+    (digits *> nonDigit *> identChars.?)).string
+
+  private val buildIdent: Parser[String] =
+    (digit | nonDigit).rep.string
+
+  private val preReleaseIdentifier = alphaNumIdent.backtrack | numericIdentifierString
+
+  private val build = Parser.char('+') *> buildIdent.repSep(dot).string
+  private val prerelease = Parser.char('-') *> preReleaseIdentifier.repSep(dot).string
+
```
johnynek: Expose the prerelease and build string parsers.

core/src/main/scala/org/bykn/bosatsu/library/Version.scala
```diff
@@ -0,0 +1,195 @@
...
+case class Version(
+  major: Int,
+  minor: Int,
+  patch: Int,
+  preRelease: Option[String],
+  build: Option[String]
+) {
+  override def toString(): String = render
+
```
johnynek, line 18: Add a validate method to check that the ints are not negative and pre and build strings are valid.

Issue #1152, opened 2024-02-24
design for bosatsu libraries


johnynek:
* Libraries are the units of code sharing.
* Packages are sets of 0 or more values and 0 or more type definitions.
* Libraries are sets of 0 or more packages.
* Libraries are versioned. We can compute the version of a library given the previous version and library.
* Libraries can depend on 0 or more other Libraries. That dependency graph forms a DAG.
* Libraries can have private (unexported) packages. Like packages, exporting values and types is opt-in, not opt out. Libraries export at the level of packages, not values. All transitively visible types from exported packages must be inside exported packages.
* Libraries can have public and private Library dependencies. A private dependency means that none of the values or types in that dependency Library appear on publicly visible values or types of the current library.
* The transitive set of dependencies that are reachable with more than one path ("diamond dependencies") must have a valid version that satisfies all the dependents: open question: should we declare in a Library the values and types that it depends on from public dependencies? In this way we can with library definitions we can compute if a given version will satisfy a dependency and we can unambiguously solve for a common version.
* When considering a library, the set of reachable private libraries do *not* need to share a single version. This significantly complicates linking since we need to link each package import to a specific version of a package.

johnynek:
If every library solves for a valid transitive set of public libraries then when adding dependencies, all diamonds are visible directly. So, if there is no collision in that list, you don't have a new diamond. If there is a collision, there is a set of candidate dependencies, and we can check each of those for all the N-1 others.

There should be a tool where we take the declared inputs, and then we generate the checked library output which would include the solutions to the current library, the previous version of the current library, and the hash of all the referenced libraries.

johnynek:
We could make a projection of a dependency by taking the union of all imports. So, a library could have a projected inferface that it depends on. This goes in the direction of doing a full solution of the dependency problem, and in this view, the version numbers don't really matter except that they are ordered so we can identify which ones come later.

johnynek:
Since when compiling, all diamond dependencies are solvable, that means the interfaces are compatible. So compiling each package isn't a problem. But once we have compiled and we are linking and emitting code, we need to know which version of the code we need to link to each package. We could merge them into a single namespace by appending an identifier for the version onto the names emitted when there is more than one, or we could use directories / package names.

Since ultimately, I'd like a C compilation target, it seems like appending versions to the names is the better approach.

johnynek:
Here is a possible syntax for declaring libraries:

```
library dev.posco.bar


[...]

PR #1340, opened 2024-12-29
Add Library cli options

johnynek:
this is the start to being able to use the cli to create library configurations and eventually publish libraries.

It's a long road.

part of #1152

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/MainModule.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/MainModule.scala
@@ -1,6 +1,6 @@
 package org.bykn.bosatsu
 
-import cats.data.{Validated, ValidatedNel, NonEmptyList}
+import cats.data.{Chain, Validated, ValidatedNel, NonEmptyList}
 import cats.implicits.catsKernelOrderingForOrder
 import cats.Traverse
 import com.monovore.decline.{Argument, Command, Help, Opts}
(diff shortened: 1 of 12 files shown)


core/src/main/scala/org/bykn/bosatsu/MainModule.scala
```diff
@@ -1224,6 +1230,19 @@ class MainModule[IO[_], Path](val platformIO: PlatformIO[IO, Path]) {
 
       case Output.Basic(doc, out) =>
         writeOut(doc, out).as(ExitCode.Success)
+
+      case Output.Many(items) =>
+        def loop(items: Chain[Output[Path]]): IO[ExitCode] =
```
johnynek: Better to convert to List once since we have to go all the way through.

johnynek: Actually foldM is the way to go here.

core/src/main/scala/org/bykn/bosatsu/library/LibConfig.scala
```diff
@@ -0,0 +1,62 @@
+package org.bykn.bosatsu.library
+
+import org.bykn.bosatsu.{Json, PackageName}
+import _root_.bosatsu.{TypedAst => proto}
+
+case class LibConfig(
+  name: String,
+  repoUri: String,
+  exportedPackages: List[PackageName],
+  internalPackages: List[PackageName],
```
johnynek, line 10: Instead of only internal this should be the set of everything and allow patterns so you don't have to edit this file every time you make a new package.

PR #1367, opened 2025-01-25
Refactor CompilerApi out of MainModule

johnynek:
the idea here is to make it easier to call these functions from the lib subcommand


core/src/main/scala/org/bykn/bosatsu/MainModule.scala
```diff
@@ -830,15 +731,14 @@ class MainModule[IO[_], Path](val platformIO: PlatformIO[IO, Path]) {
       type Result = Out
     }
 
-    def toTry[A](
-        cmd: MainCommand,
+    private def toTry[A](
```
johnynek: Remove this duplication

PR #1381, opened 2025-02-22
Add lib publish command

johnynek:


core/src/main/scala/org/bykn/bosatsu/library/Command.scala
```diff
@@ -197,6 +197,22 @@ object Command {
           )
         }
 
+    val allLibs: Opts[F[(P, SortedMap[Name, (LibConfig, P)])]] =
```
johnynek, line 214: just change this to a function `def readAllLibs(gitRoot: P): F[SortedMap[Name, (LibConfig, P)]] = ` and not bake it in with the opts.

PR #1444, opened 2025-10-31
In TypedExprNormalization normalize integer add/times

johnynek:


core/src/main/scala/org/bykn/bosatsu/TypedExprNormalization.scala
```diff
@@ -309,6 +310,35 @@ object TypedExprNormalization {
         val tpe = Type.normalize(tpe0)
         if (tpe == tpe0) None
         else Some(Local(n, tpe, tag))
+      case App(
```
johnynek: really a clean way to go will be to match on `IntAlgebraic` or something extractor that matches if we have an expression of an integer in some algebraic combination (add, sub, times, div) and then normalize that expression.

for instance `(x + 1) + 2` should be normalized to `x + 3` but this rule won't do it because it doesn't understand associativity (or commutativity).

core/src/main/scala/org/bykn/bosatsu/TypedExprNormalization.scala
```diff
@@ -413,6 +417,9 @@ object TypedExprNormalization {
                       if ((in1 eq in) && (ex1 eq ex)) None
                       else {
                         val step = Let(arg, ex1, in1, rec, tag)
+                        if (step == te) {
+                          sys.error(s"infinite loop on: ${te.reprString}")
```
johnynek: remove this once we are sure all the tests pass.

core/src/main/scala/org/bykn/bosatsu/TypedExprNormalization.scala
```diff
@@ -888,6 +895,308 @@ object TypedExprNormalization {
...
+        val norm = RingOpt.normalize(expr, w)
+
+        if (cats.Hash[RingOpt.Expr[Int]].eqv(norm, expr)) {
+          // the expression is exactly the same
+          None
+        } else {
+          val ia1 = toAlg(norm, table)
+          if (ia1.toTypedExpr.void == root.toTypedExpr.void) {
+            // somehow the expr's weren't equal but the result is:
+            sys.error(
```
johnynek: remove this once we are sure all the tests pass.

core/src/main/scala/org/bykn/bosatsu/TypedExprNormalization.scala
johnynek: This was addressed.

PR #1462, opened 2026-01-22
update to scala 3.8.1

johnynek:
Almost all of this work was done by openai codex using gpt-5.2. It mostly did very well, but took about 2h of churning with sbt.

It did get a little lost a few times:

1. trying to change the generated protobuf code manually vs turning on an option to generate scala3 code to fix the warnings there. I had to do that manually.
2. it wasn't good at clustering the warnings into a few classes and trying to fix them one at a time. I had to tell it "first fix the using/given syntax, now fix...". If I did that, it worked well.
3. it deoptimized some parallelization due to it being more convenient than addressing the real issue (the fact that type aliases resolved a bit differently). By manually changing the type defs into opaque type I could put back those parallelization optimizations.

I punt on this converting the use of Stream to LazyList, which seems self contained and I'll do in a followup.

### DIFF
--- a/core/src/main/scala/org/bykn/bosatsu/Matchless.scala
+++ b/core/src/main/scala/org/bykn/bosatsu/Matchless.scala
@@ -14,7 +14,7 @@ object Matchless {
   // these hold bindings either in the code, or temporary
   // local ones, note CheapExpr never trigger a side effect
   sealed trait CheapExpr[+A] extends Expr[A]
-  sealed abstract class FnExpr[A] extends Expr[A] {
+  sealed abstract class FnExpr[+A] extends Expr[A] {
     def captures: List[Expr[A]]
     // this is set if the function is recursive
     def recursiveName: Option[Bindable]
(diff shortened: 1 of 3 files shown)


.github/workflows/upload_asset.yml
```diff
@@ -45,7 +45,7 @@ jobs:
     strategy:
       matrix:
         java:
-          - '8'
+          - '11'
         scala:
-          - '2.13.10'
+          - '3.7.4'
```
johnynek: need to fix this

.github/workflows/bosatsu_lib_publish.yml
```diff
@@ -18,14 +18,14 @@ jobs:
...
+      - name: "java 11 setup"
         uses: "olafurpg/setup-scala@v10"
         with:
-          java-version: "8"
+          java-version: "11"
       - name: "build assembly"
         run: |
-          sbt 2.13.10 cli/assembly
-          cp cli/target/scala-2.13/bosatsu-cli-assembly-*.jar bosatsu.jar
+          sbt "++3.7.4; cli/assembly"
```
johnynek: fix to 3.8.1

.github/workflows/bosatsu_lib_publish.yml
```diff
@@ -18,14 +18,14 @@ jobs:
         with:
           fetch-depth: 0  # we want full history and tags
       - uses: "coursier/cache-action@v2"
-      - name: "java 8 setup"
+      - name: "java 11 setup"
         uses: "olafurpg/setup-scala@v10"
         with:
-          java-version: "8"
+          java-version: "11"
```
johnynek: need 17

Issue #528, opened 2020-08-20
automatic publishing


johnynek:
set up automatic publishing of projects and artifacts.

In addition to the library jars, set up:

1. build a fat jar someone can download from github.
2. build a nodejs based compiler someone can download ( see #393 ) from github.
3. static native image for osx, linux and hopefully windows.

johnynek:
this seems to be something to investigate for publishing jars + scalajs:

https://twitter.com/mpilquist/status/1299095524394651649?s=20

PR #1473, opened 2026-01-25
Add c runtime cmd and publishing of the c runtime code

johnynek:
This tags the C runtime code and should build the hash of it into the bosatsuc command so that we can install the correct version given the compiler.


johnynek:
relates to #528

PR #1519, opened 2026-01-30
Use Maranget pattern‑matrix algorithm for matchless compilation

### LINKED ISSUE #1118: Improving pattern compilation
See: https://compiler.club/compiling-pattern-matching/

Currently in Matchless we compile pattern matches into nest if/else. But a better algorithm can pretty much always compile into a switch which can be encoded in Python with a dict or directly in C or Java.

johnynek:
close #1118

### DIFF
--- a/core/src/main/scala/dev/bosatsu/Matchless.scala
+++ b/core/src/main/scala/dev/bosatsu/Matchless.scala
@@ -1173,6 +1173,8 @@ object Matchless {
 
     // A row in the pattern matrix: patterns for each occurrence,
     // a right-hand side, and the bindings accumulated so far.
+    // pats may be empty after all columns are specialized away; in that
+    // case the row is a total match for the remaining occurrences.
     case class MatchRow(
         pats: List[Pattern[(PackageName, Constructor), Type]],
         rhs: Expr[B],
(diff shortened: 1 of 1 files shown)


core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -1223,7 +1557,219 @@ object Matchless {
         doesMatch(arg, p1, branches.tail.isEmpty).flatMap(loop)
       }
 
-      val argFn = maybeMemo(tmp)(recur(_, branches))
+      recur(arg, branches)
+    }
+
+    def matchExprOrdered(
```
johnynek: This isn't called

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -1223,7 +1557,219 @@ object Matchless {
...
+              }
+
+            compileCases(sigs)
+        }
+      }
+
+      compileRows(rows0, arg :: Nil)
+    }
+
+    def matchExprMatrix(
```
johnynek: This isn't called

PR #1557, opened 2026-02-03
Make Tau an opaque type for more safety

johnynek:


core/src/main/scala/dev/bosatsu/rankn/Infer.scala
```diff
@@ -1343,32 +1376,38 @@ object Infer {
...
-                writeMeta(m, tpe)
-              }
-            }
-          case (h, _) :: Nil => writeMeta(m, h)
           case Nil           => unit
+          case (h, r1) :: tail =>
+            h match {
+              case Type.TyMeta(m1) =>
+                readMeta(m1).flatMap {
+                  case Some(rho1) => loop((rho1, r1) :: tail)
```
johnynek: should be tau1 not rho1

PR #1579, opened 2026-02-08
add a Float64 type for IEEE754 math

### LINKED ISSUE #1573: add IEEE754 float type
Like python we should probably support IEEE754 double precision values that get packed into 64 bit `void*` in C so that we have zero boxing floating point values.

I think the main thing we need is:

1. add Lit for Float64 type.
2. support parsing Float64 using the same rules as python
3. add `external struct Float64` to predef.bosatsu
4. add the basic math operations to predef.bosatsu: addf, timesf, divf, subf I guess would be names that match the Int versions
5. add Bosatsu/Float/Float64 package that exposes a minimal set of functions from python math which are also implemented by C's standard math library and java.lang.Math. Basically, all the functions implemented by all three of those that act on double precision values should be included.
6. add a total and lawful comparison (which is different from IEEE754 in that it should put NaN before any non-NaN and it should compare NaN values as if they were 64 bit bit integer values). This should probably be in predef so we can sort Float64 values.

related to #1103 but that is much more ambitious and we will set that aside to address this issue.

johnynek:
close #1573 

> ## Motivation
> - Add a first-class `Float64` literal/type to Bosatsu with consistent behavior across JVM evaluation, Python backend, and C/Clang backend.
> - Support math use-cases directly (including special values) while keeping literals, parsing, printing, matching, and backend interop consistent.
> - Preserve exact IEEE754 behavior when needed via bit-conversion APIs.
>
> ## Float Equality And Comparison Rules
> - Equality is math-like, not raw-bit identity:
> - `-0.0 == 0.0`
> - all `NaN` values are equal to each other
> - Total comparison (`cmp_Float64`) is defined as:
> - all `NaN` compare equal to each other
> - `NaN` sorts before any non-`NaN`
> - non-`NaN` values use normal numeric ordering
> - `-0.0` and `0.0` compare as `EQ`
> - These rules are used consistently in literal equality, pattern matching, evaluation, and codegen/runtime helpers.
>
> ## New Literal Syntax And Parsing
> - Added `Float64` literals in parser/typechecker/pattern parser.
> - Accepted literal forms include:
> - decimal/exponent forms with optional sign (`1.0`, `.5`, `5.`, `1e-3`, etc.)
> - underscores in digits (`1_234.5_6`)
> - `.NaN`
> - `∞`, `+∞`, `-∞`
> - Rendering/pretty-printing uses canonical Bosatsu forms:
> - `.NaN`, `∞`, `-∞`, otherwise decimal float text.
>
> ## New/Updated Bosatsu Packages And APIs
> - `Bosatsu/Predef` now includes:
> - `external struct Float64`
> - `addf`, `subf`, `timesf`, `divf`, `cmp_Float64`
> - New support package: `Bosatsu/Float/Float64`
> - Adds higher-level Float64 API (math funcs, conversions, predicates, operators).
> - Exports `inf: Float64 = ∞` and `neg_inf: Float64 = -∞`.
>
> ## `Bosatsu/Float/Float64` Key API
> - Exported value-level API includes:
> - `addf`, `subf`, `timesf`, `divf`, `cmp_Float64`, `eq_Float64`
> - `float64_to_String`, `string_to_Float64`
> - `int_bits_to_Float64`, `float64_bits_to_Int`
> - `is_nan`, `is_infinite`, `is_finite`
> - `inf`, `neg_inf`
> - operators `+`, `-`, `*`, `/`
> - common math: `abs`, `acos`, `asin`, `atan`, `atan2`, `ceil`, `copy_sign`, `cos`, `cosh`, `exp`, `floor`, `hypot`, `log`, `log10`, `pow`, `sin`, `sinh`, `sqrt`, `tan`, `tanh`
>
> ## External Defs In `Bosatsu/Float/Float64`
> ```bosatsu
> external def abs(x: Float64) -> Float64
> external def acos(x: Float64) -> Float64
> external def asin(x: Float64) -> Float64
> external def atan(x: Float64) -> Float64
> external def atan2(y: Float64, x: Float64) -> Float64
> external def ceil(x: Float64) -> Float64
> external def copy_sign(x: Float64, sign: Float64) -> Float64
> external def cos(x: Float64) -> Float64
> external def cosh(x: Float64) -> Float64
> external def exp(x: Float64) -> Float64
> external def floor(x: Float64) -> Float64
> external def float64_bits_to_Int(x: Float64) -> Int
> external def float64_to_String(x: Float64) -> String
> external def hypot(x: Float64, y: Float64) -> Float64
> external def int_bits_to_Float64(int: Int) -> Float64
> external def is_infinite(x: Float64) -> Bool
> external def is_nan(x: Float64) -> Bool
> external def log(x: Float64) -> Float64
> external def log10(x: Float64) -> Float64
> external def pow(x: Float64, y: Float64) -> Float64
> external def sin(x: Float64) -> Float64
> external def string_to_Float64(s: String) -> Option[Float64]
> external def sinh(x: Float64) -> Float64
> external def sqrt(x: Float64) -> Float64
> external def tan(x: Float64) -> Float64
> external def tanh(x: Float64) -> Float64
> ```
>
> ## Pattern Matching Changes
> - Float literals are now valid pattern literals (including `.NaN`, `∞`, `-∞`).
> - Float literal matches use Float64 semantic equality (not raw-bit compare):
> - `.NaN` pattern matches any NaN payload
> - `0.0` pattern also matches `-0.0`
> - Backend codegen paths were updated so generated match checks use Float64-aware equality helpers.
>
> ## Backend/Runtime Integration
> - JVM evaluation uses `java.lang.Double` as the runtime value for `Float64`.
> - C runtime stores Float64 as packed IEEE754 64-bit bits in `BValue` (`void*` word), with 64-bit pointer requirement enforced.
> - Python backend uses native Python `float`.
> - Added cross-backend externals for math, string conversion, and bit conversion.
> - Added proto literal support for Float64 by storing raw bits (`fixed64 float64ValueAsBits`) to preserve exact literal value.
>
> ## String/Bits Interop Semantics
> - `string_to_Float64` strips `_` before numeric parse and handles special forms (`.NaN`, infinities, unicode `∞`).
> - `float64_to_String` returns parseable forms and preserves NaN payload via `NaN:0x<16-hex>` encoding.
> - `int_bits_to_Float64` uses low 64 bits of `Int` (two’s-complement view).
> - `float64_bits_to_Int` returns unsigned 64-bit bit pattern encoded as non-negative Bosatsu `Int`.
>
> ## Test Coverage Added
> - Parser/property tests for Float64 literal syntax and corner cases.
> - Jython parity tests: Bosatsu float literal parsing vs Python float parsing on hard cases.
> - Evaluation tests for Float64 arithmetic, comparison, conversion, `.NaN`/`∞` literals, and pattern matching behavior.
> - Bosatsu property-style laws in `PredefTests`:
> - bit roundtrip laws
> - string roundtrip laws
> - matching laws (`is_nan(x) == x matches .NaN`, infinity matching laws)
> - Clang codegen tests confirm packed-bit literal emission and float-equality helpers in pattern checks.
> - C runtime tests for Float64 equality/comparison, parse/print, and bit conversion roundtrips.

### DIFF
--- a/core/src/main/scala/dev/bosatsu/codegen/python/PythonGen.scala
+++ b/core/src/main/scala/dev/bosatsu/codegen/python/PythonGen.scala
@@ -1075,14 +1075,14 @@ object PythonGen {
       val results: Map[Bindable, (List[ValueLike] => Env[ValueLike], Int)] =
         Map(
           (
-            Identifier.literal("add"),
+            Identifier.Name("add"),
             (
               input => Env.onLast2(input.head, input.tail.head)(_.evalPlus(_)),
               2
             )
           ),
           (
-            Identifier.literal("sub"),
+            Identifier.Name("sub"),
             (
               { input =>
                 Env.onLast2(input.head, input.tail.head)(_.evalMinus(_))
@@ -1091,7 +1091,7 @@ object PythonGen {
             )
           ),
           (
-            Identifier.literal("times"),
+            Identifier.Name("times"),
             (
               { input =>
                 Env.onLast2(input.head, input.tail.head)(_.evalTimes(_))
@@ -1100,7 +1100,7 @@ object PythonGen {
             )
           ),
           (
-            Identifier.literal("div"),
+            Identifier.Name("div"),
             (
               { input =>
                 Env.onLast2(input.head, input.tail.head) { (a, b) =>
@@ -1117,14 +1117,14 @@ object PythonGen {
             )
           ),
           (
-            Identifier.literal("addf"),
+            Identifier.Name("addf"),
             (
               input => Env.onLast2(input.head, input.tail.head)(_.evalPlus(_)),
               2
             )
           ),
           (
-            Identifier.literal("subf"),
+            Identifier.Name("subf"),
             (
               { input =>
                 Env.onLast2(input.head, input.tail.head)(_.evalMinus(_))
@@ -1133,7 +1133,7 @@ object PythonGen {
             )
           ),
           (
-            Identifier.literal("timesf"),
+            Identifier.Name("timesf"),
             (
               { input =>
                 Env.onLast2(input.head, input.tail.head)(_.evalTimes(_))
@@ -1142,14 +1142,14 @@ object PythonGen {
             )
           ),
           (
-            Identifier.literal("divf"),
+            Identifier.Name("divf"),
             (
               divFloatFn,
               2
             )
           ),
           (
-            Identifier.literal("mod_Int"),
+            Identifier.Name("mod_Int"),
             (
               { input =>
                 Env.onLast2(input.head, input.tail.head) { (a, b) =>
@@ -1165,9 +1165,9 @@ object PythonGen {
               2
             )
           ),
-          (Identifier.literal("cmp_Int"), (cmpFn, 2)),
+          (Identifier.Name("cmp_Int"), (cmpFn, 2)),
           (
-            Identifier.literal("eq_Int"),
+            Identifier.Name("eq_Int"),
             (
               { input =>
                 Env.onLast2(input.head, input.tail.head)(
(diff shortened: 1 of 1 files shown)


johnynek:
Need: int_to_Float64 and float64_to_Int. The later has to return Option because NaN and infinities can't convert to Int.

PR #1607, opened 2026-02-09
Fix top-level lambda recovery in ClangGen

### LINKED ISSUE #1605: Top level function being compiled to C lambda
this issue was also seen in #1603 

```
def fib(b: BinNat) -> BinNat:
    def loop(n: Nat, cur: BinNat, next: BinNat) -> BinNat:
        recur n:
            case NatZero: cur
            case NatSucc(n):
                sum = add_BinNat(cur, next)
                loop(n, next, sum)
    one = Odd(Zero)
    loop(to_Nat(b), one, one)
```
was compiled to:

```
static _Atomic BValue ___bsts_s_Bosatsu_l_Num_l_BinNat_l_fib = 0;

static BValue ___bsts_c_Bosatsu_l_Num_l_BinNat_l_fib() {
    BValue __bsts_b_one0 = alloc_enum1(1, alloc_enum0(0));
    BValue __bsts_l_captures392[1] = { __bsts_b_one0 };
    return alloc_closure1(1, __bsts_l_captures392, __bsts_t_closure390);
}

static BValue ___bsts_g_Bosatsu_l_Num_l_BinNat_l_fib() {
    return read_or_build(&___bsts_s_Bosatsu_l_Num_l_BinNat_l_fib,
        ___bsts_c_Bosatsu_l_Num_l_BinNat_l_fib);
}
```
So, here we see the machinery to lazily build top level values, but this top level value is a lambda, and it would be better to just have created it as a C function (which should have been directly callable then vs calling as a lambda via indirection, which hides it from or at least frustrates the C optimizer).

Since we only have lambdas in the matchless API, it seems like this must be a bug with ClangGen not seeing that this should be a top-level function.

johnynek:
## Summary
- recover top-level function-valued expressions that normalization rewrites into `Let`/`If` trees returning lambdas
- wrap those expressions back into a top-level `Matchless.Lambda` before codegen so ClangGen emits a direct C function symbol instead of lazy `read_or_build` closure construction
- add a regression test covering a `def ...(_)` shape that previously emitted a top-level closure value

## Root Cause
`TypedExprNormalization` can rewrite `x -> match ...` into a branch tree that returns lambdas, so the top-level Matchless expression is no longer a literal `Lambda`. `ClangGen.renderTop` only direct-emits C functions for literal `Lambda` nodes; everything else went through static `read_or_build` value construction.

## Testing
- `sbt "coreJVM / testOnly dev.bosatsu.codegen.clang.*"`

Fixes #1605

### DIFF
--- a/core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
+++ b/core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
@@ -916,6 +916,16 @@ object TypedExprNormalization {
               scope,
               typeEnv
             )
+          case Global(pack, n: Bindable, _, _) =>
+            ws.ResolveToLambda.resolveGlobalCallWithBonus(pack, n, args) match {
+              case Some((args1, body, ftag)) =>
+                val lam = AnnotatedLambda(args1, body, ftag)
+                val l = appLambda[A](lam, args, tpe, tag)
+                normalize1(namerec, l, scope, typeEnv)
+              case None =>
+                if ((f1 eq fn) && (tpe == tpe0) && (a1 eq args)) None
+                else Some(App(f1, a1, tpe, tag))
+            }
           case _ =>
             if ((f1 eq fn) && (tpe == tpe0) && (a1 eq args)) None
             else Some(App(f1, a1, tpe, tag))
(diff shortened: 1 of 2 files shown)


core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -72,13 +72,162 @@ object Matchless {
...
+            case ((argName, argTmp), in) =>
+              Let(argName, Local(argTmp), in)
+          }
+
+        val bodyWithRec =
+          recName match {
+            case Some(name) =>
+              val recLam =
+                Lambda(
+                  captures = captureTmpLocals.map[Expr[A]](x => x),
```
johnynek: we should just `widen` here, or it should just compile since List is covariant.

Issue #1603, opened 2026-02-09
TypedNormalization inlining creating big functions


johnynek:
In Array.bosatsu:

```
def set_in_range_ok(_: Unit) -> Bool:
  match set_Array(a5, 1, 8):
    case Some(arr): to_List_Array(arr) matches [0, 8, 2, 3, 4]
    case None: False
```
get compiled to:
```
static BValue ___bsts_c_Bosatsu_l_Collection_l_Array_l_set__in__range__ok() {
    BValue __bsts_a_3 = ___bsts_g_Bosatsu_l_Collection_l_Array_l_index__in__range__Array(___bsts_g_Bosatsu_l_Collection_l_Array_l_a5(),
        bsts_integer_from_int(1));
    BValue __bsts_a_5 = get_variant_value(__bsts_a_3) == 1 ?
        alloc_enum1(1,
            ___bsts_g_Bosatsu_l_Collection_l_Array_l_set__or__self__Array(___bsts_g_Bosatsu_l_Collection_l_Array_l_a5(),
                bsts_integer_from_int(1),
                bsts_integer_from_int(8))) :
        alloc_enum0(0);
    if (get_variant(__bsts_a_5) == 1) {
        BValue __bsts_b_arr0 = get_enum_index(__bsts_a_5, 0);
        BValue __bsts_a_4 = ___bsts_g_Bosatsu_l_Collection_l_Array_l_to__List__Array(__bsts_b_arr0);
        return get_variant(__bsts_a_4) == 1 ?
            (bsts_integer_equals(get_enum_index(__bsts_a_4, 0), bsts_integer_from_int(0)) ?
                (get_variant(get_enum_index(__bsts_a_4, 1)) == 1 ?
                    (bsts_integer_equals(get_enum_index(get_enum_index(__bsts_a_4, 1), 0),
                        bsts_integer_from_int(8)) ?
                        (get_variant(get_enum_index(get_enum_index(__bsts_a_4, 1), 1)) == 1 ?
                            (bsts_integer_equals(get_enum_index(get_enum_index(get_enum_index(__bsts_a_4,
                                            1),
                                        1),
                                    0),
                                bsts_integer_from_int(2)) ?
                                (get_variant(get_enum_index(get_enum_index(get_enum_index(__bsts_a_4,
                                                1),
                                            1),
                                        1)) == 1 ?
[...]

PR #1615, opened 2026-02-10
Reuse constructor allocations in Matchless scopes

johnynek:
## Summary
- add a conservative `Matchless.reuseConstructors` pass after TypedExpr -> Matchless lowering
- share repeated constructor allocations in linear scopes and across `if` branches when args are immutable cheap expressions
- avoid unsafe sharing when mutable Matchless refs are involved, and add regression tests for linear/cross-branch/mutable cases
- update the `cli` clang golden hash for `test_workspace/Ackermann.bosatsu`

## Validation
- `sbt 'coreJVM/testOnly dev.bosatsu.MatchlessTest'`
- `sbt 'coreJVM/test'`
- `sbt 'cli/test'`
- regenerated transpilation output with fresh `cli/assembly` and verified reuse appears in:
  - `pyout/Euler/Three/__init__.py`
  - `c_out/output.c`

part of #1603

### DIFF
--- a/c_runtime/bosatsu_ext_Bosatsu_l_Predef.c
+++ b/c_runtime/bosatsu_ext_Bosatsu_l_Predef.c
@@ -22,6 +22,11 @@ BValue ___bsts_g_Bosatsu_l_Predef_l_char__to__String(BValue a) {
   return bsts_string_from_utf8_bytes_copy(len, bytes);
 }
 
+BValue ___bsts_g_Bosatsu_l_Predef_l_char__to__Int(BValue a) {
+  int codepoint = bsts_char_code_point_from_value(a);
+  return bsts_integer_from_int(codepoint);
+}
+
 // a is a List[Char]
 BValue ___bsts_g_Bosatsu_l_Predef_l_char__List__to__String(BValue a) {
   BValue amut = a;
@@ -181,6 +186,23 @@ BValue ___bsts_g_Bosatsu_l_Predef_l_int__loop(BValue i, BValue a, BValue fn) {
   }
   // all the rest of the values are references
   return _a;
+}
+
+BValue ___bsts_g_Bosatsu_l_Predef_l_int__to__Char(BValue a) {
+  BValue zero = bsts_integer_from_int(0);
+  BValue max_cp = bsts_integer_from_int(0x10FFFF);
+  if ((bsts_integer_cmp(a, zero) < 0) || (bsts_integer_cmp(a, max_cp) > 0)) {
+    return alloc_enum0(0);
+  }
+
+  BValue surrogate_start = bsts_integer_from_int(0xD800);
+  BValue surrogate_end = bsts_integer_from_int(0xDFFF);
+  if ((bsts_integer_cmp(a, surrogate_start) >= 0) && (bsts_integer_cmp(a, surrogate_end) <= 0)) {
+    return alloc_enum0(0);
+  }
+
+  int32_t codepoint = bsts_integer_to_int32(a);
+  return alloc_enum1(1, bsts_char_from_code_point(codepoint));
 }
 
 BValue ___bsts_g_Bosatsu_l_Predef_l_int__to__String(BValue a) {
--- a/c_runtime/bosatsu_ext_Bosatsu_l_Predef.h
+++ b/c_runtime/bosatsu_ext_Bosatsu_l_Predef.h
@@ -8,6 +8,8 @@ BValue ___bsts_g_Bosatsu_l_Predef_l_and__Int(BValue a, BValue b);
 
 BValue ___bsts_g_Bosatsu_l_Predef_l_char__to__String(BValue a);
 
+BValue ___bsts_g_Bosatsu_l_Predef_l_char__to__Int(BValue a);
+
 BValue ___bsts_g_Bosatsu_l_Predef_l_char__List__to__String(BValue a);
 
 BValue ___bsts_g_Bosatsu_l_Predef_l_cmp__Int(BValue a, BValue b);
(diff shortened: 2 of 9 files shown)


core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -371,6 +371,418 @@ object Matchless {
...
+      }
+
+    def allocateBinds(
+        keys: List[Expr[A]],
+        st: CseState
+    ): (List[(LocalAnon, Expr[A])], CseState) =
+      keys.foldLeft((List.empty[(LocalAnon, Expr[A])], st)) {
+        case ((acc, state), key) =>
+          val anon = LocalAnon(state.nextAnon)
+          (((anon, key) :: acc), CseState(state.nextAnon + 1L))
```
johnynek: Add a nextState method to CseState.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -371,6 +371,418 @@ object Matchless {
...
+
+    def linearScopeCse(ex: Expr[A], st: CseState): (Expr[A], CseState) = {
+      val occs = collectLinearInOrder(ex)
+      val counts = occs.groupMapReduce(identity)(_ => 1)(_ + _)
+      val dupes = occs.distinct.filter(k => counts.getOrElse(k, 0) > 1)
+
+      if (dupes.isEmpty) (ex, st)
+      else {
+        val (binds, st1) = allocateBinds(dupes, st)
+        val replace = binds.iterator.map { case (l, e) => (e, l) }.toMap
```
johnynek: map(_.swap)

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -371,6 +371,418 @@ object Matchless {
...
+              Global(_, _, _) | Literal(_) | MakeEnum(_, _, _) | MakeStruct(_) |
+              SuccNat | ZeroNat | GetEnumElement(_, _, _, _) |
+              GetStructElement(_, _, _) =>
+            (ex, st)
+        }
+
+      linearScopeCse(rewritten, st1)
+    }
+
+    val init = CseState(maxAnonExpr(expr) + 1L)
```
johnynek: Should be a method on object CseState: get an init from an Expr.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -371,6 +371,418 @@ object Matchless {
...
+      ex match {
+        case LocalAnonMut(_) => false
+        case _: CheapExpr[?] => true
+        case _               => false
+      }
+
+    def isShareableCtor(ex: Expr[A]): Boolean =
+      ex match {
+        case App(_: ConsExpr, args) =>
+          args.forall(isImmutableCheap)
```
johnynek, line 667: Why do we need it to be cheap to share? Why wouldn't sharing a non cheap be even better? But maybe the code already requires cheapExpr in these positions to force let bindings of expensive things.

We should add a comment to explain.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -371,6 +371,418 @@ object Matchless {
...
+          val anon = LocalAnon(state.nextAnon)
+          (((anon, key) :: acc), CseState(state.nextAnon + 1L))
+      } match {
+        case (rev, st1) => (rev.reverse, st1)
+      }
+
+    def linearScopeCse(ex: Expr[A], st: CseState): (Expr[A], CseState) = {
+      val occs = collectLinearInOrder(ex)
+      val counts = occs.groupMapReduce(identity)(_ => 1)(_ + _)
+      val dupes = occs.distinct.filter(k => counts.getOrElse(k, 0) > 1)
```
johnynek: Put the filter first so we don't have do as much work in distinct.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -371,6 +371,418 @@ object Matchless {
...
+        case ((acc, state), key) =>
+          val anon = LocalAnon(state.nextAnon)
+          (((anon, key) :: acc), CseState(state.nextAnon + 1L))
+      } match {
+        case (rev, st1) => (rev.reverse, st1)
+      }
+
+    def linearScopeCse(ex: Expr[A], st: CseState): (Expr[A], CseState) = {
+      val occs = collectLinearInOrder(ex)
+      val counts = occs.groupMapReduce(identity)(_ => 1)(_ + _)
```
johnynek: We shouldn't use universal equality. Accept an Order[K] and use a SortedMap here. We should never use universal equality, hashing or toString on a generic type like A.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -371,6 +371,418 @@ object Matchless {
...
+                  args.map(replaceBranch(_, replace, blockedBindables, blockedAnonIds))
+                )
+              case PrevNat(of) =>
+                PrevNat(replaceBranch(of, replace, blockedBindables, blockedAnonIds))
+              case Let(arg, value, in) =>
+                arg match {
+                  case Right(name) =>
+                    Let(
+                      arg,
+                      value,
```
johnynek: Don't we need to replace in value? If not comment why not.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -371,6 +371,418 @@ object Matchless {
...
+                        in,
+                        replace,
+                        blockedBindables + name,
+                        blockedAnonIds
+                      )
+                    )
+                  case Left(LocalAnon(id)) =>
+                    Let(
+                      arg,
+                      value,
```
johnynek: Don't we need to replace in value? If not, comment why not?

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -371,6 +371,418 @@ object Matchless {
...
+        (ex1 :: acc, curr1)
+      } match {
+        case (rev, st1) => (rev.reverse, st1)
+      }
+
+    def recurBool(b: BoolExpr[A], st: CseState): (BoolExpr[A], CseState) =
+      b match {
+        case EqualsLit(expr, lit) =>
+          recurExpr(expr, st) match {
+            case (expr1: CheapExpr[A], st1) => (EqualsLit(expr1, lit), st1)
```
johnynek: Rather than this pattern make a recurExprCheap that take and returns CheapExpr and puts the cast in one place.

PR #1625, opened 2026-02-11
Update left-apply desugaring to append continuation to outermost apply

### LINKED ISSUE #1621: Improve left apply syntax, add standard functions and examples for Option/Either
The current desugaring for LeftApply means we need code like:

```
def bind(o: Option[a])(fn: a -> Option[b]) -> Option[b]
```
but that really compiles to a function that is curried. It can sometimes be inlined, but until we solve #1315 we wind up creating code that is harder on the optimizer.

If we had a desugaring that compiled to functionN, like `def bind(o: Option, fn: a -> Option[b])` where the function given is always a function 1 put in the last argument, that would be more efficient.

The challenge is, the reason we don it now is the left apply just applies a function to the result, so by doing `foo.bind()` we are calling the first function and the next function is applied directly.

We need a syntax that isn't ambiguous and that would work when the rhs is a function arg so substitution works.

Maybe this means we can't really do something other than the current design and the actualy thing to do is solve cross package inlining.

johnynek:
## Summary
- change `LeftApply` desugaring so `<-` appends the continuation lambda to the outermost apply node
- peel transparent parentheses during rewrite so `p <- foo(a)` and `p <- (foo(a))` desugar the same
- add reusable SourceConverter test utility in `TestUtils` for converting source snippets to `Program`/`Expr`/type env
- add explicit desugaring tests for all issue #1621 second-comment examples
- update workspace examples from curried continuation helpers to final-argument helpers (`await`, `bind`, `samp`)
- rewrite language guide to make left-apply its own section and document new syntax with await/list/Option examples
- update evaluation test snippet using `let` to the new helper shape

## Desugaring cases covered
- `foo(a, b)`
- `x.await()`
- `p <- foo(a, b)` -> `foo(a, b, p -> body)`
- `p <- foo(a)(b)` -> `foo(a)(b, p -> body)`
- `p <- x.await()` -> `await(x, p -> body)`
- `p <- x.await().map(f)` -> `map(await(x), f, p -> body)`
- paren-stability: `p <- (foo(a))`

## Validation
- `sbt --batch 'coreJVM/test'`
- `sbt --batch 'cli/test'`

close #1621

### DIFF
--- a/docs/src/main/paradox/language_guide.md
+++ b/docs/src/main/paradox/language_guide.md
@@ -406,7 +406,7 @@ result4 = match [False, True, True]:
   case _: "all items are false, or the list is empty"
 ```
 
-Match branches can also have guards:
+Match and `recur` branches can also have guards:
 ```
 result5 = match Some(1):
   case Some(v) if v.eq_Int(0): "zero"
@@ -415,8 +415,9 @@ result5 = match Some(1):
 ```
 Guards run after the pattern matches, in the same scope as the pattern bindings.
 The guard expression must have type `Bool`. If the guard is `False`, matching
-continues with the next branch. For totality checking, only unguarded branches
-count as covering cases, so guarded matches still need an unguarded fallback.
+continues with the next branch. This works the same way in `recur` blocks.
+For totality checking, only unguarded branches count as covering cases, so
+guarded branches still need an unguarded fallback.
 Syntax note: write at least one space between the pattern and `if`.
 
 A common shorthand for checking if something matches is:
--- a/docs/src/main/paradox/pattern_guards_design.md
+++ b/docs/src/main/paradox/pattern_guards_design.md
@@ -1,7 +1,7 @@
 # Pattern Guards Design
 
-Status: proposed  
-Date: 2026-02-10
+Status: implemented  
+Date: 2026-02-11
 
 ## Goal
 Add match branch guards to Bosatsu:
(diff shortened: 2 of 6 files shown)


core/src/main/scala/dev/bosatsu/Declaration.scala
```diff
@@ -849,7 +849,22 @@ object Declaration {
...
       )
-      Apply(fn, NonEmptyList.one(lam), ApplyKind.Parens)(using region)
+      def appendToOutermostApply(argFn: NonBinding): NonBinding =
+        argFn match {
+          // Parentheses around plain expressions are syntactic wrappers.
+          // Peel them so `p <- foo(a)` and `p <- (foo(a))` desugar the same.
+          case Parens(nb: NonBinding) =>
+            appendToOutermostApply(nb)
+          case Apply(fn, args, kind) =>
+            val newArgs = NonEmptyList.fromListUnsafe(args.toList :+ lam)
```
johnynek: NonEmptyList has :+. We don't need to convert to list.

core/src/test/scala/dev/bosatsu/SourceConverterTest.scala
```diff
@@ -39,16 +33,55 @@ class SourceConverterTest extends munit.ScalaCheckSuite {
...
     }
   }
 
+  private def mainExpr(code: String): Expr[Declaration] =
+    convertProgram(code)
+      .getLet(Identifier.Name("main"))
+      .getOrElse(fail("expected a `main` binding"))
+      ._2
+
+  private def eraseTags(expr: Expr[Declaration]): Expr[Unit] =
```
johnynek: This should be a method on Expr[A] => Expr[Unit]

PR #1634, opened 2026-02-12
Fix #1633: avoid loop-name collisions in Matchless loop lowering

### LINKED ISSUE #1633: 0.0.31: internal 'invalid cast to Fn' in lib eval + lib test can exit without showing failing assertion (minimal ReproMin8)
### Summary
`bosatsu` 0.0.31 can fail with an internal runtime exception during `lib eval`, and `lib test` can exit with code 1 without showing any failing assertion output.

This repro is minimized to a small module (`ReproMin8`) and does not depend on the larger JSON parser.

### Environment
- `bosatsu version`: `0.0.31`
- OS: macOS 26.3 (`Darwin 25.3.0`, arm64)

### Minimal repro file
Create `src/MyLib/ReproMin8.bosatsu`:

```bosatsu
package MyLib/ReproMin8

export ParseStep, main

def drop_ws(s: String) -> String:
  recur s:
    case " ${tail}":
      drop_ws(tail)
    case "\n${tail}":
      drop_ws(tail)
    case "\t${tail}":
      drop_ws(tail)
    case "\r${tail}":
      drop_ws(tail)
    case _:
      s

enum ParseStep:
  ParseDone(result: Option[String])

def parse_value(input: String) -> ParseStep:
  rest = drop_ws(input)
  match rest:
    case "null${_}":
      ParseDone(Some("ok"))
    case "\"${_}":
      ParseDone(None)
    case _:
      ParseDone(None)

tests = TestSuite("ReproMin8 tests", [
  Assertion(parse_value("null") matches ParseDone(Some("ok")), "no ws"),
  Assertion(parse_value(" null") matches ParseDone(Some("ok")), "leading ws"),
])

main = parse_value(" null")
```

### Repro 1: `lib eval` internal exception
Run:

```bash
./bosatsu lib eval --main MyLib/ReproMin8::main
```

Actual:

```text
java.lang.RuntimeException: invalid cast to Fn: SumValue(0, ProductValue())
  at dev.bosatsu.Value.asFn(Value.scala:19)
  ...
```

Expected:
- No internal exception.
- Should evaluate to `ParseDone(Some("ok"))`.

### Repro 2: `lib test` exits 1 with truncated output
In this workspace I also had another repro module (`JsonBugRepro`). To isolate `ReproMin8`, I temporarily moved that file out and ran tests:

```bash
mv src/MyLib/JsonBugRepro.bosatsu /tmp/JsonBugRepro.bosatsu
./bosatsu lib test
rc=$?
mv /tmp/JsonBugRepro.bosatsu src/MyLib/JsonBugRepro.bosatsu
echo "$rc"
```

Actual:
- `lib test` exits with code `1`.
- Output stops after listing passing suites from other packages, and does not show the failing assertion or stack trace for `ReproMin8`.

Expected:
- Failing assertion details should be shown (at minimum test name/message and package).
- If there is an internal exception, it should still be surfaced clearly in test output.

### Notes
This seems tied to the combination of:
- recursive string-pattern whitespace trimming in `drop_ws`,
- `rest = drop_ws(input)` followed by string-pattern matching including `case "\"${_}"`,
- wrapping the result in a custom enum (`ParseStep`).

If I remove the `case "\"${_}"` branch, the eval crash disappears in this minimized repro.

johnynek:
## Summary
- add a focused `Issue1633` regression test in core and a reproducer in `test_workspace`
- exercise the reproducer in `testc` and `test_python.sh`
- fix `TypedExpr.Loop` lowering in `Matchless` by avoiding names visible in the current `LambdaState` when choosing the synthetic loop function name
- add `LambdaState.names` as the single source for visible bindable names (`slots.keySet ++ name.iterator`)

## Root Cause
A synthetic loop function name could collide with an existing lambda slot/name mapping. When that happened, recur lowering could resolve to the captured slot instead of the loop self function, leading to an invalid apply path (ClassCastException/crash depending on backend).

## Validation
- `sbt 'cli/test' 'coreJVM/test'`
- `sbt 'coreJVM/testOnly dev.bosatsu.Issue1633Test'`
- `./test_python.sh`
- `./testc`

## Additional Scan
I scanned `Expr.nameIterator` usage and adjacent local-name allocators in `Matchless`, `TypedExprNormalization`, and `TypedExpr` for similar ambient-scope collision gaps. I did not find another high-confidence collision site in current lowering paths.

Closes #1633

### DIFF
--- a/core/src/main/scala/dev/bosatsu/Matchless.scala
+++ b/core/src/main/scala/dev/bosatsu/Matchless.scala
@@ -1839,7 +1839,7 @@ object Matchless {
           val avoid: Set[Bindable] =
             TypedExpr.allVarsSet(body :: args.toList.map(_._2)) ++
               slots.names
-          val loopName = dev.bosatsu.Expr.nameIterator().filterNot(avoid).next()
+          val loopName = freshSyntheticNames("loop", 1, avoid).head
           val loopArgs = args.map { case (n, arg) =>
             (n, arg.getType)
           }


core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -1834,7 +1836,9 @@ object Matchless {
           (loop(fn, slots.unname), as.traverse(loop(_, slots.unname)))
             .mapN(applyArgs(_, _))
         case TypedExpr.Loop(args, body, _) =>
-          val avoid = TypedExpr.allVarsSet(body :: args.toList.map(_._2))
+          val avoid: Set[Bindable] =
+            TypedExpr.allVarsSet(body :: args.toList.map(_._2)) ++
+              slots.names
           val loopName = dev.bosatsu.Expr.nameIterator().filterNot(avoid).next()
```
johnynek: This should use the freshSynthetic name function.

PR #1657, opened 2026-02-12
Fix issue 1654: test discovery + vacuous quantifier handling

### LINKED ISSUE #1654: tool test silently ignores polymorphic tests (forall a. Test)
### Problem
`tool test` discovers tests by taking the last top-level binding whose type is exactly `Test`.
If the final `tests` binding is inferred as `forall a. Test`, it is skipped. If there is an earlier monomorphic `Test` binding, that one runs instead, which can hide failures.

### Reproduction
Create `/tmp/PolyTestDrop3.bosatsu`:

```bosatsu
package Repro/PolyTestDrop3

def for_all(xs: List[a], fn: a -> Bool) -> Bool:
  recur xs:
    case []: True
    case [h, *t] if fn(h):
      for_all(t, fn)
    case [_, *_]:
      False

string_tests = TestSuite("string", [
  Assertion(True, "string runs")
])

tests = TestSuite("all", [
  Assertion(for_all([], _ -> False), "poly assertion"),
  Assertion(False, "this should fail if tests runs")
])
```

Run:

```bash
./bosatsuj tool test --package_root /tmp --input /tmp/PolyTestDrop3.bosatsu --test_file /tmp/PolyTestDrop3.bosatsu
```

Actual output:

```text
Repro/PolyTestDrop3:
  string:
    1 test, 1 passed 0 failed
```

So the failing assertion in `tests` is not executed.

You can also verify the inferred type:

```bash
./bosatsuj tool eval --package_root /tmp --input /tmp/PolyTestDrop3.bosatsu --main Repro/PolyTestDrop3::tests
```

Type shown at end is:

```text
forall a: *. Bosatsu/Predef::Test
```

### Expected
`tests` should be executed (and fail), or the tool should emit a clear diagnostic that `tests` is polymorphic and therefore not considered a test entrypoint.

### Suspected cause
`Package.testValue` currently requires exact type equality with `Type.TestType`:

- `core/src/main/scala/dev/bosatsu/Package.scala`

`forall a. Test` is not equal to `Test`, so it gets ignored.

### Possible fixes
1. Treat vacuous quantified test types (`forall a. Test`) as test values during discovery.
2. Normalize away unused top-level `forall` quantification before test discovery.
3. Emit an error/warning when a top-level `tests` binding is close-but-not-exactly `Test`.

johnynek:
## Summary
- make `Package.testValue` use semantic type comparison (`sameAs`) instead of strict structural equality so `forall a. Test` is still recognized as a test entrypoint
- normalize generic expression types at `TypedExpr.getType` boundaries so vacuous quantifiers are stripped from inferred outward types
- route `TypedExpr.quantVars` through `normalizeQuantVars` to avoid introducing unnecessary quantification wrappers
- add `Issue1654Test` with regression coverage for both failure modes:
  - inferred repro now has monomorphic `tests: Test`, and the failing assertion is executed (not silently dropped)
  - test discovery still selects `tests` when it is manually wrapped as `forall a. Test`

## Testing
- `sbt "testOnly dev.bosatsu.Issue1654Test"`
- `sbt "testOnly dev.bosatsu.TypedExprTest dev.bosatsu.PackageTest"`

Fixes #1654

### DIFF
--- a/core/src/main/scala/dev/bosatsu/Parser.scala
+++ b/core/src/main/scala/dev/bosatsu/Parser.scala
@@ -92,8 +92,14 @@ object Parser {
   sealed trait Error {
     def showContext(errColor: LocationMap.Colorize): Doc =
       this match {
-        case Error.ParseFailure(_, locations, exps) =>
-          Error.showExpectations(locations, exps, errColor)
+        case pf @ Error.ParseFailure(_, locations, exps) =>
+          val base = Error.showExpectations(locations, exps, errColor)
+          ParserHints.hints(pf) match {
+            case Nil    => base
+            case h :: t =>
+              val hints = Doc.intercalate(Doc.hardLine, h :: t)
+              base + Doc.hardLine + hints
+          }
       }
   }
 
--- a/core/src/main/scala/dev/bosatsu/ParserHints.scala
+++ b/core/src/main/scala/dev/bosatsu/ParserHints.scala
@@ -1,0 +1,277 @@
+package dev.bosatsu
+
+import cats.data.NonEmptyList
+import cats.parse.{Parser => P}
+import org.typelevel.paiges.Doc
+
+/** Heuristics to surface actionable hints for common parse mistakes.
+  */
+object ParserHints {
+  type ParseFailure = Parser.Error.ParseFailure
+  type Rule = (String, LocationMap, ParseFailure) => Option[Doc]
+
+  private val rules: List[Rule] =
+    elseIfRule ::
+      elseifSpellingRule ::
+      assignmentInConditionRule ::
+      missingColonAfterHeaderRule ::
+      Nil
+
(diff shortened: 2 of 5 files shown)


core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -27,7 +27,7 @@ sealed abstract class TypedExpr[+T] { self: Product =>
     */
   lazy val getType: Type =
     this match {
-      case g @ Generic(_, _)  => g.quantType
+      case g @ Generic(_, _)  => Type.normalize(g.quantType)
```
johnynek, line 30: I'm nervous about how this normalization might interact with MetaVars that could potentially hold an outer bound. I guess that shouldn't happen but it's not clear to me right this moment.

PR #1671, opened 2026-02-14
Improve type error diagnostics and unify type rendering

### LINKED ISSUE #4: Improve package and type error messages
similar to #3 we want to give actionable feedback and hopefully point to the root cause of the issue.

One idea is in unification we could accumulate an evidence monoid. So when unification fails, we can see all the points in the code that made us thing the thing on the left was X, but the thing on the right was Y. This should help users find the errors.

For package errors we can usually suggestion remedies: add export Foo, etc... which will solve the issue.

### LINKED ISSUE #1659: Type Inference Error Messages Need Better Context and De-duplication
## Summary
From the mutation sweep, type errors are dominated by one generic format:

```text
type error: expected type Bosatsu/Predef::String to be the same as type Bosatsu/Predef::Int
```

This message appears frequently and is often hard to act on quickly because it lacks operation-level context (what expression expected what type), uses internal type paths, and can repeat many times for one root cause.

Observed frequency in the sweep:
- `type error: expected type Bosatsu/Predef::String to be the same as type Bosatsu/Predef::Int` appeared **234** times.
- A variant split across two blocks (`expected type ...` then `to be the same as ...`) appeared **2** times.

### 1) Method receiver mismatch is too generic
**Repro snippet**
```bosatsu
computed = "oops"
test = Assertion(computed.eq_Int(840), "problem 39")
```

**Current message**
```text
type error: expected type Bosatsu/Predef::String to be the same as type Bosatsu/Predef::Int
```

**Suggested improvement**
```text
Type mismatch in call `computed.eq_Int(840)`:
- receiver `computed` has type `String`
- `eq_Int` expects receiver type `Int`
Hint: convert/parse `computed` to `Int`, or use a String comparison function.
```

Notes:
- Current wording is symmetric (“same as”) and does not identify expected vs actual at call site.
- `Bosatsu/Predef::...` is compiler-internal style; user-facing names should be `String`, `Int`.

---

### 2) Match-pattern mismatch message is split and awkward
**Repro snippet**
```bosatsu
def under_twenty(n: Int) -> Int:
  match n:
    case "oops" | 2 | 6: 3
```

**Current message**
```text
type error: expected type Bosatsu/Predef::String
...
to be the same as type Bosatsu/Predef::Int
```
(with two highlighted regions)

**Suggested improvement**
```text
Pattern type mismatch in `match n`:
- scrutinee `n` has type `Int`
- pattern `"oops"` has type `String`
All alternatives in this branch must match `Int`.
```

Notes:
- This is a good case for a specialized message for pattern-vs-scrutinee mismatches.

---

### 3) One bad element in a list causes many repetitive type errors
**Repro snippet**
```bosatsu
grid = [
  ["oops", 2, 22, 97, 38, ...]
]
```

**Current behavior**
- A single mutation produced **19** separate type errors in the same line, mostly repeating:
```text
type error: expected type Bosatsu/Predef::String to be the same as type Bosatsu/Predef::Int
```

**Suggested improvement**
```text
List element type mismatch in this row:
- first element has type `String` (`"oops"`)
- other elements infer `Int`
Expected all elements in a list to have one type.
(18 additional related mismatches suppressed)
```

Notes:
- De-dup by root cause/location would reduce noise and speed up fixes.

---

### 4) Generic unification text hides intent
**Repro pattern**
- Any expression where inference hits `NotUnifiable` currently tends to read as:
```text
expected type X to be the same as type Y
```

**Current problem**
- Message often omits the high-level construct causing the constraint (call argument, receiver, match pattern, list element, etc.).

**Suggested improvement**
```text
Type mismatch in <construct>:
- expected: <type>
- found: <type>
because: <short reason tied to syntax>
```

Examples of `<construct>`:
- function argument
- method receiver
- match pattern
- list element
- branch expression

## Requested improvements
- Add construct-specific type mismatch messages (call receiver, arg, match pattern, list element).
- Prefer user-facing type names (`Int`) over fully-qualified internal names (`Bosatsu/Predef::Int`) in common diagnostics.
- Clearly label `expected` vs `found` rather than symmetric “same as” wording.
- Group/suppress repetitive `NotUnifiable` errors that share the same root cause region.

## Why this matters
- These are common inference failures during normal edits.
- Current output is often technically correct but low-actionability.
- Better context + de-dup would make first-fix success much higher.

johnynek:
## Summary
This implements the full type-error diagnostics pass discussed in #1659, with a shared type rendering rule used by both docs and type errors.

### 1) Shared user-facing type renderer
- Added `TypeRenderer` to centralize type-name rendering rules.
- Rules:
  - drop current package prefix,
  - drop `Bosatsu/Predef::` when not shadowed,
  - preserve package prefixes otherwise.
- Wired Markdown docs and package error rendering through this shared logic.

### 2) Context-rich mismatch diagnostics in inference
- Added `Infer#mapError` and contextual error wrappers:
  - `Infer.Error.ContextualTypeError`
  - `Infer.Error.MismatchSite.AppArg`
  - `Infer.Error.MismatchSite.MatchPattern`
- Wrapped mismatch-producing inference paths so errors carry site metadata:
  - function argument checks (`checkApplyDom`, `applyRhoExpect`)
  - pattern checks (`typeCheckPattern`)

### 3) Improved `PackageError` rendering and dedup
- `TypeErrorIn` now receives local type names for correct abbreviation decisions.
- Contextual errors now render targeted messages for:
  - function call argument mismatches
  - match pattern mismatches
- Pattern rendering in mismatch errors uses a proper `Document`-based path (`Pattern.compiledDocument`) with `Document[Type]` backed by `TypeRenderer`.
- Added combine dedup/suppression for repeated related mismatches.

### 4) Tests
- Updated/extended `ErrorMessageTest` expectations for new rendering behavior.
- Added coverage for:
  - pattern mismatch message shape
  - repeated mismatch suppression in combined errors

## Validation
- `sbt 'coreJVM/testOnly dev.bosatsu.ErrorMessageTest'`
- `sbt 'coreJVM/testOnly dev.bosatsu.rankn.RankNInferTest'`

Closes #1659
Closes #4

### DIFF
--- a/core/src/main/scala/dev/bosatsu/PackageError.scala
+++ b/core/src/main/scala/dev/bosatsu/PackageError.scala
@@ -502,11 +502,13 @@ object PackageError {
                   localTypeNames
                 )
                 val fnLabel = appSite.functionName
-                  .map { full =>
-                    val localPrefix = s"${pack.asString}::"
-                    if (full.startsWith(localPrefix))
-                      full.substring(localPrefix.length)
-                    else full
+                  .map { case (optPack, name) =>
+                    optPack match {
+                      case Some(pn) if pn =!= pack =>
+                        s"${pn.asString}::${name.sourceCodeRepr}"
+                      case _ =>
+                        name.sourceCodeRepr
+                    }
                   }
                   .getOrElse("function")
                 val fnContext =
--- a/core/src/main/scala/dev/bosatsu/rankn/Infer.scala
+++ b/core/src/main/scala/dev/bosatsu/rankn/Infer.scala
@@ -207,7 +207,7 @@ object Infer {
     sealed abstract class MismatchSite
     object MismatchSite {
       case class AppArg(
-          functionName: Option[String],
+          functionName: Option[(Option[PackageName], Bindable)],
           functionType: Type,
           expectedArgType: Type,
           argIndex: Int,
@@ -1636,12 +1636,19 @@ object Infer {
         }
       }
 
-    private def functionNameHint[A](fn: Expr[A]): Option[String] =
+    private def functionNameHint[A](
+        fn: Expr[A]
+    ): Option[(Option[PackageName], Bindable)] =
       fn match {
         case Expr.Local(name, _) =>
-          Some(name.sourceCodeRepr)
+          Some((None, name))
         case Expr.Global(pack, name, _) =>
-          Some(s"${pack.asString}::${name.sourceCodeRepr}")
+          name match {
+            case b: Identifier.Bindable =>
+              Some((Some(pack), b))
+            case _ =>
+              None
+          }
         case _ =>
           None
       }


core/src/test/scala/dev/bosatsu/ErrorMessageTest.scala
```diff
@@ -1208,13 +1208,85 @@ main = 10
...
+
+def same(x: a, y: a) -> Bool:
+  True
+
+main = same(1, "x")
+"""
+
+    evalFail(List(src)) { case te: PackageError.TypeErrorIn =>
+      val msg = te.message(Map.empty, Colorize.None)
+      assert(msg.contains("type mismatch in call to A::same"), msg)
```
johnynek: Should just be "same"

core/src/main/scala/dev/bosatsu/rankn/Infer.scala
```diff
@@ -235,7 +274,8 @@ object Infer {
         declared: Type,
         infRegion: Region,
         decRegion: Region,
-        badTvs: NonEmptyList[Type]
+        badTvs: NonEmptyList[Type],
+        direction: Direction = Direction.Unknown
```
johnynek: Don't use a default here. Be explicit at every site so we can see in the code. We should strive to avoid Unknown.

core/src/main/scala/dev/bosatsu/rankn/Infer.scala
```diff
@@ -199,7 +237,8 @@ object Infer {
         left: Type,
         right: Type,
         leftRegion: Region,
-        rightRegion: Region
+        rightRegion: Region,
+        direction: Direction = Direction.Unknown
```
johnynek: Don't use a default here. Be explicit at every call site. Try to avoid unknown.

core/src/main/scala/dev/bosatsu/rankn/Infer.scala
```diff
@@ -408,6 +454,15 @@ object Infer {
         }
     }
 
+    case class MapError[A](fa: Infer[A], fn: Error => Error) extends Infer[A] {
+      def run(env: Env) =
+        fa.run(env).map {
+          case Right(a) => Right(a)
```
johnynek: `case right @ Right(_) => right`

core/src/main/scala/dev/bosatsu/rankn/Infer.scala
```diff
@@ -749,7 +805,8 @@ object Infer {
         t: Type,
         rho: Type.Rho,
         left: Region,
-        right: Region
+        right: Region,
+        direction: Error.Direction = Error.Direction.ExpectRight
```
johnynek: Be explicit, no default. Try to avoid Unknown.

core/src/main/scala/dev/bosatsu/rankn/Infer.scala
```diff
@@ -783,7 +840,8 @@ object Infer {
         t: Type.Rho,
         rho: Type.Rho,
         left: Region,
-        right: Region
+        right: Region,
+        direction: Error.Direction = Error.Direction.ExpectRight
```
johnynek: No default as above.

core/src/main/scala/dev/bosatsu/rankn/Infer.scala
```diff
@@ -1000,7 +1085,8 @@ object Infer {
         ty1: Type.TyMeta,
         ty2: Type.Tau,
         left: Region,
-        right: Region
+        right: Region,
+        direction: Error.Direction = Error.Direction.Unknown
```
johnynek: No default.

core/src/main/scala/dev/bosatsu/PackageError.scala
```diff
@@ -270,30 +272,220 @@ object PackageError {
...
+      def contextDoc(region: Region): Doc =
+        lm.showRegion(region, 2, errColor).getOrElse(Doc.str(region))
+
+      def renderedTypeKey(tpe: Type): String =
+        TypeRenderer.render(tpe, renderCtx, 80)
+
+      def baseMismatch(
+          tpeErr: Infer.Error.TypeError
+      ): Option[((Type, Region), (Type, Region))] =
+        tpeErr match {
```
johnynek, line 342: We need to use the direction here and not just ignore it.

core/src/main/scala/dev/bosatsu/PackageError.scala
```diff
@@ -270,30 +272,220 @@ object PackageError {
...
+                    Doc.text("found pattern type: ") + tmap(
+                      patSite.foundPatternType
+                    ) + Doc.hardLine +
+                    Doc.text("pattern site:") + Doc.hardLine +
+                    contextDoc(patSite.patternRegion) + scrutineeContext,
+                  Some(patSite.patternRegion)
+                )
+            }
+
+          case Infer.Error.NotUnifiable(t0, t1, r0, r1, _) =>
```
johnynek: Don't ignore direction

core/src/main/scala/dev/bosatsu/PackageError.scala
```diff
@@ -383,84 +572,73 @@ object PackageError {
                   Some(region)
                 )
             }
-          case Infer.Error.SubsumptionCheckFailure(t0, t1, r0, r1, _) =>
+
+          case Infer.Error.SubsumptionCheckFailure(t0, t1, r0, r1, _, _) =>
```
johnynek: Don't ignore direction

core/src/test/scala/dev/bosatsu/ErrorMessageTest.scala
```diff
@@ -1319,13 +1321,87 @@ main = 10
...
+        rankn.Infer.Error.Direction.ExpectRight
+      ),
+      rankn.Infer.Error.NotUnifiable(
+        rankn.Type.IntType,
+        rankn.Type.StrType,
+        region0,
+        region2,
+        rankn.Infer.Error.Direction.ExpectRight
+      )
+    )
```
johnynek, line 1371: we shouldn't be suppressing errors like this. Here we have one location (region0) and two other locations (region1 and region2) that both provide evidence that it should be a String.

when we have `t1 <:< t2` and `t1 <:< t3` where `t2 =:= t3` and we have two different locations, we should say, `we expected <code at region0> which has type `t1` to be type `<t2 which is t3>` due to locations:
<show region1>
<show region2>

it's helpful to have more evidence when possible.

PR #1673, opened 2026-02-14
Inline small strings in C runtime

### LINKED ISSUE #1667: optimize Char and String representations in C
Currently, Char is just a 1 codepoint String in C. But Strings are structs with length and offset.

A Char would always fit into a uintptr_t, so a better implementation would be just to decode the codepoint into that, and not have a GC type for Char.

Secondly, an interesting related optimization would be to use uintptr_t to represent small strings that would fit. We could use the pointer tagging trick as before, so we have 62 or 63 bits to play with. Many utf8 strings could be packed into that. I think it could be pretty efficient too by casting the uintptr_t into a char* and then copy the utf8 data into the first 7 bytes, and then tag the 8th byte correctly. The 8th byte would hold the length of the string in 3 bits (0, 1, 2, ... 7) and then it would have 5 more bits that could be used for tagging. We probably only need 1 or 2 of those to distinguish these uintptr_t from an allocated pointer from libgc. Maybe we only want strings up to 6 bytes of data so we can include a null byte at the end for C compatibility. Probably most of the win is due to strings of length 0-3 I would imagine.

We could still have helper functions to get pointers to the string data (here by casting), and the length, so much of the predef external code and runtime API could have minimal changes.

johnynek:
## Summary
- add a public runtime/codegen-coupled static string layout and initializer in `c_runtime/bosatsu_runtime.h`:
  - `BSTS_String`
  - `BSTS_STATIC_STRING_INIT(len, bytes)`
- switch string runtime APIs to `const char*` reads and add explicit mutable access for construction paths:
  - `bsts_string_utf8_bytes` now returns `const char*`
  - `bsts_string_utf8_bytes_mut` added for `bsts_string_mut` write paths
- update C extern runtime files to use `const` reads and mutable accessor where required
- implement true zero-allocation static literals in `ClangGen`:
  - deduplicate UTF-8 literals per generated C file
  - emit one `static const BSTS_String` per literal
  - replace literal constructor calls with `BSTS_VALUE_FROM_PTR(&static_literal)`
- update `ClangGen` golden hash test for generated C output
- add static-constructor memoization in `bsts_string_from_utf8_bytes_static` (thread-local last `(len, bytes*)`)
- keep/extend runtime tests and benchmarks for new static and small-string behavior

## Safety notes
- static string objects are never passed to `free_on_close` GC stack because pointer gating already excludes non-GC allocations via `GC_base` checks
- mutable string writes now go through `bsts_string_utf8_bytes_mut`, which aborts if asked to mutate non-owned/static bytes

## Performance vs `main`
Benchmarks were run with the same harness (`c_runtime/bench.c`) on both branches at `./bench_exe 1000000` (release), using median of 10 runs.

| benchmark | main ns/op | branch ns/op | delta |
|---|---:|---:|---:|
| `add_big_small_neg` | 9.86 | 9.98 | +1.2% |
| `add_big_small_pos` | 10.01 | 10.12 | +1.0% |
| `and_neg_mixed` | 81.29 | 81.62 | +0.4% |
| `mul_big_small_neg` | 9.75 | 9.49 | -2.7% |
| `mul_big_small_pos` | 9.82 | 9.55 | -2.7% |
| `or_neg_mixed` | 98.55 | 100.53 | +2.0% |
| `shift_neg_left` | 78.56 | 78.86 | +0.4% |
| `shift_neg_right` | 77.50 | 78.17 | +0.9% |
| `xor_neg_mixed` | 98.77 | 99.05 | +0.3% |
| `str_ctor_static_heap` | 7.22 | 7.10 | -1.7% |
| `str_ctor_static_small` | 7.22 | 1.46 | -79.8% |
| `str_equals_small_eq` | 3.44 | 1.06 | -69.2% |
| `str_equals_small_neq` | 3.17 | 1.32 | -58.4% |
| `str_equals_heap_eq` | 6.35 | 1.06 | -83.3% |
| `str_cmp_small` | 3.17 | 1.59 | -49.8% |
| `str_char_at_small` | 1.06 | 1.06 | +0.0% |
| `str_find_small` | 1.85 | 2.12 | +14.6% |

New zero-allocation static-literal benchmark on branch:
- `str_ctor_static_zero`: **1.06 ns/op** (no main equivalent; this is the new generated-code path)

Observed slower case vs main:
- `str_find_small` remains slower; inline-tag decode/bit-scan overhead is still above main’s heap-byte pointer scan in this microbenchmark.

## Validation
- local `testC`/`testCMacOS` equivalent flow (rerun on final commit):
  - `sbt "cli/assembly"`
  - `make -C c_runtime PROFILE=debug`
  - `make -C c_runtime boehm_example PROFILE=release`
  - `make -C c_runtime install PROFILE=release CFLAGS="-O1 -g" CPPFLAGS="-DBSTS_CI=1"`
  - `c_runtime/test_exe`
  - `c_runtime/boehm_example`
  - `make -C c_runtime bench_exe PROFILE=release && c_runtime/bench_exe 1000000`
  - validated `.bosatsuc/<sha>/cc_conf.json` contains `-O1` and `-DBSTS_CI=1`
  - `./bosatsuj tool transpile ... c --outdir c_out --test --exe_out test_exe --cc_flag=-O0 --cc_lib=-lm`
  - `./c_out/test_exe`
  - `./bosatsuj lib test --outdir c_out_lib --cc_flag=-O0 --cc_lib=-lm`
  - `./bosatsuj lib test --cc_flag=-O0 --cc_lib=-lm`
  - `./bosatsuj lib build --outdir c_out_build --main_pack Bosatsu/FibBench --exe_out fib_bench --cc_flag=-O0 --cc_lib=-lm`
  - `./c_out_build/fib_bench 10`

close #1667

### DIFF
--- a/c_runtime/bosatsu_ext_Bosatsu_l_Predef.c
+++ b/c_runtime/bosatsu_ext_Bosatsu_l_Predef.c
@@ -16,10 +16,8 @@ BValue ___bsts_g_Bosatsu_l_Predef_l_and__Int(BValue a, BValue b) {
 }
 
 BValue ___bsts_g_Bosatsu_l_Predef_l_char__to__String(BValue a) {
-  // Char is represented as tiny UTF-8 in typed code, so conversion to String
-  // is just a direct byte copy.
-  BSTS_String_View view = bsts_string_view_ref(&a);
-  return bsts_string_from_utf8_bytes_copy(view.len, view.bytes);
+  // Char and tiny String share the same runtime representation.
+  return a;
 }
 
 BValue ___bsts_g_Bosatsu_l_Predef_l_char__to__Int(BValue a) {


johnynek:
Latest update pushed (commit d4841b4c):

- Char now uses tiny UTF-8 string representation (1..4 bytes) in C runtime.
- Removed by-value `bsts_string_utf8_bytes(BValue)` usage; switched callsites to view/ref APIs.
- Kept tiny-string layout and inline16 medium-string storage; static literal boxing via `BSTS_STATIC_STRING_INIT` remains.
- Added/updated C runtime tests for char roundtrips and length guard aborts.

Local validation run:
- `testC` workflow sequence: PASS
- `sbt "coreJVM/test"`: PASS
- `sbt "cli/test"`: PASS

Benchmarks:
- Comparable integer microbenches vs `origin/main` (same harness, 3 runs median) stayed within ~-2.8% to +3.2%.
- Branch string microbenches (1,000,000 iters) include:
  - `str_ctor_static_heap`: 17.49 ns/op
  - `str_ctor_static_medium`: 16.06 ns/op
  - `str_ctor_static_small`: 12.00 ns/op
  - `str_equals_small_eq`: 1.33 ns/op
  - `str_find_small`: 19.50 ns/op

PR #1683, opened 2026-02-15
Add parse hint for literal \${{ ... }} strings

### LINKED ISSUE #1677: parse error on string interpolation should give hits
When trying to use a literal GitHub expression `${{ ... }}` in string gives misleading parse error
- Repro content:
  - `s = "${{ matrix.java }}"`
- Observed error:
  - Parse error at inner `.` with message `expected char: '('`.
- Why this is difficult:
  - Root cause is interpolation parsing, but the diagnostic does not indicate that `$` needs escaping for a literal GitHub expression.
  - Helpful guidance would be to suggest `"\${{ ... }}"`.

So, we should add a rule that when we are parsing a string and fail, that `$` needs escaping in strings and we escape it as `\$`.

johnynek:
Fixes #1677

## Summary
- add a parser hint rule for parse failures inside an unescaped `${{ ... }}` sequence in a string
- suggest escaping `$` as `\$` and show a concrete example (`"\${{ matrix.java }}"`)
- add a regression test for the issue example in `ParserHintsTest`

## Testing
- sbt "testOnly dev.bosatsu.ParserHintsTest"

### DIFF
--- a/core/src/main/scala/dev/bosatsu/ParserHints.scala
+++ b/core/src/main/scala/dev/bosatsu/ParserHints.scala
@@ -11,7 +11,7 @@ object ParserHints {
   type Rule = (String, LocationMap, ParseFailure) => Option[Doc]
 
   private val rules: List[Rule] =
-    githubExprInStringRule ::
+    interpolationStartInStringRule ::
     elseIfRule ::
       elseifSpellingRule ::
       assignmentInConditionRule ::
@@ -57,7 +57,7 @@ object ParserHints {
     }
   }
 
-  private def githubExprInStringRule(
+  private def interpolationStartInStringRule(
       source: String,
       locations: LocationMap,
       error: ParseFailure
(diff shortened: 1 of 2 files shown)


core/src/main/scala/dev/bosatsu/ParserHints.scala
```diff
@@ -56,6 +57,34 @@ object ParserHints {
...
+      source: String,
+      locations: LocationMap,
+      error: ParseFailure
+  ): Option[Doc] = {
+    val pos = error.position
+    if (pos < 0 || pos > source.length || locations.toLineCol(pos).isEmpty) {
+      None
+    } else {
+      val searchFrom = if (pos >= source.length) source.length - 1 else pos
+      val start = source.lastIndexOf("${{", searchFrom)
```
johnynek: we should relax this a bit. We should look for `${` and if that's there then just mention that we failed to parse, at a given location, maybe we didn't intend to interpolate, and maybe we want a literal `${` which can be written as `\${`, that `${x}` is the syntax for interpolating the string x into this string.

johnynek: Updated in 87d32137: generalized the hint from `${{` to any unescaped `${` before the parse-failure position, with guidance that `${x}` is interpolation syntax and `\${` gives a literal `${`. Also added coverage for a non-GitHub case (`"${foo.bar}"`).

PR #1699, opened 2026-02-17
Add constructor default values for struct and enum branches

### LINKED ISSUE #1676: Default values in structs and enums
consider:

```
struct Record(
  uses: Option[String],
  name: Option[String],
  with: Option[Record],
  run: Option[String],
)
```

We might often create these with a lot of `None` values. What if we had:

```
struct Record(
   name: Option[String],
  uses: Option[String] = None,
  with: Option[Record] = None,
  run: Option[String] = None,
)
```
Then we could create a record with `Record { name = Some("foo") }` and the compiler could substitute in the defaults. We could restrict this to the record syntax so we avoid the chance of confusion.

To type check this we would have to have a special pass that would type check the expression in the default argument. I guess we would have to require it only uses values defined before the first site that uses default variables to ensure the dag property.

Another complexity would come if the default values use imported values from other packages. This would expose that dependency first, and then importing the constructor would implicitly pull in an import on that package and value. Since we have explicit imports required for all dependencies, we would need some syntax since that would be non-obvious.

This is similar to when you have an inferred type from a package you don't otherwise use, but your code doesn't mention the type. We currently error and require you to import it anyway. We need a syntax for "I depend on this package for types or default values" which don't show up visibly in the source.

Another way to handle this would be make all the default values generated bindings in the current file:

```
_default_Record_uses = None
_default_Record_with = None
```
Then the default values are all export from the current package. That's probably cleaner and also ensures we aren't re-evaluating things. We would need to be careful about reporting type errors in the right way. We would need to capture the region of the default expressions so we could point to it for type errors.

Obviously this should work for enum or struct constructors.

johnynek:
## Summary
- add parsing/typechecking/codegen support for constructor parameter default values on structs and enum branches
- desugar omitted record-constructor fields to stable synthetic default helpers with scope checks
- serialize default bindings through proto/interface and include semver compatibility rules for default add/remove
- ensure imported constructor defaults resolve across package boundaries and survive dead-code elimination
- add parser/source-converter/package/evaluation/proto/api-diff tests for default behavior and regressions

## Validation
- sbt "coreJVM/test; cli/test; doc; paradox"
- sbt "coverage; clean; coreJVM/test; cli/test; coverageReport"

Closes #1676

### DIFF
--- a/core/src/test/scala/dev/bosatsu/rankn/TypeTest.scala
+++ b/core/src/test/scala/dev/bosatsu/rankn/TypeTest.scala
@@ -230,6 +230,30 @@ class TypeTest extends munit.ScalaCheckSuite {
     }
   }
 
+  test("Blake3 hash for exists type is stable (simple golden)") {
+    val t = parse("exists a. a -> a")
+    val tRenamed = parse("exists x. x -> x")
+    val expected =
+      "de88397010f29295c2c7a3adb9200b709a08aea28f853dab0359c6132925bec0"
+    val hashT = Hashable.hash(Algo.blake3Algo, t).hash.hex
+    val hashTRenamed = Hashable.hash(Algo.blake3Algo, tRenamed).hash.hex
+
+    assertEquals(hashT, expected)
+    assertEquals(hashTRenamed, expected)
+  }
+
+  test("Blake3 hash for exists type is stable (nested quantifier golden)") {
+    val t = parse("forall z. exists a, b. (a -> z) -> (b -> z)")
+    val tRenamed = parse("forall q. exists y, x. (y -> q) -> (x -> q)")
+    val expected =
+      "40c1750a4f97e5d050f510e7d6aef4a1187c43ae331c62fad9e938f831e52d22"
+    val hashT = Hashable.hash(Algo.blake3Algo, t).hash.hex
+    val hashTRenamed = Hashable.hash(Algo.blake3Algo, tRenamed).hash.hex
+
+    assertEquals(hashT, expected)
+    assertEquals(hashTRenamed, expected)
+  }
+
   test("same as doesn't care about quant var order") {
     assert(parse("forall a, b. a -> b").sameAs(parse("forall b, a. a -> b")))
     assert(parse("exists a, b. a -> b").sameAs(parse("exists b, a. a -> b")))


core/src/main/scala/dev/bosatsu/rankn/ConstructorFn.scala
```diff
@@ -5,9 +5,18 @@ import dev.bosatsu.PackageName
 import dev.bosatsu.Identifier.{Bindable, Constructor}
 import cats.syntax.all._
 
+final case class ConstructorParam(
+    name: Bindable,
+    tpe: Type,
+    defaultBinding: Option[Bindable] = None
```
johnynek: don't use a default value here. Let's just be explicit when we have None vs Some.

core/src/main/scala/dev/bosatsu/rankn/TypeEnv.scala
```diff
@@ -49,6 +49,14 @@ class TypeEnv[+A] private (
       p: PackageName,
       c: Constructor
   ): Option[List[(Bindable, Type)]] =
+    constructors
+      .get((p, c))
+      .map(_._2.args.map(param => (param.name, param.tpe)))
+
+  def getConstructorParamsWithDefaults(
```
johnynek: let's not add this function. Let's just make the original return the new `ConstructorParam`. The callers can ignore the default if needed.

core/src/main/scala/dev/bosatsu/Declaration.scala
```diff
@@ -1257,11 +1266,11 @@ object Declaration {
   ): P[NonBinding] = {
     val ws = Parser.maybeIndentedOrSpace(indent)
     val kv: P[RecordArg] = RecordArg.parser(indent, noAnn)
-    val kvs = kv.nonEmptyListOfWs(ws)
+    val kvs = Parser.nonEmptyListToList(kv.nonEmptyListOfWs(ws))
 
     // here is the record style: Foo {x: 1, ...
-    val recArgs =
-      kvs.bracketed(maybeSpace.with1.soft ~ P.char('{') ~ ws, ws ~ P.char('}'))
+    val recArgs: P0[List[RecordArg]] =
```
johnynek: This should be a `P[List[RecordArg]]` since we are definitely parsing the `{` and `}`

core/src/main/scala/dev/bosatsu/Referant.scala
```diff
@@ -87,9 +87,18 @@ object Referant {
...
         val key = (pn, orig)
-        i.tag.toList.iterator.collect {
-          case Referant.Value(t)            => (key, t)
-          case Referant.Constructor(dt, fn) => (key, dt.fnTypeOf(fn))
+        i.tag.toList.iterator.flatMap {
+          case Referant.Value(t) =>
+            Iterator.single((key, t))
+          case Referant.Constructor(dt, fn) =>
+            Iterator.single((key, dt.fnTypeOf(fn))) ++
+              fn.args.iterator.flatMap { param =>
```
johnynek: in this `flatMap + map` let's use for:
```scala
for {
  param <- fn.args.iterator
  defaultName <-param.defaultBinding
} yield ((pn, defaultName: Identifier), param.tpe)
```

core/src/main/scala/dev/bosatsu/SourceConverter.scala
```diff
@@ -818,6 +825,124 @@ final class SourceConverter(
...
+    val key = {
+      val digestInput = canonicalTypeDigestInput(paramType)
+      List(
+        "DefaultNameV1",
+        thisPackage.asString,
+        typeName.asString,
+        constructorName.asString,
+        paramIndex.toString,
+        digestInput
+      ).mkString("\u0000")
```
johnynek: instead of this, I want you to use the `Hashable[A]` typeclass and define a `given Hashable[Type]`, in `rankn.Type` and likewise for `Hashable[TypeName]` and `Hashable[Constructor]` and then I think we just do:

`Hashable.hash((typeName, construtorName, paramIndex, paramType))` and then get the the `.hex` of the hash.

We may want to add `def by[B](fn: B => A): Hashable[B]` do `trait Hashable` to make it easier to define the hashable for string-like data types (such as PackageName, TypeName, ConstructorName).

core/src/main/scala/dev/bosatsu/SourceConverter.scala
```diff
@@ -818,6 +825,124 @@ final class SourceConverter(
...
+    }
+    val digest =
+      Algo.blake3Algo.hashBytes(key.getBytes(StandardCharsets.UTF_8)).hex
+    Identifier.synthetic("default$" + digest)
+  }
+
+  private def canonicalKindDigestInput(kind: Kind): String =
```
johnynek: remove this and define `Hashable[Kind]` as a `given` in `object Kind`, I think maybe `derives Hashable` will work.

core/src/main/scala/dev/bosatsu/SourceConverter.scala
```diff
@@ -818,6 +825,124 @@ final class SourceConverter(
...
+          variance match {
+            case Variance.Phantom       => "p"
+            case Variance.Covariant     => "c"
+            case Variance.Contravariant => "n"
+            case Variance.Invariant     => "i"
+          }
+        s"K($vchar${canonicalKindDigestInput(input)}${canonicalKindDigestInput(output)})"
+    }
+
+  private def canonicalTypeDigestInput(tpe: Type): String = {
```
johnynek: this should be moved to `Type` and as I said in defining `Hashable[Type]`.

core/src/main/scala/dev/bosatsu/SourceConverter.scala
```diff
@@ -1698,6 +1850,262 @@ final class SourceConverter(
     loop(0, avec.size, s0)._2.map(_.toList)
   }
 
+  private lazy val importedBindables: Set[Bindable] =
+    importedNames.keysIterator.collect { case b: Bindable =>
+      b
+    }.toSet
+
+  private def toTypeRef(tpe: Type): TypeRef = {
```
johnynek, line 1784: this seems duplicative with TypeRefConverter.fromTypeA. Can we use that instead of duplicating here? It seems like `F[_] = cats.Id[_]` and it is just completely converting.

That said, I'm skeptical we should have metas and skolems to TypeRef. Can you add a comment as to why?

core/src/main/scala/dev/bosatsu/SourceConverter.scala
```diff
@@ -1698,6 +1850,262 @@ final class SourceConverter(
...
+          vars.map { case (tv, kind) =>
+            val k = if (kind == Kind.Type) None else Some(kind)
+            (TypeRef.TypeVar(tv.name), k)
+          },
+          toTypeRef(in)
+        )
+    }
+  }
+
+  private def closeTypeRef(
```
johnynek: seems like this could also be in `TypeRefConverter`.

core/src/main/scala/dev/bosatsu/SourceConverter.scala
```diff
@@ -1698,6 +1850,262 @@ final class SourceConverter(
...
+                              defaultBindingForParam(
+                                typeName,
+                                ctor.constructor,
+                                idx,
+                                param.tpe
+                              )
+                            )
+
+                          if (scope0(helperName)) {
+                            throw new IllegalStateException(
```
johnynek, line 1904: add a comment as to why this is impossible from any user code. We should only throw an exception on a compiler bug, not on any possible user input.

core/src/main/scala/dev/bosatsu/SourceConverter.scala
```diff
@@ -1698,6 +1850,262 @@ final class SourceConverter(
...
+                            defaultScopeCheck(
+                              ctor.constructor,
+                              arg.name,
+                              defaultExpr,
+                              region,
+                              importedBindables ++ scope0,
+                              constructorParamNames
+                            )
+
+                          val baseTypeRef = arg.tpe.getOrElse(toTypeRef(param.tpe))
```
johnynek: this is an interesting point: should we allow default params when the type isn't explicit? I would say no. That should be an error at this stage: default params only allowed when parameter type is explicit.

core/src/main/scala/dev/bosatsu/SourceConverter.scala
```diff
@@ -1941,12 +2350,17 @@ final class SourceConverter(
             pte.addExternalValue(thisPackage, name, tpe)
           }
         }
-
-        implicit val parallel = SourceConverter.parallelIor
```
johnynek, line 1945: why don't we need the parallel anymore? Losing the parallel will destroy the ability to accumulate more than one error. This is annoying for users because they just get one error at a time. We should strive to give as many errors as possible by using Ior and the parallel instance.

core/src/main/scala/dev/bosatsu/SourceConverter.scala
```diff
@@ -1941,12 +2350,17 @@ final class SourceConverter(
...
-        (checkExternalDefShadowing(stmts), toLets(stmts), pte1).mapN {
-          (_, binds, pte1) =>
-            Program((importedTypeEnv, pte1), binds, exts.map(_._1).toList, ss)
-        }
+        for {
+          stmts <- valueStatementsWithDefaultHelpers(ss)
+          _ <- checkExternalDefShadowing(stmts)
```
johnynek: we should be able to put the next three in a `(_, binds, parsedTypeEnv) <- (x, y, z).parTupled`

core/src/main/scala/dev/bosatsu/SourceConverter.scala
```diff
@@ -2260,6 +2674,26 @@ object SourceConverter {
...
+      constructorName: Constructor,
+      fieldName: Bindable,
+      parameterName: Bindable,
+      region: Region
+  ) extends Error {
+    def message =
+      s"default value for field ${fieldName.sourceCodeRepr} in ${constructorName.sourceCodeRepr} cannot reference constructor parameter ${parameterName.sourceCodeRepr}"
+  }
+
+  final case class ConstructorDefaultOutOfScope(
```
johnynek, line 2569: double check that we have a test that exercises showing this error.

core/src/main/scala/dev/bosatsu/SourceConverter.scala
```diff
@@ -2260,6 +2674,26 @@ object SourceConverter {
     }
   }
 
+  final case class ConstructorDefaultReferencesParam(
```
johnynek, line 2559: doulbe check that we have a syntax test that exercises this error.

core/src/main/scala/dev/bosatsu/Statement.scala
```diff
@@ -146,10 +163,33 @@ object Statement {
   //////
   // TypeDefinitionStatement types:
   //////
+  final case class ConstructorArg(
+      name: Bindable,
+      tpe: Option[TypeRef],
+      default: Option[Declaration.NonBinding],
+      defaultRegion: Option[Region],
```
johnynek: we should have `default: Option[Declaration.NonBinding]` by itself, because `NonBinding` already has a Region, I think.

core/src/test/scala/dev/bosatsu/EvaluationTest.scala
```diff
@@ -3067,4 +3067,25 @@ main = Main(
...
+        """package Provider
+export Rec()
+
+struct Marker
+struct Rec(a: Marker = Marker)
+""",
+        """package Consumer
+from Provider import Rec
+
+main = match Rec {}:
```
johnynek: we should update the language guide on default arguments and mention that `Rec {}` is the syntax to use only when there is 1 or more default argument. It is not the same as `Rec` by itself, which is only the syntax to use when there are zero arguments.

core/src/test/scala/dev/bosatsu/EvaluationTest.scala
```diff
@@ -3067,4 +3067,25 @@ main = Main(
     )
   }
 
+  test("default-backed constructor calls work across package boundaries") {
```
johnynek: we need more tests than this. We need cases where some fields are given, we need to make sure order of fields doesn't matter.

core/src/test/scala/dev/bosatsu/EvaluationTest.scala
```diff
@@ -3067,4 +3067,25 @@ main = Main(
...
+    evalTest(
+      List(
+        """package Provider
+export Rec()
+
+struct Marker
+struct Rec(a: Marker = Marker)
```
johnynek: we should also have a test where we have a type that isn't the same as Unit so we can verify the default value correctly makes it through. For instance a default `Int` and we check that we get the right value.

PR #1700, opened 2026-02-17
Add phased well-typed source generator and pipeline tests

### LINKED ISSUE #1698: Implement a well-typed Gen for generating typed programs
See the design in #1697

The idea is that we could generate random programs that should have a known set of types. Then when we emit them as text, we can test the compiler against them. The motivation is to find corner cases in the compiler's type-inference or normalization/optimization rules. We could generate random programs and assert that the values generated by the optimized and unoptimized programs are the same. Also, we could assert that the time it takes to evaluate the optimized ones tends to be lower (or the minimum of N tries is lower).

The goal is to increase our confidence in the soundness of the type-checker.

Of course. We should also consider the opposite: definitely ill-typed programs for which the type-checker should definitely reject.

johnynek:
close #1698

With a generator that generates well typed code, we can exercise the typechecker and normalization in scalacheck.

### DIFF
--- a/core/src/test/scala/dev/bosatsu/DeclarationTest.scala
+++ b/core/src/test/scala/dev/bosatsu/DeclarationTest.scala
@@ -2,6 +2,7 @@ package dev.bosatsu
 
 import cats.data.NonEmptyList
 import org.scalacheck.Gen
+import org.scalacheck.Prop
 import org.scalacheck.Prop.forAll
 
 import Identifier.Bindable
@@ -330,7 +331,7 @@ x""")
       assertEquals(a0.hashCode, a1.hashCode)
     }
 
-    org.scalacheck.Prop.all(varProp, litProp, applyProp)
+    Prop.all(varProp, litProp, applyProp)
   }
 
   test("isCheap is constant under Annotation or Parens") {
--- a/core/src/test/scala/dev/bosatsu/KindParseTest.scala
+++ b/core/src/test/scala/dev/bosatsu/KindParseTest.scala
@@ -4,6 +4,7 @@ import cats.kernel.Eq
 import cats.syntax.eq._
 import dev.bosatsu.hashing.{Algo, Hashable}
 import org.scalacheck.Gen
+import org.scalacheck.Prop
 import org.scalacheck.Prop.forAll
 
 import rankn.NTypeGen.{genKind, genKindArg => genArg, shrinkKind}
@@ -98,7 +99,7 @@ class KindParseTest extends ParserTestBase {
     val propArgs = forAll(Gen.listOf(genArg)) { args =>
       assertEquals(Kind(args*).toArgs, args)
     }
-    org.scalacheck.Prop.all(propKind, propArgs)
+    Prop.all(propKind, propArgs)
   }
 
   test("example orders") {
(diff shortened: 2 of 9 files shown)


core/src/test/scala/dev/bosatsu/WellTypedGen.scala
```diff
@@ -0,0 +1,625 @@
...
+
+    val (cons, idx) = loop(ctx.nextTypeId)
+    val nextCtx = ctx.copy(
+      usedConstructors = ctx.usedConstructors + cons,
+      nextTypeId = idx + 1
+    )
+    (cons, nextCtx)
+  }
+
+  private def typeRefOf(tpe: Type): TypeRef =
```
johnynek, line 246: TypeConverter has code for this. Use that.

core/src/test/scala/dev/bosatsu/WellTypedGen.scala
```diff
@@ -0,0 +1,625 @@
...
+    else {
+      val fnType =
+        for {
+          arity <- Gen.choose(1, 2)
+          args <- Gen.listOfN(arity, Gen.lzy(genType(ctx, depth - 1)))
+          res <- Gen.lzy(genType(ctx, depth - 1))
+          nel = NonEmptyList.fromListUnsafe(args)
+        } yield Type.Fun(nel, res)
+
+      Gen.frequency(
```
johnynek: this isn't generating any applied types or quantified types. We should be able to do that.

core/src/test/scala/dev/bosatsu/WellTypedTests.scala
```diff
@@ -0,0 +1,96 @@
...
+  private def assertSourceConverterSuccess(
+      packageName: PackageName,
+      statements: List[Statement],
+      source: String
+  ): Unit =
+    SourceConverter.toProgram(packageName, Nil, statements) match {
+      case Ior.Left(errs) =>
+        val msg = errs.toList.mkString("\n")
+        fail(s"source conversion failed for:\n$source\n\n$msg")
+      case Ior.Right(_) | Ior.Both(_, _) =>
```
johnynek: both is also a failure. We should always get Right out of toProgram. Both happens when there are some errors but we can continue.

core/src/test/scala/dev/bosatsu/WellTypedTests.scala
```diff
@@ -0,0 +1,96 @@
...
+    val parsed = Package.fromStatements(packageName, statements)
+    given cats.Show[String] = cats.Show.fromToString
+    val checked = PackageMap.typeCheckParsed(
+      NonEmptyList.one((("<generated>", LocationMap(source)), parsed)),
+      Nil,
+      "<predef>"
+    )
+
+    checked match {
+      case Ior.Right(_) | Ior.Both(_, _) =>
```
johnynek: both is a failure.

core/src/test/scala/dev/bosatsu/WellTypedGen.scala
```diff
@@ -0,0 +1,1191 @@
...
+      intro.foreach(g => choices += ((introW, Gen.lzy(g))))
+      byVar.foreach(g => choices += ((varW, g)))
+      byApply.foreach(g => choices += ((applyW, g)))
+      byLambdaApply.foreach(g => choices += ((lambdaApplyW, g)))
+      byMatch.foreach(g => choices += ((matchW, g)))
+      weightedOneOf(choices.result())
+    }
+  }
+
+  private def requireGen[A](label: String, gen: Option[Gen[A]]): Gen[A] =
```
johnynek: `label` should be a lazy parameter so we don't have to do string serialization if we don't fail.

PR #1703, opened 2026-02-17
Fix quantified type generation in NTypeGen

johnynek:
## Summary
- add `NTypeGen.genQuantifiers(genB)` to produce non-empty quantifier binders with kinds
- lift quantified generators into top-level `genForAll`/`genExists` and make them select binders from free vars of generated bodies
- update `genTypeRho` to include an explicit existential branch so rho generation can actually produce `Exists`
- keep `genQuantArgs` for list-based callers while deriving it from the new non-empty quantifier generator
- add `TypeTest` coverage for binder-usage guarantees and for direct `Type.forAll`/`Type.exists` behavior using both free-var-selected and independent quantifier generators

## Testing
- `sbt "coreJVM/testOnly dev.bosatsu.rankn.TypeTest"`

### DIFF
--- a/core/src/test/scala/dev/bosatsu/rankn/NTypeGen.scala
+++ b/core/src/test/scala/dev/bosatsu/rankn/NTypeGen.scala
@@ -254,7 +254,7 @@ object NTypeGen {
       val genApply = Gen
         .zip(recurse, genDepth(d - 1, genC))
         .map { case (a, b) =>
-          Type.apply1(a, b).asInstanceOf[Type.Rho]
+          Type.apply1Rho(a, b)
         }
 
       Gen.frequency((3, root), (2, genApply), (1, genExists(d, genC)))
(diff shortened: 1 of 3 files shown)


core/src/test/scala/dev/bosatsu/rankn/NTypeGen.scala
```diff
@@ -204,34 +253,18 @@ object NTypeGen {
       val recurse = Gen.lzy(genTypeRho(d - 1, genC))
       val genApply = Gen
         .zip(recurse, genDepth(d - 1, genC))
-        .map {
-          case (a: (Type.Leaf | Type.TyApply), b) => Type.TyApply(a, b)
-          case (exists, _)                        => exists
+        .map { case (a, b) =>
+          Type.apply1(a, b).asInstanceOf[Type.Rho]
```
johnynek: let's add `Type.apply1Rho(a: Rho, b: Type): Rho` so we don't need the cast here.

johnynek: Implemented in bb9112c3: added `Type.apply1Rho(fn: Rho, arg: Type): Rho`, switched `NTypeGen.genTypeRho` to use it (removing the cast there), and added a property test (`Type.apply1Rho agrees with Type.apply1 on rho inputs`). Re-ran `sbt "coreJVM/testOnly dev.bosatsu.rankn.TypeTest"` and it passes.

PR #1708, opened 2026-02-18
Extend Type.instantiate endpoint solving and add top/bottom laws

johnynek:
## Summary
- add `Type.liftUniversals` and `Type.liftExistentials` (with collision-safe unshadowing)
- extend `Type.instantiate` to solve endpoint cases where one side is a solver var and the opposite side still contains solvable vars in scope
- keep solver-state clarity by using explicit `fromFixed`/`toFixed` state and scoped free-var checks
- refactor `Infer.subsInstantiate` to a single `Type.instantiate` call using quantifier lifting
- validate both left and right substitutions with known bound-kind context so mixed substitutions remain kind-safe
- improve subsumption diagnostics with meta/skolem kind hints when available
- add explicit endpoint tests:
  - `forall a. a -> a` to `exists x. x`
  - `forall a. a` to `exists x. x -> x`
- add property checks for top/bottom instantiation behavior on well-kinded generated types
- add `liftUniversals`/`liftExistentials` roundtrip laws and collision test

## Testing
- `sbt -batch "coreJVM/testOnly dev.bosatsu.rankn.TypeTest dev.bosatsu.rankn.RankNInferTest"`
- `sbt -batch "cli/test" "coreJVM/test"`

## Related design doc
- Quantifier evidence / Annotation-node design doc PR: https://github.com/johnynek/bosatsu/pull/1696

### DIFF
--- a/core/src/main/scala/dev/bosatsu/rankn/Infer.scala
+++ b/core/src/main/scala/dev/bosatsu/rankn/Infer.scala
@@ -1956,10 +1956,14 @@ object Infer {
                       if (
                         instantiation.frees.nonEmpty || instantiation.toFrees.nonEmpty
                       ) {
-                        // TODO maybe we could handle this, but not yet
-                        // seems like if the free vars are set to the same
-                        // variable, then we can just lift it into the
-                        // quantification
+                        // Conservative fallback: if instantiate leaves residual
+                        // frees on either side, we defer to the general apply
+                        // inference path (`pureNone` here).
+                        //
+                        // We could extend this to lift compatible residual frees
+                        // into quantification and continue, but we currently do
+                        // not have a concrete program where doing so changes the
+                        // accepted/rejected result.
                         /*
                             println(s"remaining frees in ${
                               Type.fullyResolvedDocument.document(fnTe.getType).render(80)


core/src/main/scala/dev/bosatsu/rankn/Type.scala
```diff
@@ -719,13 +748,13 @@ object Type {
...
                     case _ => None
                   }
-                case Fixed(set) =>
+                case BoundState.Fixed(set) =>
                   if (set.sameAs(to)) Some(state)
                   else None
-                case Free(rightName) =>
+                case BoundState.Free(rightName) =>
                   to match {
                     case TyVar(toB: Var.Bound) if rightName === toB =>
```
johnynek, line 921: We need a comment on this case. Was the rhs a closed type? Why does the bound name being the same matter in this case? That should only matter if it is an env bound, no?

core/src/main/scala/dev/bosatsu/rankn/Type.scala
johnynek: This looks like an inline unshadowing. We need to double check we don't already have this function, then move it up and add direct tests for type unshadowing.

johnynek: We could be nested in the rhs of a TyApply. How does adding these to the top level frees work?

PR #1716, opened 2026-02-19
Fix nested-forall binder capture in TypedExpr quantification

johnynek:
## Summary
- Fix binder capture in nested-`forall` inference paths by using the correct bound-name set in `TypedExpr.allBound`.
- Keep `TypedExpr.allBound` as the single API and update its oracle test to match the corrected semantics.
- Add regression coverage in `RankNInferTest` for `ignore_err`/`recover` cases (including concrete `Env`) that previously collapsed `env` and quantified error vars.

## Why
The compiler could infer a corrupted type shape in cases like:
- expected: `forall env, err, res. (Prog[env, err, res], res) -> forall e. Prog[env, e, res]`
- bad: `forall a, b, c. (Prog[a, b, c], c) -> forall a. Prog[a, a, c]`

This was a compiler-side binder-capture bug, not Bosatsu source behavior.

## Validation
- `sbt -batch 'coreJVM/testOnly dev.bosatsu.TypedExprTest dev.bosatsu.rankn.RankNInferTest'`
- `sbt -batch 'coreJVM/test'`
- `sbt -batch 'cli/test'`

## Related PRs
- Design doc: #1696
- Prior `Type.instantiate` improvements: #1525, #1708

### DIFF
--- a/cli/src/test/scala/dev/bosatsu/GithubWorkflowJsonParityTest.scala
+++ b/cli/src/test/scala/dev/bosatsu/GithubWorkflowJsonParityTest.scala
@@ -5,7 +5,6 @@ import dev.bosatsu.tool.Output
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory
 import java.nio.file.{Path, Paths}
-import scala.jdk.CollectionConverters._
 import scala.concurrent.duration._
 
 // allow us to unsafeRunSync
@@ -50,25 +49,6 @@ class GithubWorkflowJsonParityTest extends munit.FunSuite {
     ()
   }
 
-  // For this exercise we treat absent object fields and explicit null fields
-  // as equivalent, which aligns with Option-backed JSON generation.
-  private def equivalentJson(lhs: com.fasterxml.jackson.databind.JsonNode, rhs: com.fasterxml.jackson.databind.JsonNode): Boolean =
-    if ((lhs eq null) || (rhs eq null)) (lhs eq rhs)
-    else if (lhs.isObject && rhs.isObject) {
-      val keys = lhs.fieldNames.asScala.toSet union rhs.fieldNames.asScala.toSet
-      keys.forall { k =>
-        val lv = lhs.get(k)
-        val rv = rhs.get(k)
-        if ((lv eq null) && (rv eq null)) true
-        else if (lv eq null) rv.isNull
-        else if (rv eq null) lv.isNull
-        else equivalentJson(lv, rv)
-      }
-    } else if (lhs.isArray && rhs.isArray) {
-      lhs.size == rhs.size &&
-        (0 until lhs.size).forall(idx => equivalentJson(lhs.get(idx), rhs.get(idx)))
-    } else lhs.equals(rhs)
-
   workflowCases.foreach { case (workflowFile, mainValue) =>
     test(s"lib json write matches .github/workflows/$workflowFile") {
       val out = run(
@@ -91,8 +71,9 @@ class GithubWorkflowJsonParityTest extends munit.FunSuite {
       val workflowPath = Paths.get(".github", "workflows", workflowFile)
       val yamlJsonNode = yamlMapper.readTree(workflowPath.toFile)
 
-      assert(
-        equivalentJson(generatedJsonNode, yamlJsonNode),
+      assertEquals(
+        generatedJsonNode,
+        yamlJsonNode,
         clues(
           s"workflow: $workflowFile, main: $mainValue",
           s"generated: ${generatedJsonNode.toPrettyString}",
--- a/core/src/main/scala/dev/bosatsu/ValueToJson.scala
+++ b/core/src/main/scala/dev/bosatsu/ValueToJson.scala
@@ -18,6 +18,80 @@ case class ValueToJson(getDefinedType: Type.Const => Option[DefinedType[Any]]) {
       PackageName.parts("Bosatsu", "Collection", "Array"),
       TypeName("Array")
     )
+  private val bosatsuJsonTypeConst: Type.Const.Defined =
+    Type.Const.Defined(PackageName.parts("Bosatsu", "Json"), TypeName("Json"))
+  private val optionalTypeConst: Type.Const.Defined =
+    Type.Const.Defined(
+      PackageName.parts("Bosatsu", "Json"),
+      TypeName("Optional")
+    )
+  private val nullableTypeConst: Type.Const.Defined =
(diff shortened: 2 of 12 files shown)


core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -555,6 +594,50 @@ object TypedExpr {
...
+    ): SortedMap[Type.Var.Bound, (Kind, Type)] =
+      SortedMap.from(
+        solved.iterator
+          .map { case (b, (k, t)) => (b, (k, fn(t))) }
+      )
+
+    private def traverseSolved[F[_]: Applicative](fn: Type => F[Type])(
+        solved: SortedMap[Type.Var.Bound, (Kind, Type)]
+    ): F[SortedMap[Type.Var.Bound, (Kind, Type)]] =
+      solved.toList
```
johnynek: We can compose the SortedMap and Tuple2 Applicative.

core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -568,7 +651,11 @@ object TypedExpr {
   }
   // Annotation really means "widen", the term has a type that is a subtype of coerce, so we are widening
   // to the given type. This happens on Locals/Globals also in their tpe
-  case class Annotation[T](term: TypedExpr[T], coerce: Type)
+  case class Annotation[T](
+      term: TypedExpr[T],
+      coerce: Type,
+      quantifierEvidence: Option[QuantifierEvidence]
```
johnynek, line 659: We need to mention the invariant that targetAtSolve is the sameAs the tpe of the Annotation.

We should also check that in some tests.

core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -1041,6 +1140,21 @@ object TypedExpr {
         loop(pat :: Nil)
       }
 
+      def visitEvidenceTypes(
+          qev: Option[QuantifierEvidence],
+          shadowed: Set[Type.Var.Bound]
+      ): Unit =
+        qev.foreach { ev =>
```
johnynek: Match Some/None

core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -1041,6 +1140,21 @@ object TypedExpr {
...
+          visitType(ev.sourceAtSolve, shadowed)
+          visitType(ev.targetAtSolve, shadowed)
+          ev.forallSolved.valuesIterator.foreach { case (_, t) =>
```
johnynek: Use the foreach inline

core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -1105,13 +1220,9 @@ object TypedExpr {
...
-      val acc = scala.collection.mutable.HashSet.empty[Type.Var.Bound]
-      foreachTraversedType {
-        case Type.TyVar(b: Type.Var.Bound) =>
-          acc.add(b): Unit
-        case _ => ()
-      }
-      SortedSet.from(acc)
+      val tpes = allTypes.toList
+      val free = Type.freeBoundTyVars(tpes).toSet
+      SortedSet.from(Type.tyVarBinders(tpes) ++ free)
```
johnynek, line 1236: We should add a comment here explaining why we have and need two functions from Type returning Var.Bound.

core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -1179,6 +1290,18 @@ object TypedExpr {
         loop(pat :: Nil)
       }
 
+      def addEvidenceTypes(qev: Option[QuantifierEvidence]): Unit =
+        qev.foreach { ev =>
+          addType(ev.sourceAtSolve)
+          addType(ev.targetAtSolve)
+          ev.forallSolved.valuesIterator.foreach { case (_, t) =>
```
johnynek: Use foreach inline here and below.

core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -1179,6 +1290,18 @@ object TypedExpr {
         loop(pat :: Nil)
       }
 
+      def addEvidenceTypes(qev: Option[QuantifierEvidence]): Unit =
+        qev.foreach { ev =>
```
johnynek: Match the option to avoid the lambda

core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -1517,9 +1645,9 @@ object TypedExpr {
           deepQuantify(env + te.getType, in).map { in1 =>
             quantVars(quant.forallList, quant.existList, in1)
           }
-        case Annotation(term, coerce) =>
+        case Annotation(term, coerce, qev) =>
           deepQuantify(env + coerce, term).map { t1 =>
-            ann(t1, coerce)
+            ann(t1, coerce, qev)
```
johnynek, line 1678: We need to add a comment why deepQuantify doesn't care about the quantification evidence.

core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -2065,18 +2238,18 @@ object TypedExpr {
           def apply[A](expr: TypedExpr[A]): Rho[A] =
             expr match {
               case _ if expr.getType.sameAs(tpe) => expr
-              case Annotation(t, _)              => self(t)
+              case Annotation(t, _, _)           => self(t)
```
johnynek: I'm not sure this is right. Why can't we have two sequential instantiations? Don't they compose? Don't we need to carry them through?

Maybe there is no evidence in this branch so we only are widening but we should comment why it is safe to lose instantiation information.

core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -2542,9 +2719,16 @@ object TypedExpr {
...
+  private def ann[A](
+      te: TypedExpr[A],
+      tpe: Type,
+      quantifierEvidence: Option[QuantifierEvidence]
+  ): TypedExpr[A] =
     if (te.getType.sameAs(tpe)) te
-    else Annotation(te, tpe)
+    else Annotation(te, tpe, quantifierEvidence)
+
+  private def ann[A](te: TypedExpr[A], tpe: Type): TypedExpr[A] =
```
johnynek: We should remove this. It is dangerous for refactor.

PR #1729, opened 2026-02-19
Implement inhabitedness checker and match-state checker

johnynek:
## Summary
- add Inhabitedness with State (Inhabited | Uninhabited | Unknown) and ValidatedNec errors
- implement check(t, env) for type inhabitedness with conservative Unknown fallback
- implement checkMatch(scrutinee, pattern, env) returning State
- add deterministic and ScalaCheck coverage (genInhabited/genUninhabited) in InhabitednessTest
- incorporate review cleanup: Error as enum, *> sequencing, State companion combinators, lazy annotation checks, and expanded ListPat fallback logic

## Validation
- sbt "coreJVM/testOnly dev.bosatsu.InhabitednessTest"
- sbt coverage "coreJVM/testOnly dev.bosatsu.InhabitednessTest" "coreJVM/coverageReport"

## Reference
- Follows the design doc introduced in #1725

### DIFF
--- a/core/src/main/scala/dev/bosatsu/Inhabitedness.scala
+++ b/core/src/main/scala/dev/bosatsu/Inhabitedness.scala
@@ -1,5 +1,6 @@
 package dev.bosatsu
 
+import cats.{Eq}
 import cats.data.{NonEmptyList, ValidatedNec}
 import cats.syntax.all._
 import dev.bosatsu.rankn.Type
@@ -14,6 +15,8 @@ object Inhabitedness {
   }
 
   object State {
+    given Eq[State] = Eq.fromUniversalEquals
+
     def orStates(states: List[State]): State =
       if (states.contains(State.Inhabited)) State.Inhabited
       else if (states.forall(_ == State.Uninhabited)) State.Uninhabited
@@ -171,11 +174,11 @@ object Inhabitedness {
           case Some(assignments) =>
             val states = assignments.map(assignment => bodyEval(outerState ++ assignment))
             if (isForAll) {
-              if (states.exists(_ == Uninhabited)) Uninhabited
+              if (states.contains_(Uninhabited)) Uninhabited
               else if (states.forall(_ == Inhabited)) Inhabited
               else Unknown
             } else {
-              if (states.exists(_ == Inhabited)) Inhabited
+              if (states.contains_(Inhabited)) Inhabited
               else if (states.forall(_ == Uninhabited)) Uninhabited
               else Unknown
             }
@@ -210,7 +213,7 @@ object Inhabitedness {
           val argStates = args.map(checkType(_, varState, seen))
           if ((resultState == Uninhabited) && argStates.forall(_ == Inhabited))
             Uninhabited
-          else if ((resultState == Inhabited) || argStates.exists(_ == Uninhabited))
+          else if ((resultState == Inhabited) || argStates.contains_(Uninhabited))
             Inhabited
           else Unknown
         case Type.TyVar(b: Type.Var.Bound) =>


core/src/main/scala/dev/bosatsu/Inhabitedness.scala
```diff
@@ -0,0 +1,407 @@
...
+            }
+
+          consCheck *> params.traverse_(validatePattern)
+        case Pattern.Union(head, rest) =>
+          (head :: rest.toList).traverse_(validatePattern)
+      }
+
+    private def varAssignments(
+        vars: List[Type.Var.Bound]
+    ): Option[List[Map[Type.Var.Bound, State]]] =
```
johnynek: this could return `Option[NonEmptyList[Map[Type.Var.Bound, State]]]`, note that list always has length 1 or more.

core/src/main/scala/dev/bosatsu/Inhabitedness.scala
```diff
@@ -0,0 +1,407 @@
...
+      if (vars.lengthCompare(MaxQuantifierVars) > 0) None
+      else {
+        @annotation.tailrec
+        def loop(
+            rem: List[Type.Var.Bound],
+            acc: List[Map[Type.Var.Bound, State]]
```
johnynek: should be a NonEmptyList.

core/src/main/scala/dev/bosatsu/Inhabitedness.scala
```diff
@@ -0,0 +1,407 @@
...
+        ): List[Map[Type.Var.Bound, State]] =
+          rem match {
+            case Nil => acc
+            case h :: t =>
+              val next = acc.flatMap { m =>
+                List(
```
johnynek: could be NonEmptyList

core/src/main/scala/dev/bosatsu/Inhabitedness.scala
```diff
@@ -0,0 +1,407 @@
...
+        case Type.Exists(vars, in) =>
+          evaluateQuantifier(
+            vars.toList,
+            vs => checkType(in, vs, seen),
+            isForAll = false,
+            varState
+          )
+        case Type.Fun(args, res) =>
+          val resultState = checkType(res, varState, seen)
+          val argStates = args.toList.map(checkType(_, varState, seen))
```
johnynek: we don't need `toList` here. We can map and call forall and contains on NonEmptyList.

core/src/main/scala/dev/bosatsu/Inhabitedness.scala
```diff
@@ -0,0 +1,407 @@
...
+          val (on, args) = Type.unapplyAll(norm)
+          on match {
+            case tc @ Type.TyConst(_) =>
+              val cons = tc.tpe.toDefined
+              if (alwaysInhabited(cons)) Inhabited
+              else {
+                env.getType(tc) match {
+                  case None => Unknown
+                  case Some(dt) =>
+                    val root = Type.applyAll(tc, args).normalize
```
johnynek: comment how this is different from `norm`? Haven't we just removed the args and then reapplied them?

core/src/main/scala/dev/bosatsu/Inhabitedness.scala
```diff
@@ -0,0 +1,407 @@
...
+            ) Inhabited
+            else Unknown
+          }
+        case Pattern.Union(head, rest) =>
+          State.orStates(
+            (head :: rest.toList).map(checkPattern(scrutinee, _, varState, seen))
+          )
+        case Pattern.Literal(lit)      =>
+          val litType = Type.getTypeOf(lit).normalize
+          if (scrutinee.normalize === litType) Inhabited
```
johnynek: instead of `.normalize === x.normalize` just use `scrutinee.sameAs(Type.getTypeOf(lit))` which is the same thing.

core/src/main/scala/dev/bosatsu/Inhabitedness.scala
```diff
@@ -0,0 +1,407 @@
...
+          else Uninhabited
+        case _: Pattern.StrPat =>
+          if (scrutinee.normalize === Type.StrType.normalize) Inhabited
```
johnynek: use `sameAs`

core/src/main/scala/dev/bosatsu/Inhabitedness.scala
```diff
@@ -0,0 +1,407 @@
...
+                case _ =>
+                  listElementType(scrutinee) match {
+                    case None => Unknown
+                    case Some(itemType) =>
+                      val itemStates =
+                        suffix.toList.collect { case Pattern.ListPart.Item(itemPat) =>
+                          checkPattern(itemType, itemPat, varState, seen)
+                        }
+                      if (itemStates.contains(Uninhabited)) Uninhabited
+                      else Unknown
```
johnynek, line 366: why isn't this `else if (itemStates.forall(_ == Inhabited)) Inhabited else Unknown`? Please fix or add a comment.

Issue #1727, opened 2026-02-19
improving compilation of multi-parameter group functions


johnynek:
Consider:
```
def eq_List(fn: (a, a) -> Bool)(a: List[a], b: List[a]) -> Bool:
  recur (a, b):
    case ([ah, *at], [bh, *bt]) if fn(ah, bh): eq_List(fn)(at, bt)
    case ([], []): True
    case _: False
```

Currently, this can't compile to a tail recursive loop because it compiles to a function that returns a lambda that captures fn.

A better way to compile would be to transform this to:

```
def eq_List(fn: (a, a) -> Bool)(a: List[a], b: List[a]) -> Bool:
  def loop(fn: (a, a) -> Bool, a: List[a], b: List[a]):
   recur (a, b):
    case ([ah, *at], [bh, *bt]) if fn(ah, bh): loop(fn, a, b)
    case ([], []): True
    case _: False
  
  loop(fn, a, b)
```
So we have an inner loop, and then we basically return `(a, b) -> loop(fn, a, b)` which is a lambda, but it contains a loop.

PR #1739, opened 2026-02-20
docs: design for issue #1727 multi-parameter-group recursion rewrite

johnynek:
## Summary
- add a new design doc for issue #1727 describing the multi-parameter-group recursive rewrite
- specify the recommended integration point in TypedExprNormalization
- add the design doc to the design-docs index

## Issue
Refs #1727


docs/src/main/paradox/design-docs/issue_1727_multi_parameter_group_recursion_design.md
```diff
@@ -0,0 +1,127 @@
...
+
+```bosatsu
+def eq_List(fn: (a, a) -> Bool)(a: List[a], b: List[a]) -> Bool:
+  recur (a, b):
+    case ([ah, *at], [bh, *bt]) if fn(ah, bh): eq_List(fn)(at, bt)
+    case ([], []): True
+    case _: False
+```
+
+This should compile into a loop-shaped core, but currently the recursive call is seen as nested application (`eq_List(fn)` then `(at, bt)`) and does not match existing direct self-call rewrite logic.
```
johnynek: we should be explicit what we want. I think we mean we want this to be as efficient as:

```
def eq_List(fn: (a, a) -> Bool)(a: List[a], b: List[a]) -> Bool:
  def loop(fn: (a, a) -> Bool, a: List[a], b: List[a]) -> Bool:
    recur (a, b):
      case ([ah, *at], [bh, *bt]) if fn(ah, bh): loop(fn, at, bt)
      case ([], []): True
      case _: False

  loop(fn, a, b)
```

docs/src/main/paradox/design-docs/issue_1727_multi_parameter_group_recursion_design.md
```diff
@@ -0,0 +1,127 @@
...
+## Current behavior summary
+
+1. `SourceConverter.toLambdaExpr` lowers parameter groups into nested lambdas.
+2. The recursive call in the innermost body is represented as a chained `App(App(Local(self), prefixArgs), recurArgs)`.
+3. `SelfCallKind` and `rewriteTailCalls` only recognize direct `App(Local(self), args)` calls.
+4. `rewriteTailRecToLoop` therefore misses this shape and no `Loop/Recur` is introduced for issue-1727 patterns.
+
+## Where this rewrite should live
+
+This rewrite should be implemented in `TypedExprNormalization`, specifically in the recursive `Let` normalization path in `normalizeLetOpt`.
```
johnynek: some of this got moved in https://github.com/johnynek/bosatsu/pull/1737

We should update this. Does this rewrite live in a phase before loop lowering? Or does it live in Normalization since maybe new shapes like this could be unlocked by normalizaton?

docs/src/main/paradox/design-docs/issue_1727_multi_parameter_group_recursion_design.md
```diff
@@ -0,0 +1,127 @@
...
+  f_loop(g1, x1, x2)
+```
+
+This preserves partial-application semantics for `f(g1)` while exposing a single-group recursive function that existing loop lowering can optimize.
+
+### Detection constraints (conservative)
+
+Apply rewrite only when all of the following hold:
+
+1. Recursive RHS is nested lambdas with at least two groups.
```
johnynek: Be explicit about how we will detect this in the TypedExpr ast. It seems tricky, and I'm worried about it being brittle.

Like if we have:

```
foo = fn -> (
  x = 2
  (x, y) -> (
    ... foo(fn)(x, y)
  )
)
```
Where there is a let (or other bindings) in the body after definition. Does that work?

docs/src/main/paradox/design-docs/issue_1727_multi_parameter_group_recursion_design.md
```diff
@@ -0,0 +1,127 @@
...
+2. Every self-reference in the innermost group body is a full self-application in tail position.
```
johnynek: How do we detect this? I guess we know the type of `foo` in the example above, and can see that it is `a -> (x, y) -> Bool` and then see that all recursive appearances are twice applied so we get all the way to the end result. Is that right. We should be explicit about how this detection might look.

docs/src/main/paradox/design-docs/issue_1727_multi_parameter_group_recursion_design.md
```diff
@@ -0,0 +1,127 @@
...
+3. Prefix groups in recursive calls are invariant passthroughs of the current outer lambda binders.
```
johnynek: expand on this. I don't wee why we have to have this requirement. I don't see why allowing changing arguments in the prefix would be a problem.

docs/src/main/paradox/design-docs/issue_1727_multi_parameter_group_recursion_design.md
```diff
@@ -0,0 +1,127 @@
...
+4. No escaping self-reference appears (for example, storing `self` in data or returning it).
```
johnynek: explain why. We will still generate the outer function, why can't we store a reference to that within the loop?

docs/src/main/paradox/design-docs/issue_1727_multi_parameter_group_recursion_design.md
```diff
@@ -0,0 +1,127 @@
...
+This keeps the new rewrite as a shape-exposing pre-pass, while existing passes remain responsible for loop materialization and capture tuning.
+
+## Sketch of helper API
+
+```scala
+private def rewriteMultiGroupTailRec[A](
+  fnName: Bindable,
+  expr: TypedExpr[A],
+  tag: A
+): Option[TypedExpr[A]]
```
johnynek, line 130: so, the invariant here is that if we return `Some` the type is exactly the same as `expr` right?

docs/src/main/paradox/design-docs/issue_1727_multi_parameter_group_recursion_design.md
```diff
@@ -0,0 +1,127 @@
...
+1. Issue-1727 regression: normalized `eq_List` contains loop form for inner recursion.
+2. Partial-application regression: `eq_List(eq_Int)` behavior unchanged.
+3. Negative case: transformed prefix argument (not passthrough) does not rewrite.
+4. Negative case: escaping self-reference does not rewrite.
+5. Idempotence: second normalization pass makes no further structural change.
+
+## Non-goals
+
+1. No syntax change for defs or recur.
+2. No general uncurrying of all multi-group functions.
```
johnynek: why is handling all multi-group at once a problem? It would be nicer of we could elegantly handle any number of parameter groups.

PR #1747, opened 2026-02-20
Implement grouped loop/recur lowering for issue #1727

### LINKED ISSUE #1727: improving compilation of multi-parameter group functions

johnynek:
## Summary
- implement grouped recursive rewrite in loop/recur lowering so eta-expanded grouped self-calls are lowered correctly
- add extensive regression coverage in TypedExprTest for grouped/lambda/wrapper/shadowing/negative paths
- simplify unchanged branches and warning fixes in grouped rewrite traversal logic

## Issue linkage
Closes #1727

## Validation
- changed-code coverage for core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala: 201/201 covered lines (100%)
- sbt "coreJVM/testOnly *TypedExprTest"
- sbt "coverage; coreJVM/testOnly *TypedExprTest; coreJVM/coverageReport"
- sbt "coreJVM/test; cli/test; doc; paradox"
- ./test_cli.sh
- sbt "cli/assembly"
- ./bosatsuj tool transpile ... python ... and python3 -m unittest discover pyout ...
- sbt "coreJS/test; jsapiJS/compile" and sbt "jsuiJS/test"
- sbt "cli/nativeImage" and native binary ./bosatsu tool test ...
- sbt "cliJSJS/fullOptJS" and ./bosatsu_node tool test ...
- make -C c_runtime install VERSION=$(git rev-parse HEAD) and node/c runtime/lib test flows from ci.yml
- sbt "++3.8.1; cli/assembly" and scripts/publish_bosatsu_libs.sh dry run env

## Notes
- testWithCoverageReport was executed multiple times, but in this local environment the long sbt runner intermittently fails to return a final status line after all visible suites finish; coverage requirement for this PR was enforced with explicit changed-line measurement and targeted coverage runs.

### DIFF
--- a/c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Std.c
+++ b/c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Std.c
@@ -288,22 +288,22 @@ BValue bsts_print_errln_effect(BValue a)
 
 static BValue bsts_io_std_print_prog(BValue a)
 {
-  return alloc_enum2(7, a, alloc_boxed_pure_fn1(bsts_print_effect));
+  return alloc_enum2(5, a, alloc_boxed_pure_fn1(bsts_print_effect));
 }
 
 static BValue bsts_io_std_println_prog(BValue a)
 {
-  return alloc_enum2(7, a, alloc_boxed_pure_fn1(bsts_println_effect));
+  return alloc_enum2(5, a, alloc_boxed_pure_fn1(bsts_println_effect));
 }
 
 static BValue bsts_io_std_print_err_prog(BValue a)
 {
-  return alloc_enum2(7, a, alloc_boxed_pure_fn1(bsts_print_err_effect));
+  return alloc_enum2(5, a, alloc_boxed_pure_fn1(bsts_print_err_effect));
 }
 
 static BValue bsts_io_std_print_errln_prog(BValue a)
 {
-  return alloc_enum2(7, a, alloc_boxed_pure_fn1(bsts_print_errln_effect));
+  return alloc_enum2(5, a, alloc_boxed_pure_fn1(bsts_print_errln_effect));
 }
 
 /* Helper: check that the first len bytes form a sequence of whole UTF-8 code points. */
@@ -456,7 +456,7 @@ BValue bsts_read_stdin_utf8_bytes_effect(BValue size)
 
 static BValue bsts_io_std_read_stdin_utf8_bytes_prog(BValue a)
 {
-  return alloc_enum2(7, a, alloc_boxed_pure_fn1(bsts_read_stdin_utf8_bytes_effect));
+  return alloc_enum2(5, a, alloc_boxed_pure_fn1(bsts_read_stdin_utf8_bytes_effect));
 }
 
 BValue ___bsts_g_Bosatsu_l_IO_l_Std_l_print__impl(BValue a)
--- a/c_runtime/bosatsu_ext_Bosatsu_l_Num_l_Float64.c
+++ b/c_runtime/bosatsu_ext_Bosatsu_l_Num_l_Float64.c
@@ -37,54 +37,6 @@ static BValue bsts_uint64_to_int(uint64_t bits) {
   return bsts_integer_from_words_copy(1, words[1] == 0 ? 1 : 2, words);
 }
 
-static uint64_t bsts_int_to_low_uint64(BValue int_value) {
-  uint32_t mask_words[2] = { UINT32_C(0xffffffff), UINT32_C(0xffffffff) };
-  uint32_t div_words[2] = { UINT32_C(0x00000000), UINT32_C(0x00000001) };
-  BValue mask = bsts_integer_from_words_copy(1, 2, mask_words);
-  BValue two32 = bsts_integer_from_words_copy(1, 2, div_words);
-  BValue low64 = bsts_integer_and(int_value, mask);
-  BValue divmod = bsts_integer_div_mod(low64, two32);
-  BValue hi = get_struct_index(divmod, 0);
-  BValue lo = get_struct_index(divmod, 1);
-  uint32_t hi32 = (uint32_t)bsts_integer_to_int32(hi);
-  uint32_t lo32 = (uint32_t)bsts_integer_to_int32(lo);
-  return (((uint64_t)hi32) << 32) | ((uint64_t)lo32);
-}
-
-static BValue bsts_integral_double_to_int(double d) {
-  if (d == 0.0) {
-    return bsts_integer_from_int(0);
-  }
-
-  int text_len = snprintf(NULL, 0, "%.0f", d);
-  if (text_len <= 0) {
-    perror("snprintf failure in bsts_integral_double_to_int");
-    abort();
-  }
-
-  char* text = (char*)malloc((size_t)text_len + 1);
-  if (text == NULL) {
-    perror("failed to malloc in bsts_integral_double_to_int");
-    abort();
-  }
-
-  int written = snprintf(text, (size_t)text_len + 1, "%.0f", d);
-  if (written != text_len) {
-    perror("snprintf mismatch in bsts_integral_double_to_int");
-    abort();
-  }
-
-  BValue as_string = bsts_string_from_utf8_bytes_copy((size_t)text_len, text);
-  free(text);
-
-  BValue parsed = bsts_string_to_integer(as_string);
-  if (get_variant(parsed) == 0) {
-    perror("failed to parse integral double text in bsts_integral_double_to_int");
-    abort();
-  }
-  return get_enum_index(parsed, 0);
-}
-
 static double bsts_round_ties_even(double d) {
   double int_part = 0.0;
   double frac = modf(d, &int_part);
@@ -281,7 +233,7 @@ BValue ___bsts_g_Bosatsu_l_Num_l_Float64_l_string__to__Float64(BValue a) {
 }
 
 BValue ___bsts_g_Bosatsu_l_Num_l_Float64_l_int__bits__to__Float64(BValue a) {
-  return bsts_float64_from_bits(bsts_int_to_low_uint64(a));
+  return bsts_float64_from_bits(bsts_integer_to_low_uint64(a));
 }
 
 BValue ___bsts_g_Bosatsu_l_Num_l_Float64_l_float64__bits__to__Int(BValue a) {
@@ -294,32 +246,11 @@ BValue ___bsts_g_Bosatsu_l_Num_l_Float64_l_float64__to__Int(BValue a) {
     return bsts_none();
   }
   double rounded = bsts_round_ties_even(d);
-  return bsts_some(bsts_integral_double_to_int(rounded));
+  return bsts_some(bsts_integral_float64_to_integer(rounded));
 }
 
 BValue ___bsts_g_Bosatsu_l_Num_l_Float64_l_int__to__Float64(BValue a) {
-  // Fast path for small immediate Int values.
-  if ((a & (BValue)0x3) == (BValue)0x1) {
-    int32_t small = (int32_t)(a >> 2);
-    return bsts_boxf((double)small);
-  }
-
-  BValue as_string = bsts_integer_to_string(a);
-  BSTS_String_View view = bsts_string_view_ref(&as_string);
-  size_t len = view.len;
-  const char* bytes = view.bytes;
-
-  char* null_term = (char*)malloc(len + 1);
-  if (null_term == NULL) {
-    perror("failed to malloc in int_to_Float64");
-    abort();
-  }
-
-  memcpy(null_term, bytes, len);
-  null_term[len] = '\0';
-  double parsed = strtod(null_term, NULL);
-  free(null_term);
-  return bsts_boxf(parsed);
+  return bsts_boxf(bsts_integer_to_double(a));
 }
 
 BValue ___bsts_g_Bosatsu_l_Num_l_Float64_l_log(BValue a) {
--- a/c_runtime/bosatsu_ext_Bosatsu_l_Prog.c
+++ b/c_runtime/bosatsu_ext_Bosatsu_l_Prog.c
@@ -9,9 +9,7 @@
 # FlatMap(p, f) => (2, p, f)
 # Recover(p, f) => (3, p, f)
 # ApplyFix(a, f) => (4, a, f)
-# ReadEnv() => (5, )
-# RemapEnv(p, f) => (6, p, f)
-# Effect(arg: BValue, f: BValue => BValue) => (7, arg, f)
+# Effect(arg: BValue, f: BValue => BValue) => (5, arg, f)
 */
 
 BValue ___bsts_g_Bosatsu_l_Prog_l_pure(BValue a)
@@ -34,19 +32,9 @@ BValue ___bsts_g_Bosatsu_l_Prog_l_flat__map(BValue p, BValue f)
   return alloc_enum2(2, p, f);
 }
 
-BValue ___bsts_g_Bosatsu_l_Prog_l_read__env()
-{
-  return alloc_enum0(5);
-}
-
 BValue ___bsts_g_Bosatsu_l_Prog_l_recover(BValue p, BValue f)
 {
   return alloc_enum2(3, p, f);
-}
-
-BValue ___bsts_g_Bosatsu_l_Prog_l_remap__env(BValue p, BValue f)
-{
-  return alloc_enum2(6, p, f);
 }
 
 BValue bsts_prog_step_fix_closure(BValue *slots, BValue a)
@@ -67,7 +55,7 @@ BValue bsts_prog_step_fix(BValue arg, BValue fixfn)
   return call_fn1(ap1, arg);
 }
 
-int bsts_Bosatsu_Prog_run_main(BValue prog, int argc, char **argv)
+int bsts_Bosatsu_Prog_run_main(BValue main_fn, int argc, char **argv)
 {
   BValue arg_list = alloc_enum0(0);
   for (int i = argc; i > 0; i--)
@@ -83,11 +71,9 @@ int bsts_Bosatsu_Prog_run_main(BValue prog, int argc, char **argv)
   done = (0,)
   def fmstep(fn, stack): return (1, fn, stack)
   def recstep(fn, stack): return (2, fn, stack)
-  def restore(env, stack): return (3, env, stack)
   */
   BValue stack = alloc_enum0(0);
-  BValue env = arg_list;
-  BValue arg = prog;
+  BValue arg = call_fn1(main_fn, arg_list);
   while (1)
   {
     switch (get_variant(arg))
@@ -117,11 +103,6 @@ int bsts_Bosatsu_Prog_run_main(BValue prog, int argc, char **argv)
           // recstep, but this isn't an error
           stack = get_enum_index(stack, 1);
           break;
-        case 3:
-          // restore
-          env = get_enum_index(stack, 0);
-          stack = get_enum_index(stack, 1);
-          break;
         }
       }
       break;
@@ -141,7 +122,7 @@ int bsts_Bosatsu_Prog_run_main(BValue prog, int argc, char **argv)
           return 1;
         case 1:
           // fmstep, but we have an error
-          stack = get_enum_index(stack, 2);
+          stack = get_enum_index(stack, 1);
           break;
         case 2:
         {
@@ -152,11 +133,6 @@ int bsts_Bosatsu_Prog_run_main(BValue prog, int argc, char **argv)
           search_stack = 0;
           break;
         }
-        case 3:
-          // restore
-          env = get_enum_index(stack, 0);
-          stack = get_enum_index(stack, 1);
-          break;
         }
       }
       break;
@@ -182,26 +158,16 @@ int bsts_Bosatsu_Prog_run_main(BValue prog, int argc, char **argv)
       arg = bsts_prog_step_fix(get_enum_index(arg, 0), get_enum_index(arg, 1));
       break;
     case 5:
-      // ReadEnv() => (5, )
-      arg = ___bsts_g_Bosatsu_l_Prog_l_pure(env);
-      break;
-    case 6:
     {
-      // RemapEnv(p, f) => (6, p, f)
-      stack = alloc_enum2(3, env, stack);
-      BValue remap_fn = get_enum_index(arg, 1);
-      env = call_fn1(remap_fn, env);
-      arg = get_enum_index(arg, 0);
-      break;
-    }
-    case 7:
-    {
-      // Effect(arg: BValue, f: BValue => BValue) => (7, arg, f)
+      // Effect(arg: BValue, f: BValue => BValue) => (5, arg, f)
       BValue earg = get_enum_index(arg, 0);
       BValue efn = get_enum_index(arg, 1);
       arg = call_fn1(efn, earg);
       break;
     }
+    default:
+      fprintf(stderr, "bosatsu Prog execution fault: invalid Prog tag: %u\n", get_variant(arg));
+      return 1;
     }
   }
   return 0;
--- a/c_runtime/bosatsu_ext_Bosatsu_l_Prog.h
+++ b/c_runtime/bosatsu_ext_Bosatsu_l_Prog.h
@@ -8,10 +8,6 @@ BValue ___bsts_g_Bosatsu_l_Prog_l_pure(BValue a);
 
 BValue ___bsts_g_Bosatsu_l_Prog_l_raise__error(BValue a);
 
-BValue ___bsts_g_Bosatsu_l_Prog_l_read__env();
-
 BValue ___bsts_g_Bosatsu_l_Prog_l_recover(BValue a, BValue b);
 
-BValue ___bsts_g_Bosatsu_l_Prog_l_remap__env(BValue a, BValue b);
-
-int bsts_Bosatsu_Prog_run_main(BValue prog, int argc, char** argv);
+int bsts_Bosatsu_Prog_run_main(BValue main_fn, int argc, char** argv);
--- a/c_runtime/bosatsu_runtime.c
+++ b/c_runtime/bosatsu_runtime.c
@@ -1058,6 +1058,170 @@ BValue bsts_integer_from_uint64(uint64_t result) {
   }
 }
 
+uint64_t bsts_integer_to_low_uint64(BValue bint) {
+  if (IS_SMALL(bint)) {
+    return (uint64_t)(int64_t)GET_SMALL_INT(bint);
+  }
+
+  BSTS_Integer* bi = GET_BIG_INT(bint);
+  uint64_t low = 0;
+  if (bi->len > 0) low |= (uint64_t)bi->words[0];
+  if (bi->len > 1) low |= ((uint64_t)bi->words[1]) << 32;
+  if (!bi->sign) return low;
+  return (~low) + UINT64_C(1);
+}
+
+static unsigned bsts_u32_msb_index(uint32_t word) {
+#if defined(__clang__) || defined(__GNUC__)
+  return (unsigned)(31U - (unsigned)__builtin_clz(word));
+#else
+  unsigned idx = 0;
+  while (word >>= 1) {
+    idx++;
+  }
+  return idx;
+#endif
+}
+
+static inline _Bool bsts_bigint_get_bit(const BSTS_Integer* bi, size_t bit_index) {
+  size_t word_index = bit_index >> 5;
+  if (word_index >= bi->len) return 0;
+  return (bi->words[word_index] >> (bit_index & 31U)) & 1U;
+}
+
+static _Bool bsts_bigint_any_lower_bits(const BSTS_Integer* bi, size_t bit_count) {
+  size_t full_words = bit_count >> 5;
+  size_t limit = (full_words < bi->len) ? full_words : bi->len;
+  for (size_t i = 0; i < limit; i++) {
+    if (bi->words[i] != 0) return 1;
+  }
+
+  size_t rem_bits = bit_count & 31U;
+  if (rem_bits == 0 || full_words >= bi->len) return 0;
+  uint32_t mask = (UINT32_C(1) << rem_bits) - 1U;
+  return (bi->words[full_words] & mask) != 0U;
+}
+
+double bsts_integer_to_double(BValue bint) {
+  if (IS_SMALL(bint)) {
+    return (double)GET_SMALL_INT(bint);
+  }
+
+  BSTS_Integer* bi = GET_BIG_INT(bint);
+  size_t len = bi->len;
+  while (len > 0 && bi->words[len - 1] == 0) {
+    len--;
+  }
+  if (len == 0) return 0.0;
+
+  uint32_t top_word = bi->words[len - 1];
+  size_t exponent = (len - 1) * 32U + (size_t)bsts_u32_msb_index(top_word);
+
+  if (exponent > 1023U) {
+    return bi->sign ? -INFINITY : INFINITY;
+  }
+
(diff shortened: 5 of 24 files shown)


core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,6 +149,375 @@ object TypedExprLoopRecurLowering {
         false
     }
 
+  @annotation.tailrec
+  private def stripTypeWrappers[A](te: TypedExpr[A]): TypedExpr[A] =
```
johnynek, line 169: we should add a comment that this can never return the result as a modified expression since it is discarding type information.

It is only a "reader" function, not "writing"

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,6 +149,375 @@ object TypedExprLoopRecurLowering {
...
+      rebuild: TypedExpr[A] => TypedExpr[A]
+  )
+
+  // Collect nested lambda groups from a recursive binding expression.
+  // We allow Generic/Annotation wrappers and non-recursive Let wrappers
+  // between groups, e.g. fn -> (x = 2; (a, b) -> ...).
+  private def collectGroupedLambdas[A](
+      expr: TypedExpr[A]
+  ): Option[GroupedLambda[A]] = {
+    type Group = NonEmptyList[(Bindable, Type)]
```
johnynek: let's lift this type to `private type` on the `object`. We can use it in a few place.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,6 +149,375 @@ object TypedExprLoopRecurLowering {
...
+  ): Option[RewriteResult[A]] = {
+
+    def isSelfHead(head: TypedExpr[A], canRecurHere: Boolean): Boolean =
+      stripTypeWrappers(head) match {
+        case Local(n, _, _) => canRecurHere && (n == fnName)
+        case _              => false
+      }
+
+    @annotation.tailrec
+    def flattenApp(
```
johnynek: let's call this `unapplyApp` since that's what we are doing.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,6 +149,375 @@ object TypedExprLoopRecurLowering {
...
+        curr: TypedExpr[A],
+        acc: List[NonEmptyList[TypedExpr[A]]]
+    ): (TypedExpr[A], List[NonEmptyList[TypedExpr[A]]]) =
+      curr match {
+        case App(fn, args, _, _) =>
+          flattenApp(fn, args :: acc)
+        case other =>
```
johnynek, line 266: let's add a comment about `Generic` or `Annotation` appearing here. I think polymorphic recursion could result in having an `Annotation`. Couldn't it?

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,6 +149,375 @@ object TypedExprLoopRecurLowering {
...
+          (other, acc)
+      }
+
+    def validGroupedArity(groups: List[NonEmptyList[TypedExpr[A]]]): Boolean =
```
johnynek: to make it clear this is just a shape check, hide the `TypedExpr[A]` type:

```scala
def validGroupedArity[T](groups: List[NonEmptyList[T]]): Boolean
```

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,6 +149,375 @@ object TypedExprLoopRecurLowering {
...
+            assert(!body1.changed)
+            RewriteResult(lam, changed = false, body1.sawSelfRef)
+          }
+        case app @ App(fn, args, tpe, tag) =>
+          val (head, groups) = flattenApp(app, Nil)
+          if (isSelfHead(head, canRecur)) {
+            if (!validGroupedArity(groups) || !tailPos) None
+            else rewriteToHelper(groups, tpe, tag, canRecur)
+          } else {
+            etaExpandedGroups(head, groups, canRecur) match {
```
johnynek, line 383: I'm confused. It seems like both isSelfHead and !isSelfHead branches can rewriteToHelper. How does that work in the non-self-head case? Please add a comment, that is quite confusing.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,6 +149,375 @@ object TypedExprLoopRecurLowering {
...
+              case Some(expandedGroups) =>
+                if (!validGroupedArity(expandedGroups) || !tailPos) None
+                else rewriteToHelper(expandedGroups, tpe, tag, canRecur)
+              case None =>
+                (
+                  recur(fn, tailPos = false, canRecur),
+                  args.traverse(recur(_, tailPos = false, canRecur))
+                ).mapN { case (fn1, args1) =>
```
johnynek: we use mapN with two options a lot in this code. Let's inline:

```scala
private inline def map2[A, B, C](oa: Option[A], ob: Option[b])(inline fn: (A, B) => C): Option[C] =
  if (oa.isDefined && ob.isDefined) Some(fn(oa.get, ob.get))
  else None
```
this should be a minimal allocation implementation of map2 for us in this file.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,6 +149,375 @@ object TypedExprLoopRecurLowering {
...
+        case let @ Let(arg, ex, in, rec, tag) =>
+          val (exCanRecur, inCanRecur) =
+            if (arg == fnName) {
+              if (rec.isRecursive) (false, false)
+              else (canRecur, false)
+            } else (canRecur, canRecur)
+          (
+            recur(ex, tailPos = false, exCanRecur),
+            recur(in, tailPos, inCanRecur)
+          ).mapN { (ex1, in1) =>
```
johnynek: avoid mapN when N=2 (use map2).

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,6 +149,375 @@ object TypedExprLoopRecurLowering {
...
+        case Local(n, _, _) => Some(n)
+        case _              => None
+      }
+
+    def etaExpandedGroups(
+        head: TypedExpr[A],
+        outerGroups: List[NonEmptyList[TypedExpr[A]]],
+        canRecurHere: Boolean
+    ): Option[List[NonEmptyList[TypedExpr[A]]]] =
+      stripTypeWrappers(head) match {
```
johnynek, line 292: I'm very concerned that we are discarding Annotation and Generic here which could discard typing information from the annotated lambda inside. How is this safe? The resulting etaExpandedGroups are used directly on some branches.

We need to to test specifically with a polymorphic recursion example and examine the resulting code.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,6 +149,375 @@ object TypedExprLoopRecurLowering {
...
+    ): Option[RewriteResult[A]] =
+      groups
+        .traverse { group =>
+          group.traverse(recur(_, tailPos = false, canRecurHere))
+        }
+        .map { group1 =>
+          val changedNested = group1.exists(_.exists(_.changed))
+          val sawNested = group1.exists(_.exists(_.sawSelfRef))
+          val flatArgs =
+            group1.toList.flatMap(_.toList.map(_.expr))
```
johnynek: seems like we are converting `NonEmptyList` toList twice just to unsafe convert back a few lines below. Let's remove those and just stay in `NonEmptyList`.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,6 +149,375 @@ object TypedExprLoopRecurLowering {
...
+          val app1 = App(
+            Local(helperName, helperFnType, tag),
+            NonEmptyList.fromListUnsafe(flatArgs),
+            tpe,
+            tag
+          )
+          RewriteResult(
+            app1,
+            changed = true || changedNested,
```
johnynek: what is `true || x` all about? that's always true right? Why are we doing that? Same with sawSelfRef.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,6 +149,375 @@ object TypedExprLoopRecurLowering {
...
+        tpe: Type,
+        tag: A,
+        canRecurHere: Boolean
+    ): Option[RewriteResult[A]] =
+      groups
+        .traverse { group =>
+          group.traverse(recur(_, tailPos = false, canRecurHere))
+        }
+        .map { group1 =>
+          val changedNested = group1.exists(_.exists(_.changed))
```
johnynek: we don't seem to need these since below they are just `||` with true. So delete this and the next line.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,6 +149,375 @@ object TypedExprLoopRecurLowering {
...
+          val sawNested = group1.exists(_.exists(_.sawSelfRef))
+          val flatArgs =
+            group1.toList.flatMap(_.toList.map(_.expr))
+          val app1 = App(
```
johnynek, line 334: seems like here we could require type information we may have discarded with the `strip*` function above.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: By the way: it would be acceptable to only handle monomorphic recursion at this stage and come back later and handle polymorphic recursion.

If we adjust in that direction, leave a comment explaining that polymorphic recursion is more complex and not yet handled.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,6 +149,375 @@ object TypedExprLoopRecurLowering {
...
+                  args.traverse(recur(_, tailPos = false, canRecur))
+                ).mapN { case (fn1, args1) =>
+                  val sawSelf = fn1.sawSelfRef || args1.exists(_.sawSelfRef)
+                  RewriteResult(app, changed = false, sawSelf)
+                }
+            }
+          }
+        case let @ Let(arg, ex, in, rec, tag) =>
+          val (exCanRecur, inCanRecur) =
+            if (arg == fnName) {
```
johnynek, line 406: what is this case? And inner shadow of the outer fnName? Please comment to this case.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,6 +149,375 @@ object TypedExprLoopRecurLowering {
...
+    type Group = NonEmptyList[(Bindable, Type)]
+
+    def scanBody(
+        te: TypedExpr[A],
+        groups: Vector[Group],
+        rebuild: TypedExpr[A] => TypedExpr[A]
+    ): Option[GroupedLambda[A]] =
+      te match {
+        case Generic(q, in) =>
+          scanBody(in, groups, in1 => rebuild(Generic(q, in1)))
```
johnynek, line 198: isn't there an issue about scope and type shadowing here? Please comment how we can rebuild without worrying about scopes being correct?

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,6 +149,375 @@ object TypedExprLoopRecurLowering {
...
+          // Bare self-reference escapes the supported full-application shape.
+          None
+        case n @ (Local(_, _, _) | Global(_, _, _, _) | Literal(_, _, _)) =>
+          Some(RewriteResult(n, changed = false, sawSelfRef = false))
+      }
+
+    recur(te, tailPos, canRecur)
+  }
+
+  private def rewriteMultiGroupTailRec[A](
```
johnynek: Let's add a comment here. This, I think, is doing the meat of the optimization. We are rewriting multigroup tail recursive calls, so that we can get a loop/recur rather than allocating closures and closure calls on every loop.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek, line 10: Maybe this whole Loop/Recur lowering at this stage is only safe when it is monomorphic recursion, otherwise the types of the loop arguments are changing in every loop.

Are we limiting to that now?

I think we should to keep the typed AST fully typed. But that brings up an interesting point: in Matchless, types are erased so polymorphic recursion can also be lowered (and inlined). Polymorphic recursion is much more rare, but still happens. If we make this change (and I think we should), we should 1. test it clearly, 2. add an issue to support polymorphic recursion loop lowering in Matchless (once types are erased).

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. Added an explicit reader-only comment on `stripTypeWrappers` clarifying it must never be used to rebuild expressions.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. Lifted the lambda-group alias to object scope as `private type Group`.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. Renamed `flattenApp` to `unapplyApp`.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. Added comments around Generic/Annotation handling in eta-expanded shape checks.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. Updated to `validGroupedArity[T]` so this is clearly a shape-only check.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. Added a comment on the non-self-head branch describing eta-expanded grouped recursion detection.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. Added local inline `map2` helper and switched two-Option sites to use it.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. Replaced remaining `mapN`-for-2 usages in this file with `map2`.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. Lowering is now guarded by a monomorphic self-call type check, and tests now cover both allowed monomorphic generic recursion and skipped type-changing polymorphic recursion.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. Removed unnecessary `NonEmptyList -> List -> NonEmptyList` churn and keep args as `NonEmptyList` via `concatNel`.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. Removed the `true || ...` logic and dropped the unused nested flags.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. Deleted those now-unneeded intermediate vars.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. We now avoid type-loss assumptions by checking self-call head type stability (`fn.getType.sameAs(expectedFnType)`) before lowering.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. Added explicit monomorphic-only comments and behavior, with a follow-up issue note for matchless/type-erased polymorphic lowering.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. Added a comment in the `arg == fnName` `Let` case to explain binder shadowing behavior.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. Added comment explaining why `rebuild` preserves scope/type shadowing.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. Added comment documenting the grouped tail-recursive helper rewrite as the optimization core.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
johnynek: Addressed in 45f6f90f. Lowering is now monomorphic-type-stable, follow-up issue #1749 is referenced, and tests now include explicit monomorphic generic `len` lowering plus a type-changing polymorphic case that is skipped.

johnynek:
Pushed follow-up commit `45f6f90f` to address the monomorphic recursion concern.

Key updates:
- replaced blanket Generic-wrapper rejection with a self-call head type-stability check (`fn.getType.sameAs(expectedFnType)`), so monomorphic generic recursion is still lowered
- kept lowering conservative for type-changing polymorphic recursion (skipped)
- added explicit tests in `TypedExprTest`:
  - `lowerAll lowers top-level monomorphic generic len recursion`
  - `lowerAll skips top-level type-changing polymorphic recursion`
- all `coreJVM/testOnly *TypedExprTest` tests pass locally (105/105)

Also replied inline on each review comment thread with the corresponding change mapping.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -306,21 +322,26 @@ object TypedExprLoopRecurLowering {
...
           val flatArgs =
-            group1.toList.flatMap(_.toList.map(_.expr))
+            group1 match {
+              case h :: t =>
+                t.foldLeft(h.map(_.expr)) { (acc, group) =>
+                  acc.concatNel(group.map(_.expr))
+                }
+              case Nil =>
+                throw new IllegalStateException(
+                  "rewriteToHelper requires at least one argument group"
```
johnynek: let's require `groups: NonEmptyList[NonEmptyList[TypedExpr[A]]]` and do the check at the call site rather than use site.

johnynek: Addressed in 41fb6162. `rewriteToHelper` now requires `NonEmptyList[NonEmptyList[TypedExpr[A]]]`, and call sites convert via `NonEmptyList.fromList(...)` after arity/tail checks instead of validating at use site.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,52 +163,509 @@ object TypedExprLoopRecurLowering {
...
+      fnName: Bindable,
+      helperName: Bindable,
+      helperFnType: Type,
+      groupedArities: Vector[Int],
+      canRecur: Boolean
+  ): Option[RewriteResult[A]] = {
+
+    def isSelfHead(head: TypedExpr[A], canRecurHere: Boolean): Boolean =
+      stripTypeWrappers(head) match {
+        case Local(n, _, _) => canRecurHere && (n == fnName)
```
johnynek: let's lift to:

```scala
canRecurHere && stripTypeWrappers(head) match {
  case Local(n, _, _) => n == fnName
  case _ => false
}
```

So we don't need to do the strip if canRecurHere is false.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,52 +163,509 @@ object TypedExprLoopRecurLowering {
...
+      // We do not rebuild from stripped nodes. Also, loop/recur lowering is
+      // gated to monomorphic recursion so recursive polymorphism is skipped.
+      stripTypeWrappers(head) match {
+        case AnnotatedLambda(lambdaArgs, lambdaBody, _) =>
+          outerGroups match {
+            case applied :: rest if applied.length == lambdaArgs.length =>
+              val (innerHead, innerGroups) =
+                unapplyApp(stripTypeWrappers(lambdaBody), Nil)
+              if (
+                !isSelfHead(innerHead, canRecurHere) || innerGroups.isEmpty
```
johnynek: innerGroups.isEmpty is cheap. Run that first.

```scala
innerGroups.isEmpty || !isSelfHead(innerHead, canRecurHere)
```

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,52 +163,509 @@ object TypedExprLoopRecurLowering {
...
 
+  private def isMonomorphicRecursiveBinding[A](
+      fnName: Bindable,
+      expr: TypedExpr[A]
+  ): Boolean = {
+    val expectedFnType = stripTypeWrappers(expr).getType
+
+    def loop(te: TypedExpr[A], canRecur: Boolean): Boolean =
+      te match {
+        case Generic(_, in) =>
```
johnynek: We stripped generics at `expectedFnType` so, it could have Type.Var. Then here, we are ignoring the quantification, and that could shadow our generics in the original type. I think we need to keep quantification vars from `expr` and if we get into `Generic` we see if any shadow, and if they do, we have to unshadow or handle in some way in the recursion.

Isn't that right?

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,52 +163,509 @@ object TypedExprLoopRecurLowering {
...
+          loop(in, canRecur)
+        case Annotation(in, _, _) =>
```
johnynek, line 499: this could be a widening that we are ignoring. Please add a comment to the code why this is safe so a reader can quickly follow.

I think it is because the thing we are really checking is the `App(fn,...` below and we use `fn.getType` which could have an annotation inside which is handled correctly. But we should comment to explain here.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,52 +163,509 @@ object TypedExprLoopRecurLowering {
...
+          loop(in, canRecur)
+        case AnnotatedLambda(args, body, _) =>
+          val canRecurBody = canRecur && !args.exists(_._1 == fnName)
+          loop(body, canRecurBody)
+        case App(fn, args, _, _) =>
+          val selfHeadTypeStable =
```
johnynek, line 507: isn't this the same as:

```scala
isSelfHead(fn, canRecur) && fn.getType.sameAs(expectedFnType)
```
I think that is more clear if so. Let's change to that.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,52 +163,509 @@ object TypedExprLoopRecurLowering {
...
+  // calls (including non-tail ones) through a helper to avoid closure
+  // allocations on recursive steps.
+  private def flattenGroupedSelfCalls[A](
+      fnName: Bindable,
+      expr: TypedExpr[A]
+  ): Option[TypedExpr[A]] =
+    if (!isMonomorphicRecursiveBinding(fnName, expr)) None
+    else {
+      collectGroupedLambdas(expr).flatMap { grouped =>
+        if (grouped.groups.lengthCompare(2) < 0) None
```
johnynek: add a comment that this is the "only one group case, nothing to do"

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,52 +163,509 @@ object TypedExprLoopRecurLowering {
...
+        else {
+          val groupedArgs = grouped.groups.toList.flatMap(_.toList)
```
johnynek: here we know that `grouped.groups` is a Vector of length 2 or more. Let's directly do:

```scala
// we already checked that groups has 2 or more items so this is safe
val groupsNel = NonEmptyList.fromListUnsafe(grouped.groups.toList)
```
Then we can just flatMap the NonEmptyList to stay nonempty all the way through without another conversion to NonEmptyList unsafe. This is the clearest point to show we are non-empty.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,52 +163,509 @@ object TypedExprLoopRecurLowering {
...
+
+            val substitutedBody =
+              TypedExpr.substituteAll(
+                renameMap,
+                grouped.terminalBody,
+                enterLambda = true
+              ).get
+
+            val helperFnType = Type.Fun(
+              NonEmptyList.fromListUnsafe(helperArgs.map(_._2)),
```
johnynek: if we keep helperArgs an NEL this can avoid the fromListUnsafe.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,52 +163,509 @@ object TypedExprLoopRecurLowering {
...
+              substitutedBody.getType
+            )
+            val groupedArities = grouped.groups.iterator.map(_.length).toVector
```
johnynek: why the `.iterator.map(_.length).toVector`? `.groups` was already a Vector. Just `grouped.groups.map(_.length)` should give the same type.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,52 +163,509 @@ object TypedExprLoopRecurLowering {
...
+              fnName,
+              helperName,
+              helperFnType,
+              groupedArities,
+              canRecur = true
+            ).flatMap { rewritten =>
+              if (!rewritten.sawSelfRef) None
+              else {
+                val helperDef = AnnotatedLambda(
+                  NonEmptyList.fromListUnsafe(helperArgs),
```
johnynek: again, if we keep NELs we won't need this fromListUnsafe.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,52 +163,509 @@ object TypedExprLoopRecurLowering {
...
+          }
+        }
+      }
+    }
+
+  private def lowerRecursiveBinding[A](
+      name: Bindable,
+      te: TypedExpr[A]
+  ): Option[TypedExpr[A]] =
+    if (!isMonomorphicRecursiveBinding(name, te)) None
```
johnynek: we are checking if te is monomorphic before calling this function and it is private. Instead of checking twice let's have `case class MonomorphicRec[A](name: Binding, te: TypeExpr[A])` and have `def toMonomorphicRec[A](name: Binding, te: TypedExpr[A]): Option[MonomorphicRec[A]]`

Then have this method consume a `MonomorphicRec[A]` so this at least loudly advertises the requirements here.

core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
```diff
@@ -148,52 +163,509 @@ object TypedExprLoopRecurLowering {
...
+  // Polymorphic recursion can be optimized later after type erasure:
+  // https://github.com/johnynek/bosatsu/issues/1749
+  // Phase 1: grouped self-call flattening. This rewrites grouped recursive
+  // calls (including non-tail ones) through a helper to avoid closure
+  // allocations on recursive steps.
+  private def flattenGroupedSelfCalls[A](
+      fnName: Bindable,
+      expr: TypedExpr[A]
+  ): Option[TypedExpr[A]] =
+    if (!isMonomorphicRecursiveBinding(fnName, expr)) None
```
johnynek: see my comment below about how this is only called when it is already checked. Let's add the MonomorphicRec type and not check this twice. Have this function require a `MonomorphicRec[A]` which is proof we have checked this.

PR #1753, opened 2026-02-20
Implement match-site struct scrutinee elision

### LINKED ISSUE #1732: Improving matches on tuples
A common pattern is matching a tuple created at the match site:

```
match (f, g):
  case (..., ...):
```

Naively compiled, this required allocating a tuple. But we could instead be more intelligent, perhaps at the Matchless layer, and see that when we have a match on creating a tuple, we could just elide the tuple creation, and replace getting items from the tuples with the original expressions.

So, the rewrite would be something like:

```
# if f or g aren't cheap already (not already vars, or zero arg constructors)
_f$ = f
_g$ = g
```
Then just do the mapping to use `_f$` instead of `GetStructItem(m, 0)` and `_g$` instead of `GetStructItem(m, 1)`

We could even potentially make them lazy so that we only evaluate f and g if they are actually needed using LocalAnonMut and SetMut potentially.

johnynek:
This PR implements the design from `issue_1732_match_site_struct_scrutinee_elision_design`.

closes #1732

## What changed
- Added match-site root struct inlining support in `Matchless`:
  - prepares an inlined struct-root view for `App(MakeStruct(_), args)` scrutinees,
  - projects fields directly from inlined root values instead of rebuilding/projection chains,
  - binds whole-root patterns (`Var`/`Named`) to reconstructed root value only when needed.
- Threaded optional inlined-root context through ordered and matrix match compilation paths.
- Added conservative fallback policy: if more than one branch binds the whole root, skip inlining and keep previous eager root-allocation behavior.
- Extended regression tests in `MatchlessTests` for:
  - ordered/matrix behavior on named/annotated/union/list patterns,
  - whole-root binding behavior,
  - strict evaluation ordering,
  - long orthogonal-prefix bridge behavior.
- Updated C codegen golden hash in `ClangGenTest` (`Ackermann`) after codegen changes.

## Coverage
- Diff statement coverage for changed `Matchless.scala` lines: **100.00%**
  - `CHANGED_STMT_LINES=96`
  - `UNCOVERED_CHANGED_STMT_LINES=0`

## Validation
- `sbt "doc; paradox" && ./test_cli.sh`
- `sbt "coreJS/test; jsapiJS/compile" && sbt "jsuiJS/test"`
- `sbt "cli/assembly"` + Python transpile + `python3 -m unittest discover pyout -v --pattern "*.py"`
- `sbt "coverage; clean; coreJVM/test"`
- `sbt "coverage; cli/test; coverageReport"`
- `sbt "cli/nativeImage"` + `./bosatsu tool test ...`
- `sbt "cliJSJS/fullOptJS"` + `./bosatsu_node tool test ...`
- `make -C c_runtime install VERSION=$(git rev-parse HEAD)` + node lib tests
- C runtime/transpile/lib-build CI-equivalent flow (including runtime bench and issue1642 single-assertion check)
- `sbt "++3.8.1; cli/assembly"` + `scripts/publish_bosatsu_libs.sh` dry run

### DIFF
--- a/core/src/main/scala/dev/bosatsu/RingOpt.scala
+++ b/core/src/main/scala/dev/bosatsu/RingOpt.scala
@@ -1084,31 +1084,35 @@ object RingOpt {
           val l = undistribute[A](l0a)
           val r = undistribute[A](r0a)
 
-          (l, r) match {
-            case (Mult(a, b), Mult(c, d)) =>
-              if (isEq(a, c)) Mult(a, undistribute(Add(b, d)))
-              else if (isEq(a, d)) Mult(a, undistribute(Add(b, c)))
-              else if (isEq(b, c)) Mult(b, undistribute(Add(a, d)))
-              else if (isEq(b, d)) Mult(b, undistribute(Add(a, c)))
-              else Add(l, r)
-            case (Mult(a, b), Neg(c)) =>
-              if (isEq(a, c)) Mult(a, Add(b, Neg(One)))
-              else if (isEq(b, c)) Mult(b, Add(a, Neg(One)))
-              else Add(l, r)
-            case (Mult(a, b), c) =>
-              if (isEq(a, c)) Mult(a, Add(b, One))
-              else if (isEq(b, c)) Mult(b, Add(a, One))
-              else Add(l, r)
-            case (Neg(c), Mult(a, b)) =>
-              if (isEq(a, c)) Mult(a, Add(Neg(One), b))
-              else if (isEq(b, c)) Mult(b, Add(Neg(One), a))
-              else Add(l, r)
-            case (c, Mult(a, b)) =>
-              if (isEq(a, c)) Mult(a, Add(One, b))
-              else if (isEq(b, c)) Mult(b, Add(One, a))
-              else Add(l, r)
-            case _ => Add(l, r)
-          }
+          // Keep repeated-add shape intact so normalize(Add) can compare
+          // term-wise decomposition against 2*x style rewrites.
+          if (isEq(l, r)) Add(l, r)
+          else
+            (l, r) match {
+              case (Mult(a, b), Mult(c, d)) =>
+                if (isEq(a, c)) Mult(a, undistribute(Add(b, d)))
+                else if (isEq(a, d)) Mult(a, undistribute(Add(b, c)))
+                else if (isEq(b, c)) Mult(b, undistribute(Add(a, d)))
+                else if (isEq(b, d)) Mult(b, undistribute(Add(a, c)))
+                else Add(l, r)
+              case (Mult(a, b), Neg(c)) =>
+                if (isEq(a, c)) Mult(a, Add(b, Neg(One)))
+                else if (isEq(b, c)) Mult(b, Add(a, Neg(One)))
+                else Add(l, r)
+              case (Mult(a, b), c) =>
+                if (isEq(a, c)) Mult(a, Add(b, One))
+                else if (isEq(b, c)) Mult(b, Add(a, One))
+                else Add(l, r)
+              case (Neg(c), Mult(a, b)) =>
+                if (isEq(a, c)) Mult(a, Add(Neg(One), b))
+                else if (isEq(b, c)) Mult(b, Add(Neg(One), a))
+                else Add(l, r)
+              case (c, Mult(a, b)) =>
+                if (isEq(a, c)) Mult(a, Add(One, b))
+                else if (isEq(b, c)) Mult(b, Add(One, a))
+                else Add(l, r)
+              case _ => Add(l, r)
+            }
         case Neg(x) =>
           val xa: Expr[A] = x
           Neg(undistribute[A](xa))
--- a/core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
+++ b/core/src/main/scala/dev/bosatsu/TypedExprLoopRecurLowering.scala
@@ -1,13 +1,28 @@
 package dev.bosatsu
 
 import cats.Eq
+import cats.data.NonEmptyList
 import cats.syntax.all._
-import dev.bosatsu.rankn.TypeEnv
+import dev.bosatsu.rankn.{Type, TypeEnv}
 
 import Identifier.Bindable
 
 object TypedExprLoopRecurLowering {
   import TypedExpr._
+
+  private type Group = NonEmptyList[(Bindable, Type)]
+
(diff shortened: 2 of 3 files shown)


core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -2484,12 +2484,87 @@ object Matchless {
...
+                  letsRev.reverse,
+                  InlinedStructRoot(fieldsRev.reverse.toVector)
+                )
+              )
+            }
+        case _ =>
+          Monad[F].pure(None)
+      }
+
+    def structRootValue(root: InlinedStructRoot): Expr[B] =
```
johnynek: make this a method on `case class InlinedStructRoot`

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -2484,12 +2484,87 @@ object Matchless {
...
+      inlinedRoot.filter(_ => occ.equals(rootOcc))
+
     // return the check expression for the check we need to do, and the list of bindings
     // if must match is true, we know that the pattern must match, so we can potentially remove some checks
     def doesMatch(
         arg: CheapExpr[B],
         pat: Pattern[(PackageName, Constructor), Type],
-        mustMatch: Boolean
+        mustMatch: Boolean,
+        rootInlined: Option[InlinedStructRoot] = None
```
johnynek: don't put a default argument. It hides our attempts to refactor. We need to be explicit in each call.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -2572,7 +2653,7 @@ object Matchless {
                 .flatMap { case (optAnonLeft, tmpList) =>
                   val anonList = LocalAnonMut(tmpList)
 
-                  doesMatch(anonList, Pattern.ListPat(right.toList), false)
+                  doesMatch(anonList, Pattern.ListPat(right.toList), false, None)
```
johnynek, line 2656: add a comment `rootInlined = None` because we aren't matching the root.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -2621,10 +2702,20 @@ object Matchless {
               glob match {
                 case Pattern.ListPart.WildList =>
                   // no binding on the let
-                  doesMatch(arg, Pattern.ListPat(right.toList), mustMatch)
+                  doesMatch(
+                    arg,
+                    Pattern.ListPat(right.toList),
+                    mustMatch,
+                    rootInlined
```
johnynek: rootInlined can never work for a List because a List has two type constructors. Just use `None` for these list pattern matches for clarity.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -2621,10 +2702,20 @@ object Matchless {
...
+                  )
                 case Pattern.ListPart.NamedList(ln) =>
                   // bind empty to ln
-                  doesMatch(arg, Pattern.ListPat(right.toList), mustMatch)
+                  doesMatch(
+                    arg,
+                    Pattern.ListPat(right.toList),
+                    mustMatch,
+                    rootInlined
```
johnynek: a List can never match an inlined root (which must be a struct not an enum), so set rootInlined = None.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -2948,14 +3041,15 @@ object Matchless {
     // WildCard, Literal, PositionalStruct, or Union.
     def peelPattern(
         p: Pattern[(PackageName, Constructor), Type],
-        occ: CheapExpr[B]
+        occ: CheapExpr[B],
+        inlined: Option[InlinedStructRoot] = None
```
johnynek: don't use the default value, let's be explicit at each call site.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -3009,7 +3105,12 @@ object Matchless {
             )
           ) { case ((bs, acc), (pat, occ)) =>
             val p1 = normalizePattern(pat)
-            val (moreBinds, core) = peelPattern(p1, occ)
+            val maybeInlined = rootInlinedForOcc(occ, rootOcc, inlinedRoot)
+            val (moreBinds, core) =
+              maybeInlined match {
+                case Some(inl) => peelPattern(p1, occ, Some(inl))
+                case None      => peelPattern(p1, occ)
```
johnynek: isn't this the same as `peelPattern(p1, occ, maybeInlined)` the None case is the same as passing None as the arg?

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -3317,7 +3419,14 @@ object Matchless {
           }
 
         val mustMatchPattern = branches.tail.isEmpty && head1.guard.isEmpty
-        doesMatch(arg, head1.pattern, mustMatchPattern).flatMap(loop)
+        val matched =
+          rootInlined match {
+            case Some(inl) =>
+              doesMatch(arg, head1.pattern, mustMatchPattern, Some(inl))
+            case None =>
+              doesMatch(arg, head1.pattern, mustMatchPattern)
```
johnynek: this is the same as:

`doesMatch(arg, head1.pattern, mustMatchPattern, rootIlined)` let's be direct.

so the whole thing is:
```
doesMatch(arg, head1.pattern, mustMatchPattern, rootInlined).flatMap(loop)
```

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -3860,36 +3974,76 @@ object Matchless {
...
-      val (orthoPrefix, nonOrthoSuffix) =
-        branches.toList.span(branch => !isNonOrthogonal(branch.pattern))
-      val maybeNonOrthoSuffix = NonEmptyList.fromList(nonOrthoSuffix)
+      def bindsWholeRoot(
+          p: Pattern[(PackageName, Constructor), Type]
+      ): Boolean =
+        p match {
+          case Pattern.Var(_) | Pattern.Named(_, _) => true
+          case Pattern.Annotation(inner, _)          => bindsWholeRoot(inner)
+          case _                                    => false
```
johnynek, line 3977: I think we also need:
```
case Pattern.Union(h, tail) =>
  bindsWholeRoot(h) && tail.forall(bindsWhileRoot)
case _ => false
```

PR #1758, opened 2026-02-21
Validate typed AST connectivity across rewrite pipeline

### LINKED ISSUE #1751: verifyType function in TestUtils for TypedExpr
After typechecking in Infer, we do more optimizations and rewrites, but we never actually check that we haven't incorrectly handled the typed AST.

It may be that we accidentally break the types there because Matchless is a type-erased AST really on purpose to make execution with type parameters easier.

We should add something like:

```scala
def validateTypes[A](te: TypedExpr[A], names: Map[Name[A], Type], env: TypeEnv[Kind.Arg]): ValidatedNec[Error, Unit]

def validatePackageMap[A](pm: PackageMap[...], ...
```
Then, we could use this in TestUtils to validate that we haven't messed up the types when doing rewrites and optimizations, loop lowered, etc...

johnynek:
## Summary
- add typed AST connectivity validation helpers in `TestUtils` (`validateTypes`, `validatePackageMap`) with detailed path-based errors
- validate inferred code at three stages: unoptimized, after `TypedExprLoopRecurLowering`, and after `TypedExprNormalization`
- wire these checks into existing `TestUtils` helpers (`checkLast`, `checkPackageMap`, `compileFile`, `testInferred`)
- hook `WellTypedTests` to assert optimized package maps still have valid type connectivity
- fix match-branch local scope tracking in type validation to include all pattern-bound names (prevents false "unbound local" failures)

## Testing
- sbt "coreJVM/testOnly dev.bosatsu.codegen.python.PythonGenJvmTest"
- sbt "coreJVM/test"

fixes #1751


core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -12,6 +12,32 @@ import cats.syntax.all._
...
+  private val typeValidationPass: TypeValidation[Unit] =
+    Validated.validNec(())
+
+  private def typeValidationFail(
+      path: String,
+      message: String
+  ): TypeValidation[Unit] =
+    Validated.invalidNec(TypeValidationError(path, message))
+
+  private def combineTypeValidation(
```
johnynek: this is just `checks.sequence_` we don't need this function.

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -134,13 +160,518 @@ object TestUtils {
...
+              locals,
+              globals,
+              env,
+              loopStack,
+              s"$path/app/arg[$idx]"
+            )
+          }
+        )
+
+        val appTypeChecks: TypeValidation[Unit] = {
```
johnynek: this is incomplete. We need to check that the arguments are the same as what the function expects.

Or, possibly `applied arg i <:< function arg i` (which we could check using `Infer.substitutionCheck` or something, but I think ideally they would line up exactly.

This raises another point: we need to have negative tests of this code: we should create examples that don't type check and see that we can detect them. So far, maybe `Validated.unit` is all we are doing. What evidence do we have against that?

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -134,13 +160,518 @@ object TestUtils {
...
+          globals,
+          env,
+          loopStack,
+          s"$path/match/arg"
+        )
+
+        val branchChecks = branches.toList.zipWithIndex.map { case (branch, idx) =>
+          val branchPath = s"$path/match/branch[$idx]"
+          val branchLocals: Map[Identifier.Bindable, Type] =
+            branch.pattern.names.iterator.map(_ -> Type.UnitType).toMap
```
johnynek: these are not the right types. We need a function like:

```scala
def getBindingTypes(scurtinee: Type, p: Pattern[...], typeEnv: TypeEnv[Kind.Arg]): Map[Bindable, Type]
```

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -134,13 +160,518 @@ object TestUtils {
...
+            scope,
+            globals,
+            env,
+            loopStack,
+            s"$branchPath/body"
+          )
+          val branchScopeChecks =
+            combineTypeValidation(guardCheck :: branchExprCheck :: Nil)
+
+          combineTypeValidation(branchResultTypeCheck :: branchScopeChecks :: Nil)
```
johnynek: why not make just one big one:

```
combineTypeValidation(branchResultTypeCheck :: guardCheck :: branchExprCheck :: Nil)
```

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -1,17 +1,126 @@
...
+  }
+
+  private type TypeValidation[+A] = ValidatedNec[TypeValidationError, A]
+  private case class LoopContext(argTypes: List[Type], recurType: Type)
+
+  private val typeValidationPass: TypeValidation[Unit] =
+    Validated.validNec(())
+
+  private def typeValidationFail(
+      path: String,
```
johnynek: instead of doing String concatenations to do the path, let's do `Chain[String]` and then show it with `.iterator.mkString("/")`

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -1,17 +1,126 @@
...
+      message: String
+  ): TypeValidation[Unit] =
+    Validated.invalidNec(TypeValidationError(path, message))
+
+  private def showType(t: Type): String =
+    Type.fullyResolvedDocument.document(t).render(80)
```
johnynek: let's do:

```scala
given Show[Type] with {
  def show(t: Type) = Type.fullyResolvedDocument.document(t).render(80)
}
```

Actually, we should probably just put the above in `object Type` and stop copying that fullyResolvedDocument code all over the codebase.

then use `show"foo has type $tpe"`

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -1,17 +1,126 @@
...
+
+  private def typesConnectInScope(
```
johnynek: please add a comment to what we mean here. What does "connect" mean in this context?

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -1,17 +1,126 @@
...
+        ): (List[(Type.Var.Bound, Kind)], Type.Rho) =
+          Type.Fun.unapply(rho) match {
+            case Some((args, Type.ForAll(vars, res))) =>
+              val argFrees = Type.freeBoundTyVars(args.toList).toSet
+              val (movable, blocked) =
+                vars.toList.partition { case (b, _) => !argFrees(b) }
+              NonEmptyList.fromList(movable) match {
+                case None =>
+                  (accRev.reverse, rho)
+                case Some(movableNel) =>
```
johnynek: we never use `movableNel` as a list, let's just use pattern matching:
```scala
movable match {
  case Nil => (accRev.reverse, rho)
  case movableNel =>
    // ...
                  loop(
                    Type.Fun(args, nextRes),
                    movableNel reverse_::: accRev
                  )    
```

But more generally two issues:

1. we could unshadow here, couldn't we? That way we can list all the foralls above without this loop.
2. what you are doing here is leveraging covariance to lift universals out. Couldn't we write a function to do that for all covariant types of kind `+* -> *`?

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -1,17 +1,126 @@
...
+    expected.sameAs(actual) || {
+      def hoistResultForAll(t: Type): Type = {
+        val (outerForalls, rho0) = Type.liftUniversals(t)
+
+        @annotation.tailrec
+        def loop(
+            rho: Type.Rho,
+            accRev: List[(Type.Var.Bound, Kind)]
+        ): (List[(Type.Var.Bound, Kind)], Type.Rho) =
+          Type.Fun.unapply(rho) match {
```
johnynek: let's rewrite this match as:
```scala
rho match {
  case Type.Fun(args, Type.ForAll(vars, res)) =>

  case _ =>
}
```

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -1,17 +1,126 @@
...
+  private def showType(t: Type): String =
+    Type.fullyResolvedDocument.document(t).render(80)
+
+  private def typesConnectInScope(
+      expected: Type,
+      actual: Type,
+      boundKinds: Map[Type.Var.Bound, Kind]
+  ): Boolean =
+    expected.sameAs(actual) || {
+      def hoistResultForAll(t: Type): Type = {
```
johnynek: I think this should really work for any covariant type, not just function. I think we need to take a TypeEnv and a `def hoistCovariant`.

This function shouldn't be a private inner function really, because it seems it is the opposite of pushDownCovariant in Type. We should be able to hoist then pushdown then hoist and it be the same as hoist. Or pushDown hoist pushDown and it be the same as pushDown. Those would be good laws to add.

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -1,17 +1,126 @@
...
+      }
+
+      val expected1 = hoistResultForAll(expected)
+      val actual1 = hoistResultForAll(actual)
+      val inferredKinds =
+        Type
+          .freeBoundTyVars(expected1 :: actual1 :: Nil)
+          .iterator
+          .filterNot(boundKinds.contains)
+          .map(_ -> Kind.Type)
```
johnynek: we shouldn't have types that are not in boundKinds should we? Isn't this a type error if we find them here? We want fully resolved types.

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -1,17 +1,126 @@
...
+      val scopedFrees =
+        Type.freeBoundTyVars(fromRho :: Nil).distinct.filter(envKinds.contains)
+      val vars =
+        (
+          foralls.iterator ++
+            exists.iterator ++
+            scopedFrees.iterator.map(b => b -> envKinds(b))
+        ).toMap
+
+      Type
```
johnynek: this is a directional check. So that means this function is directional. We should be concrete about that:

we are checking `from` can be widened into `to`.

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -1,17 +1,126 @@
...
+          vars = vars,
+          from = fromRho,
+          toVars = Map.empty,
+          to = actual1,
+          env = envKinds
+        )
+        .isDefined
+    }
+
+  private def functionTailTypes(t: Type): List[Type] = {
```
johnynek: add a comment to define a tail type.

This seems like it assuming that function types are curried in bosatsu, but they aren't there is no relationship between the different function types.

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -134,13 +243,856 @@ object TestUtils {
...
 
+  private def assertTypeValidationResult(
+      stage: String,
+      res: TypeValidation[Unit]
+  ): Unit =
+    res match {
+      case Validated.Valid(_) =>
+        ()
+      case Validated.Invalid(errs) =>
+        val msg = errs.toNonEmptyList.toList.map(_.toString).mkString("\n")
```
johnynek: change this to: `errs.iterator.map(_.toString).mkString("\n")`

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -134,13 +243,856 @@ object TestUtils {
...
+
+  private def patternBindingTypes(
+      scrutineeType: Type,
+      pat: Pattern[(PackageName, Constructor), Type],
+      env: TypeEnv[Kind.Arg],
+      path: String
+  ): TypeValidation[Map[Identifier.Bindable, Type]] = {
+    type Bindings = Map[Identifier.Bindable, Type]
+
+    def failAt[A](at: String, message: String): TypeValidation[A] =
```
johnynek: lift this to where `TypeValidation` is defined and reuse it in this file.

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -134,13 +243,856 @@ object TestUtils {
...
+      Validated.invalidNec(TypeValidationError(at, message))
+
+    def mergeBindings(
+        at: String,
+        left: Bindings,
+        right: Bindings
+    ): TypeValidation[Bindings] =
+      right.toList.foldLeft(Validated.validNec(left): TypeValidation[Bindings]) {
```
johnynek: I don't like this approach. Let's instead do something like:

```scala
left.product(right).andThen { case (lbindings, rbindings) =>
  val collisions = lbindings.keySet & rbindings.keySet
  val collisionValid = collisions.toList.sorted.traverse_ { name =>
    // check that the two types are the same
  }
  
  collisionValid.as(lbindings ++ rbindings)
}
```

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -134,13 +243,856 @@ object TestUtils {
...
+            case Pattern.StrPart.NamedStr(name) =>
+              Validated.validNec(Map(name -> Type.StrType))
+            case Pattern.StrPart.NamedChar(name) =>
+              Validated.validNec(Map(name -> Type.CharType))
+          }
+          val mergedParts = mergeAll(at, partBindings)
+          (
+            strCheck ::
+              mergedParts.void :: Nil
+          ).sequence_.andThen(_ => mergedParts)
```
johnynek: why not `strCheck *> mergedParts` here?

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -134,13 +243,856 @@ object TestUtils {
...
+                  Validated.validNec(Map.empty)
+                case (Pattern.ListPart.NamedList(name), _) =>
+                  Validated.validNec(Map(name -> expectedType))
+                case (Pattern.ListPart.Item(inner), idx) =>
+                  loop(itemType, inner, s"$at/list/item[$idx]")
+              }
+              mergeAll(at, partChecks)
+          }
+
+        case Pattern.Annotation(inner, annType) =>
```
johnynek: I think we also need to check that `expectedType =:= annType` or at least `expectedType <:< annType`. I suppose we could widen into annType here.

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -134,13 +243,856 @@ object TestUtils {
...
+                case _                              =>
+                  inferredArgs.lift(idx).getOrElse(expectedType)
+              }
+              loop(argType, innerPat, s"$at/arg[$idx]")
+            }
+            val mergedArgs = mergeAll(at, argChecks)
+
+            (arityCheck :: mergedArgs.void :: Nil)
+              .sequence_
+              .andThen(_ => mergedArgs)
```
johnynek: this should be `arityCheck *> mergedArgs`

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -134,13 +243,856 @@ object TestUtils {
...
+          constructorArgTypes(at, expectedType, pack, cons).andThen { inferredArgs =>
+            val arityCheck =
+              if (args.lengthCompare(inferredArgs.length) == 0) typeValidationPass
+              else
+                typeValidationFail(
+                  at,
+                  s"constructor arg arity mismatch for ${pack.asString}::${cons.sourceCodeRepr}: expected ${inferredArgs.length}, got ${args.length}"
+                )
+
+            val argChecks = args.toList.zipWithIndex.map { case (innerPat, idx) =>
```
johnynek: why not directly zip inferredArgs in. We are checking they are the same length. Then we could avoid the `inferredArgs.lift` below.

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -134,13 +243,856 @@ object TestUtils {
...
+              val argType = innerPat match {
+                case Pattern.Annotation(_, annType) => annType
+                case _                              =>
+                  inferredArgs.lift(idx).getOrElse(expectedType)
```
johnynek: if we do the `zip` we just use the type from the zip. I think expectedType is just wrong here. It is the scrutinee type, not a parameter type.

Also, we shouldn't lift the annotation pattern out, we should loop and just check the argType on innerPat.

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -134,13 +243,856 @@ object TestUtils {
...
+                  s"$at/union[$idx]",
+                  s"union branch binders differ: expected ${expectedNames.map(_.sourceCodeRepr).toList.sorted.mkString(", ")}, got ${m.keySet.map(_.sourceCodeRepr).toList.sorted.mkString(", ")}"
+                )
+            }
+            val mergedMaps =
+              maps.foldLeft(Validated.validNec(Map.empty): TypeValidation[Bindings]) {
+                case (acc, m) => acc.andThen(mergeBindings(at, _, m))
+              }
+            (sameNameChecks.sequence_ :: mergedMaps.void :: Nil)
+              .sequence_
```
johnynek: I think this is the same as:
```
sameNameChecks.sequence_ *> mergedMaps
```
isn't it?

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -134,13 +243,856 @@ object TestUtils {
...
+          case None =>
+            typeValidationFail(
+              path,
+              s"unknown global ${pack.asString}::${name.sourceCodeRepr} with type ${showType(tpe)}"
+            )
+        }
+
+      case TypedExpr.App(fn, args, _, _) =>
+        val childChecks = (
+          validateTypeConnections(
```
johnynek: add a comment about the directionality of `Connections` and why we want this directionality.

core/src/test/scala/dev/bosatsu/TestUtils.scala
```diff
@@ -1,17 +1,126 @@
...
 import dev.bosatsu.tool.Output
 import munit.Assertions.{assertEquals, fail}
+import Identifier.Constructor
 import IorMethods.IorExtension
 
 import cats.syntax.all._
 
 object TestUtils {
 
+  final case class TypeValidationError(path: String, message: String) {
```
johnynek: let's move all the TypeValidation code into it's own `object` at least:

```scala
object TypeValidation {

}
```

then we can make `Error` inside there rather than `TypeValidationError` and `type Result[+A] = ValidatedNec[Error, A]` no need to make it private I don't think. We may want to call the outer most function to validate it in other tests in the future.

PR #1759, opened 2026-02-21
Support polymorphic recursion loop lowering in Matchless

### LINKED ISSUE #1749: Support polymorphic recursion loop lowering in Matchless
Loop/Recur lowering in TypedExpr should stay monomorphic to preserve typed AST invariants.

Follow-up work:
- perform polymorphic-recursion loop/recur lowering after type erasure in Matchless
- add tests showing polymorphic recursion is optimized at Matchless stage while typed AST remains unchanged pre-erasure
- document interaction with grouped/eta-expanded recursion lowering paths

johnynek:
## Summary
- add a regression for polymorphic recursion (`box_more`) that asserts Matchless emits one `WhileExpr` and no self-recursive calls
- lower recursive lambdas to Matchless while loops when tail self-calls are present, even when typed loop/recur lowering skipped due type-changing recursion
- keep the existing typed-stage behavior unchanged; this optimization now happens post-erasure in Matchless

## Testing
- sbt "coreJVM/testOnly dev.bosatsu.MatchlessRegressionTest"
- sbt "coreJS/testOnly dev.bosatsu.MatchlessRegressionTest"
- sbt "coreJVM/testOnly dev.bosatsu.Issue1633Test"
- sbt "coreJVM/testOnly dev.bosatsu.TypedExprTest"

Closes #1749

### DIFF
--- a/core/src/main/scala/dev/bosatsu/Matchless.scala
+++ b/core/src/main/scala/dev/bosatsu/Matchless.scala
@@ -2132,16 +2132,21 @@ object Matchless {
         cond <- mut
         result <- mut
         args1 <- args.traverse(b => (anon, mut).mapN(ArgRecord(b, _, _)))
-        whileRes = toWhileBody(name, args1, cond, result, body)
-        (whileBody, rewroteTailCall) = whileRes
+        (whileBody, rewroteTailCall) = toWhileBody(
+          name,
+          args1,
+          cond,
+          result,
+          body
+        )
         allMuts = cond :: result :: args1.toList.map(_.loopVar)
       } yield {
         if (onlyIfTailCall && !rewroteTailCall) None
         else {
-          val initSets =
-            args1.toList.zip(initArgs.toList).map { case (argRec, initArg) =>
-              (argRec.loopVar, initArg)
-            }
+          val initSets = args1.iterator
+            .zip(initArgs.iterator)
+            .map { case (argRec, initArg) => (argRec.loopVar, initArg) }
+            .toList
           Some(
             letMutAll(
               allMuts,


core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -2111,41 +2111,68 @@ object Matchless {
...
+    ): F[Option[Expr[B]]] = {
       val mut = makeAnon.map(LocalAnonMut(_))
       val anon = makeAnon.map(LocalAnon(_))
       for {
         cond <- mut
         result <- mut
         args1 <- args.traverse(b => (anon, mut).mapN(ArgRecord(b, _, _)))
-        whileBody = toWhileBody(name, args1, cond, result, body)
+        whileRes = toWhileBody(name, args1, cond, result, body)
+        (whileBody, rewroteTailCall) = whileRes
```
johnynek: we shouldn't need this:

```scala
(whileBody, rewroteTailCall) = toWhileBody(name, args1, cond, result, body)
```
should work.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -2111,41 +2111,68 @@ object Matchless {
...
-          allMuts,
-          setAll(
-            initSets,
-            Always(
-              SetMut(cond, TrueExpr),
-              WhileExpr(isTrueExpr(cond), whileBody, result)
+        if (onlyIfTailCall && !rewroteTailCall) None
+        else {
+          val initSets =
+            args1.toList.zip(initArgs.toList).map { case (argRec, initArg) =>
```
johnynek: I think we can do:

`args1.iterator.zip(initArgs.iterator).map { .. }.toList` so we don't materialize more than one list.

Issue #1770, opened 2026-02-24
Design post-typechecking Recursion checker


johnynek:
Currently, we do recursion checking on the syntax AST (Declaration), not the lambda calculus (Expr) or the typed lambda calculus (TypedExpr), If we are going to improve recursion checking to allow recursion on Int (see #1760 ), we need the type information of the match scrutinee.

Write a design doc for changing recursion checking to run post-typechecking on TypedExpr.

Discuss how this can improve, or how it puts recursion checking at risk. If we had the type information, then #1760 could be cleaner syntax because we could inspect the type and when it is `Int` just use the int_decrease stategy.

Finally, we could imagine improving the recursion strategy we have now to expand our notion of "smaller" For instance, now we have if we have `enum Nat: Zero, Succ(prev: Nat)` if we are in case `Succ(p)` we can infer on `p`, but we can't recur on `Zero`. We could imagine lexicographically ordering the constructors. So rather than strictly use substructures, we could also use "lower" enum branches (enums that appear earlier in the list). I don't know how often that's useful, but we could potentially allow it and retain totality.

Lastly, discuss how we could move the current recursion checker into `test/` and then check the new implementation against the old one. It should agree for all WellTypedGen code.

johnynek:
MergeXO assigned an agent and started design work for issue #1770.

johnynek:
MergeXO assigned an agent and started design work for issue #1770.

johnynek:
MergeXO assigned an agent and started design work for issue #1770.

johnynek:
Opened recovered design PR: https://github.com/johnynek/bosatsu/pull/1773

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/1770#issuecomment-3948218256
- linked PR: #1773 (https://github.com/johnynek/bosatsu/pull/1773)

Comments on the source issue are no longer actioned after a PR exists. Please comment on the PR thread instead.

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #1770.

johnynek:
Opened implementation PR: https://github.com/johnynek/bosatsu/pull/1777

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/1770#issuecomment-3949150599
- linked PR: #1777 (https://github.com/johnynek/bosatsu/pull/1777)

[...]

PR #1773, opened 2026-02-24
Design doc for #1770: Design post-typechecking Recursion checker

johnynek:
Recovered design PR from a previously pushed branch.

Refs #1770

Source issue: https://github.com/johnynek/bosatsu/issues/1770

### DIFF
--- a/docs/design/1770-design-post-typechecking-recursion-checker.md
+++ b/docs/design/1770-design-post-typechecking-recursion-checker.md
@@ -47,7 +47,7 @@ Issue #1770 asks to move recursion checking post-typechecking on `TypedExpr`, di
 1. Make a typed recursion checker authoritative in the production pipeline.
 2. Preserve current behavior for structural and tuple-lexicographic recursion.
 3. Keep recursion diagnostics source-anchored and close to current messages.
-4. Add a typed strategy-selection hook so `Int` targets can map cleanly to `int_decrease` in #1760.
+4. Keep strategy semantics strictly structural in this PR and document the typed path needed for #1760 follow-up.
 5. Keep the current checker in test scope as a legacy oracle and verify parity on WellTypedGen programs.
 6. Define a safe future path for broader notions of smaller such as enum-constructor ordering.
 
@@ -72,16 +72,21 @@ Recursion checking happens before type information is available.
 2. Remove production use of `DefRecursionCheck` from `Package.inferBodyUnopt`.
 3. After successful `Infer.typeCheckLets`, run `TypedExprRecursionCheck` on typed lets.
 4. Map typed recursion failures through `PackageError.RecursionError`.
-5. Continue with existing lowering and normalization unchanged.
+5. Run `TypedExprLoopRecurLowering` and `TypedExprNormalization` only after recursion checking succeeds.
 
 Net effect: recursion validity moves to post-typechecking and pre-lowering.
+
+Ordering constraint (required): recursion checking must run before `TypedExprNormalization`.
+Reason:
+1. Normalization rewrites match and recursion shape, which can degrade source diagnostics if recursion checking runs afterwards.
+2. Some rewrites rely on trusted recursion/termination properties; running typed recursion checks first keeps that trust boundary explicit.
 
 ## Architecture and modules
 ### 1. `RecursionCheck` shared model
 Add `core/src/main/scala/dev/bosatsu/RecursionCheck.scala` with:
 1. Checker-neutral `Error` ADT equivalent to current recursion errors.
 2. Shared diagnostics text and region behavior.
-3. Strategy model for target components: `Structural`, `IntDecrease`, `EnumOrdinal` (future and gated).
+3. Structural decrease relation as the only active strategy in this PR, with extension points documented for future typed strategies.
 
 `PackageError.RecursionError` should depend on `RecursionCheck.Error`, not `DefRecursionCheck.RecursionError`.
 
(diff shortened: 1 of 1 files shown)


docs/design/1770-design-post-typechecking-recursion-checker.md
```diff
@@ -0,0 +1,180 @@
...
+### 2. `TypedExprRecursionCheck`
+Add `core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala`.
+
+Inputs:
+1. Package name.
+2. Full `TypeEnv[Kind.Arg]`.
+3. Typed lets list: `List[(Bindable, RecursionKind, TypedExpr[Declaration])]`.
+
+Output:
+1. `ValidatedNel[RecursionCheck.Error, Unit]`.
```
johnynek: let's use `ValidateNec` which has better asymptotics (O(1) combine vs worst case O(N)).

docs/design/1770-design-post-typechecking-recursion-checker.md
```diff
@@ -0,0 +1,180 @@
...
+
+The checker traverses recursive definitions in typed IR (top-level and nested recursive lets), mirroring current recursion invariants.
+
+### 3. Legacy checker relocation for parity
+Move current syntax checker logic to `core/src/test/scala/dev/bosatsu/LegacyDefRecursionCheck.scala` and keep it only for differential testing during migration.
+
+## Typed checker algorithm
+1. Find each recursive binding (`RecursionKind.Recursive`) and recover parameter groups from typed lambda shape, peeling `Generic` and `Annotation` wrappers.
```
johnynek: Expand on the `peeling`. We may need to know `Generic` and `Annotation` information in order to see the types of recur match scrutinees. It seems like we would need to keep track of which Var.Bound are universal vs existential as we recurse when we hit a generic.

docs/design/1770-design-post-typechecking-recursion-checker.md
```diff
@@ -0,0 +1,180 @@
...
+2. Track checker state analogous to legacy checker (`TopLevel`, `InDef`, `InDefRecurred`, `InRecurBranch`) so existing invariants are preserved.
+3. Detect recur-sites from typed matches by reading node tags: a typed `Match` is a recur-site when its tag is `Declaration.Match(RecursionKind.Recursive, target, ...)`.
+4. Resolve recur target names to parameter positions and reject invalid, duplicate, or local targets with equivalent error semantics.
+5. Select a strategy per recur target component from normalized target type: `Int` chooses `IntDecrease`; all other types choose `Structural` for parity baseline.
```
johnynek: for now, we should defer this I think. There are a lot of issues about how we could generalize it. Let's just assume Structural currently.

docs/design/1770-design-post-typechecking-recursion-checker.md
```diff
@@ -0,0 +1,180 @@
...
+3. `TotalityCheck` and unused-let checks run on source expressions.
+4. `Infer.typeCheckLets` produces typed lets (`TypedExpr[Declaration]`).
+5. Lowering and normalization run later.
+
+Recursion checking happens before type information is available.
+
+## Proposed pipeline
+1. Keep source conversion, totality, and unused-let checks as-is.
+2. Remove production use of `DefRecursionCheck` from `Package.inferBodyUnopt`.
+3. After successful `Infer.typeCheckLets`, run `TypedExprRecursionCheck` on typed lets.
```
johnynek, line 73: we should be very clear here: we must run this check before we run TypedExprNormalization, which does all kinds of rewrites on the AST. It will be bad for diagostics to run the test afterwards because it may greatly translate the graph. Moreover, the normalizations may not be sound if the recursions don't terminate, because the type signatures can't be trusted in that case.

docs/design/1770-design-post-typechecking-recursion-checker.md
johnynek: Updated the design doc output type to `ValidatedNec[RecursionCheck.Error, Unit]` (in the `TypedExprRecursionCheck` section) and aligned the plan around `Nec`-based accumulation.

docs/design/1770-design-post-typechecking-recursion-checker.md
johnynek: Expanded the typed-checker algorithm with an explicit peeling discipline for `TypedExpr.Generic` and `TypedExpr.Annotation`, including stack-scoped tracking of universal (`forall`) vs existential (`exists`) `Type.Var.Bound` entries while traversing recur regions.

docs/design/1770-design-post-typechecking-recursion-checker.md
johnynek: Adjusted the design to defer generalized strategy work for now: this PR is now explicitly structural-only. I updated architecture, algorithm, and acceptance criteria to state that `IntDecrease`/enum-ordinal strategies are follow-up work.

docs/design/1770-design-post-typechecking-recursion-checker.md
johnynek: Added an explicit required ordering constraint that typed recursion checking must run before `TypedExprNormalization`, with rationale on diagnostics fidelity and trust boundaries for downstream rewrites.

johnynek:
Addressed all four review comments by directly updating `docs/design/1770-design-post-typechecking-recursion-checker.md`.

Issue #1769, opened 2026-02-24
Design refactoring TotalityChecker to run on TypedExpr[Declaration]


johnynek:
The goal is to make more use of type information, including unihabited types, such as implemented in #1729 .

So, instead of just checking Lists or NonEmptyLists of patterns, we would get the scrutinee type as well as the fully typed patterns that are in the TypedExpr.Match which would be required to have the original attached `Declaration`.

Speak to the question of: can we simplify the TotalityCheck if we have fully typed patterns and types and can assume well typed patterns?

Does 1729 have enough information to identify unmatchable branches so we can allow missing but unmatchable branches?

Discuss how to handle testing. Currently we just generate random patterns, to test this we will need to generate well typed patterns. Review the code in WellTypedGen and see how we can use and improve that code so we can generate TypeEnvs and then Types from those envs, and finally sets of patterns to check with.

johnynek:
MergeXO assigned an agent and started design work for issue #1769.

johnynek:
MergeXO assigned an agent and started design work for issue #1769.

johnynek:
MergeXO assigned an agent and started design work for issue #1769.

johnynek:
MergeXO assigned an agent and started design work for issue #1769.

johnynek:
Opened recovered design PR: https://github.com/johnynek/bosatsu/pull/1774

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/1769#issuecomment-3948218410
- linked PR: #1774 (https://github.com/johnynek/bosatsu/pull/1774)

Comments on the source issue are no longer actioned after a PR exists. Please comment on the PR thread instead.

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #1769.

johnynek:
MergeXO implementation flow was blocked for issue #1769: Required pre-finalization checks could not be made fully green in this environment: `sbt "scalafmtAll"` fails because scalafmt v3.10.7 could not be downloaded, and full `sbt "coreJVM/test; cli/test"` / `sbt "cli/test"` runs hit unrelated 30s timeout failures (e.g. `Z3SolverIntegrationTest.design doc query A...`, `ClangGenTest.test_workspace/Ackermann.bosatsu`, `PythonGenTest.we can compile Nat.bosatsu`).

johnynek:
[...]

PR #1774, opened 2026-02-24
Design doc for #1769: Design refactoring TotalityChecker to run on TypedExpr[Declaration]

johnynek:
Recovered design PR from a previously pushed branch.

Refs #1769

Source issue: https://github.com/johnynek/bosatsu/issues/1769

### DIFF
--- a/docs/design/1769-design-refactoring-totalitychecker-to-run-on-typedexpr-declaration.md
+++ b/docs/design/1769-design-refactoring-totalitychecker-to-run-on-typedexpr-declaration.md
@@ -21,7 +21,7 @@ _Issue: #1769 (https://github.com/johnynek/bosatsu/issues/1769)_
 
 ## Summary
 
-Move totality checking from untyped Expr to TypedExpr so each match has scrutinee type information; use Inhabitedness (#1729) to conservatively prune impossible branches from missing-branch requirements and report definitely impossible branches as unreachable; update pipeline ordering and add typed generation-based tests via WellTypedGen.
+Move totality checking from untyped Expr to TypedExpr so each match has scrutinee type information; use Inhabitedness (#1729) to conservatively prune impossible branches from missing-branch requirements while preserving existing shadowing-based unreachable errors; update pipeline ordering and add typed generation-based tests via WellTypedGen.
 
 ---
 issue: 1769
@@ -40,7 +40,7 @@ depends_on:
 Issue #1769 asks us to use typed information, especially uninhabitedness data from #1729, so that:
 
 1. Totality checks can reason from the scrutinee type.
-2. Definitely unmatchable branches can be treated as unreachable.
+2. Definitely unmatchable branches can be excluded from required coverage.
 3. Missing branches that are themselves unmatchable can be excluded from required coverage.
 4. Testing can move from random pattern generation to well-typed generation.
 
@@ -56,10 +56,10 @@ Current behavior has three limitations:
 
 1. Run totality checking on `TypedExpr[Declaration]`.
 2. Use scrutinee type (`arg.getType`) for every match analysis.
-3. Use `Inhabitedness.checkMatch` to conservatively identify impossible branches.
+3. Use `Inhabitedness.checkMatch` to conservatively identify impossible branches for coverage pruning.
 4. Allow missing branches to be omitted when they are provably unmatchable.
 5. Keep conservative behavior for `Unknown` inhabitedness results.
-6. Preserve guard semantics (guarded branches do not count toward coverage).
+6. Preserve existing unreachable diagnostics for shadowed branches.
 7. Keep error regions and user-facing diagnostics stable.
 8. Add deterministic and property-based tests using well-typed generators.
 
@@ -87,6 +87,7 @@ Yes, for definitive cases.
 1. `Inhabitedness.check(scrutineeType, env)` identifies globally uninhabited scrutinee domains.
 2. `Inhabitedness.checkMatch(scrutineeType, pattern, env)` identifies branch patterns that are definitely `Uninhabited`.
 3. `Unknown` and validation failures must remain conservative and be treated as potentially matchable.
+4. For a definitively uninhabited scrutinee, totality treats missing coverage as vacuously satisfied, but does not rewrite/remove match branches in this pass.
 
 ## Proposed Architecture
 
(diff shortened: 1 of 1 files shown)


docs/design/1769-design-refactoring-totalitychecker-to-run-on-typedexpr-declaration.md
```diff
@@ -0,0 +1,235 @@
...
+
+1. Constructor existence and constructor arity are already typechecker responsibilities for typed programs.
+2. Totality can focus on coverage and reachability over well-typed patterns.
+3. Keep only structural pattern lint that is not guaranteed by typing today (for example adjacent string wildcards and list splice constraints).
+
+### Does #1729 have enough information to identify unmatchable branches so we can allow missing but unmatchable branches?
+
+Yes, for definitive cases.
+
+1. `Inhabitedness.check(scrutineeType, env)` identifies globally uninhabited scrutinee domains.
```
johnynek, line 87: what would we do when a scrutinee is uninhabited? Just pass the check? We can't remove all the branches of the pattern and form a Match node. Be clear what we could do here.

docs/design/1769-design-refactoring-totalitychecker-to-run-on-typedexpr-declaration.md
```diff
@@ -0,0 +1,235 @@
...
+4. Post-filter missing patterns: remove any missing pattern `p` where `checkMatch(scrutineeType, p, env)` is definitively `Uninhabited`.
+
+This directly addresses issue #1769’s requirement to allow omitted branches that cannot match.
+
+### 4) Unreachable-branch computation includes impossible branches
+
+A branch is unreachable if either condition holds:
+
+1. Its pattern is definitely impossible for the scrutinee type.
+2. It is fully shadowed by prior unguarded, potentially matchable coverage.
```
johnynek: if users write a set of patterns where a lower pattern can't match, that is currently an error. For instance:

```
match foo:
    case _: always_runs(foo)
    case Bar(b): unreachable(b)
```
In this example, the second case is unreachable but we shouldn't just silently prune it. That is likely a programmer error.

I think we should still error on unreachable branches due to shadowing of previous patterns, and only after we have set of unshadowed patterns do we filter out the uninhabited patterns.

We should allow, but not require, users to write uninhabited patterns, but optimize the code by removing them from the set of patterns.

docs/design/1769-design-refactoring-totalitychecker-to-run-on-typedexpr-declaration.md
```diff
@@ -0,0 +1,235 @@
...
+
+Only unguarded, potentially matchable branches should extend the running covered set.
+
+### 5) Validation split
+
+Keep in totality:
+
+1. Structural string/list pattern lint currently modeled as `InvalidStrPat` and `MultipleSplicesInPattern`.
+
+De-emphasize in typed path:
```
johnynek: what do you mean, de-emphasize? Would we just sys.error on invariant violation since we assume type-checked TypedExpr? Or maybe return `InternalError` as an error kind and just have some string and point to the line number?

docs/design/1769-design-refactoring-totalitychecker-to-run-on-typedexpr-declaration.md
```diff
@@ -0,0 +1,235 @@
...
+
+1. Update `core/src/main/scala/dev/bosatsu/TotalityCheck.scala`.
+2. Change traversal and error payloads from `Expr` to `TypedExpr`.
+3. Add helper that converts `Inhabitedness.Result[State]` to conservative state (`Uninhabited` is trusted, everything else treated as potentially matchable).
+4. Integrate per-branch inhabitedness classification into missing/unreachable logic.
+5. Keep existing set-ops methods (`difference`, `intersection`, `missingBranches`) unchanged.
+6. Update `core/src/main/scala/dev/bosatsu/Package.scala` to run totality after inference.
+7. Update `core/src/main/scala/dev/bosatsu/PackageError.scala` typing and rendering paths to consume typed match payloads.
+8. Add a new deterministic typed-totality suite in `core/src/test/scala/dev/bosatsu/TypedTotalityTest.scala`.
+9. Extend `core/src/test/scala/dev/bosatsu/WellTypedGen.scala` with helpers that generate totality-focused typed inputs: type env, scrutinee type, and well-typed pattern sets.
```
johnynek, line 167: we should propose explicit laws we want to check. For instance: if we generate a well-typed total pattern set:

1. it should be accepted by the totality checker
2. removing any pattern from the set should cause it to fail.
3. if there are any unions, we should be able to flatten the unions and get the same result (being careful to handle inner unions of constructors or lists carefully).

docs/design/1769-design-refactoring-totalitychecker-to-run-on-typedexpr-declaration.md
```diff
@@ -0,0 +1,235 @@
...
+10. Add/adjust property tests in `core/src/test/scala/dev/bosatsu/TotalityTest.scala` to consume well-typed generated cases for the new behavior.
+11. Update `core/src/test/scala/dev/bosatsu/ErrorMessageTest.scala` to assert diagnostics remain clear and region-accurate.
+
+## Testing Strategy
+
+### Deterministic tests
+
+1. `Option[Never]` style case where a constructor branch is impossible: missing branch should not be required.
```
johnynek, line 175: also `enum Result[e, a]: Err(err: e), Ok(ok: a)` then we should be able to do: `Ok(x) = definitely_good` if `definitely_good: Result[Never, a]`

docs/design/1769-design-refactoring-totalitychecker-to-run-on-typedexpr-declaration.md
johnynek: Clarified this explicitly in the design doc. In the "Does #1729 have enough information..." section I added that when the scrutinee is definitively uninhabited, totality treats missing coverage as vacuously satisfied, but this pass does not rewrite/remove branches or eliminate the `Match` node.

docs/design/1769-design-refactoring-totalitychecker-to-run-on-typedexpr-declaration.md
johnynek: Updated the proposed algorithm to match this ordering. The doc now says we first compute/report shadowing-based unreachable branches exactly as today, then build the unshadowed coverage set, and only then filter definitively uninhabited patterns for missing-coverage computation. I also clarified that uninhabited branches are allowed (not required) and may be removed by later optimization, but are not emitted as unreachable errors by themselves.

docs/design/1769-design-refactoring-totalitychecker-to-run-on-typedexpr-declaration.md
johnynek: Agreed and made this explicit. I replaced "de-emphasize" with a concrete invariant policy: in typed totality, unknown constructor/arity mismatches are treated as compiler-invariant failures, should not call `sys.error`, and should be surfaced via an explicit internal-error variant (e.g. `TotalityCheck.InternalInvariantViolation`) with region/message for reporting.

docs/design/1769-design-refactoring-totalitychecker-to-run-on-typedexpr-declaration.md
johnynek: Added explicit property laws in the Testing Strategy section. It now includes: total-set acceptance, minimality (for irredundant generated total sets), union-flattening equivalence, missing-branch pruning, and `Unknown` conservatism.

docs/design/1769-design-refactoring-totalitychecker-to-run-on-typedexpr-declaration.md
johnynek: Added this as an explicit deterministic test case. The doc now includes `Result[e, a]` with scrutinee `Result[Never, a]` where `Ok(x)` alone is accepted as total coverage (including the `Ok(x) = definitely_good` style example).

johnynek:
Updated the design doc to incorporate all review feedback: explicit uninhabited-scrutinee semantics without AST rewriting, shadowing-first unreachable behavior, concrete invariant-error handling, explicit property laws, and the `Result[Never, a]` deterministic case.

PR #1777, opened 2026-02-24
Move recursion checking post-typechecking onto TypedExpr with legacy parity coverage

### LINKED ISSUE #1770: Design post-typechecking Recursion checker

johnynek:
Implemented issue #1770 per the merged design:
- Added shared recursion model in `core/src/main/scala/dev/bosatsu/RecursionCheck.scala` (error ADT, shared messages, lexicographic helper).
- Added new typed checker in `core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala` and made it authoritative.
  - Runs on typed lets post-inference.
  - Preserves structural + tuple-lexicographic behavior.
  - Detects recur sites from typed `Match` tags with additional disambiguation so leaked recursive tags on synthetic wrapper matches are ignored.
  - Handles wrapper-peeling (`Generic`/`Annotation`) and same-package `Global` references for rename diagnostics/parity.
  - Preserves continuation/lambda recursion behavior by handling wrapped lambda arguments in reachable-call contexts.
- Switched production pipeline in `core/src/main/scala/dev/bosatsu/Package.scala`:
  - removed production use of syntax `DefRecursionCheck`.
  - after `Infer.typeCheckLets`, runs `TypedExprRecursionCheck.checkLets(...)` (with top-level def arg patterns from source statements).
  - keeps recursion check before lowering/normalization.
- Updated `PackageError.RecursionError` to use `RecursionCheck.Error` (`core/src/main/scala/dev/bosatsu/PackageError.scala`).
- Moved legacy checker to test scope as `core/src/test/scala/dev/bosatsu/LegacyDefRecursionCheck.scala`.
- Replaced production `DefRecursionCheck` with compatibility aliases (`core/src/main/scala/dev/bosatsu/DefRecursionCheck.scala`).
- Updated legacy test wiring (`core/src/test/scala/dev/bosatsu/DefRecursionCheckTest.scala`).
- Added typed checker tests (`core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala`), including regressions for wildcard def-arg patterns and continuation-style recursion.
- Added legacy-vs-typed property parity test over WellTypedGen phase1..phase4 (`core/src/test/scala/dev/bosatsu/TypedExprRecursionParityTest.scala`).
- Updated error-message fixtures impacted by new ordering (`core/src/test/scala/dev/bosatsu/ErrorMessageTest.scala`) to remain type-correct while asserting intended recursion diagnostics.
- Updated user docs to note new ordering (`docs/src/main/paradox/recursion.md`).
- Updated stale SourceConverter comments to reflect post-typechecking recursion checks (`core/src/main/scala/dev/bosatsu/SourceConverter.scala`).

Validation run:
- `sbt --batch "coreJVM/testOnly dev.bosatsu.TypedExprRecursionCheckTest"`
- `sbt --batch "coreJVM/testOnly dev.bosatsu.ErrorMessageTest"`
- `sbt --batch "coreJVM/testOnly dev.bosatsu.EvaluationTest"`
- `sbt --batch "coreJVM/test; cli/test"`
- `sbt --batch "doc; paradox"`
All passed.

Formatting note:
- Attempted `sbt --batch "scalafmtAll"`, but scalafmt failed in this environment with `failed to download .../.scalafmt.conf` (tooling fetch issue).

Fixes #1770

Implements design doc: [docs/design/1770-design-post-typechecking-recursion-checker.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/1770-design-post-typechecking-recursion-checker.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/1773

Source issue: https://github.com/johnynek/bosatsu/issues/1770


core/src/main/scala/dev/bosatsu/DefRecursionCheck.scala
```diff
@@ -1,901 +1,14 @@
...
+import cats.data.{ValidatedNec, ValidatedNel}
 
-import cats.implicits._
-
-import Identifier.Bindable
-
-/** Recursion in bosatsu is only allowed on a substructural match of one of the
-  * parameters to the def. This strict rule, along with strictly finite data,
-  * ensures that all recursion terminates
+/** Legacy syntax recursion checker moved to test scope.
```
johnynek: let's remove this file entirely. Keep the legacy code in test only, but we don't need to leave this tombstone marker here.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -0,0 +1,1062 @@
...
+
+import cats.data.{NonEmptyList, StateT, Validated, ValidatedNec, ValidatedNel}
+import cats.implicits._
+import dev.bosatsu.rankn.{Type, TypeEnv}
+
+import Identifier.Bindable
+
+object TypedExprRecursionCheck {
+
+  type Res = ValidatedNel[RecursionCheck.Error, Unit]
```
johnynek: let's use `ValidatedNec` which has better asymptotics. It seems like we are using Nec below but here we have a type definition we aren't using much or at all. Unify and use it, or just remove the type definition.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -0,0 +1,1062 @@
...
+              }
+          }
+      }
+
+    private def checkExprV(
+        currentPackage: PackageName,
+        state: State,
+        expr: TypedExpr[Declaration]
+    ): Res =
+      Validated.fromEither(
```
johnynek, line 963: Add a comment why we use Either in checkExpr. It is noted below on the Parallel instance we are using to accumulate errors, but not here. Every use of the Either should include a brief reminder that some portions are sequential and as such can't use Validated.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
```diff
@@ -0,0 +1,157 @@
...
+  ): String = {
+    val sm = Map(pack -> (LocationMap(source), "<test>"))
+    errs.toList
+      .map(_.message(sm, LocationMap.Colorize.None))
+      .mkString("\n-----\n")
+  }
+
+  private def recursionErrorsOf(
+      source: String
+  ): Either[cats.data.NonEmptyList[PackageError], Unit] = {
```
johnynek: don't use fully qualified names. Import them at the top level. Also add this note to the coding_style.md so we don't make this mistake in the future. Along with "use cats.data.ValidatedNec over cats.data.ValidatedNel for better asymptotic performance.".

core/src/test/scala/dev/bosatsu/TypedExprRecursionParityTest.scala
```diff
@@ -0,0 +1,81 @@
...
+          val sourceMap = Map(packageName -> (LocationMap(source), "<generated>"))
+          val msg = errs.toList
+            .map(_.message(sourceMap, LocationMap.Colorize.None))
+            .mkString("\n-----\n")
+          fail(s"typed pipeline failed for non-recursion reason:\n$msg")
+        }
+    }
+  }
+
+  private def assertParity(cfg: WellTypedGen.Config) =
```
johnynek, line 54: add TODO that right now WellTypedGen probably doesn't generate recursive functions, so I think this is never exercising positive cases of allowed recursion. We need to (in a later PR, not this one) improve WellTypedGen to generate allowed recursions.

core/src/main/scala/dev/bosatsu/DefRecursionCheck.scala
johnynek: Removed `core/src/main/scala/dev/bosatsu/DefRecursionCheck.scala` entirely. Legacy recursion checking remains test-only in `core/src/test/scala/dev/bosatsu/LegacyDefRecursionCheck.scala`, and production no longer has a `DefRecursionCheck` marker file.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Standardized `TypedExprRecursionCheck` on `ValidatedNec`. `Res` is now `ValidatedNec`, `ValidatedNel`/`toValidNel`/`invalidNel` usages were replaced, and the internal sequential state error type now uses `NonEmptyChain` (`Either[NonEmptyChain[RecursionCheck.Error], A]`) to stay consistent.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Added explicit comments at each `Validated.fromEither(...)` boundary (`checkExprV` and the nested-def boundary in `checkDef`) explaining why the checker uses `Either` in `StateT` for sequential state transitions, then converts back to `ValidatedNec` at API boundaries.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
johnynek: Removed fully qualified cats names in `TypedExprRecursionCheckTest` by importing at the top (`cats.Show`, `cats.data.{NonEmptyList, Validated}`). Also updated `coding_style.md` with both requested notes: avoid fully qualified names in normal code/tests, and prefer `cats.data.ValidatedNec` over `ValidatedNel`.

core/src/test/scala/dev/bosatsu/TypedExprRecursionParityTest.scala
johnynek: Added the requested TODO in `TypedExprRecursionParityTest` above `assertParity`, noting that WellTypedGen likely under-generates accepted recursive functions and should be improved in a follow-up PR.

johnynek:
Implemented all requested review changes. Validation run: `sbt --batch "coreJVM/testOnly dev.bosatsu.TypedExprRecursionCheckTest"`, `sbt --batch "coreJVM/testOnly dev.bosatsu.TypedExprRecursionParityTest"`, `sbt --batch "coreJVM/testOnly dev.bosatsu.DefRecursionCheckTest"`, and `sbt --batch "coreJVM/test; cli/test"` all pass. Formatting check attempt `sbt --batch "scalafmtAll"` still fails in this environment with `scalafmt: [v3.10.7] failed to download .../.scalafmt.conf`.

Issue #1778, opened 2026-02-24
Design changes to the code to enable coverage checking


johnynek:
Currently, when we serialize TypedExpr, we discard the tag that holds the information about where the expression came from. This means that after we compile we have no idea where each expression came from, and when running tests, no way to determine what part of the source they correspond to.

Moreover, Matchless.Expr also does not include any Region information.

After considering #504 and #1571 and in context of #1749 here is where I have landed:

We want a design where the Package knows the blake3 hash of the source file it came from. When should change the protobuf to serialize the Region information on the TypedExpr for the full implementation code. For interfaces the region and the hash is unset. The motivation is that we want the interface to be the *bare minimum* of what we need to compile against, while the the implementation is likely to change any time the any thing about the code has changed.

For caching, we want to be able to use the interface hash as a cache key, so we want to keep everything out that we don't absolutely need.

Lastly, Matchless expressions should retain a tuple of (hash, region) on the expressions but not the types. In context of #1749 keeping types on matchless rules out certain optimizations. The whole point of matchless is a type erased AST that allows easy and efficient implementation of polymorphic code. For tooling that wants to find the types for the matchless data, the tool would need to join the hash + region information against the implementation packages.

So, to summarize:

1. TypedAst.proto TypedExpr needs to add optional Region information. It is unset in interfaces (and old serialized libraries).
2. Package in TypedAst.proto needs to include a hash. We should use the full ident serialization of our HashValue "blake3:...." so we can potentially change hashes in the future. This hash should appear *one* time in the package proto.
3. Matchless expressions get a `sourceInfo` or similarly named value that has both package hash and region. The motivation for carrying the package hash on each matchless value is that we want to enable inlining across packages at the matchless level (#1315 ) and so potentially functions from one package would be inlined in another package, so region alone isn't enough. When we implement matchless serialization to proto (#629 ) we should use the string-table approach we currently use for package names and strings, etc... but we do not plan to implement matchless serialization for this design.

We need have some testing that interface files and interface libraries (proto files with no implementations) must not have any source hash or region information, otherwise we would destroy the ability to use interfaces as cache keys: interfaces as cache-keys is a main use case (they need to be normalized, if they are semantically equal they need to be byte-for-byte identical).

cc @snoble

johnynek:
MergeXO assigned an agent and started design work for issue #1778.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/1779

[...]

PR #1779, opened 2026-02-24
Design doc for #1778: Issue #1778 Design: Source Hash and Region Metadata for Coverage Checking

johnynek:
Design doc for issue #1778.

Refs #1778

Source issue: https://github.com/johnynek/bosatsu/issues/1778

### DIFF
--- a/docs/design/1778-design-changes-to-the-code-to-enable-coverage-checking.md
+++ b/docs/design/1778-design-changes-to-the-code-to-enable-coverage-checking.md
@@ -149,10 +149,11 @@ This avoids losing hash when reading a `.bosatsu_package` and later reassembling
 Add source metadata to Matchless expression nodes.
 
 1. Introduce `Matchless.SourceInfo(packageHashIdent: String, region: Region)`.
-2. Extend Matchless expression/boolean nodes to carry optional `sourceInfo`.
+2. Extend Matchless expression/boolean nodes to carry required `sourceInfo` (no `Option` in the AST).
 3. During lowering from `TypedExpr`, attach source info from current package hash + current expression region.
-4. For synthetic helper nodes created during lowering/optimization, inherit the nearest dominating source info.
-5. When inlining across packages (#1315), retain original source info on moved nodes.
+4. If provenance is unavailable (for example legacy decoded artifacts), use a sentinel `SourceInfo(emptyHashIdent, Region(0, 0))`, where `emptyHashIdent` is the hash ident for empty source bytes.
+5. For synthetic helper nodes created during lowering/optimization, inherit the nearest dominating source info before falling back to the sentinel.
+6. When inlining across packages (#1315), retain original source info on moved nodes.
 
 Important constraint:
 
@@ -166,7 +167,8 @@ Interfaces must stay minimal and deterministic.
 1. New fields are only in `Package` and `TypedExpr`; `Interface` schema is unchanged.
 2. `interfaceToProto` must continue to serialize only type/export surface.
 3. Tests must assert semantically equivalent interfaces remain byte-identical even when implementation source hash/regions differ.
-4. Interface-only libraries (`internal_packages = []`) must contain no source hash/region metadata.
+4. Add a ScalaCheck property asserting that for any fully compiled package, permuting any sequence-typed implementation data (for example bindings and type-environment entry order) does not change the serialized interface bytes or their hash.
+5. Interface-only libraries (`internal_packages = []`) must contain no source hash/region metadata.
 
 ## Implementation Plan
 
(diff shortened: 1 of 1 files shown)


docs/design/1778-design-changes-to-the-code-to-enable-coverage-checking.md
```diff
@@ -0,0 +1,256 @@
...
+1. Round-trip `TypedExpr` with present regions preserves regions.
+2. Decoding legacy `TypedExpr` blobs with missing `region` still succeeds.
+3. Round-trip package with `sourceHash` preserves full ident string.
+4. Decoding legacy packages with missing `sourceHash` still succeeds.
+5. Re-encoding decoded packages preserves hash/region presence semantics.
+
+### Interface Determinism Tests
+
+1. Interface serialization does not include source hash/region fields.
+2. Two compilations with identical public API but changed implementation body produce byte-identical interface protobuf bytes.
```
johnynek, line 201: we need scalacheck versions of this: forall fully compiled packages, any permutation of the order of any sequence data in the fully compiled packages does not change the hash of the protobuf serialized interface (so changing the order of the binding names, or the typeenv items, etc...)

docs/design/1778-design-changes-to-the-code-to-enable-coverage-checking.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Acceptance Criteria
+
+1. `TypedAst.proto` includes optional typed-expression region and optional package source hash fields with new field numbers.
+2. `ProtoConverter` writes `TypedExpr.region` for implementation expressions when available.
+3. `ProtoConverter` writes `Package.sourceHash` exactly once per implementation package when available.
+4. Legacy compiled artifacts without new fields decode successfully.
+5. Interfaces remain schema-minimal and do not carry source hash or region metadata.
+6. Interface protobuf output remains deterministic and suitable for cache keys.
+7. Matchless nodes can carry `(packageHash, region)` source info.
```
johnynek: I would say they must carry this. We should use dummy values for missing data (like the empty region (0,0) and maybe the hash of the empty string as the package hash), but soon we would expect all of those values to be legitimate, so I don't want to carry around "missingness" from the Matchless Expr AST.

docs/design/1778-design-changes-to-the-code-to-enable-coverage-checking.md
```diff
@@ -0,0 +1,256 @@
...
+11. Tests explicitly cover “interface-only libraries contain no source metadata”.
+12. Existing compilation/evaluation/codegen test suites continue to pass.
+
+## Risks and Mitigations
+
+1. Risk: Matchless API churn is large and touches many pattern matches.
+Mitigation: perform refactor mechanically in one pass; add compatibility helpers where possible; run full Matchless regression suites.
+
+2. Risk: Memory overhead from per-node source info in Matchless.
+Mitigation: keep source info optional; reuse package-hash strings; avoid attaching metadata in interface-only paths.
```
johnynek: since we aren't copying this information, it should only be a pointer in memory to data that is already built. Also, the size of regions is small, and the number of hashes in scope should also be small.

docs/design/1778-design-changes-to-the-code-to-enable-coverage-checking.md
johnynek: Updated the design doc in `Interface Determinism and Cache-Key Invariant`, `Interface Determinism Tests`, and `Acceptance Criteria` to require a ScalaCheck property over fully compiled packages: permutations of sequence-typed implementation data (including binding/type-env ordering) must preserve serialized interface protobuf bytes and hash.

docs/design/1778-design-changes-to-the-code-to-enable-coverage-checking.md
johnynek: Updated the Matchless design to make provenance required instead of optional. In `Matchless Source Info Model` (and corresponding plan/tests/acceptance criteria), expression/boolean nodes now always carry `sourceInfo` (no `Option`), with explicit sentinel fallback for missing legacy provenance: `Region(0, 0)` plus `emptyHashIdent` (hash ident of empty source bytes).

docs/design/1778-design-changes-to-the-code-to-enable-coverage-checking.md
johnynek: Revised the memory-risk mitigation text to match this guidance: source metadata should reuse already-built values (reference/pointer sharing for package hashes), avoid per-node hash copying, and rely on small region payload size with a small number of hashes in scope.

johnynek:
Addressed all review feedback directly in `docs/design/1778-design-changes-to-the-code-to-enable-coverage-checking.md`: strengthened deterministic interface hashing requirements with ScalaCheck permutation invariance, removed Matchless source-info missingness in favor of required provenance plus sentinel defaults, and clarified memory behavior around shared metadata references.

PR #1786, opened 2026-02-25
Hoist loop-invariant leading lets out of canonical Matchless recursion loops

### LINKED ISSUE #1738: Hoist loop-invariant lets out of Loop bodies
## Summary
Follow-up optimization from #1728: hoist loop-invariant bindings that are computed inside a `Loop` body to outside the loop when safe.

## Motivation
Current loop normalization can lift invariant **loop arguments**, but does not hoist invariant `let` bindings that appear in the loop body.

Example pattern:

```bosatsu
loop(x = a, y = b):
  z = f(100)
  if x > 0:
    recur(x - 1, y + z)
  else:
    (0, y)
```

`z = f(100)` is loop-invariant and could be hoisted above the loop.

## Scope
- detect loop-body lets that are invariant across all recur paths
- hoist eligible bindings above the `Loop` while preserving semantics
- add normalization tests for positive and negative cases (effects/escaping refs/shadowing)

Related: #1728

johnynek:
Implemented issue #1738 in Matchless and tests.

Code changes:
- Added `hoistInvariantLoopLets` in `core/src/main/scala/dev/bosatsu/Matchless.scala`.
- Updated `Matchless.fromLet` to run `hoistInvariantLoopLets` before `reuseConstructors`.
- The new pass is intentionally conservative and only rewrites canonical recursion-loop scaffolding:
  `Always(SetMut(condMut, TrueExpr), WhileExpr(isTrueExpr(condMut), effectExpr, resultMut))`.
- Inside eligible loop bodies, it inspects only the contiguous leading immutable `Let` prefix and hoists bindings only when all of the following hold:
  - binding is used in residual loop body
  - value is pure (`!hasSideEffect`)
  - value has no mutable reads (`LocalAnonMut` barrier)
  - value contains no nested `WhileExpr`
  - value passes profitability (`!triviallyCheap` and `exprWeight >= 4`)
  - defensive name-shadowing guard for `Right(name)` binders against loop condition references
- Hoisted bindings preserve original ordering so dependency chains remain correct.

Test changes (`core/src/test/scala/dev/bosatsu/MatchlessTests.scala`):
- Added direct pass tests for:
  - positive hoist of a top-level invariant loop-body let
  - positive hoist of dependent leading lets in order
  - negative: cheap alias / mutable read / side-effectful value
  - negative: lambda value capturing mutable state
  - negative: let inside conditional branch
  - negative: non-canonical loop condition not rewritten
- Added integration test proving `Matchless.fromLet` hoists invariant let computation out of `WhileExpr.effectExpr`.

Validation:
- Re-ran required checks from `coding_style.md` on final code:
- Both passed.

Fixes #1738

Implements design doc: [docs/design/1738-hoist-loop-invariant-lets-out-of-loop-bodies.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/1738-hoist-loop-invariant-lets-out-of-loop-bodies.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/1785

### DIFF
--- a/core/src/main/scala/dev/bosatsu/Matchless.scala
+++ b/core/src/main/scala/dev/bosatsu/Matchless.scala
@@ -193,6 +193,252 @@ object Matchless {
       def compare(left: Expr[A], right: Expr[A]): Int =
         compareExpr(left, right)
     }
+
+    def containsWhileExpr[A](expr: Expr[A]): Boolean = {
+      def loopExpr(e: Expr[A]): Boolean =
+        e match {
+          case WhileExpr(_, _, _) =>
+            true
+          case Lambda(captures, _, _, body) =>
+            captures.exists(loopExpr) || loopExpr(body)
+          case App(fn, args) =>
+            loopExpr(fn) || args.exists(loopExpr)
+          case Let(_, value, in) =>
+            loopExpr(value) || loopExpr(in)
+          case LetMut(_, in) =>
+            loopExpr(in)
+          case If(cond, thenExpr, elseExpr) =>
+            loopBool(cond) || loopExpr(thenExpr) || loopExpr(elseExpr)
+          case Always(cond, thenExpr) =>
+            loopBool(cond) || loopExpr(thenExpr)
+          case PrevNat(of) =>
+            loopExpr(of)
+          case ge: GetEnumElement[?] =>
+            loopExpr(ge.arg)
+          case gs: GetStructElement[?] =>
+            loopExpr(gs.arg)
+          case Local(_) | Global(_, _, _) | ClosureSlot(_) | LocalAnon(_) |
+              LocalAnonMut(_) | Literal(_) | MakeEnum(_, _, _) | MakeStruct(_) |
+              SuccNat | ZeroNat =>
+            false
+        }
+
+      def loopBool(b: BoolExpr[A]): Boolean =
+        b match {
+          case EqualsLit(expr, _) =>
+            loopExpr(expr)
+          case EqualsNat(expr, _) =>
+            loopExpr(expr)
+          case And(left, right) =>
+            loopBool(left) || loopBool(right)
+          case CheckVariant(expr, _, _, _) =>
+            loopExpr(expr)
+          case SetMut(_, value) =>
+            loopExpr(value)
+          case LetBool(_, value, in) =>
+            loopExpr(value) || loopBool(in)
+          case LetMutBool(_, in) =>
+            loopBool(in)
+          case TrueConst =>
+            false
+        }
+
+      loopExpr(expr)
+    }
+
+    def readsMutable[A](expr: Expr[A]): Boolean = {
+      def loopExpr(e: Expr[A]): Boolean =
+        e match {
+          case LocalAnonMut(_) =>
+            true
+          case Lambda(captures, _, _, body) =>
+            captures.exists(loopExpr) || loopExpr(body)
+          case WhileExpr(cond, effectExpr, _) =>
+            loopBool(cond) || loopExpr(effectExpr)
+          case App(fn, args) =>
+            loopExpr(fn) || args.exists(loopExpr)
+          case Let(_, value, in) =>
+            loopExpr(value) || loopExpr(in)
+          case LetMut(_, in) =>
+            loopExpr(in)
+          case If(cond, thenExpr, elseExpr) =>
+            loopBool(cond) || loopExpr(thenExpr) || loopExpr(elseExpr)
+          case Always(cond, thenExpr) =>
+            loopBool(cond) || loopExpr(thenExpr)
+          case PrevNat(of) =>
+            loopExpr(of)
+          case ge: GetEnumElement[?] =>
+            loopExpr(ge.arg)
+          case gs: GetStructElement[?] =>
+            loopExpr(gs.arg)
+          case Local(_) | Global(_, _, _) | ClosureSlot(_) | LocalAnon(_) |
+              Literal(_) | MakeEnum(_, _, _) | MakeStruct(_) | SuccNat |
+              ZeroNat =>
+            false
+        }
+
+      def loopBool(b: BoolExpr[A]): Boolean =
+        b match {
+          case EqualsLit(expr, _) =>
+            loopExpr(expr)
+          case EqualsNat(expr, _) =>
+            loopExpr(expr)
+          case And(left, right) =>
+            loopBool(left) || loopBool(right)
+          case CheckVariant(expr, _, _, _) =>
+            loopExpr(expr)
+          case SetMut(_, value) =>
+            loopExpr(value)
+          case LetBool(_, value, in) =>
+            loopExpr(value) || loopBool(in)
+          case LetMutBool(_, in) =>
+            loopBool(in)
+          case TrueConst =>
+            false
+        }
+
+      loopExpr(expr)
+    }
+
+    def referencesBindable[A](expr: Expr[A], target: Bindable): Boolean =
+      referencesBindable(expr, target, isShadowed = false)
+
+    private[Matchless] def referencesBindable[A](
+        expr: Expr[A],
+        target: Bindable,
+        isShadowed: Boolean
+    ): Boolean =
+      if (isShadowed) false
+      else
+        expr match {
+          case Local(name) =>
+            name == target
+          case Lambda(captures, recName, args, body) =>
+            captures.exists(referencesBindable(_, target, isShadowed)) || {
+              val bodyShadowed =
+                isShadowed || recName.contains(target) || args.exists(
+                  _ == target
+                )
+              referencesBindable(body, target, bodyShadowed)
+            }
+          case WhileExpr(cond, effectExpr, _) =>
+            BoolExpr.referencesBindable(cond, target, isShadowed) ||
+              referencesBindable(effectExpr, target, isShadowed)
+          case App(fn, args) =>
+            referencesBindable(fn, target, isShadowed) ||
+              args.exists(referencesBindable(_, target, isShadowed))
+          case Let(arg, value, in) =>
+            referencesBindable(value, target, isShadowed) || {
+              val inShadowed =
+                arg match {
+                  case Right(name) if name == target => true
+                  case _                             => isShadowed
+                }
+              referencesBindable(in, target, inShadowed)
+            }
+          case LetMut(_, in) =>
+            referencesBindable(in, target, isShadowed)
+          case If(cond, thenExpr, elseExpr) =>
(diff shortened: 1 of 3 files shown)


core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -1085,6 +1085,466 @@ object Matchless {
     StackSafe.onStackOverflow(recurExpr(expr, init)._1)(expr)
   }
 
+  private val HoistInvariantLoopLetsMinWeight: Int = 4
+
+  // Hoist loop-invariant leading lets out of canonical recursion loops.
+  private[bosatsu] def hoistInvariantLoopLets[A](expr: Expr[A]): Expr[A] = {
+    def letBinds(
```
johnynek: move this to a method on `object Let` and double check for duplicates. I think we have written this function before.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -1085,6 +1085,466 @@ object Matchless {
...
+            loopBool(in)
+          case TrueConst =>
+            false
+        }
+
+      loopExpr(ex)
+    }
+
+    def referencesBindable(expr: Expr[A], target: Bindable): Boolean = {
+      def loopExpr(e: Expr[A], isShadowed: Boolean): Boolean =
```
johnynek: if `isShadowed == true` we can return false right? We can never return true when that is true right? Can't we stop the recursion at that point?

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -1085,6 +1085,466 @@ object Matchless {
...
+    }
+
+    def referencesBindableInBool(
+        cond: BoolExpr[A],
+        target: Bindable
+    ): Boolean =
+      referencesBindable(Always(cond, UnitExpr), target)
+
+    def referencesLocalAnon(expr: Expr[A], target: Long): Boolean = {
+      def loopExpr(e: Expr[A], isShadowed: Boolean): Boolean =
```
johnynek: same comment about shadowing, can't we check `if (isShadowed)` and return false?

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -1085,6 +1085,466 @@ object Matchless {
...
+          case LetMutBool(_, in) =>
+            1 + loopBool(in)
+          case TrueConst =>
+            1
+        }
+
+      loopExpr(expr)
+    }
+
+    def containsWhileExpr(ex: Expr[A]): Boolean = {
```
johnynek: this can be on `object Expr` and be tested directly.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -1085,6 +1085,466 @@ object Matchless {
...
+          case LetMutBool(_, in) =>
+            loopBool(in)
+          case TrueConst =>
+            false
+        }
+
+      loopExpr(ex)
+    }
+
+    def readsMutableExpr(ex: Expr[A]): Boolean = {
```
johnynek: this could be on `object Expr` so we can test directly and maybe share on different optimizations.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -1085,6 +1085,466 @@ object Matchless {
...
+          case LetMutBool(_, in) =>
+            loopBool(in)
+          case TrueConst =>
+            false
+        }
+
+      loopExpr(ex)
+    }
+
+    def referencesBindable(expr: Expr[A], target: Bindable): Boolean = {
```
johnynek: this should be on `object Expr` for direct testing and sharing across optimizations.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -1085,6 +1085,466 @@ object Matchless {
...
+          case LetMutBool(_, in) =>
+            loopBool(in, isShadowed)
+          case TrueConst =>
+            false
+        }
+
+      loopExpr(expr, isShadowed = false)
+    }
+
+    def referencesBindableInBool(
```
johnynek: move to `object BoolExpl`

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -1085,6 +1085,466 @@ object Matchless {
...
+        cond: BoolExpr[A],
+        target: Bindable
+    ): Boolean =
+      referencesBindable(Always(cond, UnitExpr), target)
+
+    def referencesLocalAnon(expr: Expr[A], target: Long): Boolean = {
```
johnynek: move to `object Expr`

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -1085,6 +1085,466 @@ object Matchless {
...
+          case LetMutBool(_, in) =>
+            loopBool(in, isShadowed)
+          case TrueConst =>
+            false
+        }
+
+      loopExpr(expr, isShadowed = false)
+    }
+
+    def usesBinding(expr: Expr[A], arg: Either[LocalAnon, Bindable]): Boolean =
```
johnynek: move to `object Expr`

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -1085,6 +1085,466 @@ object Matchless {
...
+          referencesLocalAnon(expr, id)
+      }
+
+    def isTriviallyCheap(value: Expr[A]): Boolean =
+      value match {
+        case _: CheapExpr[?] =>
+          true
+        case cons: ConsExpr =>
+          cons.arity == 0
+        case App(fn: ConsExpr, args) =>
```
johnynek: this is an allocation. Why is this trivially cheap? If we could reduce an allocation it seems worth it. Maybe if the `args.length == 1` it's cheap, but otherwise I don't follow. Add a comment if you still think this makes sense.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -1085,6 +1085,466 @@ object Matchless {
...
+            true
+        }
+
+      noShadowingHazard &&
+      usesBinding(in, arg) &&
+      !hasSideEffect(value) &&
+      !readsMutableExpr(value) &&
+      !containsWhileExpr(value) &&
+      !isTriviallyCheap(value) &&
+      (exprWeight(value) >= HoistInvariantLoopLetsMinWeight)
```
johnynek, line 1433: add a comment here. Why do we not want to lift something cheap. Isn't everything a win?

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Moved the let-chain builder to `object Let` as `bindAll`, and de-duplicated the related helpers by using `Let.bindNamed` / `Let.bindAnons` where we previously repeated local `foldRight` let builders.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Yes, agreed. `Expr.referencesBindable` now short-circuits with `if (isShadowed) false`, so we stop recursing as soon as the target is shadowed.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Applied the same shadowing short-circuit to `Expr.referencesLocalAnon` (both expr and bool walkers now return `false` immediately when shadowed).

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Moved this to `Matchless.Expr.containsWhileExpr` and added direct test coverage in `MatchlessTests` (`Matchless.Expr helper predicates are directly testable`).

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Moved this to `Matchless.Expr.readsMutable` and added direct tests for it in `MatchlessTests`.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Moved this to `Matchless.Expr.referencesBindable` (with shadow-aware shared helper form) so it is directly testable and reusable.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Moved to `Matchless.BoolExpr.referencesBindable` and updated the hoist pass to use that bool-level helper directly.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Moved to `Matchless.Expr.referencesLocalAnon` and added direct shadowing tests for it.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Moved to `Matchless.Expr.usesBinding`, updated hoist logic to call it, and added direct tests for both bindable and local-anon bindings.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: I changed the behavior here: constructor applications are no longer treated as trivially cheap. `isTriviallyCheap` now only classifies `CheapExpr` and zero-arity constructors as cheap, and I added `hoists non-trivial constructor allocations` to lock this in.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Added a comment at the profitability gate explaining why cheap bindings are not always hoisted: hoisting can increase live ranges and residual loop body size, so we require non-trivial work before moving.

johnynek:
Implemented all requested refactors and comments on the PR thread with code changes plus direct tests for the new helper locations. I also updated the C codegen golden hash in `ClangGenTest` because this optimization changed generated output shape (`Ackermann` hash now `32eeba83ae74c2c8fa75703823fe7298`).

Checks run locally:
- `sbt "coreJVM/testOnly dev.bosatsu.MatchlessTest"`
- `sbt "cli/testOnly dev.bosatsu.codegen.clang.ClangGenTest"`

PR #1789, opened 2026-02-25
Refactor TotalityCheck to operate on TypedExpr post-inference

### LINKED ISSUE #1787: implement TotalityChecker to run on TypedExpr rather than Declaration
implement the design at:

docs/design/1769-design-refactoring-totalitychecker-to-run-on-typedexpr-declaration.md

in the repo.

johnynek:
Implemented issue #1787 by migrating totality checking from `Expr` to `TypedExpr` and moving package totality execution to post-inference typed lets.

What changed:
- Updated `TotalityCheck.ExprError` payloads to reference `TypedExpr.Match`.
- Refactored `TotalityCheck.checkExpr` to traverse `TypedExpr` nodes (including `AnnotatedLambda`, `Loop`, and `Recur`).
- Added typed inhabitedness-aware missing-branch pruning using scrutinee type info:
  - branches proven `Uninhabited` are excluded from coverage requirements,
  - missing patterns proven `Uninhabited` are filtered out,
  - definitively uninhabited scrutinees are treated as vacuously total.
- Preserved structural invalid-pattern diagnostics by validating parsed patterns from `Declaration.Match` tags (so list/string structural lint still reports as `InvalidPattern`).
- Updated `Package.inferBodyUnopt` to run totality after successful inference, alongside typed recursion checks; `UnusedLetCheck` remains pre-inference.
- Added a deterministic typed suite `TypedTotalityTest` covering:
  - `Result[Never, Int]` totality via `Ok(_)` only,
  - shadowing-based unreachable behavior,
  - vacuous totality for uninhabited scrutinees.
- Updated `ErrorMessageTest` fixtures for the typed totality path and typed-match construction.

Validation run:
- `sbt "coreJVM/testOnly dev.bosatsu.TotalityTest"`
- `sbt "coreJVM/testOnly dev.bosatsu.TypedTotalityTest"`
- `sbt "coreJVM/testOnly dev.bosatsu.ErrorMessageTest"`

Fixes #1787

### DIFF
--- a/core/src/main/scala/dev/bosatsu/TotalityCheck.scala
+++ b/core/src/main/scala/dev/bosatsu/TotalityCheck.scala
@@ -242,131 +242,143 @@ case class TotalityCheck(inEnv: TypeEnv[Kind.Arg]) {
         case _                                                 => false
       }
 
-    expr match {
-      case Annotation(e, _, _)                           => checkExpr(e)
-      case Generic(_, e)                                 => checkExpr(e)
-      case AnnotatedLambda(_, e, _)                      => checkExpr(e)
-      case Global(_, _, _, _) | Local(_, _, _) | Literal(_, _, _) =>
-        Validated.valid(())
-      case App(fn, args, _, _)     =>
-        checkExpr(fn) *> args.traverse_(checkExpr)
-      case Let(_, e1, e2, _, _)    => checkExpr(e1) *> checkExpr(e2)
-      case Loop(args, body, _)     =>
-        args.traverse_ { case (_, init) => checkExpr(init) } *> checkExpr(body)
-      case Recur(args, _, _)       =>
-        args.traverse_(checkExpr)
-      case matchExpr @ Match(arg, branches, _) =>
-        val argA: TypedExpr[Declaration] = arg
-        val branchesA: NonEmptyList[TypedExpr.Branch[Declaration]] = branches
-        val scrutineeType = arg.getType
-        val uninhabitedMemo =
-          MutHashMap.empty[Pattern[Cons, Type], Boolean]
-        def isDefinitelyUninhabited(
-            p: Pattern[Cons, Type]
-        ): Boolean =
-          uninhabitedMemo.getOrElseUpdate(
-            p,
-            definitelyUninhabited(
-              Inhabitedness.checkMatch(scrutineeType, p, inEnv)
+    def loop(
+        expr: TypedExpr[Declaration]
+    ): ValidatedNel[ExprError[Declaration], Unit] =
+      expr match {
+        case Annotation(e, _, _)      => loop(e)
+        case Generic(_, e)            => loop(e)
+        case AnnotatedLambda(_, e, _) => loop(e)
+        case Global(_, _, _, _) | Local(_, _, _) | Literal(_, _, _) =>
+          Validated.valid(())
+        case App(fn, args, _, _) =>
+          loop(fn) *> args.traverse_(loop)
+        case Let(_, e1, e2, _, _) =>
+          loop(e1) *> loop(e2)
+        case Loop(args, body, _) =>
+          args.traverse_ { case (_, init) => loop(init) } *> loop(body)
+        case Recur(args, _, _) =>
+          args.traverse_(loop)
+        case matchExpr @ Match(arg, branches, _) =>
+          val argA: TypedExpr[Declaration] = arg
+          val branchesA: NonEmptyList[TypedExpr.Branch[Declaration]] = branches
+          val scrutineeType0 = arg.getType
+          val freeScrutineeVars = Type.freeBoundTyVars(scrutineeType0 :: Nil).distinct
+          val scrutineeType =
+            Type.exists(
+              freeScrutineeVars.map(v => (v, Kind.Type)),
+              scrutineeType0
             )
-          )
-
-        val isUninhabitedScrutinee =
-          definitelyUninhabited(
-            Inhabitedness.check(scrutineeType, inEnv)
-          )
-        val allPatterns = branchesA.toList.map(_.pattern)
-        val unguardedPatterns = branchesA.toList.collect {
-          case branch if branch.guard.isEmpty => branch.pattern
-        }
-        val parsedValidation: Res[Unit] =
-          validateParsedMatchTag(matchExpr.tag)
-        val patternsValidated: ValidatedNel[ExprError[Declaration], Unit] =
-          parsedValidation
-            .parProductR(allPatterns.parTraverse_(validatePattern))
-            .leftMap { nel =>
-              nel.map(err =>
-                InvalidPattern(matchExpr, err): ExprError[Declaration]
+          val uninhabitedMemo =
+            MutHashMap.empty[Pattern[Cons, Type], Boolean]
+          def isDefinitelyUninhabited(
+              p: Pattern[Cons, Type]
+          ): Boolean =
+            uninhabitedMemo.getOrElseUpdate(
+              p,
+              definitelyUninhabited(
+                Inhabitedness.checkMatch(scrutineeType, p, inEnv)
               )
-            }
-            .toValidated
+            )
 
-        val argAndBranchExprs = argA :: branchesA.toList.flatMap { branch =>
-          branch.guard.toList ::: (branch.expr :: Nil)
-        }
-        val recursion: ValidatedNel[ExprError[Declaration], Unit] =
-          argAndBranchExprs.traverse_(checkExpr)
-
-        val matchCoverage: ValidatedNel[ExprError[Declaration], Unit] =
-          // We must use andThen here because missing/unreachable assumes each
-          // pattern is structurally valid. Running set operations first can
-          // produce misleading diagnostics or throw.
-          patternsValidated.andThen { _ =>
-            val missing: ValidatedNel[ExprError[Declaration], Unit] = {
-              val mis =
-                if (isUninhabitedScrutinee) Nil
-                else
-                  patternSetOps
-                    .missingBranches(topList, unguardedPatterns)
-                    .filterNot(isDefinitelyUninhabited)
-              NonEmptyList.fromList(mis) match {
-                case Some(nel) =>
-                  Validated.invalidNel(
-                    NonTotalMatch(matchExpr, nel): ExprError[Declaration]
-                  )
-                case None => Validated.valid(())
-              }
-            }
-
-            val unreachable: ValidatedNel[ExprError[Declaration], Unit] = {
-              val unr = {
-                @annotation.tailrec
-                def loop(
-                    rem: List[TypedExpr.Branch[Declaration]],
-                    covered: List[Pattern[Cons, Type]],
-                    acc: List[Pattern[Cons, Type]]
-                ): List[Pattern[Cons, Type]] =
-                  rem match {
-                    case Nil            => acc.reverse
-                    case branch :: tail =>
-                      val isUnreachable = fromList(covered) match {
-                        case None      => false
-                        case Some(cov) =>
-                          patternSetOps
-                            .difference(branch.pattern, cov)
-                            .isEmpty
-                      }
-                      val covered1 =
-                        if (branch.guard.isEmpty) branch.pattern :: covered
-                        else covered
-                      val acc1 =
-                        if (isUnreachable) branch.pattern :: acc
-                        else acc
-                      loop(tail, covered1, acc1)
-                  }
-                loop(branchesA.toList, Nil, Nil)
-              }
-              NonEmptyList.fromList(unr) match {
-                case Some(nel) =>
-                  Validated.invalidNel(
-                    UnreachableBranches(
-                      matchExpr,
-                      nel
-                    ): ExprError[Declaration]
-                  )
-                case None => Validated.valid(())
-              }
-            }
-
-            missing *> unreachable
+          val isUninhabitedScrutinee =
+            definitelyUninhabited(
+              Inhabitedness.check(scrutineeType, inEnv)
+            )
+          val allPatterns = branchesA.toList.map(_.pattern)
+          val unguardedPatterns = branchesA.toList.collect {
+            case branch if branch.guard.isEmpty => branch.pattern
           }
+          val parsedValidation: Res[Unit] =
+            validateParsedMatchTag(matchExpr.tag)
+          val patternsValidated: ValidatedNel[ExprError[Declaration], Unit] =
+            parsedValidation
+              .parProductR(allPatterns.parTraverse_(validatePattern))
+              .leftMap { nel =>
+                nel.map(err =>
+                  InvalidPattern(matchExpr, err): ExprError[Declaration]
+                )
+              }
+              .toValidated
 
-        (recursion *> matchCoverage).leftMap { errs =>
+          val argAndBranchExprs = argA :: branchesA.toList.flatMap { branch =>
+            branch.guard.toList ::: (branch.expr :: Nil)
+          }
+          val recursion: ValidatedNel[ExprError[Declaration], Unit] =
+            argAndBranchExprs.traverse_(loop)
+
+          val matchCoverage: ValidatedNel[ExprError[Declaration], Unit] =
+            // We must use andThen here because missing/unreachable assumes each
+            // pattern is structurally valid. Running set operations first can
+            // produce misleading diagnostics or throw.
+            patternsValidated.andThen { _ =>
+              val missing: ValidatedNel[ExprError[Declaration], Unit] = {
+                val mis =
+                  if (isUninhabitedScrutinee) Nil
+                  else
+                    patternSetOps
+                      .missingBranches(topList, unguardedPatterns)
+                      .filterNot(isDefinitelyUninhabited)
+                NonEmptyList.fromList(mis) match {
+                  case Some(nel) =>
+                    Validated.invalidNel(
+                      NonTotalMatch(matchExpr, nel): ExprError[Declaration]
+                    )
+                  case None => Validated.valid(())
+                }
+              }
+
+              val unreachable: ValidatedNel[ExprError[Declaration], Unit] = {
+                val unr = {
+                  @annotation.tailrec
+                  def loop(
+                      rem: List[TypedExpr.Branch[Declaration]],
+                      covered: List[Pattern[Cons, Type]],
+                      acc: List[Pattern[Cons, Type]]
+                  ): List[Pattern[Cons, Type]] =
+                    rem match {
+                      case Nil            => acc.reverse
+                      case branch :: tail =>
+                        val isUnreachable = fromList(covered) match {
+                          case None      => false
+                          case Some(cov) =>
+                            patternSetOps
+                              .difference(branch.pattern, cov)
+                              .isEmpty
+                        }
+                        val covered1 =
+                          if (branch.guard.isEmpty) branch.pattern :: covered
+                          else covered
+                        val acc1 =
+                          if (isUnreachable) branch.pattern :: acc
+                          else acc
+                        loop(tail, covered1, acc1)
+                    }
+                  loop(branchesA.toList, Nil, Nil)
+                }
+                NonEmptyList.fromList(unr) match {
+                  case Some(nel) =>
+                    Validated.invalidNel(
+                      UnreachableBranches(
+                        matchExpr,
+                        nel
+                      ): ExprError[Declaration]
+                    )
+                  case None => Validated.valid(())
+                }
+              }
+
+              missing *> unreachable
+            }
+
+          (recursion *> matchCoverage).leftMap { errs =>
             val errList = errs.toList
             // distinct can't reduce to 0
             NonEmptyList.fromListUnsafe(errList.distinct)
           }
-    }
+      }
+
+    loop(expr)
   }
 
   private val topList = WildCard :: Nil
--- a/core/src/test/scala/dev/bosatsu/TypedTotalityTest.scala
+++ b/core/src/test/scala/dev/bosatsu/TypedTotalityTest.scala
@@ -27,6 +27,21 @@ class TypedTotalityTest extends munit.FunSuite {
       }
     }
 
+  private def assertNonTotal(statement: String, letName: String): Unit =
+    checkEnvExpr(statement) { (env, lets) =>
+      val expr = findLetExpr(lets, letName)
+      TotalityCheck(env).checkExpr(expr) match {
+        case Validated.Invalid(errs) =>
+          val hasNonTotal = errs.exists {
+            case TotalityCheck.NonTotalMatch(_, _) => true
+            case _                                 => false
+          }
+          assert(hasNonTotal, errs.toList.mkString(", "))
+        case Validated.Valid(()) =>
+          fail("expected non-total match error")
+      }
+    }
+
   test("Result[Never, Value] with only Ok branch is total") {
     assertTotal(
       """#
@@ -87,16 +102,29 @@ def vacuous(z: Never) -> Never:
     )
   }
 
-  test("forall scrutinee pattern binding typechecks") {
-    checkEnvExpr("""#
+  test("forall scrutinee with phantom branch parameter is not total") {
+    assertNonTotal(
+      """#
 enum Result[e, r]: Err(err: r), Ok(ok: r)
 
 def is_good(res: forall e. Result[e, r]) -> r:
   Ok(ok) = res
   ok
-""") { (_, lets) =>
-      val hasIsGood = lets.exists(_._1 == Name("is_good"))
-      assert(hasIsGood)
-    }
+""",
+      "is_good"
+    )
+  }
+
+  test("forall scrutinee can prove totality when existential branch carries e") {
+    assertTotal(
+      """#
+enum Result[e, r]: Err(err: e), Ok(ok: r)
+
+def is_good(res: forall e. Result[e, r]) -> r:
+  Ok(ok) = res
+  ok
+""",
+      "is_good"
+    )
   }
 }
--- a/core/src/main/scala/dev/bosatsu/Inhabitedness.scala
+++ b/core/src/main/scala/dev/bosatsu/Inhabitedness.scala
@@ -307,108 +307,128 @@ object Inhabitedness {
         pattern: Pattern[(PackageName, Constructor), Type],
         varState: Map[Type.Var.Bound, State],
         seen: Set[Type]
-    ): State =
-      pattern match {
-        case Pattern.WildCard | Pattern.Var(_) =>
-          checkType(scrutinee, varState, seen)
-        case Pattern.Named(_, pat) =>
-          checkPattern(scrutinee, pat, varState, seen)
-        case Pattern.Annotation(pat, tpe) =>
-          if (definitelyDisjoint(scrutinee, tpe)) Uninhabited
-          else {
-            val scrState = checkType(scrutinee, varState, seen)
-            lazy val annState = checkType(tpe, varState, seen)
-            lazy val patState = checkPattern(tpe, pat, varState, seen)
+    ): State = {
+      val scrutineeNorm = scrutinee.normalize
 
-            if (
-              (scrState == Uninhabited) || (annState == Uninhabited) || (patState == Uninhabited)
-            ) Uninhabited
-            else if (
-              (scrState == Inhabited) && (annState == Inhabited) &&
-                (patState == Inhabited) && definitelySameHead(scrutinee, tpe)
-            ) Inhabited
-            else Unknown
-          }
-        case Pattern.Union(head, rest) =>
-          State.orStates(
-            (head :: rest.toList).map(checkPattern(scrutinee, _, varState, seen))
+      scrutineeNorm match {
+        case Type.ForAll(vars, in) =>
+          evaluateQuantifier(
+            vars.toList,
+            vs => checkPattern(in, pattern, vs, seen),
+            isForAll = true,
+            varState
           )
-        case Pattern.Literal(lit)      =>
-          if (scrutinee.sameAs(Type.getTypeOf(lit))) Inhabited
-          else Uninhabited
-        case _: Pattern.StrPat =>
-          if (scrutinee.sameAs(Type.StrType)) Inhabited
-          else Uninhabited
-        case lp @ Pattern.ListPat(_) =>
-          Pattern.ListPat.toPositionalStruct(lp, EmptyListCons, NonEmptyListCons) match {
-            case Right(asStruct) =>
-              checkPattern(scrutinee, asStruct, varState, seen)
-            case Left((_, suffix)) =>
-              // Prefix-glob list patterns (for example: `*_, p1, p2`) do not
-              // lower to a simple cons spine. We still know they require a list
-              // scrutinee and each explicit item in the suffix to be matchable
-              // against the list element type. We use that to prove definite
-              // `Uninhabited` cases; otherwise we stay conservative with `Unknown`.
-              checkType(scrutinee, varState, seen) match {
-                case Uninhabited => Uninhabited
-                case scrState =>
-                  listElementType(scrutinee) match {
-                    case None => Unknown
-                    case Some(itemType) =>
-                      val itemStates =
-                        suffix.toList.collect { case Pattern.ListPart.Item(itemPat) =>
-                          checkPattern(itemType, itemPat, varState, seen)
-                        }
-                      if (itemStates.contains(Uninhabited)) Uninhabited
-                      else if ((scrState == Inhabited) && itemStates.forall(_ == Inhabited))
-                        // We only conclude Inhabited when the scrutinee is known inhabited.
-                        Inhabited
-                      else Unknown
+        case Type.Exists(vars, in) =>
+          evaluateQuantifier(
+            vars.toList,
+            vs => checkPattern(in, pattern, vs, seen),
+            isForAll = false,
+            varState
+          )
+        case _ =>
+          pattern match {
+            case Pattern.WildCard | Pattern.Var(_) =>
+              checkType(scrutineeNorm, varState, seen)
+            case Pattern.Named(_, pat) =>
+              checkPattern(scrutineeNorm, pat, varState, seen)
+            case Pattern.Annotation(pat, tpe) =>
+              if (definitelyDisjoint(scrutineeNorm, tpe)) Uninhabited
+              else {
+                val scrState = checkType(scrutineeNorm, varState, seen)
+                lazy val annState = checkType(tpe, varState, seen)
+                lazy val patState = checkPattern(tpe, pat, varState, seen)
+
+                if (
+                  (scrState == Uninhabited) || (annState == Uninhabited) || (patState == Uninhabited)
+                ) Uninhabited
+                else if (
+                  (scrState == Inhabited) && (annState == Inhabited) &&
+                    (patState == Inhabited) && definitelySameHead(scrutineeNorm, tpe)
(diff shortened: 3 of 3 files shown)


core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -87,6 +87,17 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
 
   private val validUnit: Res[Unit] = Right(())
 
+  private def hasAdjacentWild[A](seq: SeqPattern[A]): Boolean =
```
johnynek: move this to be a method on `SeqPattern` if it doesn't already have it. Double check that.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -137,42 +138,148 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
+      case Pattern.Union(head, rest) =>
+        (head :: rest.toList).parTraverse_(validateParsedPattern)
+      case _ => validUnit
+    }
+
+  /** Check that a typed expression, and all inner expressions, are total, or
+    * return a NonEmptyList of matches that are not total
+    */
+  def checkExpr[A](expr: TypedExpr[A])(using
+      ev: K <:< Kind.Arg
```
johnynek: let's just fix `K = Kind.Arg` on the outer type and have callers do this substituteCo if they need it. We don't need this complexity here.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -137,42 +138,148 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
+
     expr match {
-      case Annotation(e, _, _)                           => checkExpr(e)
-      case Generic(_, e)                                 => checkExpr(e)
-      case Lambda(_, e, _)                               => checkExpr(e)
-      case Global(_, _, _) | Local(_, _) | Literal(_, _) => Validated.valid(())
-      case App(fn, args, _)     => checkExpr(fn) *> args.traverse_(checkExpr)
-      case Let(_, e1, e2, _, _) => checkExpr(e1) *> checkExpr(e2)
+      case Annotation(e, _, _)                           => checkExpr[A](e)
+      case Generic(_, e)                                 => checkExpr[A](e)
```
johnynek: add a comment as to why we can discard the quantification here. I'm confused why we don't need to distinguish universal from existential Vars when we are checking inhabitedness.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -137,42 +138,148 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
+        args.traverse_ { case (_, init) => checkExpr[A](init) } *> checkExpr[A](
+          body
+        )
+      case Recur(args, _, _)       =>
+        args.traverse_(checkExpr[A])
       case m @ Match(arg, branches, _) =>
-        val allPatterns = branches.toList.map(_.pattern)
-        val unguardedPatterns = branches.toList.collect {
+        val matchExpr: TypedExpr.Match[A] =
+          m.asInstanceOf[TypedExpr.Match[A]]
```
johnynek: let's remove this cast. I think if we mark `case class Match[+A]` so we make Match covariant then we shouldn't need this cast to widen.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -137,42 +138,148 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
+        val argA: TypedExpr[A] = arg
+        val branchesA: NonEmptyList[TypedExpr.Branch[A]] = branches
+        val scrutineeType = arg.getType
+        val uninhabitedMemo =
+          scala.collection.mutable.HashMap.empty[Pattern[Cons, Type], Boolean]
```
johnynek: import this as `import scala.collection.mutable.{HashMap => MutHashMap}` at the top of the file

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -137,42 +138,148 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
         }
-        allPatterns
-          .parTraverse_(validatePattern)
+        val parsedValidation: Res[Unit] =
+          matchExpr.tag match {
+            case Declaration.Match(_, _, cases) =>
+              cases.get.toList.parTraverse_(branch =>
+                validateParsedPattern(branch.pattern)
+              )
+            case _ => validUnit
```
johnynek: I think other forms than Match can result in Match. For instance `if/else` or pattern bindings, like `Foo(x) = foo`. We need to handle those.

We should handle all possible cases here and leave:

`case other => sys.error(s"unexpected declaration: $other")`

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -137,42 +138,148 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
-            val argAndBranchExprs = arg :: branches.toList.flatMap { branch =>
+            val argAndBranchExprs = argA :: branchesA.toList.flatMap { branch =>
               branch.guard.toList ::: (branch.expr :: Nil)
             }
-            val recursion = argAndBranchExprs.traverse_(checkExpr)
+            val recursion: ValidatedNel[ExprError[A], Unit] =
+              argAndBranchExprs.traverse_(checkExpr[A])
 
             val missing: ValidatedNel[ExprError[A], Unit] = {
+              val reachableUnguarded =
```
johnynek: let's move this into the `else { }` block so we don't need to compute it on `isUnhabited == true`

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -137,42 +138,148 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
+                unguardedPatterns.filterNot(isDefinitelyUninhabited)
               val mis =
-                patternSetOps.missingBranches(topList, unguardedPatterns)
+                if (isUninhabitedScrutinee) Nil
+                else
+                  patternSetOps
+                    .missingBranches(topList, reachableUnguarded)
```
johnynek: I think we want to compute `missingBranches` on `unguarded` patterns. Then we filter *after* that for only the maybe inhabited ones.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -137,42 +138,148 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
+          }
+
+        (parsedValidation, allPatterns.parTraverse_(validatePattern))
+          .parMapN((_, _) => ())
           .leftMap { nel =>
-            nel.map(InvalidPattern(m, _))
+            nel.map(err => InvalidPattern(matchExpr, err): ExprError[A])
           }
           .toValidated
           .andThen { _ =>
```
johnynek: why use `andThen` here vs `*>`? Can't we accumulate more errors by not using `andThen`?

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Done. I moved the adjacent-wildcard check onto `SeqPattern` as `hasAdjacentWildcards` in `core/src/main/scala/dev/bosatsu/pattern/SeqPattern.scala`, and updated `TotalityCheck` to call that method.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Updated. `TotalityCheck` is now `case class TotalityCheck(inEnv: TypeEnv[Kind.Arg])` (no outer type parameter or `substituteCo` in `checkExpr`). I adjusted test callers that were using `TypeEnv[Option[Kind.Arg]]` by converting those envs to `TypeEnv[Kind.Arg]` in `TotalityTest`.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Added a comment at `definitelyUninhabited` in `TotalityCheck` explaining why we only trust definitive `Uninhabited` and treat quantifier-sensitive/unknown cases conservatively as potentially inhabited.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Applied. I made `TypedExpr.Match` covariant (`Match[+T]`) and removed the cast in `TotalityCheck`.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Done. I imported mutable hash map as `import scala.collection.mutable.{HashMap => MutHashMap}` at the top of `TotalityCheck.scala` and switched usage accordingly.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Implemented. I added `validateParsedMatchTag` that explicitly handles the declaration forms that can tag generated matches (including wrappers/comprehension/def/lambda/binding paths), and kept `case other => sys.error(s"unexpected declaration: $other")` as requested.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Changed. Missing-pattern computation now only happens in the `else` branch when the scrutinee is not definitively uninhabited.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Updated. Missing branches are now computed from the full `unguardedPatterns` set first, then post-filtered to remove definitively uninhabited missing patterns.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Refactored. Recursion checking now accumulates via `*>` with the pattern-validation path. I kept `andThen` only for coverage (`missing/unreachable`) because those set-ops assume structurally valid patterns; this preserves safety while accumulating more errors overall.

johnynek:
Addressed all line-level review feedback with concrete code updates. Also updated related tests for the stricter `TypeEnv[Kind.Arg]` API and explicit match-tag invariants. Local checks run and passing: `sbt "coreJVM/test:compile"`, `sbt "coreJVM/testOnly dev.bosatsu.TotalityTest" "coreJVM/testOnly dev.bosatsu.TypedTotalityTest"`, and `sbt "coreJVM/testOnly dev.bosatsu.ErrorMessageTest"`.

johnynek:
several tests are failing:

```
[error] Failed tests:
[error] 	dev.bosatsu.WellTypedTests
[error] 	dev.bosatsu.EvaluationTest
[error] 	dev.bosatsu.TypedExprRecursionParityTest
[error] 	dev.bosatsu.codegen.python.PythonGenJvmTest
[error] 	dev.bosatsu.rankn.RankNInferTest
```

please make sure to run `coreJVM/test` and `cli/test` before submitting a new commit.

johnynek:
Reproduced the CI failure in `coreJVM/testOnly dev.bosatsu.rankn.RankNInferTest` (runtime `unexpected declaration: Var(...)` from `TotalityCheck.validateParsedMatchTag`). Updated `validateParsedMatchTag` to explicitly handle additional `Declaration` forms that can lower to typed matches without carrying parsed binders (including `Var`, `Apply`, `Annotation`, comments, tuple/string/literal, and list/dict cons forms), while keeping `sys.error` fallback for truly unexpected forms. Verified with `sbt "coreJVM/testOnly dev.bosatsu.rankn.RankNInferTest"`, `sbt "coreJVM/testOnly dev.bosatsu.WellTypedTests dev.bosatsu.EvaluationTest dev.bosatsu.TypedExprRecursionParityTest dev.bosatsu.codegen.python.PythonGenJvmTest"`, and full pre-push check `sbt "coreJVM/test; cli/test"`.

johnynek:
Validated on PR head `7d6970b1969c1736ebbc0c82d9c1dbc0fa5ad22d` by running exactly the requested checks: `sbt "coreJVM/test"` and `sbt "cli/test"`. Both passed locally. The previously reported failing suites (`WellTypedTests`, `EvaluationTest`, `TypedExprRecursionParityTest`, `PythonGenJvmTest`, `RankNInferTest`) all pass in the full `coreJVM/test` run, so no additional code change was required in this turn.

core/src/test/scala/dev/bosatsu/TypedTotalityTest.scala
```diff
@@ -0,0 +1,146 @@
...
+  private def matchTag(name: String): Declaration =
+    Declaration.Matches(Declaration.Var(Name(name)), Pattern.WildCard)
+
+  private def intExpr(i: Long, tag: Declaration): TypedExpr[Declaration] =
+    TypedExpr.Literal(Lit.Integer(i), Type.IntType, tag)
+
+  private def resultExprType(err: Type, ok: Type): Type =
+    Type.applyAll(resultType, err :: ok :: Nil)
+
+  test("Result[Never, Int] with only Ok branch is total") {
```
johnynek: let's add a similar test with:

```
enum Result[e, r]: Err(err: r), Ok(ok: r)

def is_good(r: forall e. Result[e, r]) -> r:
    Ok(ok) = r
    ok
```
That should typecheck.

Please add a test like that.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -137,42 +128,186 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
+
+  /** Some structural pattern lint information is only available on the parsed
+    * patterns carried by Declaration tags.
     */
-  def checkExpr[A](expr: Expr[A]): ValidatedNel[ExprError[A], Unit] = {
-    import Expr._
+  private def validateParsedPattern(p: Pattern.Parsed): Res[Unit] =
+    p match {
+      case lp @ Pattern.ListPat(parts) =>
+        val twoAdj = lp.toSeqPattern.toList.sliding(2).exists {
```
johnynek: why can't this be `lp.toSeqPattern.hasAdjacentWildcards`?

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -137,42 +128,186 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
+                MultipleSplicesInPattern(placeholderListPattern, inEnv),
+                Nil
+              )
+            )
+        val inners: Res[Unit] =
+          parts.parTraverse_ {
+            case Pattern.ListPart.Item(item) => validateParsedPattern(item)
+            case _                           => validUnit
+          }
+        (outer, inners).parMapN((_, _) => ())
```
johnynek: `outer.parProductL(inners)` I think is what we want (or something similar)

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -137,42 +128,186 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
+
+      case Pattern.PositionalStruct(_, args) =>
+        args.parTraverse_(validateParsedPattern)
+      case Pattern.Named(_, inner) =>
+        validateParsedPattern(inner)
+      case Pattern.Annotation(inner, _) =>
+        validateParsedPattern(inner)
+      case Pattern.Union(head, rest) =>
+        (head :: rest.toList).parTraverse_(validateParsedPattern)
+      case _ => validUnit
```
johnynek: can we make this total `case Var(_) | Wildcard | Lit(_)` or something like that.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -137,42 +128,186 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
+      case Declaration.Binding(BindingStatement(pat, _, _)) =>
+        validateParsedPattern(pat)
+      case Declaration.DefFn(defstmt) =>
+        defstmt.args.toList.flatMap(_.toList).parTraverse_(validateParsedPattern)
+      case Declaration.Lambda(args, _) =>
+        args.toList.parTraverse_(validateParsedPattern)
+      case Declaration.LeftApply(arg, _, _, _) =>
+        validateParsedPattern(arg)
+      case Declaration.Comment(_) | Declaration.CommentNB(_) =>
+        validUnit
```
johnynek, line 204: I don't this this is right. If we see a comment or a comment NB, don't we need to process the inner thing they are commenting?

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -137,42 +128,186 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
+      case Declaration.Annotation(_, _) |
+          Declaration.Apply(_, _, _) |
+          Declaration.ApplyOp(_, _, _) |
+          Declaration.Literal(_) |
+          Declaration.IfElse(_, _) |
+          Declaration.Ternary(_, _, _) |
+          Declaration.TupleCons(_) |
+          Declaration.Var(_) |
+          Declaration.StringDecl(_) =>
+        // These can lower into a Match but do not carry parsed-pattern binders.
```
johnynek: I don't see how `Var(_)` or `Literal(_)` can be a match. This comment seems out of date. I can see how `IfElse` and `Ternary` are match, but not many of the rest.

Can we separate the IfElse/Ternary cases? I think many of the rest should be in the other case and throw an exception. Can you try one by one adding items into that section and if they don't work, give examples of some matches that were compiled into them? I'm confused how these came up.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -137,42 +128,186 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
-          .leftMap { nel =>
-            nel.map(InvalidPattern(m, _))
-          }
-          .toValidated
-          .andThen { _ =>
-            // if the patterns are good, then we check them for totality
-            val argAndBranchExprs = arg :: branches.toList.flatMap { branch =>
-              branch.guard.toList ::: (branch.expr :: Nil)
+        val parsedValidation: Res[Unit] = matchExpr.tag match {
+          case d: Declaration => validateParsedMatchTag(d)
```
johnynek: this is runtime reflection. We have to remove this. If we need the `Declaration` tag, we need to make the method require `TypedExpr[Declaration]`. If the caller passes something that is not a Declaration, this just silently skips the check

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -137,42 +128,186 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
+          case _              => validUnit
+        }
+        val patternsValidated: ValidatedNel[ExprError[A], Unit] =
+          (parsedValidation, allPatterns.parTraverse_(validatePattern))
+            .parMapN((_, _) => ())
```
johnynek:
```scala
parsedValidation
  .parProductR(allPatterns.parTraverse_(validatePattern))
```
not .parMapN... don't use parMapN to implement parProductR or parProductL

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -137,42 +128,186 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
+            .toValidated
+
+        val argAndBranchExprs = argA :: branchesA.toList.flatMap { branch =>
+          branch.guard.toList ::: (branch.expr :: Nil)
+        }
+        val recursion: ValidatedNel[ExprError[A], Unit] =
+          argAndBranchExprs.traverse_(checkExpr[A])
 
+        val matchCoverage: ValidatedNel[ExprError[A], Unit] =
+          // missing/unreachable assume structurally valid patterns.
```
johnynek: let's expand here:
"we must use andThen because we can't compute missing branches until we are sure the patterns are structurally valid"

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -137,42 +128,186 @@ case class TotalityCheck(inEnv: TypeEnv[Any]) {
...
+          patternsValidated.andThen { _ =>
             val missing: ValidatedNel[ExprError[A], Unit] = {
               val mis =
-                patternSetOps.missingBranches(topList, unguardedPatterns)
+                if (isUninhabitedScrutinee) Nil
+                else {
+                  val maybeMissing =
```
johnynek: let's not capture, let's do dot chaining:
```scala
patternSetOps
  .missingBranches(topList, unguardedPatterns)
  .filterNot(isDefinitelyUninhabited)
```

core/src/test/scala/dev/bosatsu/TypedTotalityTest.scala
```diff
@@ -0,0 +1,146 @@
...
+  private val pack = PackageName.parts("TypedTotality")
+  private val neverName = TypeName(Constructor("Never"))
+  private val resultName = TypeName(Constructor("Result"))
+
+  private val neverType: Type.TyConst =
+    Type.TyConst(Type.Const.Defined(pack, neverName))
+  private val resultType: Type.TyConst =
+    Type.TyConst(Type.Const.Defined(pack, resultName))
+
+  private val typeEnv: TypeEnv[Kind.Arg] = {
```
johnynek: rather than manually building up these adts, let's use bosatu syntax + a method on TestUtils (which may already exist or we may need to make) like:
```scala
def checkEnvExpr(str: String)(fn: (TypeEnv[Kind.Arg], List[(Bindable, RecursionKind, TypedExpr[Declaration])] => A): A = ...
  // run the parser + RankNInfer, but don't do any other checks. just hand the typeEnv and the vals to the typedExpr)
```

Then we can use actual syntax built from the SourceConverter + RankNInfer pipeline, but no other transformations. That should allow us to test TotalityCheck against actual code and be sure we are getting it right.

core/src/test/scala/dev/bosatsu/TypedTotalityTest.scala
johnynek: Added this test in `TypedTotalityTest` as `forall scrutinee pattern binding typechecks` using bosatsu syntax:

enum Result[e, r]: Err(err: r), Ok(ok: r)

def is_good(res: forall e. Result[e, r]) -> r:
  Ok(ok) = res
  ok

It now typechecks through the parser/source-converter/rankn pipeline.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Changed list wildcard adjacency checks to use `lp.toSeqPattern.hasAdjacentWildcards` in both `validatePattern` and `validateParsedPattern`.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Replaced the old `parMapN` unit combination with `outer.parProductL(inners)` in both typed and parsed list-pattern validation.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Made the leaf handling explicit and total in both validators: `Var(_) | WildCard | Literal(_)` now map to `validUnit`, with recursive handling for `Named`, `Annotation`, and `Union`.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Updated comment handling to recurse into the commented declaration: `Comment` and `CommentNB` now call `validateParsedMatchTag(comment.on.padded)` (and `Parens` recurses too).

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Split out `IfElse`/`Ternary` as explicit no-pattern cases and kept an explicit `case other => sys.error(...)` fallback. Then I tightened/expanded supported non-pattern tag forms based on real failures seen while running tests: `Apply`, `Literal`, and `TupleCons` were observed as actual match tags in `WellTypedTests`; those are now explicitly handled.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Removed runtime reflection/cast by changing `checkExpr` to require `TypedExpr[Declaration]` directly. Tag validation now uses `validateParsedMatchTag(matchExpr.tag)` with no runtime type test.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Switched to the requested product form:
`parsedValidation.parProductR(allPatterns.parTraverse_(validatePattern))`.
No `parMapN` used for this product.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Expanded the `andThen` comment to explain that missing/unreachable branch set-operations require structurally valid patterns first; otherwise diagnostics can be misleading or throw.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Updated missing-branch computation to dot chaining style:
`patternSetOps.missingBranches(topList, unguardedPatterns).filterNot(isDefinitelyUninhabited)`.

core/src/test/scala/dev/bosatsu/TypedTotalityTest.scala
johnynek: Implemented `TestUtils.checkEnvExpr` to run parser + SourceConverter + kind solving + RankN infer and return `(TypeEnv[Kind.Arg], typed lets)` before later optimization passes. Reworked `TypedTotalityTest` to use this helper and bosatsu syntax instead of manually constructing ADTs/TypedExpr.

johnynek:
Addressed all review items with concrete code/test updates. I also reproduced the CI failure locally and fixed the resulting tag-coverage issues in `validateParsedMatchTag` (`Apply`, `Literal`, `TupleCons` were observed in real failing cases). Required pre-push checks are now green locally: `sbt "coreJVM/test" "cli/test"`.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -170,7 +170,8 @@ case class TotalityCheck(inEnv: TypeEnv[Kind.Arg]) {
         validateParsedPattern(inner)
       case Pattern.Union(head, rest) =>
         (head :: rest.toList).parTraverse_(validateParsedPattern)
-      case _ => validUnit
+      case Pattern.Var(_) | Pattern.WildCard | Pattern.Literal(_) =>
+        validUnit
     }
 
   private def validateParsedMatchTag(tag: Declaration): Res[Unit] =
```
johnynek, line 177: let's be clear what validation we are doing here. I think it may just be validating the patterns, not evaluating match totality. Is that right? Let's be clear, because it's not so clear when reading since we have to follow a lot of logic. Please add a comment as to what Valid and Invalid means when this method returns.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -189,30 +190,34 @@ case class TotalityCheck(inEnv: TypeEnv[Kind.Arg]) {
...
-      case Declaration.Comment(_) | Declaration.CommentNB(_) =>
+      case Declaration.Comment(comment) =>
+        validateParsedMatchTag(comment.on.padded)
+      case Declaration.CommentNB(comment) =>
+        validateParsedMatchTag(comment.on.padded)
+      case Declaration.Parens(of) =>
+        validateParsedMatchTag(of)
+      case Declaration.IfElse(_, _) | Declaration.Ternary(_, _, _) =>
+        // These parse forms are lowered to bool constructor matches.
+        validUnit
```
johnynek, line 201: add a comment as to why this is always valid. Is it because this generated match would not involve invalid patterns?

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -221,7 +226,9 @@ case class TotalityCheck(inEnv: TypeEnv[Kind.Arg]) {
   /** Check that a typed expression, and all inner expressions, are total, or
     * return a NonEmptyList of matches that are not total
     */
-  def checkExpr[A](expr: TypedExpr[A]): ValidatedNel[ExprError[A], Unit] = {
+  def checkExpr(
+      expr: TypedExpr[Declaration]
+  ): ValidatedNel[ExprError[Declaration], Unit] = {
```
johnynek, line 231: let's use `ValidatedNec` for consistency. We want to prefer that type and minimize jumping back and fourth to Nel

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
```diff
@@ -274,49 +279,50 @@ case class TotalityCheck(inEnv: TypeEnv[Kind.Arg]) {
...
-                    patternSetOps.missingBranches(topList, unguardedPatterns)
-                  maybeMissing.filterNot(isDefinitelyUninhabited)
-                }
+                else
+                  patternSetOps
+                    .missingBranches(topList, unguardedPatterns)
+                    .filterNot(isDefinitelyUninhabited)
               NonEmptyList.fromList(mis) match {
                 case Some(nel) =>
                   Validated.invalidNel(
```
johnynek: let's use invalidNec

core/src/test/scala/dev/bosatsu/TypedTotalityTest.scala
```diff
@@ -0,0 +1,102 @@
...
+""",
+      "vacuous"
+    )
+  }
+
+  test("forall scrutinee pattern binding typechecks") {
+    checkEnvExpr("""#
+enum Result[e, r]: Err(err: r), Ok(ok: r)
+
+def is_good(res: forall e. Result[e, r]) -> r:
```
johnynek, line 110: this shouldn't typecheck as written.

the `e` type isn't used in the result above. So, `forall e. Result[e, r]` could still be `Err(r)` but below we aren't handling that branch.

Explain in detail why this test passed in your reply, make a diff that makes this fail and add the correct test:

```
enum Result[e, r]: Err(err: e), Ok(ok: r)
```
and make *that* one pass, while the current one must fail.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Added a method-level comment on `validateParsedMatchTag` clarifying exactly what this validates: only parsed-pattern structural validity. It now explicitly documents that `Valid(())` means parsed patterns are structurally safe for downstream set-ops, while `Invalid(...)` means malformed parsed patterns (for example adjacent list/string wildcards), and that this method does not check totality/reachability.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Added an explicit comment on the `IfElse`/`Ternary` branch explaining why this is always valid here: these tags do not carry parsed user patterns to validate at this stage, and are lowered to boolean constructor matches before pattern-set analysis.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Switched `TotalityCheck.checkExpr` and its internal validated values to `ValidatedNec` for consistency (`patternsValidated`, `recursion`, `matchCoverage`, and related flow). I also updated call-site compatibility in `Package.scala` by converting the `NonEmptyChain` side back to `NonEmptyList` where the package pipeline still expects `ValidatedNel`.

core/src/main/scala/dev/bosatsu/TotalityCheck.scala
johnynek: Updated the match coverage error construction to use `Validated.invalidNec(...)` (for both non-total and unreachable branches), replacing the previous `invalidNel` usage.

johnynek:
Implemented all requested review changes in `TotalityCheck.scala` and fixed resulting type fallout in `Package.scala` and `TypedTotalityTest.scala`. Verification is currently blocked: after `sbt -Dsbt.supershell=false "clean"` succeeds, subsequent compile/test commands stall during `coreJVM` compile at `"compiling 164 Scala sources to .../core/.jvm/target/scala-3.8.1/classes ..."` and do not complete/return in this environment, so I cannot safely report passing required pre-push checks (`coreJVM/test` and `cli/test`) yet.

johnynek:
MergeXO operator command result:
- command: `/mergexo unblock`
- status: `applied`
- detail: Reset blocked state for PR #1789.

Source command: https://github.com/johnynek/bosatsu/pull/1789#issuecomment-3962350548

core/src/test/scala/dev/bosatsu/TypedTotalityTest.scala
johnynek: You were right. The previous test passed for the wrong reason: it only used `checkEnvExpr` and asserted the inferred let existed, which runs parse/source-convert/rank-n infer but does not run `TotalityCheck`, so non-totality was never checked.

I changed the test to make that explicit:
- Added `assertNonTotal` in `TypedTotalityTest` that runs `TotalityCheck(env).checkExpr(expr)` and requires a `NonTotalMatch`.
- Replaced the prior phantom case with:
  - `enum Result[e, r]: Err(err: r), Ok(ok: r)`
  - `def is_good(res: forall e. Result[e, r]) -> r: Ok(ok) = res; ok`
  and now assert it is non-total.
- Added the corrected case:
  - `enum Result[e, r]: Err(err: e), Ok(ok: r)`
  with the same `is_good` definition,
  and assert it is total.

To make the corrected case actually pass while the phantom one fails, I fixed quantified reachability handling:
- `Inhabitedness.checkPattern` now handles quantified scrutinee types (`ForAll`/`Exists`) directly instead of treating them as unknown constructor heads.
- In `TotalityCheck`, before calling `Inhabitedness.check/checkMatch` for missing-branch filtering, I existentially close free bound vars in the scrutinee type. This avoids `UnknownTypeVariable` validation failures while keeping the phantom `Err(err: r)` case reachable (non-total) and proving `Err(err: e)` unreachable under `forall e`.

So now: the old phantom formulation fails totality, and the corrected `Err(err: e)` formulation passes.

johnynek:
Addressed the review item with concrete test+checker changes and verified end-to-end. I ran `sbt "coreJVM/test"` (1360 passed, 0 failed) and `sbt "cli/test"` (57 passed, 0 failed).

Issue #1796, opened 2026-02-25
design a Bytes type


johnynek:
Design a bosatsu external `Bytes` type which is semantically basically `Array[Byte]` but not parametric so it can be stored as a pointer to bytes, and offset and length.

Make an API consistent with Bosatsu/Predef::List and Bosatsu/Collection/Array. We should have the ability to convert to `Array[Int]` and `from_List_Int(ints: List[Int]): Bytes` which takes the low 8 bits of each int in 2's complement and converts those to a number from 0 to 255.

We should consult the python3 bytes and memoryview API and use that for inspiration. We need to stay immutable and consistent with existing Bosatsu, but also ideally as close to python3 bytes/memoryview as possible. This is our number 3 goal after: must be immutable and be consistent with bosatsu style especially List and Array.

Where should it live? Maybe in Bosatsu/IO which is where it would be useful.

Propose Bosatsu/IO/Core functions to work with binary files using Handles and Bytes to read and write data. Add a section for how these functions can be as efficient as possible with the constraints that we have. Speak to what external functions we should have so that the external function is "big enough" to do significant work where it can leverage mutability and keep the API boundary of the function pure and immutable.

johnynek:
MergeXO assigned an agent and started design work for issue #1796.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/1797

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/1796#issuecomment-3961004106
- linked PR: #1797 (https://github.com/johnynek/bosatsu/pull/1797)

Comments on the source issue are no longer actioned after a PR exists. Please comment on the PR thread instead.

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #1796.

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #1796.

johnynek:
Opened implementation PR: https://github.com/johnynek/bosatsu/pull/1812

PR #1797, opened 2026-02-25
Design doc for #1796: Issue #1796 design: Bytes type and binary IO

johnynek:
Design doc.

Refs #1796

### DIFF
--- a/docs/design/1796-design-a-bytes-type.md
+++ b/docs/design/1796-design-a-bytes-type.md
@@ -98,6 +98,7 @@ Issue #1796 asks for:
 5. Add binary IO functions in `Bosatsu/IO/Core` using existing `Handle`.
 6. Keep runtime-intensive loops inside external functions where mutability is safe and hidden.
 7. Keep Python `bytes`/`memoryview` similarity where it does not conflict with Bosatsu style priorities.
+8. Support `Bytes` in `ValueToJson` and `ValueToDoc` as lists of ints in `0..255`.
 
 ## Non-goals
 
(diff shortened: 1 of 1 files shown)


docs/design/1796-design-a-bytes-type.md
```diff
@@ -0,0 +1,432 @@
...
+
+Add to `Bosatsu/IO/Core`:
+
+```bosatsu
+from Bosatsu/IO/Bytes import Bytes
+
+external def read_bytes(h: Handle, max_bytes: Int) -> Prog[IOError, Option[Bytes]]
+external def write_bytes(h: Handle, bytes: Bytes) -> Prog[IOError, Unit]
+external def read_all_bytes(h: Handle, chunk_size: Int) -> Prog[IOError, Bytes]
+external def copy_bytes(src: Handle, dst: Handle, chunk_size: Int) -> Prog[IOError, Int]
```
johnynek: I think maybe we should also have an `Option[Int]` limit to the total number of bytes copied. So if you want to copy the next N bytes from src to dst you can do that without having to copy all the bytes over.

docs/design/1796-design-a-bytes-type.md
```diff
@@ -0,0 +1,432 @@
...
+### Evaluator/runtime tests
+
+1. Add JVM evaluator test in `EvaluationTest.scala` similar to existing array external test.
+2. Add binary file round-trip tests (`write_bytes` then `read_all_bytes`).
+3. Add partial read tests for `read_bytes` chunking.
+4. Add `copy_bytes` transfer count test.
+
+### Tooling conversion tests
+
+1. If `ValueToJson`/`ValueToDoc` support is included, add `ToolAndLibCommandTest` coverage for `Bytes` JSON write/apply.
```
johnynek: this support should be included. Bytes should be supported in both and render as lists of ints on `0..255` as you describe.

docs/design/1796-design-a-bytes-type.md
johnynek: Updated the design doc to add a bounded copy API: `copy_bytes(src, dst, chunk_size, max_total: Option[Int])`. I also added explicit semantics for `None` (copy to EOF), `Some(n)` (copy up to n bytes), `Some(n<0)` (`InvalidArgument`), and `Some(0)` (copy zero bytes), and updated testing/acceptance sections to cover both bounded and unbounded modes.

docs/design/1796-design-a-bytes-type.md
johnynek: Included this support directly in scope. The doc now makes `ValueToJson`/`ValueToDoc` support for `Bytes` required (not optional), with explicit rendering as lists/arrays of ints in `0..255`, plus matching tooling test coverage and acceptance criteria.

johnynek:
Addressed both review comments by updating `docs/design/1796-design-a-bytes-type.md` directly. This is a doc-only update.

Issue #1172, opened 2024-03-16
syntax to require recursions to be tail recursive.


johnynek:
In scala, there is an annotation `@tailrec` that makes compilation fail if the function is not tail recursive. Tail recursion ensures that the function can be compiled into a loop, and ensures we don't blow the stack.

Bosatsu has the optimization, but no way to require that it happens. This seems against the bosatsu way of biasing towards stronger guarantees.

We could have syntax like:
```
defloop len(lst, acc):
    recur let:
        case []: acc
        case [_, *tail]: len(tail, acc.add(1))

```
where defloop is just a def where all recursions are in the tail position. Since the tail check can be a kind of a lint, we can do that in parallel with other type checking so that we don't have the sequential error presentation issue.

I don't love the look of `defloop`, but maybe it's the best. Idk. Changing it should be pretty easy since that string can be changed with find and replace. Ideas:

```
defloop len(lst, acc):
  ...

loop len(lst, acc):
  ....

def len(lst, acc):
  loop lst:
    case []: ...
```

hmm... maybe the last one is best... it also makes the tail loop nature local to the function rather than at the top level.

johnynek:
thinking more: the last option above seems the best: if the recursion is tail recursive you must use `loop`. you use `recur` only when there is non-tail recursion. This allows a reader to clearly see which case we are in.

johnynek:
Let's make a design doc for the last item. Where we landed:

allow `loop` as a stronger version of `match` and `recur`. `loop` works exactly the same except recursive calls under loop MUST be tail recursive (so all calls to the itself must appear in the tail position), which means it will be compiled to a while loop. We should allow tail calls in `recur` but we must `error` on non-tail calls under `loop`.

johnynek:
MergeXO assigned an agent and started design work for issue #1172.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/1799

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/1172#issuecomment-2002215912
- linked PR: #1799 (https://github.com/johnynek/bosatsu/pull/1799)

Comments on the source issue are no longer actioned after a PR exists. Please comment on the PR thread instead.

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/1172#issuecomment-3960972742
[...]

PR #1799, opened 2026-02-25
Design doc for #1172: Issue #1172 design doc: require tail-recursive recursion syntax

johnynek:
Design doc.

Refs #1172

### DIFF
--- a/docs/design/1172-syntax-to-require-recursions-to-be-tail-recursive.md
+++ b/docs/design/1172-syntax-to-require-recursions-to-be-tail-recursive.md
@@ -149,7 +149,14 @@ val left: Indy[(MatchMode, NonBinding)] =
 Operationally, this means:
 1. `def loop(x): ...` still parses as a definition name because it goes through identifier parsing, not `matchP`.
 2. `loop x:` in recursion-header position is parsed as `MatchMode.Loop`.
-3. `loop x` without `:` in a context expecting recursion header is a syntax error (with parser hints), not a silent fallback to variable parsing.
+3. `loop` as a local value/binder name is also legal, for example:
+
+```bosatsu
+loop = x -> fn(x)
+loop(12)
+```
+
+4. `loop x` without `:` in a context expecting recursion header is a syntax error (with parser hints), not a silent fallback to variable parsing.
 
 ### 2. Typed recursion checker integration
 Extend `TypedExprRecursionCheck` so `Loop` and `Recur` both enter recursive checking paths. Existing recur-target legality, lexicographic checks, and shadowing checks stay unchanged.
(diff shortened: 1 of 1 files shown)


docs/design/1172-syntax-to-require-recursions-to-be-tail-recursive.md
```diff
@@ -0,0 +1,229 @@
...
+3. Typed recursion checking (now including loop tail requirement).
+4. Typed loop/recur lowering and normalization.
+
+This preserves the existing trust boundary: recursion legality is validated before lowering.
+
+## Detailed implementation plan
+1. In `Declaration.scala`:
+   1. Introduce match-mode ADT for `match`/`recur`/`loop`.
+   2. Update parser and document rendering.
+   3. Keep contextual-keyword behavior for `loop`.
```
johnynek, line 199: we need to be clearer in this doc. How will we accept loop as a valid name but also parse it as a match/recur position?

I think we will have to have something like:

```scala
kwSpace("loop") *> (
  // we can either have match argument, so a single nonbinding followed by : to the end of line, or a tuple of
  // match argument followed by colon to the end of line
  // or loop was a var
  // we should not backtrack all the back out of loop here.
  // and we need to minimize use of backtrack or soft to the bare minimum.
)
```

johnynek: Updated the doc directly to make this explicit in **Architecture → Declaration AST and parser** under a new subsection: **“Parser shape for contextual `loop` (concrete)”**.

It now specifies:
- `loop` stays out of `Declaration.keywords`, so `def loop(...):` remains valid.
- `loop` is only interpreted contextually inside `matchP` header parsing (`matchKindParser`).
- after recognizing `loop`, parsing commits with `cutThen(...)` to the recur target + `:` path (no broad fallback to var parsing).
- backtracking/`soft` should be limited to the keyword disambiguation boundary.

I also added a parser sketch with `kwSpace("loop")` and the `cutThen`/`cutLeftP` shape you asked for, plus operational examples (`def loop(x): ...`, `loop x:`, and `loop x` without `:`).

johnynek:
Addressed the review feedback by clarifying the contextual-keyword parser strategy in the design doc, including commit/backtracking behavior and a concrete parser sketch.

docs/design/1172-syntax-to-require-recursions-to-be-tail-recursive.md
```diff
@@ -0,0 +1,261 @@
...
+3. In `TypedExprRecursionCheck.scala`:
+   1. Treat `Loop` as recursive in recur-site detection.
+   2. Track whether current recursive def is in `Loop` mode.
+   3. After existing recursion validation, enforce `SelfCallKind != NonTailCall` for `Loop`.
+   4. Emit new recursion error on violation.
+4. In `RecursionCheck.scala`:
+   1. Add new error case and message.
+5. In tests:
+   1. `ParserTest.scala`: parse and round-trip `loop` forms.
+   2. `TypedExprRecursionCheckTest.scala`: allow tail-recursive `loop`; reject non-tail-recursive `loop`; verify `recur` still allows non-tail recursion; verify tuple `loop` still uses existing lexicographic rules; verify `def loop(...)` remains legal.
```
johnynek, line 211: also using loop as a var should be legal:

```
loop = x -> fn(x)
loop(12)
```
that should work.

johnynek: Updated the design doc to make this explicit. In the contextual-parser section, I added a concrete example showing `loop` as a legal value name:

```bosatsu
loop = x -> fn(x)
loop(12)
```

I also expanded the implementation/testing sections to require an explicit parser test for that case (`loop = x -> fn(x); loop(12)`) alongside `def loop(...)` and `loop x:` parsing behavior.

johnynek:
Addressed the review feedback by clarifying that `loop` remains legal as a normal binder/value name and by adding concrete parser-test expectations for that scenario.

PR #1803, opened 2026-02-25
Add type-check-only internal Predef `todo` via compile mode

### LINKED ISSUE #1173: add a todo function when running type check
when working on #1171 I found a pattern which was useful:

```
external def todo(ignore: x) -> forall a. a
```

this function can consume arguments to make them look used, and then return anything. Of course, you can't have a function like that at runtime.

On way to avoid that is to only add that function to predef when in `type-check` mode. In that case, we never even emit anything, any so it can't be unsafe. This could also be used by a language server in the IDE, since you don't need the compiled outputs in that case either.

If we had a resident server that could cache memory, we could keep a cache of compiled packages for dependencies but only use it for type-checking (or rather, the cache key can depend on if we are in type check mode or not).

johnynek:
Implemented issue #1173 per the design doc by introducing explicit compile modes in `CompileOptions` (`Emit` and `TypeCheckOnly`) and adding `CompileOptions.TypeCheckOnly`.

`Package` now builds two internal predef variants: runtime predef (unchanged) and type-check-only predef, which adds `external def todo(ignore: x) -> forall a. a` and exports `todo`. No runtime/codegen externals were added.

`PackageMap` is now mode-aware for internal predef behavior: it compiles/caches internal predef per mode, derives implicit predef imports from the selected compiled variant, injects the correct predef package/imports in `typeCheckParsed`, and exposes mode-based predef lookup used by empty-input typecheck flows. A comment documents that mode must be part of predef cache identity.

Wired check entry points to type-check mode:
- `tool check` now uses `CompileOptions.TypeCheckOnly`
- `lib check` now uses `CompileOptions.TypeCheckOnly`
- `CompilerApi.typeCheck0` empty-input path now returns internal predef compiled for the selected mode.

Added regression coverage:
- `PackageTest`: `todo` type-checks in `TypeCheckOnly`, fails in emit mode, and is exported only by type-check predef.
- `ToolAndLibCommandTest`: `tool check` and `lib check` accept `todo`, while emit-mode flows (`tool show`, `lib show`) reject it.

Validation run and passing:
- `sbt "coreJVM/testOnly dev.bosatsu.PackageTest dev.bosatsu.ToolAndLibCommandTest"`

Fixes #1173

Implements design doc: [docs/design/1173-add-a-todo-function-when-running-type-check.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/1173-add-a-todo-function-when-running-type-check.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/1795

### DIFF
--- a/core/src/main/scala/dev/bosatsu/PackageMap.scala
+++ b/core/src/main/scala/dev/bosatsu/PackageMap.scala
@@ -654,8 +654,7 @@ object PackageMap {
 
   /** Backward compatible runtime predef handle.
     */
-  val predefCompiled: Package.Inferred =
-    predefCompiledForMode(CompileOptions.Mode.Emit)
+  def predefCompiled: Package.Inferred = predefCompiledEmit
 
   private def predefImportsFromExports(
       exports: List[ExportedName[Referant[Kind.Arg]]]
--- a/core/src/test/scala/dev/bosatsu/ToolAndLibCommandTest.scala
+++ b/core/src/test/scala/dev/bosatsu/ToolAndLibCommandTest.scala
@@ -2044,6 +2044,7 @@ main = depBox
       case Left(err) =>
         val msg = Option(err.getMessage).getOrElse(err.toString)
         assert(msg.contains("todo"), msg)
+        assert(msg.contains("only available in type-check mode"), msg)
     }
   }
 
@@ -2894,6 +2895,7 @@ main = 0
       case Left(err) =>
         val msg = Option(err.getMessage).getOrElse(err.toString)
         assert(msg.contains("todo"), msg)
+        assert(msg.contains("only available in type-check mode"), msg)
     }
   }
 
--- a/core/src/main/scala/dev/bosatsu/PackageError.scala
+++ b/core/src/main/scala/dev/bosatsu/PackageError.scala
@@ -73,6 +73,16 @@ object PackageError {
             .grouped + Doc.char('?')
     }
 
+  private def checkModeTodoHint(name: Identifier): Doc =
+    name match {
+      case b: Identifier.Bindable if b.asString == "todo" =>
+        Doc.hardLine + Doc.text(
+          "hint: `todo` is only available in type-check mode (`tool check`/`lib check`) and is not available in emit commands (`show`/`build`/`transpile`/`test`)."
+        )
+      case _ =>
+        Doc.empty
+    }
+
   private def nearestConstructorsDoc(
       suggestions: List[Identifier.Constructor]
   ): Doc =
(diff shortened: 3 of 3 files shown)


core/src/main/scala/dev/bosatsu/PackageMap.scala
```diff
@@ -586,65 +594,124 @@ object PackageMap {
...
+  def predefCompiledForMode(mode: CompileOptions.Mode): Package.Inferred =
+    mode match {
+      case CompileOptions.Mode.Emit          => predefCompiledEmit
+      case CompileOptions.Mode.TypeCheckOnly => predefCompiledTypeCheckOnly
+    }
 
-  private val predefImports: Import[PackageName, Unit] =
+  /** Backward compatible runtime predef handle.
+    */
+  val predefCompiled: Package.Inferred =
```
johnynek: can this be a `def predefCompiled: Package.Inferred = predefCompiledEmit`

core/src/test/scala/dev/bosatsu/ToolAndLibCommandTest.scala
```diff
@@ -2002,6 +2002,51 @@ main = depBox
...
+        "src/Todo/Foo.bosatsu",
+        "--value",
+        "Todo/Foo::main"
+      )
+    ) match {
+      case Right(out) =>
+        fail(s"expected emit-mode rejection, got: $out")
+      case Left(err) =>
+        val msg = Option(err.getMessage).getOrElse(err.toString)
+        assert(msg.contains("todo"), msg)
```
johnynek, line 2046: we need to have a hint. Unused `todo` suggests a left around `todo` that is only valid in check mode. Let's have a nice message and test for it.

core/src/main/scala/dev/bosatsu/PackageMap.scala
johnynek: Done. I changed this to `def predefCompiled: Package.Inferred = predefCompiledEmit` in `PackageMap.scala`.

core/src/test/scala/dev/bosatsu/ToolAndLibCommandTest.scala
johnynek: Implemented. I added a targeted hint for `todo` when it is unknown in emit-mode flows (`PackageError.checkModeTodoHint`) and attached it to unknown-name rendering. I also updated `ToolAndLibCommandTest` to assert the new hint text in both emit-mode rejection paths (`tool show` and `lib show`).

johnynek:
Addressed both review comments with code changes and test updates. Validation run: `sbt "coreJVM/test:compile"`, `sbt "coreJVM/testOnly dev.bosatsu.ToolAndLibCommandTest"`, `sbt "cli/test"`, and `sbt "coreJVM/testOnly dev.bosatsu.Z3SolverIntegrationTest"` all passed. (One broad `coreJVM/test` run timed out once in `Z3SolverIntegrationTest` query A; isolated rerun passed.)

PR #1804, opened 2026-02-25
Add JVM evaluator benchmark and optimize int/JSON paths

johnynek:
## Summary
- add `EvalBenchmarkMain` to benchmark JVM evaluator workloads (function calls, tail loops, non-tail recursion, integer ops, string ops) with loop mode for async-profiler
- optimize evaluator/runtime hot paths from profiling (local env path, integer fast paths, constructor allocation path)
- follow up with mixed-int JSON compatibility fix so `Bosatsu/Json::JInt` handles both `Integer` and `BigInteger`
- avoid repeated `longValue` extraction in integer add/sub/mul overflow branches
- update `eval_optimization.md` with checkpoint/test status

## Validation
- `sbt coreJVM/test` (pass)
- `sbt cli/test` (pass)
- `sbt -batch 'coreJVM/testOnly dev.bosatsu.ToolAndLibCommandTest'` (pass; verifies direct `Bosatsu/Json::Json` path)

### DIFF
--- a/eval_optimization.md
+++ b/eval_optimization.md
@@ -57,12 +57,15 @@
 
 ### Checkpoint 2
 - Change:
-  - Investigated additional hotspots (`LongMap` mutations/lookups, dynamic dispatch stubs) with async-profiler.
-  - No further low-risk changes produced a clear significant gain beyond checkpoint 1.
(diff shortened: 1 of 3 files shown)


eval_optimization.md
```diff
@@ -0,0 +1,88 @@
+# Eval Optimization Log
```
johnynek, line 1: please remove this file from the final commit. This is a scratchpad doc we don't need in the repo.

PR #1808, opened 2026-02-25
Refactor TotalityCheck to typed matches with inhabitedness-aware coverage pruning

### LINKED ISSUE #1769: Design refactoring TotalityChecker to run on TypedExpr[Declaration]

johnynek:
Implemented issue #1769 per the merged design by moving totality checking to `TypedExpr[Declaration]` and running it after inference. `TotalityCheck` now traverses typed expressions, uses each match scrutinee type (`arg.getType`), computes inhabitedness via `Inhabitedness.check`/`checkMatch`, and prunes only definitively uninhabited branches/patterns from missing-coverage requirements while keeping `Unknown` conservative. Unreachable diagnostics remain shadowing/order-based and guard behavior is unchanged. Added typed-path invariant handling with `InternalInvariantViolation` (for constructor/arity invariant breaks) instead of crashing, and updated `PackageError` rendering accordingly.

To preserve existing structural pattern diagnostics, added `checkExprPatterns` for pre-inference string/list structural lint and wired it in `Package.inferBodyUnopt`; typed totality now runs on inferred lets (skipping lets already flagged by structural lint to avoid duplicate/misleading reports). Updated package pipeline and error plumbing in `Package.scala`/`PackageError.scala`.

Tests were updated and expanded in scope files from the design doc: added deterministic typed suite `TypedTotalityTest` (Option[Never], Result[Never,_], guarded coverage behavior, vacuous uninhabited scrutinee, shadowing unreachable, structural lint stability, internal invariant reporting), extended `WellTypedGen` with typed-match extraction (`TypedMatchCase`, `typedMatchCaseGen`), added typed totality properties in `TotalityTest`, and updated `ErrorMessageTest` expectations for invariant-reporting behavior.

Validation run results in this environment: `sbt "coreJVM/test:compile"` passed; focused suites including `TypedTotalityTest`, `TotalityTest`, `WellTypedTests`, and `ErrorMessageTest` passed; full `coreJVM/test; cli/test` execution was run with passing output across touched and broad suites.

Fixes #1769

Implements design doc: [docs/design/1769-design-refactoring-totalitychecker-to-run-on-typedexpr-declaration.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/1769-design-refactoring-totalitychecker-to-run-on-typedexpr-declaration.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/1774

### DIFF
--- a/core/src/main/scala/dev/bosatsu/Package.scala
+++ b/core/src/main/scala/dev/bosatsu/Package.scala
@@ -407,14 +407,34 @@ object Package {
           }
 
           val fullTypeEnv = importedTypeEnv ++ typeEnv
-          val totalityCheck: ValidatedNel[PackageError, Unit] =
-            lets
-              .traverse_ { case (_, _, expr) =>
-                TotalityCheck(fullTypeEnv).checkExpr(expr)
-              }
-              .leftMap { errs =>
-                errs.map(PackageError.TotalityCheckError(p, _))
-              }
+          val structuralPatternCheckByLet: List[
+            (Identifier.Bindable, ValidatedNel[PackageError, Unit])
+          ] =
+            lets.map { case (name, _, expr) =>
+              val check =
+                TotalityCheck(fullTypeEnv)
+                  .checkExprPatterns(expr)
+                  .leftMap(
+                    _.map { case (tag, err) =>
+                      PackageError.TotalityCheckError(
+                        p,
+                        TotalityCheck.InvalidPattern(
+                          TotalityCheck.syntheticMatch(tag),
+                          err
+                        )
+                      ): PackageError
+                    }
+                  )
+              (name, check)
+            }
+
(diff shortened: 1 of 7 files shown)


johnynek:
replaced by #1789

PR #1809, opened 2026-02-25
Implement #1778: persist source hash/region provenance in proto and Matchless

### LINKED ISSUE #1778: Design changes to the code to enable coverage checking

johnynek:
Implemented the accepted #1778 design to preserve source provenance for implementations while keeping interfaces metadata-free and deterministic.

What changed:
- `TypedAst.proto`:
  - Added `Region` message.
  - Added `PackageSourceHash` wrapper message.
  - Added optional `region` on `TypedExpr`.
  - Added optional `sourceHash` on `Package`.
- Package provenance plumbing:
  - Added package source metadata support in `Package` (`sourceHashIdent`, helpers, empty-hash sentinel).
  - `CompilerApi` now computes and attaches source hash idents from UTF-8 source bytes for compiled implementations.
- Proto conversion and compatibility:
  - `ProtoConverter` now encodes/decodes typed-expression regions and package source hash.
  - Legacy artifacts without new fields still decode.
  - Invalid present hash idents fail decode clearly.
  - Decoded typed package handling now preserves provenance (`Package.Typed[Any]`) through proto and IO flows.
- Matchless provenance model:
  - Added required `Matchless.SourceInfo(packageHashIdent, region)` on Matchless expressions/booleans.
  - Added sentinel fallback (`SourceInfo.empty`) for missing legacy provenance.
  - Lowering (`Matchless.fromLet` / `MatchlessFromTypedExpr`) now threads package hash + region provenance.
  - Rewrite/optimization paths preserve/inherit source info.
- Platform IO propagation:
  - Updated `PlatformIO`, JVM/JS impls, and `MemoryMain` signatures/flows to preserve decoded provenance metadata.
- Tests:
  - Expanded `ProtoConverterTest` for region/hash roundtrips, legacy decode behavior, invalid hash rejection, and interface-byte determinism/permutation invariance.
  - Added Matchless provenance tests (`MatchlessTests`, `MatchlessInterfaceTest`).
  - Added interface-only library metadata exclusion test (`LibConfigTest`).
  - Added package file provenance roundtrip assertions (`IOPlatformIOTest`).
  - Updated dependent tests for new Matchless/source metadata behavior (`Issue1633Test`, normalization helpers, etc.).
  - Updated `ClangGenTest` Ackermann golden hash to the new deterministic generated output.

Validation run:
  - `coreJVM/test` passed.
  - Initial `cli/test` failed only at `dev.bosatsu.codegen.clang.ClangGenTest` golden hash mismatch; fixed expected hash.
- Re-ran `sbt "cli/test"` -> passed.
- Re-ran `sbt "coreJVM/test:compile; coreJVM/testOnly dev.bosatsu.MatchlessTest"` after final test-file cleanup -> passed.

Fixes #1778

Implements design doc: [docs/design/1778-design-changes-to-the-code-to-enable-coverage-checking.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/1778-design-changes-to-the-code-to-enable-coverage-checking.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/1779

### DIFF
--- a/cli/src/main/scala/dev/bosatsu/IOPlatformIO.scala
+++ b/cli/src/main/scala/dev/bosatsu/IOPlatformIO.scala
@@ -226,7 +226,7 @@ object IOPlatformIO extends PlatformIO[IO, JPath] {
   def readInterfacesAndPackages(
       ifacePaths: List[Path],
       packagePaths: List[Path]
-  ): IO[(List[Package.Interface], List[Package.Typed[Unit]])] =
+  ): IO[(List[Package.Interface], List[Package.Typed[Any]])] =
     (
       ifacePaths.traverse(read[proto.Interfaces](_)),
       packagePaths.traverse(read[proto.Packages](_))
@@ -242,7 +242,7 @@ object IOPlatformIO extends PlatformIO[IO, JPath] {
   def readInterfaces(paths: List[Path]): IO[List[Package.Interface]] =
     readInterfacesAndPackages(paths, Nil).map(_._1)
 
-  def readPackages(paths: List[Path]): IO[List[Package.Typed[Unit]]] =
+  def readPackages(paths: List[Path]): IO[List[Package.Typed[Any]]] =
     readInterfacesAndPackages(Nil, paths).map(_._2)
 
   def readLibrary(path: Path): IO[Hashed[Algo.Blake3, proto.Library]] =
--- a/cli/src/test/scala/dev/bosatsu/IOPlatformIOTest.scala
+++ b/cli/src/test/scala/dev/bosatsu/IOPlatformIOTest.scala
@@ -5,10 +5,31 @@ import cats.implicits._
 import cats.effect.{IO, Resource}
 import java.io.File
 import java.nio.file.{Files, Path, Paths}
+import dev.bosatsu.hashing.Algo
+import dev.bosatsu.IorMethods.IorExtension
 import org.scalacheck.Gen
 import org.scalacheck.Prop.forAll
 
 class IOPlatformIOTest extends munit.ScalaCheckSuite {
+  private def stripTypedProvenance[A](pack: Package.Typed[A]): Package.Typed[Unit] = {
+    val noExprTags = Package.typedFunctor.map(pack)(_ => ())
+    Package.setProgramFrom(noExprTags, ())
+  }
+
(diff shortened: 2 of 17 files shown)


johnynek:
this duplicated #1790 by some race conditions in mergexo

PR #1812, opened 2026-02-26
Implement Bytes type and binary IO externals for issue #1796

### LINKED ISSUE #1796: design a Bytes type

johnynek:
Implemented the merged design for a new immutable external `Bosatsu/IO/Bytes::Bytes` type and binary handle APIs.

What changed:
- Added new package `test_workspace/Bosatsu/IO/Bytes.bosatsu` with the designed API (`from_List_Int`, `from_Array_Int`, `to_List_Int`, `to_Array_Int`, `size/get/fold/concat/slice/find/starts_with/ends_with`) plus Bytes semantics tests.
- Added binary IO extern signatures to `test_workspace/Bosatsu/IO/Core.bosatsu`: `read_bytes`, `write_bytes`, `read_all_bytes`, `copy_bytes`.
- Exported `Bosatsu/IO/Bytes` in `test_workspace/core_alpha_conf.json` (and updated `next_version` for compatibility checks).
- Implemented JVM evaluator/runtime support in `core/src/main/scala/dev/bosatsu/Predef.scala`:
  - New `BytesValue(data, offset, len)` representation and bytes extern registration.
  - Implementations for all Bytes operations.
  - Implementations for IO/Core binary ops (`read_bytes`, `write_bytes`, `read_all_bytes`, `copy_bytes`) and handle internals updated to carry binary streams.
- Implemented Python runtime support in `test_workspace/ProgExt.py` with `_BosatsuBytes` view model and all new Bytes/IO-Core extern behaviors; wired in `test_workspace/Prog.bosatsu_externals`.
- Implemented C runtime support:
  - New files `c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Bytes.c/.h`.
  - Extended `c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Core.c/.h` for binary IO externs.
  - Wired build/install in `c_runtime/Makefile`.
- Added Bytes rendering/parsing support to tooling:
  - `core/src/main/scala/dev/bosatsu/ValueToJson.scala` and tests in `ValueToJsonTest.scala`.
  - `core/src/main/scala/dev/bosatsu/ValueToDoc.scala` and tests in `ValueToDocTest.scala`.
  - CLI/library JSON coverage for Bytes in `ToolAndLibCommandTest.scala`.
- Added JVM evaluation coverage for Bytes extern behavior in `EvaluationTest.scala`.
- Updated workflow command-generation fixtures and parity tests to include new required test inputs (`Bosatsu/Collection/Array` and `Bosatsu/IO/Bytes`) in:
  - `.github/workflows/ci.yml`
  - `test_workspace/Bosatsu/Example/Json/Github/Workflows/Util.bosatsu`
  - `cli/src/test/scala/dev/bosatsu/PathModuleTest.scala`

Validation run:
- Required command passed: `scripts/test_basic.sh`
- Additional runtime build sanity check passed: `make PROFILE=debug` in `c_runtime`.

Fixes #1796

Implements design doc: [docs/design/1796-design-a-bytes-type.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/1796-design-a-bytes-type.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/1797

### DIFF
--- a/c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Bytes.c
+++ b/c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Bytes.c
@@ -500,8 +500,14 @@ BValue ___bsts_g_Bosatsu_l_IO_l_Bytes_l_find__Bytes(BValue bytes, BValue needle,
   return bsts_integer_from_int(-1);
 }
 
-BValue ___bsts_g_Bosatsu_l_IO_l_Bytes_l_uft8__bytes__from__String(BValue str)
+BValue ___bsts_g_Bosatsu_l_IO_l_Bytes_l_utf8__bytes__from__String(BValue str)
 {
+  // Tiny strings (<= 5 bytes) are immediate values and their byte view points
+  // into the current stack slot, so we must copy those bytes.
+  enum
+  {
+    BSTS_TINY_STRING_MAX_LEN = 5
+  };
   size_t raw_len = bsts_string_utf8_len_ref(&str);
   if (raw_len == 0)
   {
@@ -514,9 +520,15 @@ BValue ___bsts_g_Bosatsu_l_IO_l_Bytes_l_uft8__bytes__from__String(BValue str)
 
   int len = (int)raw_len;
   const char *src = bsts_string_utf8_bytes_ref(&str);
-  uint8_t *data = bsts_bytes_alloc_data(len);
-  memcpy(data, src, (size_t)len);
-  return bsts_bytes_wrap(data, 0, len);
+  if (raw_len <= (size_t)BSTS_TINY_STRING_MAX_LEN)
+  {
+    uint8_t *data = bsts_bytes_alloc_data(len);
+    memcpy(data, src, (size_t)len);
+    return bsts_bytes_wrap(data, 0, len);
+  }
+
+  // For non-tiny strings we can share the UTF-8 storage without copying.
+  return bsts_bytes_wrap((uint8_t *)src, 0, len);
 }
 
 BValue ___bsts_g_Bosatsu_l_IO_l_Bytes_l_utf8__bytes__to__String(BValue bytes)
@@ -540,10 +552,12 @@ BValue ___bsts_g_Bosatsu_l_IO_l_Bytes_l_utf8__bytes__to__String(BValue bytes)
     idx += width;
   }
 
-  return alloc_enum1(1, bsts_string_from_utf8_bytes_copy((size_t)b->len, (const char *)data));
+  // This constructor shares storage for long strings and only copies for
+  // small inline string representations.
+  return alloc_enum1(1, bsts_string_from_utf8_bytes_static((size_t)b->len, (const char *)data));
 }
 
-BValue ___bsts_g_Bosatsu_l_IO_l_Bytes_l_utf__Char__at(BValue bytes, BValue index)
+BValue ___bsts_g_Bosatsu_l_IO_l_Bytes_l_utf8__Char__at(BValue bytes, BValue index)
 {
   BSTS_Bytes *b = bsts_bytes_unbox(bytes);
   if (!bsts_bytes_index_in_range(index, b->len))
--- a/core/src/main/scala/dev/bosatsu/Predef.scala
+++ b/core/src/main/scala/dev/bosatsu/Predef.scala
@@ -140,8 +140,8 @@ object Predef {
       )
       .add(
         ioBytesPackageName,
-        "uft8_bytes_from_String",
-        FfiCall.Fn1(PredefImpl.uft8_bytes_from_String(_))
+        "utf8_bytes_from_String",
+        FfiCall.Fn1(PredefImpl.utf8_bytes_from_String(_))
       )
       .add(
         ioBytesPackageName,
@@ -150,8 +150,8 @@ object Predef {
       )
       .add(
         ioBytesPackageName,
-        "utf_Char_at",
-        FfiCall.Fn2(PredefImpl.utf_Char_at(_, _))
+        "utf8_Char_at",
+        FfiCall.Fn2(PredefImpl.utf8_Char_at(_, _))
       )
 
   private def addIoCoreExternals(externals: Externals): Externals =
@@ -2945,7 +2945,7 @@ object PredefImpl {
     }
   }
 
-  def uft8_bytes_from_String(strValue: Value): Value =
+  def utf8_bytes_from_String(strValue: Value): Value =
     strValue match {
       case Value.Str(str) =>
         val utf8 = str.getBytes(StandardCharsets.UTF_8)
@@ -2965,7 +2965,7 @@ object PredefImpl {
     }
   }
 
-  def utf_Char_at(bytesValue: Value, index: Value): Value = {
+  def utf8_Char_at(bytesValue: Value, index: Value): Value = {
     val bytes = asBytes(bytesValue)
     inRangeIndex(i(index), bytes.len) match {
       case Some(idx) =>
--- a/test_workspace/Bosatsu/IO/Bytes.bosatsu
+++ b/test_workspace/Bosatsu/IO/Bytes.bosatsu
@@ -20,9 +20,9 @@ export (
   starts_with_Bytes,
   ends_with_Bytes,
   find_Bytes,
-  uft8_bytes_from_String,
+  utf8_bytes_from_String,
   utf8_bytes_to_String,
-  utf_Char_at,
+  utf8_Char_at,
   index_in_range_Bytes,
 )
 
@@ -42,9 +42,9 @@ external def slice_Bytes(bytes: Bytes, start: Int, end: Int) -> Bytes
 external def starts_with_Bytes(bytes: Bytes, prefix: Bytes) -> Bool
 external def ends_with_Bytes(bytes: Bytes, suffix: Bytes) -> Bool
 external def find_Bytes(bytes: Bytes, needle: Bytes, start: Int) -> Int
-external def uft8_bytes_from_String(str: String) -> Bytes
+external def utf8_bytes_from_String(str: String) -> Bytes
 external def utf8_bytes_to_String(bytes: Bytes) -> Option[String]
-external def utf_Char_at(bytes: Bytes, idx: Int) -> Option[Char]
+external def utf8_Char_at(bytes: Bytes, idx: Int) -> Option[Char]
 
 def get_Bytes(bytes: Bytes, idx: Int) -> Option[Int]:
   get_map_Bytes(bytes, idx, _ -> None, i -> Some(i))
@@ -67,7 +67,7 @@ norm_expected = [255, 0, 1, 255, 0, 1, 255]
 norm_bytes = from_List_Int(norm_source)
 base = from_List_Int([1, 2, 3, 4, 5])
 needle = from_List_Int([3, 4])
-utf8_sample = uft8_bytes_from_String("Aé🙂")
+utf8_sample = utf8_bytes_from_String("Aé🙂")
 
 tests = TestSuite("Bytes tests", [
   Assertion(size_Bytes(empty_Bytes) matches 0, "empty size"),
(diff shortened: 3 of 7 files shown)


johnynek:
Fixed the CI failure in `test_cli.sh` by adding the missing module inputs needed by `Bosatsu/IO/Core.bosatsu`: `test_workspace/Bosatsu/Collection/Array.bosatsu` and `test_workspace/Bosatsu/IO/Bytes.bosatsu`. Verified locally that `./test_cli.sh` now passes, and the required pre-push check `scripts/test_basic.sh` also passes.

johnynek:
Implemented fixes for the CI regressions reported on PR #1812 and reproduced the failing paths locally before/after changes. Changes made: (1) `test_workspace/ProgExt.py`: `to_Array_Int` now returns Bosatsu Array shape `(tuple(data), 0, len)` instead of a raw Python list, fixing the Python import/runtime `TypeError` in `Bosatsu/IO/Bytes`. (2) `core/src/main/scala/dev/bosatsu/codegen/python/PythonGen.scala`: fixed Array preallocation codegen used by indexed writes by replacing `[] * n` generation with `[0] * n` generation at the three preallocation sites, fixing Python import/runtime `IndexError` in transpiled `Bosatsu/Collection/Array`. (3) `core/src/main/scala/dev/bosatsu/Predef.scala`: removed JVM-only gating from `addIoBytesExternals` so `Bosatsu/IO/Bytes` externals are registered on Scala.js/Node too, fixing `no External for external def: Name(empty_Bytes)` in node/coreJS paths. Validation run locally: `sbt "cli/assembly"`, CI-equivalent transpile command for PY, `python3 -m unittest discover pyout -v --pattern "*.py"`, `sbt "cliJSJS/fullOptJS"`, CI-equivalent `./bosatsu_node tool test ...`, `sbt "coreJS/testOnly dev.bosatsu.EvaluationTest; coreJS/testOnly dev.bosatsu.ToolAndLibCommandTest"`, and required pre-push check `scripts/test_basic.sh` (completed with PASS).

test_workspace/Bosatsu/IO/Core.bosatsu
```diff
@@ -111,6 +120,8 @@ external now_mono: Prog[IOError, Duration]
 external def sleep(d: Duration) -> Prog[IOError, Unit]
 
 _ = now_mono
+bytes_identity = (b: Bytes) -> b
+_ = bytes_identity
```
johnynek: we don't need this. Let's remove it.

core/src/main/scala/dev/bosatsu/ValueToDoc.scala
```diff
@@ -183,6 +188,25 @@ case class ValueToDoc(getDefinedType: Type.Const => Option[DefinedType[Any]]) {
...
                   Left(IllTyped(revPath.reverse, tpe, other))
               }
+            case Type.TyConst(`bytesTypeConst`) =>
+              {
+                case ExternalValue(bytes: PredefImpl.BytesValue) =>
+                  val items = List.newBuilder[Doc]
+                  var idx = 0
+                  while (idx < bytes.len) {
+                    val intValue = bytes.data(bytes.offset + idx).toInt & 0xff
+                    items += Doc.str(intValue)
```
johnynek: let's render these as hex, not integers, and use spaces to separate and use `<` to separate, so it would look something like `<e3 23 00>` when we have bytes.

test_workspace/Bosatsu/IO/Bytes.bosatsu
```diff
@@ -0,0 +1,106 @@
...
+external def size_Bytes(bytes: Bytes) -> Int
+external def get_map_Bytes[a](bytes: Bytes, idx: Int, default: Unit -> a, fn: Int -> a) -> a
+external def get_or_Bytes(bytes: Bytes, idx: Int, default: Unit -> Int) -> Int
+external def foldl_Bytes[a](bytes: Bytes, init: a, fn: (a, Int) -> a) -> a
+external def concat_all_Bytes(chunks: List[Bytes]) -> Bytes
+external def slice_Bytes(bytes: Bytes, start: Int, end: Int) -> Bytes
+external def starts_with_Bytes(bytes: Bytes, prefix: Bytes) -> Bool
+external def ends_with_Bytes(bytes: Bytes, suffix: Bytes) -> Bool
+external def find_Bytes(bytes: Bytes, needle: Bytes, start: Int) -> Int
+
```
johnynek, line 48: we should add:

```
external def uft8_bytes_from_String(str: String) -> Bytes
```
which ideally should be a near zero cost operation, at least on the C platform. We should be able to just cast the data into a Bytes if we use the right representation. At least for long strings.

also
```
external def utf8_bytes_to_String(b: Bytes) -> Option[String]
```
which if it is valid utf8 should also share memory. Obviously we would have to check all the bytes to validate that the data is utf8.

and finally:
```
external def utf_Char_at(b: Bytes, idx: Int) -> Option[Char]
```
So, it will return a Char if it is a valid utf8 codepoint.

test_workspace/Bosatsu/IO/Core.bosatsu
johnynek: Removed the extra `bytes_identity` helper from `Bosatsu/IO/Core.bosatsu` as requested. To keep `Bytes` imported for external signatures without adding API surface, this now uses a single minimal typed sentinel: `_ = (_: Bytes) -> now_mono`.

core/src/main/scala/dev/bosatsu/ValueToDoc.scala
johnynek: Updated `ValueToDoc` bytes rendering to hex with spaces and angle brackets (for example `<e3 23 00>`). Also updated `ValueToDocTest` expectation accordingly (`<00 0a ff>`).

test_workspace/Bosatsu/IO/Bytes.bosatsu
johnynek: Added the requested UTF-8 helpers to `Bosatsu/IO/Bytes`: `uft8_bytes_from_String`, `utf8_bytes_to_String`, and `utf_Char_at`. Implemented them across runtimes (JVM `Predef.scala`, Python `ProgExt.py` + `Prog.bosatsu_externals`, C runtime `.c/.h`) with UTF-8 validation semantics, and added Bosatsu-level tests for roundtrip, invalid UTF-8, and character-at-byte-index behavior.

johnynek:
Implemented all requested review changes and reran the required pre-push gate `scripts/test_basic.sh`; it passes on this branch.

test_workspace/Bosatsu/IO/Bytes.bosatsu
```diff
@@ -39,6 +42,9 @@ external def slice_Bytes(bytes: Bytes, start: Int, end: Int) -> Bytes
 external def starts_with_Bytes(bytes: Bytes, prefix: Bytes) -> Bool
 external def ends_with_Bytes(bytes: Bytes, suffix: Bytes) -> Bool
 external def find_Bytes(bytes: Bytes, needle: Bytes, start: Int) -> Int
+external def uft8_bytes_from_String(str: String) -> Bytes
```
johnynek: there is a typo in this name. It should be `utf8` not `uft8`, fix this here and look for other similar typos.

test_workspace/Bosatsu/IO/Core.bosatsu
```diff
@@ -119,9 +119,7 @@ external now_wall: Prog[IOError, Instant]
 external now_mono: Prog[IOError, Duration]
 external def sleep(d: Duration) -> Prog[IOError, Unit]
 
-_ = now_mono
-bytes_identity = (b: Bytes) -> b
-_ = bytes_identity
+_ = (_: Bytes) -> now_mono
```
johnynek: now_mono should be `exported` it shouldn't be here in this function.

test_workspace/Prog.bosatsu_externals
```diff
@@ -46,6 +46,9 @@
     slice_Bytes: ProgExt.slice_Bytes,
     starts_with_Bytes: ProgExt.starts_with_Bytes,
     ends_with_Bytes: ProgExt.ends_with_Bytes,
-    find_Bytes: ProgExt.find_Bytes
+    find_Bytes: ProgExt.find_Bytes,
+    uft8_bytes_from_String: ProgExt.uft8_bytes_from_String,
```
johnynek: typo

c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Bytes.c
```diff
@@ -407,3 +499,65 @@ BValue ___bsts_g_Bosatsu_l_IO_l_Bytes_l_find__Bytes(BValue bytes, BValue needle,
 
   return bsts_integer_from_int(-1);
 }
+
+BValue ___bsts_g_Bosatsu_l_IO_l_Bytes_l_uft8__bytes__from__String(BValue str)
```
johnynek: typo: uft8 should be utf8

c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Bytes.c
```diff
@@ -407,3 +499,65 @@ BValue ___bsts_g_Bosatsu_l_IO_l_Bytes_l_find__Bytes(BValue bytes, BValue needle,
...
+  }
+  if (raw_len > (size_t)INT_MAX)
+  {
+    return bsts_bytes_empty();
+  }
+
+  int len = (int)raw_len;
+  const char *src = bsts_string_utf8_bytes_ref(&str);
+  uint8_t *data = bsts_bytes_alloc_data(len);
+  memcpy(data, src, (size_t)len);
```
johnynek: we should be able to do this without a memcpy. Recall: we never mutate BValue strings, if they are already stored as a char array + offset + length, that can be the same representation as a Bytes array. So, when we have that kind of representation, we should just be able to copy the pointer or potentially even simply cast the BValue if the shape is identical in memory.

We want a cheap operation here so we can very cheaply convert a String to Bytes for sending on Handles (and in the future on the network). Maximizing our zero-copy operations is ideal.

If there is some reason zero copy can't be made to work here, add a comment explaining why.

c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Bytes.c
```diff
@@ -407,3 +499,65 @@ BValue ___bsts_g_Bosatsu_l_IO_l_Bytes_l_find__Bytes(BValue bytes, BValue needle,
...
+    int cp = 0;
+    int width = 0;
+    if (!bsts_decode_utf8_code_point(data, b->len, idx, &cp, &width))
+    {
+      return alloc_enum0(0);
+    }
+    idx += width;
+  }
+
+  return alloc_enum1(1, bsts_string_from_utf8_bytes_copy((size_t)b->len, (const char *)data));
```
johnynek: could we avoid copying? Can't we just make this a large string and copy the pointer, offset and length? Zero copy is best. If we can't for some reason, add a comment explaining why.

core/src/main/scala/dev/bosatsu/Predef.scala
```diff
@@ -2873,6 +2945,39 @@ object PredefImpl {
...
+
+  def utf8_bytes_to_String(bytesValue: Value): Value = {
+    val bytes = asBytes(bytesValue)
+    decodeUtf8Slice(bytes.data, bytes.offset, bytes.len) match {
+      case Some(str) => VOption.some(Value.Str(str))
+      case None      => VOption.none
+    }
+  }
+
+  def utf_Char_at(bytesValue: Value, index: Value): Value = {
```
johnynek: this should be utf8_Char_at

test_workspace/Prog.bosatsu_externals
```diff
@@ -46,6 +46,9 @@
...
+    utf8_bytes_to_String: ProgExt.utf8_bytes_to_String,
+    utf_Char_at: ProgExt.utf_Char_at
```
johnynek: this should be utf8_Char_at

test_workspace/Bosatsu/IO/Bytes.bosatsu
```diff
@@ -39,6 +42,9 @@ external def slice_Bytes(bytes: Bytes, start: Int, end: Int) -> Bytes
...
+external def utf8_bytes_to_String(bytes: Bytes) -> Option[String]
+external def utf_Char_at(bytes: Bytes, idx: Int) -> Option[Char]
```
johnynek: this should be utf8_Char_at. We have to be clear, there is also utf16 and utf32.

test_workspace/Bosatsu/IO/Bytes.bosatsu
johnynek: Fixed. Renamed `uft8_bytes_from_String` to `utf8_bytes_from_String` and corrected matching `uft8` typos across the related declarations and runtime/external mappings.

test_workspace/Bosatsu/IO/Core.bosatsu
johnynek: Updated. `now_mono` is no longer used in that function position; the file now keeps the import checker satisfied without embedding it in that function shape.

test_workspace/Prog.bosatsu_externals
johnynek: Fixed the typo in `test_workspace/Prog.bosatsu_externals` to match the corrected utf8 external names.

c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Bytes.c
johnynek: Fixed. Renamed the C runtime symbol usage from `uft8` to `utf8` to match the Bosatsu external names.

c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Bytes.c
johnynek: Implemented zero-copy for `utf8_bytes_from_String` on non-tiny strings in C. Tiny strings still copy because their storage shape is not safe to alias directly; I added a concise comment explaining that constraint.

c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Bytes.c
johnynek: Updated `utf8_bytes_to_String` to avoid unnecessary copying: it now validates UTF-8 and uses the static UTF-8 string constructor path so long-string storage can be shared; I added a short comment on representation constraints.

core/src/main/scala/dev/bosatsu/Predef.scala
johnynek: Fixed. Renamed the Scala Predef external mapping to `utf8_Char_at`.

test_workspace/Prog.bosatsu_externals
johnynek: Fixed. Updated `test_workspace/Prog.bosatsu_externals` to use `utf8_Char_at`.

test_workspace/Bosatsu/IO/Bytes.bosatsu
johnynek: Fixed. Renamed the Bytes API entry to `utf8_Char_at` in `Bosatsu/IO/Bytes.bosatsu` for explicit UTF-8 naming.

johnynek:
Addressed all review comments in this round: utf8 typo/name corrections across Bosatsu/Scala/Python/C, removal of `now_mono` from that function usage site, and zero-copy-oriented C utf8 bytes/string conversion improvements with comments where representation limits require copying. Re-ran `scripts/test_basic.sh` and it passes.

Issue #1818, opened 2026-02-26
Make sure synthetic names are unambiguously identifiable


johnynek:
Currently, we don't use a different Identifier subclass for Synthentic. We just put an unparseable name inside a `Name`. But it's not tested that it can't collide with `Backticked` for instance. So I think a user can write a backticked name that starts with an underscore and we may not distinguish it all the places we need to.

Secondly, in some code paths, we just use `Expr.nameIterator` to generate fresh names that don't collide with the current scope. We should have a standard way to do this that is using synthetic names any time we need to make internal names to the compiler. There should be a method for synthetics iterating up from smallest values and we can filter than like we are doing with Expr.nameIterator, but it should be in `object Bindable` I think.

We should have tests that when we serialize to and from protobuf we can handle synthetics and not get them confused with backticked names that start with an `_`.

Ideally we can do this all without breaking currently serialized data except where that code might have already been inconsistent (for instance if we are currently having collisions of backticked and synthentic names in some cases, it's probably fine to just assume those are backticked in currently serialized code, if we have to).

johnynek:
MergeXO assigned an agent and started design work for issue #1818.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/1822

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #1818.

johnynek:
Opened implementation PR: https://github.com/johnynek/bosatsu/pull/1825

PR #1822, opened 2026-02-27
Design doc for #1818: Issue #1818 Design: Make sure synthetic names are unambiguously identifiable

johnynek:
Design doc.

Refs #1818

### DIFF
--- a/docs/design/1818-make-sure-synthetic-names-are-unambiguously-identifiable.md
+++ b/docs/design/1818-make-sure-synthetic-names-are-unambiguously-identifiable.md
@@ -26,7 +26,7 @@ _Issue: #1818 (https://github.com/johnynek/bosatsu/issues/1818)_
 
 ## Summary
 
-Design doc proposing an explicit Synthetic identifier subtype, kind-aware identifier identity semantics, centralized synthetic fresh-name iteration, migration of internal name-generation call sites, protobuf round-trip coverage, and compatibility/rollout guidance for issue #1818.
+Design doc proposing an explicit Synthetic identifier subtype, synthetic-aware/backward-compatible identifier identity semantics, centralized synthetic fresh-name iteration, migration of internal name-generation call sites, protobuf round-trip coverage, and compatibility/rollout guidance for issue #1818.
 
 ---
 issue: 1818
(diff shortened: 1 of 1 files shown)


docs/design/1818-make-sure-synthetic-names-are-unambiguously-identifiable.md
```diff
@@ -0,0 +1,225 @@
...
+
+This keeps protobuf schema unchanged and preserves compatibility for valid existing artifacts. Legacy artifacts that already lost information due prior collisions are treated as best-effort decode.
+
+## Problem statement
+
+Today synthetic names are represented as `Identifier.Name("_...")` and distinguished by convention.
+
+That creates three concrete problems:
+
+1. Identifier equality and ordering are keyed by `asString` only, so `Name("_x")` and `Backticked("_x")` can collide in `Map`/`Set`/`SortedMap` paths even though they represent different source-level identities.
```
johnynek: I think current code may leverage this, especially with operators. I think we leverage in the compiler that `operator +` and `Backticked("+")` are `==`.

I think to be fully compatible we would need do something like:
```scala
def eqv(a: Identifier, b: Identifier) =
  (a, b) match {
    case (Synthetic(sa), Synthetic(sb)) => sa == sb
    case (Synthetic(_), _) | (_, Synthetic(_)) => false
    case (notSynA, notSynB) => notSynA.asString == notSynB.asString
  }
```
Or something like that.

Then if protobuf serialization uses `sourceCodeRepr` then we should be okay, because `_foo` won't parse as a operator, backticked, or name.

A similar technique can be used to implement `Order[Bindable]` and `Hash[Bindable]` which allows us to consider all the syntactic forms equal as they were before.

Re-read the relevant code and see adjust this doc as needed.

johnynek: Updated the design doc to match this compatibility constraint. In the "Make identifier identity synthetic-aware and backward compatible" section, the proposal now uses a hybrid rule: `Synthetic` is disjoint from non-synthetics, while non-synthetic identifiers keep existing `asString`-based equality/order/hash behavior (including current operator/backticked interop). I also updated the implementation plan, tests, acceptance criteria, risks, and rollout notes to reflect that exact boundary.

johnynek:
Addressed the review feedback by editing `docs/design/1818-make-sure-synthetic-names-are-unambiguously-identifiable.md` to preserve non-synthetic compatibility while isolating synthetics. Ran required pre-push check `scripts/test_basic.sh`; it passed.

PR #1823, opened 2026-02-27
Add contextual `loop` recursion syntax with enforced tail recursion

### LINKED ISSUE #1172: syntax to require recursions to be tail recursive.

johnynek:
Implemented issue #1172 per the design doc by introducing a source-level recursion mode for matches: `Declaration.MatchKind` (`match` / `recur` / `loop`), parsing `loop <target>:` contextually, and preserving `loop` as a normal identifier outside recursion headers.

Compiler changes:
- Updated `Declaration.Match` to use the new match-mode ADT and pretty-print the correct header keyword.
- Extended parser/header logic to recognize `loop` only in match-header position while allowing fallback to normal identifier parsing (`def loop(...)`, `loop = ...`, `loop(...)`).
- Added `loop`/`recur` header support to parser hints for missing-colon diagnostics.
- Updated typed recursion checking so both `recur` and `loop` enter recursion legality checks; for `loop`, enforced tail-position self calls via `SelfCallKind` after legality succeeds.
- Added recursion error `LoopRequiresTailRecursion` and made the existing “no recursive call” diagnostic mode-aware (`recur` vs `loop`).
- Updated source conversion and test generators/mechanical helpers (`SourceConverter`, `Gen`, `LegacyDefRecursionCheck`, `WellTypedGen`) for the new match mode type.

Test/doc updates:
- Added parser tests for `loop` forms, tuple loop headers, and contextual-keyword behavior.
- Added parser-hint test for missing `:` after a `loop` header.
- Added typed recursion-check tests for accepted/rejected `loop` behavior (tail, non-tail, guards, function-value passing, tuple lexicographic case) and `recur` non-tail compatibility.
- Added error-message tests for `loop` no-recursion and non-tail diagnostics.
- Updated docs (`recursion.md`, `language_guide.md`) to explain when to use `recur` vs `loop` and show `loop` examples.

Validation run:
- `sbt "coreJVM/testOnly dev.bosatsu.ParserHintsTest dev.bosatsu.ParserTest dev.bosatsu.TypedExprRecursionCheckTest dev.bosatsu.ErrorMessageTest"`
- `sbt "coreJVM/testOnly dev.bosatsu.DefRecursionCheckTest"`
- `scripts/test_basic.sh` (required pre-push command)
- `sbt "doc; paradox"`
All passed.

Fixes #1172

Implements design doc: [docs/design/1172-syntax-to-require-recursions-to-be-tail-recursive.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/1172-syntax-to-require-recursions-to-be-tail-recursive.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/1799

### DIFF
--- a/core/src/main/scala/dev/bosatsu/Declaration.scala
+++ b/core/src/main/scala/dev/bosatsu/Declaration.scala
@@ -997,7 +997,9 @@ object Declaration {
   ) extends NonBinding
 
   private def matchHeaderKeyword(keyword: String): P[Unit] =
-    (P.string(keyword) <* P.charIn(Set(' ', '\t')).peek).void
+    // Backtrack only at keyword boundary so `loop(...)` stays a normal
+    // identifier/apply expression while `loop <target>:` parses as a header.
+    ((P.string(keyword) <* P.charIn(Set(' ', '\t')).peek).backtrack).void
 
   val matchKindParser: P[MatchKind] =
     matchHeaderKeyword("match")
@@ -1260,9 +1262,10 @@ object Declaration {
     }
 
     val left: Indy[(MatchKind, NonBinding)] = Indy { indent =>
-      // Keep `loop` contextual: if header parsing fails, allow fallback to normal
-      // identifier parsing (`def loop(...)`, `loop = ...`, `loop(...)`).
-      (((matchKindParser <* spaces) ~ arg(indent)).backtrack <* maybeSpace)
+      // Keep `loop` contextual in declaration position:
+      // - `loop = ...` should remain a normal binding
+      // - malformed recursion headers should commit once the header is chosen
+      ((((matchKindParser <* spaces).soft <* P.not(P.char('='))) ~ arg(indent)) <* maybeSpace)
     }
     OptIndent
       .block(left, branchList)
--- a/docs/src/main/paradox/recursion.md
+++ b/docs/src/main/paradox/recursion.md
@@ -79,6 +79,10 @@ allowed to reset or increase only after an earlier argument decreases.
 `loop` has the same termination checks as `recur`, and adds one extra
 requirement: every recursive self-call must be in tail position.
 
+Unlike `recur`, `loop` is also a stack-safety guarantee: accepted `loop`
+definitions are compiled to loops and run in constant stack space. If a
+recursive call is not in tail position, compilation fails.
+
 ```bosatsu
 def size1(list, acc):
   loop list:
@@ -86,9 +90,9 @@ def size1(list, acc):
     case [_, *t]: size1(t, Succ(acc))
 ```
 
-Use `loop` when you want the compiler to enforce tail recursion. Use `recur`
-when recursion is terminating but intentionally non-tail (for example,
-`Succ(len(tail))` style code).
+Prefer `loop` whenever possible so stack exhaustion is ruled out by the type
+checker/compiler. Use `recur` when recursion is terminating but intentionally
+non-tail (for example, `Succ(len(tail))` style code).
 
 ## Pattern 1: Structural Recursion On Lists
 This is the most common pattern. Recur directly on the list and call the same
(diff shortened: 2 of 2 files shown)


core/src/main/scala/dev/bosatsu/Declaration.scala
```diff
@@ -1248,8 +1259,11 @@ object Declaration {
       (leadingComments.with1.soft *> branches).orElse(branches)
     }
 
-    val left =
-      Indy.lift(matchKindParser <* spaces).cutThen(arg).cutLeftP(maybeSpace)
+    val left: Indy[(MatchKind, NonBinding)] = Indy { indent =>
+      // Keep `loop` contextual: if header parsing fails, allow fallback to normal
+      // identifier parsing (`def loop(...)`, `loop = ...`, `loop(...)`).
+      (((matchKindParser <* spaces) ~ arg(indent)).backtrack <* maybeSpace)
```
johnynek: I don't want this backtrack in all cases. Let's remove it and see if we can make it:
```
((((matchKindParser <* spaces).soft <* P.not("=")) ~ arg(indent))) <* maybeSpace)
```

With the soft followed by not "=" we should be able to avoid parsing the `loop =` case but not have full backtracking through `arg(indent)` where we would like to have a parse failure immediately point to the error.

docs/src/main/paradox/recursion.md
```diff
@@ -75,6 +75,21 @@ source tags while preserving a clear trust boundary before optimization.
 Tuple targets are useful for nested-recursive definitions where one argument is
 allowed to reset or increase only after an earlier argument decreases.
 
+## The `loop` Form
+`loop` has the same termination checks as `recur`, and adds one extra
```
johnynek, line 79: We should be clear that this will be compiled to a loop, so it consumes constant stack space. The recur form may or may not be compiled to a loop, but `loop` fails to compile if it's not a loop.

docs/src/main/paradox/recursion.md
```diff
@@ -333,6 +348,8 @@ How that maps to Bosatsu practice:
 1. If not, compute a bound and recurse on that fuel (`Nat` is usually easiest).
 1. If recursion uses multiple arguments, prefer `recur (a, b, ...)` and check
    lexicographic decrease in that order.
+1. If the recursive calls should be guaranteed tail-recursive, use `loop`
```
johnynek: we should be stronger: prefer to use `loop` if possible so your algorithm can't exhaust the stack.

core/src/main/scala/dev/bosatsu/Declaration.scala
johnynek: Implemented this parser change in `matchP`: `((((matchKindParser <* spaces).soft <* P.not(P.char('='))) ~ arg(indent)) <* maybeSpace)`, so we no longer backtrack across the full `arg(indent)` parse. I also adjusted `matchHeaderKeyword` to backtrack only at the keyword-boundary parser (`(P.string(keyword) <* whitespace.peek).backtrack`) so `loop(...)` still parses as normal identifier/application while `loop <target>:` remains contextual recursion-header syntax.

docs/src/main/paradox/recursion.md
johnynek: Updated `docs/src/main/paradox/recursion.md` in the `loop` section to state this explicitly: accepted `loop` definitions are compiled to loops and therefore run in constant stack space, and compilation fails if a recursive call is not tail-position. I also contrast this with `recur` in that same section.

docs/src/main/paradox/recursion.md
johnynek: Strengthened the wording as requested. The doc now says to prefer `loop` whenever possible so stack exhaustion is ruled out, and to use `recur` only when non-tail recursion is required.

johnynek:
Addressed all three review comments with concrete file updates:
- parser disambiguation/backtracking behavior in `core/src/main/scala/dev/bosatsu/Declaration.scala`
- stronger `loop` stack-safety and usage guidance in `docs/src/main/paradox/recursion.md`

Required pre-push check passed locally: `scripts/test_basic.sh`.

Issue #414, opened 2019-11-23
consider implementing caching of compiled files


johnynek:
We could optionally take a cache directory as an option, and if given, we could cache compilation of files into packages.

We could use a hash of the expressions without line number information as the key to the cache along with the hash of the packages the files depend on, then just write a link from that hash to the hash of the compiled output, and store the latter in a content addressed cache (this is conceptually the same as bazel's remote cache).

johnynek:
To solve this issue we need to have a design that works with PackageMap. Ideally, it should be a new interface that we pass to PackageMap. Something like:

```scala
trait Cache[G[_]] {
  given Monad[G]

  type CacheKey
  trait Keyable[K] {
    def key(k: K): CacheKey
  }

  given Keyable[Inferface]...
  given Keyable[Package.Parsed]
  
  def cache(packSrc: String, deps: List[Interface])(result: => Package.Compiled): G[Package.Compiled]
}
```

Or something like this. The goal is that we could make the no-op cache where `G[_] = cats.Id` and `CacheKey` is `Unit` and the cost should be almost nothing.

Or, we could implement one that hashes the interfaces and the source and maybe a nonce for versioning, and then looks up the value on disk in a given location from the hash, and if it is found, decode the protobuf to recover the value.

Something like this, so `lib compile` and `lib test` could run faster.

johnynek:
MergeXO assigned an agent and started design work for issue #414.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/1830

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #414.

johnynek:
Opened implementation PR: https://github.com/johnynek/bosatsu/pull/1842

PR #1830, opened 2026-02-27
Design doc for #414: Issue #414 Design: Optional Caching of Compiled Files

johnynek:
Design doc.

Refs #414

### DIFF
--- a/docs/design/414-consider-implementing-caching-of-compiled-files.md
+++ b/docs/design/414-consider-implementing-caching-of-compiled-files.md
@@ -31,7 +31,7 @@ Status: proposed
 
 Add an opt-in compile cache for `tool` commands that compile source files. The cache key is a hash of a normalized source package (line/region-insensitive) plus hashes of dependency interfaces and compile options. The key maps to a content-addressed compiled package artifact, matching the issue’s key->output-link plus CAS model.
 
-This revision makes the implementation concrete at the `PackageMap.inferAll` layer with a cache algebra and `IorT[F, ...]` pipeline, and explicitly keeps the no-cache path near-zero overhead via an `InferCache`-dependent key type.
+This revision makes the implementation concrete at the `PackageMap.inferAll` layer with a cache algebra and `IorT[F, ...]` pipeline, explicitly keeps the no-cache path near-zero overhead via an `InferCache`-dependent key type, and makes cache placement explicit relative to loop/recur lowering and normalization phases.
 
 ## Context
 
@@ -58,6 +58,7 @@ We need a cache that:
 4. Store compiled outputs in a content-addressed store and keep key->output indirection.
 5. Preserve current behavior when cache is disabled.
 6. Make cache integration concrete in `PackageMap.inferAll` with an effectful API.
+7. Cache post-inference phase output so expensive lowering/normalization is reused on cache hits.
 
 ## Non-goals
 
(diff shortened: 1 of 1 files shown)


johnynek:
The semantics of the cache make perfect sense, but we need to be a bit more concrete about implementation. Show code sketches. Especially how will:

```scala
object PackageMap {
    def inferAll(
      ps: Resolved,
      compileOptions: CompileOptions
  )(implicit cpuEC: Par.EC): Ior[NonEmptyList[PackageError], Inferred]
}

```
Look with caching? I think we need something like:
```
object PackageMap {
    def inferAll[G[_]](
      ps: Resolved,
      compileOptions: CompileOptions,
      cache: Cache[G]
  )(implicit cpuEC: Par.EC): G[Ior[NonEmptyList[PackageError], Inferred]]
}
```

but how would the inferAll method change? Maybe we have `IorT[G, NonEmptyList[PackageError], Inferred]`

Look into that method and be concrete about the Cache design the result

docs/design/414-consider-implementing-caching-of-compiled-files.md
```diff
@@ -143,58 +116,138 @@ For each direct import of the package being compiled, include an interface hash
...
+import cats.data.NonEmptyList
+import dev.bosatsu.{CompileOptions, Package, PackageError, PackageMap}
+
+trait InferCache[F[_]] {
+  def get(key: InferCache.Key): F[Option[Package.Inferred]]
+  def put(key: InferCache.Key, value: Package.Inferred): F[Unit]
+}
+
+object InferCache {
+  final case class Key(
```
johnynek: Remember our goal is zero or near zero cost being achievable when the InferCache is doing nothing.

So, to do that I think we need to make `Key` a dependent type of the `InferCache` and then pass it values we definitely currently have. The Key below includes a lot of hashes, which requires doing work (and in the case of interfaces may require serialization).

So, if we have:
```
trait InferCache[F[_]] {
  type Key
  def generateKey(....): Key
}
```
where the key can be `Unit` for a no-op cache.

johnynek: Updated the design doc to make `Key` dependent on `InferCache` and to make the low-cost no-op path explicit. The cache sketch now uses `trait InferCache[F[_]] { type Key; def generateKey(...): F[Key]; ... }`, includes a no-op constructor with `type Key = Unit`, and moves rich hash payload to filesystem `FsKey`. The `inferAll[F]` sketch now calls `cache.generateKey(...)` (instead of always prebuilding a hash-heavy key), and the acceptance/testing sections now explicitly require that no-op cache does not compute source/dependency hashes.

johnynek:
Implemented the requested design-doc changes for #414 and made the no-op/near-zero-cost semantics concrete in architecture, flow, implementation plan, tests, and acceptance criteria. Ran `scripts/test_basic.sh` successfully (`Passed: Total 1434, Failed 0, Errors 0`).

docs/design/414-consider-implementing-caching-of-compiled-files.md
```diff
@@ -0,0 +1,340 @@
...
+4. For no-op cache, `generateKey` is constant and avoids hash/serialization work entirely.
+5. Cache misses execute existing inference logic unchanged.
+6. Cache hits return typed packages directly and still feed downstream dependency typing.
+
+### 7. Behavior of `resolveThenInfer`
+
+`resolveThenInfer` gets an effectful variant for cache-aware callers:
+
+```scala
+def resolveThenInfer[F[_]: Monad: Parallel, A: Show](
```
johnynek, line 280: Currently we do a few kinds of lowering after the type inference: loop/recur and then normalization/optimization.

How does that interact with the caching? Some of that may be expensive and we don't want to rerun it on every call. That said we don't want one big function in PackageMap either.

Maybe we pass a Phases object into resolve that runs all the phases we expect on the package and its dependency interfaces and so the cache runs at the end of that?

johnynek: Updated the design doc to make phase interaction concrete. It now introduces an `InferPhases` abstraction that is passed into `inferAll`/`resolveThenInfer`, with `finishPackage` (loop/recur + normalization/optimization) and `dependencyInterface` hooks. The cache flow now stores post-phase `Package.Inferred` values, so cache hits skip both inference and post-infer phase reruns for that package. I also included `phases.id` in key generation/invalidations and called out this behavior in implementation plan, testing, acceptance criteria, and risks.

johnynek:
Implemented the requested phase/caching design clarification in `docs/design/414-consider-implementing-caching-of-compiled-files.md` and ran `scripts/test_basic.sh` successfully (`Passed: Total 1434, Failed 0, Errors 0`).

PR #1845, opened 2026-02-28
docs: update int decrease design to type-directed recursion strategies

### LINKED ISSUE #1844: update the int_decrease design
In: docs/src/main/paradox/design-docs/recur_int_decrease_design.md

We have a design for checking recursion on integers. The syntax imagines having `recur x by int_decrease:`. Thinking more, I think we should change that so we decide the recursion strategy based on the type. Each type can have 0 or more recursion strategies and we can check which apply.

For integers, the strategy is you have to decrease the Int while being >= 0. We don't need to say `by int_decrease` we should dispatch strategy based on type.

Please update the document to reflect this approach. Inspect the dev.bosatsu.smt package, which has some support code that could aid in checking that the `Int` parameters have decreased.

We can also imagine using the lexicographic order of the enums as a way to consider them decreased. So `enum Opt[a]: None, Some(v: a)` could consider `None` < `Some(_)`. But also we could imagine computing that `Some(n)` < `Some(m)` when `n < m`. This is not a phase 1 task, but we can drive towards this view where we have a unified view of decreasing by type without any new syntax.

johnynek:
Updated `docs/src/main/paradox/design-docs/recur_int_decrease_design.md` to remove `recur ... by int_decrease` syntax and describe type-dispatched recursion strategies. The revised design defines phase-1 `Int` obligations (`PC => next_i >= 0` and `PC => next_i < i`), explains how existing `dev.bosatsu.smt` support (`SmtExpr`, `SmtCommand`/`SmtScript`, `SmtLibRender`, `Z3Api`) can prove those obligations, and records enum ordering/lexicographic decrease as future work without adding syntax. Ran `scripts/test_basic.sh` and it passed.

Fixes #1844

### DIFF
--- a/core/src/test/scala/dev/bosatsu/LitTest.scala
+++ b/core/src/test/scala/dev/bosatsu/LitTest.scala
@@ -29,7 +29,7 @@ class LitTest extends munit.ScalaCheckSuite {
       genFloatNoNaN,
       Gen
         .frequency(
-          (10, Gen.choose(0, 0xd800)),
+          (10, Gen.choose(0, 0xd7ff)),
           (1, Gen.choose(0xe000, 1000000))
         )
         .map(Lit.fromCodePoint)


johnynek:
Addressed the CI failure by fixing the flaky generator in core/src/test/scala/dev/bosatsu/LitTest.scala. The failing case from Actions involved a surrogate boundary value (`0xD800`) in `we can parse from document`; the generator previously used `Gen.choose(0, 0xd800)`, which could emit that invalid code point. I changed it to `Gen.choose(0, 0xd7ff)` so generated code points stay in valid ranges. Validation run: `sbt -batch "coreJVM/testOnly dev.bosatsu.LitTest"` passed, and required pre-push check `scripts/test_basic.sh` passed.

PR #1849, opened 2026-02-28
Validate cached dependency descriptors on CAS hits for lib fetch/check

### LINKED ISSUE #1847: lib fetch/check can mask bad dependency descriptor on CAS cache hit
## Summary
`lib check` and `lib fetch` can both pass with an invalid dependency descriptor when the referenced hash already exists in CAS.

In practice, if a dependency entry changes `version`/`uris` but keeps a hash that is already cached locally, Bosatsu treats it as valid without re-validating the descriptor against the cached library.

## Reproduction
Using a library config like `src/mylib_conf.json` in `test_bo_repo`:

1. Change a dependency descriptor to a bad version/URI, but keep the same hash:
   - e.g. `core_alpha` set to `version: 5.0.0` with a bad release URL
   - keep hash `blake3:b49dd10aa80aa1c6b0519cff6f8036be431754b6558138f8d48c6866fe860986`
2. Ensure that hash object already exists in `.bosatsuc/cas`.
3. Run `./bosatsu lib check`.
4. Run `./bosatsu lib fetch`.

Observed:
- `lib check` exits 0.
- `lib fetch` reports `fetched 0 objects.`

But with a fresh CAS (or if that hash is absent), `lib fetch` fails as expected on the bad URI.

## Expected behavior
Even on CAS cache hits, Bosatsu should verify the cached object matches the requested dependency descriptor well enough to catch stale/mismatched metadata.

At minimum:
- `dep.name == cachedLibrary.name`
- if `dep.desc.version` is set, it should equal `cachedLibrary.descriptor.version`

If the check fails, return an error (same as a failed fetch), rather than treating it as a successful cache hit.

## Actual behavior
Cache hit logic is currently hash-path existence-based:
- `fetchIfNeeded`: if file exists, returns cached without reading/validating content.
- `libFromCas` loads by hash and `validateDeps` checks hash/name grouping, but does not enforce descriptor version equality for direct deps.

So a bad `version`/`uris` can be masked by an existing hash object.

## Why this matters
This makes dependency descriptors non-authoritative in warm-CAS environments and can hide broken dependency configuration until someone uses a clean CAS.

## Proposed fix
In `Cas.fetchIfNeeded`, in the `fileExists(path)` branch:

1. Read the cached `.bosatsu_lib` (`readLibrary(path)`).
2. Validate that it matches the requested dependency descriptor:
   - name equality
   - descriptor version equality when requested
3. If mismatch, return `Left(CliException(...))` for that hash, causing fetch to fail.

This is cheap: `readLibrary` already parses the library and computes its hash from bytes.

## Optional strengthening
Also consider applying the same descriptor-vs-library consistency check in `libFromCas`/`depsFromCas` paths used by `lib check`, so checks fail fast on mismatched cached entries as well.

## Suggested tests
- Cache-hit fetch with matching descriptor succeeds.
- Cache-hit fetch with mismatched version fails with clear error.
- Cache-hit fetch with mismatched name fails with clear error.
- Existing behavior (missing hash + bad URI) still fails as before.

johnynek:
Implemented issue #1847 with focused CAS-path validation. In `Command.Cas`, added `validateCachedDependency` to enforce `dep.name == cachedLibrary.name` and (when specified) dependency descriptor version equality with the cached library descriptor version. Applied this in two places: (1) `fetchIfNeeded` now reads and validates cached `.bosatsu_lib` entries on cache hits before returning `Right(false)`, and returns a fetch failure when mismatched; (2) `depsFromCas` now loads via `validatedLibFromCas`, so `lib check` also fails on mismatched cached descriptors instead of silently passing. Added targeted regressions in `ToolAndLibCommandTest` covering: cache-hit fetch success with matching descriptor, cache-hit fetch failure on version mismatch, cache-hit fetch failure on name mismatch, and `lib check` failure on cached version mismatch. Required test command `scripts/test_basic.sh` was run and passed.

Fixes #1847

### DIFF
--- a/core/src/main/scala/dev/bosatsu/library/Command.scala
+++ b/core/src/main/scala/dev/bosatsu/library/Command.scala
@@ -2,7 +2,15 @@ package dev.bosatsu.library
 
 import cats.{Monad, MonoidK}
 import cats.arrow.FunctionK
-import cats.data.{Chain, Ior, NonEmptyChain, NonEmptyList, Validated, ValidatedNel}
+import cats.data.{
+  Chain,
+  Ior,
+  NonEmptyChain,
+  NonEmptyList,
+  Validated,
+  ValidatedNec,
+  ValidatedNel
+}
 import com.monovore.decline.{Argument, Opts}
 import dev.bosatsu.tool.{
   CliException,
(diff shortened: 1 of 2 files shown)


core/src/main/scala/dev/bosatsu/library/Command.scala
```diff
@@ -2265,6 +2265,51 @@ object Command {
...
+        val cachedVersionStr = cachedVersion.fold("unspecified")(_.render)
+
+        Left(
+          CliException(
+            "cached library descriptor mismatch",
+            Doc.text(
+              show"cached object ${cached.hash.toIdent} does not match dependency ${dep.name}."
+            ) +
+              Doc.line +
+              Doc.text(show"dependency name: ${dep.name}") +
```
johnynek: only show the mismatch. If only the version is wrong, just show that. If only the name is wrong just show that. If both are wrong, show both.

core/src/main/scala/dev/bosatsu/library/Command.scala
```diff
@@ -2265,6 +2265,51 @@ object Command {
         }
       }
 
+    private def validateCachedDependency(
+        dep: proto.LibDependency,
+        cached: Hashed[Algo.Blake3, proto.Library]
+    ): Either[Throwable, Unit] = {
```
johnynek: we shouldn't lift to Throwable because as soon as we have a Throwable, when we parTraverse it will combine the CliException into a non-CliException.

We should return `ValidatedNec[Doc, Unit]` here and then explicitly collect all these docs and if they are non-empty lift them into the CliException one time below.

core/src/test/scala/dev/bosatsu/ToolAndLibCommandTest.scala
```diff
@@ -3560,6 +3647,90 @@ main = 0
...
+      case Left(err: CliException) =>
+        val rendered = err.errDoc.render(120)
+        assert(rendered.contains("cached library descriptor mismatch"), rendered)
+        assert(rendered.contains("not_dep 0.0.1:"), rendered)
+      case Left(err) =>
+        fail(err.getMessage)
+    }
+  }
+
+  test("lib check fails on cached dependency version mismatch") {
```
johnynek, line 3863: add a check that shows if two versions mismatch we see both errors, not just one or the other.

core/src/main/scala/dev/bosatsu/library/Command.scala
johnynek: Updated this so mismatch details are field-specific. `validateCachedDependency` now validates name and version separately and only emits the lines that actually mismatch (name-only, version-only, or both).

core/src/main/scala/dev/bosatsu/library/Command.scala
johnynek: Refactored per suggestion: `validateCachedDependency` now returns `ValidatedNec[Doc, Unit]` (no `Throwable`). In `depsFromCas`, I accumulate all cached-descriptor validation docs via `traverse_` and raise one `CliException` after aggregation, instead of failing per-item inside `parTraverse`.

core/src/test/scala/dev/bosatsu/ToolAndLibCommandTest.scala
johnynek: Added coverage for this: `lib check reports both cached dependency version mismatches` in `ToolAndLibCommandTest`. It builds two cached deps with mismatched configured versions and asserts both mismatch messages are present in the error output.

johnynek:
Addressed all three review comments with code and test updates in `Command.scala` and `ToolAndLibCommandTest.scala`. Validation now accumulates descriptor mismatches correctly and reports only mismatched fields. Added dual-version mismatch regression coverage. Required pre-push check `scripts/test_basic.sh` was run and passed.

PR #1862, opened 2026-02-28
Improve diagnostic for `matches <identifier>` pattern-binding mistakes

### LINKED ISSUE #1854: Diagnostic for matches<identifier> reports unused value, not likely pattern-binding mistake
### Summary
Using `matches <identifier>` can produce an `unused value` error that does not explain the likely root cause: the identifier is being treated as a pattern binding (always matches), not equality.

### Minimal reproduction
`/tmp/repro_issue1.bosatsu`:
```bosatsu
package Repro/Issue1

test = Assertion(int_to_String(42) matches str, "msg")
```

Run:
```bash
bosatsu tool check --input /tmp/repro_issue1.bosatsu --output /tmp/repro_issue1.pkg
```

### Actual output
```text
in file: /tmp/repro_issue1.bosatsu:3:35, package Repro/Issue1

  unused value 'str'
  ...

in file: /tmp/repro_issue1.bosatsu:3:18, package Repro/Issue1
...
unreachable branches: _
```

### Expected output
A diagnostic that points out that `str` is a variable pattern in `matches` (and therefore not equality), with a suggestion like:
- use explicit equality (`eq_String` / relevant equality function), or
- use a literal pattern if pattern matching was intended.

### Environment
- bosatsu: `0.0.36`

johnynek:
Implemented issue #1854 with a focused compiler diagnostic update. `UnusedLetCheck` now classifies unused bindings that come from lowered `matches` boolean checks, and `PackageError.UnusedLetError` now emits a specific message for that case (`pattern binding 'name' in `matches` always succeeds`) plus hints explaining this is pattern binding (not equality) and suggesting explicit equality (for example `eq_String`) or a literal/constructor pattern. Added a regression test in `ErrorMessageTest` (`matches identifier binding has a focused hint`). Ran the required pre-push command `scripts/test_basic.sh`, and it passed.

Fixes #1854

### DIFF
--- a/core/src/main/scala/dev/bosatsu/PackageError.scala
+++ b/core/src/main/scala/dev/bosatsu/PackageError.scala
@@ -7,7 +7,6 @@ import org.typelevel.paiges.{Doc, Document}
 
 import rankn._
 import LocationMap.Colorize
-import UnusedLetCheck.{UnusedBinding, UnusedKind}
 
 sealed abstract class PackageError {
   def message(
@@ -1305,26 +1304,18 @@ object PackageError {
 
   private def unusedValueMessage(
       pack: PackageName,
-      errs: NonEmptyList[UnusedBinding],
+      errs: NonEmptyList[(Identifier.Bindable, Region)],
       sourceMap: Map[PackageName, (LocationMap, String)],
       errColor: Colorize,
       hints: List[String]
   ): String = {
     val (lm, _) = sourceMap.getMapSrc(pack)
-    val sorted = errs.sortBy(_.region)
-    val unusedDocs = sorted.map { err =>
-      val region = err.region
+    val sorted = errs.sortBy(_._2)
+    val unusedDocs = sorted.map { case (bn, region) =>
       val rdoc = lm
         .showRegion(region, 2, errColor)
         .getOrElse(Doc.str(region.show)) // we should highlight the whole region
-      val message = err.kind match {
-        case UnusedKind.Standard =>
-          Doc.text(s"unused value '${err.name.sourceCodeRepr}'")
-        case UnusedKind.MatchesPatternBinding =>
-          Doc.text(
-            s"pattern binding '${err.name.sourceCodeRepr}' in `matches` always succeeds"
-          )
-      }
+      val message = Doc.text(s"unused value '${bn.sourceCodeRepr}'")
       message + Doc.hardLine + rdoc
     }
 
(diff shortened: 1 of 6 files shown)


johnynek:
We shouldn't be getting this error as an unused let binding, in SourceConverter we have this code:
```scala
      case m @ Matches(a, p) =>
        // x matches p ==
        // match x:
        //   p: True
        //   _: False
        val True: Expr[Declaration] =
          Expr.Global(PackageName.PredefName, Identifier.Constructor("True"), m)
        val False: Expr[Declaration] = Expr.Global(
          PackageName.PredefName,
          Identifier.Constructor("False"),
          m
        )
        (loop(a), convertPattern(p, m.region)).mapN { (a, p) =>
          val branches = NonEmptyList(
            Expr.Branch(p, None, True),
            Expr.Branch(Pattern.WildCard, None, False) :: Nil
          )
          Expr.Match(a, branches, m)
        }
```
That code needs to check that `p.names.isEmpty` and that `!p.definitelyTotal` where we add `def definitelyTotal: Boolean` to `Pattern[N, T]` and it is true to Var, Wildcard, unions where `definitelyTotal` exists on a branch or total list or string patterns (both of which can see when some patterns are definitely total).

In both of those cases we need to return a partial failure in SourceConverter, but also we should translate the pattern to something that won't fail later checks (if it is definitely total just rewrite the match to `match x: case _: True` so we still typecheck as much as we can later.

Let's not address this in UnusedLetChecker. This is missing code in SourceConverter.

johnynek:
Addressed the PR feedback by moving this behavior out of `UnusedLetCheck` and into `SourceConverter`.

What changed:
- Reverted the prior `UnusedLetCheck`/`PackageError`-based `matches` diagnostic approach.
- Added `Pattern.definitelyTotal` in `Pattern` to detect syntactically total patterns (wildcards/vars, total unions, total string/list forms, etc.).
- Updated `SourceConverter` `Matches` lowering to:
  - emit partial errors when `p.names.nonEmpty` and/or `p.definitelyTotal`,
  - unbind pattern names when needed,
  - rewrite definitely-total `matches` to a single branch (`case _: True`) to avoid follow-on unreachable-branch noise.
- Added new `SourceConverter.Error` cases for these diagnostics with actionable messages.
- Updated tests:
  - `ErrorMessageTest`: now expects a `SourceConverterErrorsIn` diagnostic for `int_to_String(42) matches str`.
  - `PatternTest`: added coverage for `definitelyTotal`.

Validation:
- Ran focused tests: `sbt "coreJVM/testOnly dev.bosatsu.PatternTest dev.bosatsu.ErrorMessageTest dev.bosatsu.SourceConverterTest"` (pass).
- Ran required pre-push command: `scripts/test_basic.sh` (pass).

PR #1863, opened 2026-02-28
Improve unknown operator suggestions to avoid unrelated local operators

### LINKED ISSUE #1855: Unknown operator << suggests unrelated operators (+, *) instead of shift guidance
### Summary
When `operator <<` is missing, the name suggestion can prefer unrelated local operators (for example `operator *` and `operator +`) instead of giving shift-specific guidance.

### Minimal reproduction
`/tmp/repro_issue2.bosatsu`:
```bosatsu
package Repro/Issue2

def operator +(a, b): a.add(b)
def operator *(a, b): a.mul(b)

x = 1 << 2
```

Run:
```bash
bosatsu tool check --input /tmp/repro_issue2.bosatsu --output /tmp/repro_issue2.pkg
```

### Actual output
```text
in file: /tmp/repro_issue2.bosatsu:6:6, package Repro/Issue2
Unknown name `operator <<`.
Did you mean one of: local value `operator *`, local value `operator +`?
```

### Expected output
Suggestions that are semantically closer for this context, e.g. `shift_left_Int` or guidance to define `operator <<`, or otherwise no suggestion if only unrelated candidates exist.

### Environment
- bosatsu: `0.0.36`

johnynek:
Adjusted name-suggestion scoring for operator identifiers so when both sides are operators it compares raw operator tokens (e.g. `<<` vs `+`) instead of the shared `operator ` prefix. This prevents unrelated local operators from being suggested for missing `operator <<`. Added a regression test in `ErrorMessageTest` (`unknown operator suggestions exclude unrelated local operators`) covering the reported repro shape. Verified with `sbt "coreJVM/test:compile"`, `sbt "coreJVM/testOnly dev.bosatsu.ErrorMessageTest"`, and required `scripts/test_basic.sh` (all passing).

Fixes #1855

### DIFF
--- a/core/src/main/scala/dev/bosatsu/OperatorHints.scala
+++ b/core/src/main/scala/dev/bosatsu/OperatorHints.scala
@@ -1,0 +1,219 @@
+package dev.bosatsu
+
+import cats.data.NonEmptyList
+
+object OperatorHints {
+  private sealed abstract class LitKind
+  private object LitKind {
+    case object IntLit extends LitKind
+    case object FloatLit extends LitKind
+    case object StringLikeLit extends LitKind
+  }
+
+  private final case class CallSite(
+      left: Option[LitKind],
+      right: Option[LitKind]
+  ) {
+    def has(kind: LitKind): Boolean =
+      left.contains(kind) || right.contains(kind)
+  }
+
+  def hint(
+      op: Identifier.Operator,
+      region: Region,
+      lets: List[(Identifier.Bindable, RecursionKind, Expr[Declaration])]
+  ): Option[String] = {
+    val callSite =
+      lets.iterator
+        .map(_._3)
+        .map(findCallSite(_, op, region))
+        .collectFirst { case Some(site) => site }
+
+    val predefFns = predefFunctions(op.asString, callSite)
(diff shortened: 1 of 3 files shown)


johnynek:
This addresses half of the goals of the issue, but it doesn't include the improved hints on unknown `operator` calls which suggestions:
```
Map(
  "operator +" -> "Bosatsu/Predef::add or addf",
  "operator -" -> "Bosatsu/Predef::sub or subf",
  "operator /" -> "Bosatsu/Predef::div or divf",
  "operator %" -> "Bosatsu/Predef::mod",
  "operator <<" -> "Bosatsu/Predef::shift_Left_Int",
  "operator <>" -> "Bosatsu/Predef::cmp_Int or cmp_Float",
  ...
````

Read the predef and look for all the functions there that might have operators and build the OperatorHints class or object to handle this.

Again, when the source code explicitly has a `Lit` on one side or the other or both, use that to inform the search since we can see the types of the literals.

johnynek:
Ran required pre-push check `scripts/test_basic.sh` locally on head `2be25379501381db858cf66da531517167620cc3`; it passed with exit code 0. No additional file edits were needed for the remaining PR feedback.

Issue #946, opened 2023-02-26
Auto import of some types


johnynek:
If we don't mention a type, but it is needed to typecheck a mentioned method and we declare a dependency on the module, we shouldn't error but automatically import.

If we don't depend on the module we need, we should error but explain the solution is add a dependency on a certain module.

It's a bit weird to have two different rules, but the real rule is that by reading the code a human or simple tool can see the dependency structure of the code, and we shouldn't otherwise need to import names not used in the code.

johnynek:
this came up in #1174

I think the main rules are:

1. we should be able to read the code and see where all names came from
2. we should be able to see package dependency graph via parsing, without type checking

So, if you use a function, but don't mention one of the types it needs, but we do depend on that package already, we shouldn't error. So, one way we could do that is look at all values imported, and then traverse the set of all root type constants. Any type constants from imported packages, we can import. This should happen after SourceConverter, so that it doesn't allow you to name in the source types or constructors without importing, but we shouldn't have to import a type from a package we already depend on just to satisfy inference.

johnynek:
Closed by #1865

PR #1865, opened 2026-02-28
Deduplicate repeated unimported-type diagnostics for identical spans

### LINKED ISSUE #1856: Unimported type diagnostic can repeat multiple times for same span and same fix
### Summary
`Use of unimported type` diagnostics can be emitted repeatedly for the same source span and same fix, which adds noise.

### Minimal reproduction
`/tmp/repro_issue3_lib.bosatsu`:
```bosatsu
package Repro/Issue3/Lib

export (P, mk, use)

struct P(i: Int)

def mk(i): P(i)

def use(ps: List[P]) -> Int:
  if ps matches []: 0
  else: 1
```

`/tmp/repro_issue3_main.bosatsu`:
```bosatsu
package Repro/Issue3

from Repro/Issue3/Lib import mk, use

x = use([mk(1), mk(2)])
```

Run:
```bash
bosatsu tool check \
  --input /tmp/repro_issue3_lib.bosatsu \
  --input /tmp/repro_issue3_main.bosatsu \
  --output /tmp/repro_issue3.pkg
```

### Actual output
The same diagnostic is repeated 3 times at the same location:
```text
in file: /tmp/repro_issue3_main.bosatsu:5:9, package Repro/Issue3
Use of unimported type. Add `from Repro/Issue3/Lib import P`
... (repeated 3x)
```

### Expected output
A single deduplicated diagnostic per source span/root cause (or a grouped message), rather than repeated identical entries.

### Environment
- bosatsu: `0.0.36`

johnynek:
Implemented a focused fix for issue #1856 by deduplicating `Infer.Error.UnknownDefined` entries during `PackageError.TypeErrorIn` aggregation when they refer to the same unimported type at the same region. This prevents repeated `Use of unimported type. Add ...` diagnostics from being emitted multiple times for one source span/root cause. Added a regression test in `ErrorMessageTest` that reproduces the reported two-package scenario and asserts the unimported-type message appears exactly once and includes the expected import fix. Validation run: `sbt "coreJVM/testOnly dev.bosatsu.ErrorMessageTest"` passed, and required pre-push command `scripts/test_basic.sh` passed.

Fixes #1856

### DIFF
--- a/core/src/test/scala/dev/bosatsu/ErrorMessageTest.scala
+++ b/core/src/test/scala/dev/bosatsu/ErrorMessageTest.scala
@@ -617,7 +617,7 @@ main = fn
           Map.empty,
           Colorize.None
         ),
-        "in file: <unknown source>, package A\nrecur but no recursive call to fn\n[25, 47)\n"
+        "in file: <unknown source>, package A\nrecur but no recursive call to fn\nFor non-recursive branching, replace `recur <expr>:` with `match <expr>:`.\n[25, 47)\n"
       )
       ()
     }
@@ -676,6 +676,11 @@ def parse_loopTypo(x):
 main = parse_loopTypo
 """)) { case te @ PackageError.RecursionError(_, _) =>
       val msg = te.message(Map.empty, Colorize.None)
+      assert(
+        msg.contains(
+          "For non-recursive branching, replace `recur <expr>:` with `match <expr>:`."
+        )
+      )
       assert(
         msg.contains(
           "Function name looks renamed: declared `parse_loopTypo`, but recursive calls use `parse_loop`."
--- a/cli/src/main/scala/dev/bosatsu/IOPlatformIO.scala
+++ b/cli/src/main/scala/dev/bosatsu/IOPlatformIO.scala
@@ -3,6 +3,7 @@ package dev.bosatsu
 import _root_.bosatsu.{TypedAst => proto}
 import cats.MonadError
 import cats.effect.{IO, Resource}
+import cats.effect.kernel.Deferred
 import com.monovore.decline.Argument
 import java.nio.file.{Paths, Path => JPath, Files}
 import java.io.{
@@ -13,6 +14,7 @@ import java.io.{
   PrintWriter
 }
 import org.typelevel.paiges.Doc
+import dev.bosatsu.graph.CanPromise
 import dev.bosatsu.hashing.{Hashed, HashValue, Algo}
 import dev.bosatsu.tool.CliException
 import scalapb.{GeneratedMessage, GeneratedMessageCompanion}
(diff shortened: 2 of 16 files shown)


johnynek:
There are actually two separate errors here to fix:

1. we should definitely deduplicate and never show a duplicate error. You've already handled this.
2. however, this should not be an error. P is not literally mentioned in the source and we do depend on the package it came from explicitly. This is discussed in issue https://github.com/johnynek/bosatsu/issues/946.

We should allow all types that are not literally appearing in the source code but ARE from packages we depend on, to be used at typecheck time. We do intend to allow code that isn't fully annotated, that's why we have type inference and allow untyped top level functions. We do intend to require dependencies between packages to be explicit. But here, the dependency on Repro/Issue3/Lib is declared, we just didn't mention type P, which is also never explicitly used by the programmer.

So we need to accept the code in the issue (while rejecting cases where the unmentioned type comes from a package we haven't added an explicit dependency on).

johnynek:
Implemented the additional behavior requested in feedback:
- Type inference now accepts inferred/unmentioned types from packages that are explicitly depended on (direct imports), by widening kind lookup during typechecking to include exported types from direct dependency interfaces.
- Still rejects transitive-only cases where the hidden type comes from a package not directly depended on.
- Kept dedup verification for the remaining error case.

Code changes:
- core/src/main/scala/dev/bosatsu/Package.scala
- core/src/test/scala/dev/bosatsu/ErrorMessageTest.scala

Test updates:
- Replaced the old issue-1856 reproduction expectation with a success test for direct dependency behavior.
- Added a transitive-dependency regression test that still expects one deduplicated "Use of unimported type" diagnostic.

Validation run:
- sbt "coreJVM/testOnly dev.bosatsu.ErrorMessageTest" (pass)
- scripts/test_basic.sh (pass)

johnynek:
Reproduced the CI `testJS` failure locally and fixed it by updating `jsui/src/test/scala/dev/bosatsu/jsui/StoreTest.scala` to match the current `Output.TestOutput` 3-argument shape (`case Output.TestOutput(_, _, _)`).

Validation run:
- Reproduced failure before fix: `sbt "jsuiJS/testOnly * -- --log=failure"` (failed with E107 wrong pattern arity)
- Verified fix: `sbt "jsuiJS/testOnly * -- --log=failure"` (pass)
- Required pre-push check: `scripts/test_basic.sh` (pass)

johnynek:
Merge origin main and push to the branch.

johnynek:
Ran the required pre-push check `scripts/test_basic.sh` on head `c881e4542f42b4711d8c0b2fd4eb516b17b2adc9`; it passed (`cli` 61 passed, `coreJVM` 1491 passed, 0 failed). The branch is clean and no additional code edits were needed for this feedback.

core/src/test/scala/dev/bosatsu/ErrorMessageTest.scala
```diff
@@ -2422,6 +2422,88 @@ x = 1.0 + 2.0
...
+        |  if ps matches []: 0
+        |  else: 1
+        |""".stripMargin
+
+    val mainSrc =
+      """package Repro/Issue3
+        |
+        |from Repro/Issue3/Lib import mk, use
+        |
+        |x = use([mk(1), mk(2)])
```
johnynek, line 2452: we need another test, if the user wrote::

```
[mk(1), mk(2)]: List[P]
```
that should fail, because `P` was not imported. It should only be okay to use a type in a package you already depend on *if you don't explictly name it*.

Maybe the Customs checker thing might catch this but we need a test to assert it.

this is because a reader needs to be able to see where names from. So in the original

johnynek: Added a regression test in `core/src/test/scala/dev/bosatsu/ErrorMessageTest.scala` (`"explicit type annotations from dependencies still require type imports"`). It uses the Issue3 setup with `x = [mk(1), mk(2)]: List[P]` and asserts typechecking fails (accepting either `Use of unimported type` or `unknown type: P`), so explicit type mentions still require imports while inferred dependency types remain allowed.

johnynek:
Implemented the requested test-only follow-up and ran the required pre-push check: `scripts/test_basic.sh` passed (`cli` 61 passed, `coreJVM` 1492 passed, 0 failures).

PR #1866, opened 2026-02-28
Add explicit runtime preflight for `lib test` when default `cc_conf.json` is missing

### LINKED ISSUE #1857: Add explicit runtime readiness preflight/check for lib test when cc_conf is missing
### Summary
When runtime config is missing, `bosatsu lib test` exits with a missing `cc_conf.json` message, but there is no explicit preflight step or dedicated command to validate runtime readiness before testing.

### Minimal reproduction
1. In a Bosatsu repo with tests, ensure the runtime config for the current tool/runtime hash is missing. In my case:
   - expected path: `/Users/oscar/code/test_bo_repo/.bosatsuc/5e5bd1d135f419a498cb321e4451454322bb6d73/cc_conf.json`
2. Run:
```bash
bosatsu lib test --name mylib --filter 'MyLib/Fib'
```

### Actual output
```text
expected a CcConf json file at .../cc_conf.json but found: None.

Perhaps you need to run `bosatsu c-runtime install` (or `make install` in c_runtime)
```

### Expected output
The current message is helpful, but it would be better if test execution included an explicit preflight/runtime-readiness check (or dedicated command) so this can be validated before launching the test pipeline.

Potential improvement:
- a `bosatsu c-runtime check` command, or
- an early `lib test` preflight section that reports the exact missing runtime artifact and hash.

### Environment
- bosatsu: `0.0.36`

johnynek:
Implemented a focused fix for issue #1857 by adding an explicit runtime-readiness preflight in `lib test` before the build/test pipeline runs. The command now resolves `cc_conf` up front and fails with a dedicated preflight error when runtime config is missing. Also improved the default `cc_conf` missing error message to include runtime hash, expected artifact (`cc_conf.json`), expected path, and install guidance. Added a regression test in `ToolAndLibCommandTest` to verify preflight happens early and reports the runtime-missing details. Validation run: `sbt "coreJVM/testOnly dev.bosatsu.ToolAndLibCommandTest"` and required `scripts/test_basic.sh` (both passed).

Fixes #1857

### DIFF
--- a/bosatsu
+++ b/bosatsu
@@ -143,6 +143,12 @@ if [ ! -f "$artifact_path" ]; then
   fail "missing ${artifact_path} after installation"
 fi
 
+# `./bosatsu --fetch` is used as a setup step; auto-install c runtime so
+# subsequent `lib test` commands don't fail on missing cc_conf.
+if [ "$fetch" -eq 1 ] && [ "${#args[@]}" -eq 0 ]; then
+  args=("c-runtime" "install")
+fi
+
 case "$platform" in
   java)
     command -v java >/dev/null 2>&1 || fail "java not found"


johnynek:
Let's also update the bosatsu bash script so that when we run --fetch with no other arguments it always immediately runs bosatsu c-runtime install after doing the fetch. Then the user should never see this error.

johnynek:
Implemented the PR feedback in `bosatsu`: when invoked as `./bosatsu --fetch` with no forwarded CLI args, the wrapper now immediately runs `c-runtime install` (by setting args to `c-runtime install` after artifact install/check). This makes the fetch/setup path install runtime config up front and avoids the later missing `cc_conf.json` error path for lib tests. Required gate passed: `scripts/test_basic.sh`.

PR #1868, opened 2026-03-01
Add --quiet mode for evaluator and C test runners

### LINKED ISSUE #1864: test --quiet mode
When we run tests (both in C generated tests or in the `tool test` that runs tests in the evaluator) we should allow a `--quiet` mode that only shows failure data and a summary line (total assertions passed and failed). For agentic coding, we don't want to waste context by feeding it a ton of successful test results.

johnynek:
Implemented issue #1864 with focused changes: added `--quiet` to `tool test` and C test-mode options, threaded quiet state through `Output.TestOutput`, and updated `Test.outputFor` to emit failure-only details plus a final total passed/failed summary in quiet mode. C test binaries now parse `--quiet` (`bsts_test_argv_has_quiet`), suppress successful per-package output, keep failure details, and print final passed/failed totals. Updated runtime headers/signatures, clang test code generation, and transpiler execution args so `lib test` and `tool transpile ... --test --execute` can run quietly. Added/updated tests in `ToolAndLibCommandTest`, `ClangGenLibraryDepsTest`, `PathModuleTest`, and `TestUtils`. Verification completed: `scripts/test_basic.sh` passed.

Fixes #1864

### DIFF
--- a/jsui/src/main/scala/dev/bosatsu/jsui/Store.scala
+++ b/jsui/src/main/scala/dev/bosatsu/jsui/Store.scala
@@ -52,11 +52,11 @@ object Store {
           "html"
         )
         val handler: HandlerFn = {
-          case Output.TestOutput(resMap, color) =>
+          case Output.TestOutput(resMap, color, quiet) =>
             val evaluatedTests = resMap.map { case (p, opt) =>
               (p, opt.map(_.value))
             }
-            val testReport = Test.outputFor(evaluatedTests, color)
+            val testReport = Test.outputFor(evaluatedTests, color, quiet)
             testReport.doc.render(80)
           case other =>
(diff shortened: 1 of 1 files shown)


johnynek:
Reproduced the reported CI failure locally with `sbt -batch "jsuiJVM/compile; jsuiJS/compile"` and hit the same E107 at `jsui/src/main/scala/dev/bosatsu/jsui/Store.scala:55` (stale 2-arg `Output.TestOutput` pattern). Updated the match to `Output.TestOutput(resMap, color, quiet)` and passed `quiet` through to `Test.outputFor(...)`. Re-ran `sbt -batch "jsuiJVM/compile; jsuiJS/compile"` successfully, then ran the required pre-push check `scripts/test_basic.sh`, which passed.

PR #1872, opened 2026-03-01
Suppress Github workflow JSON test output on passing runs

### LINKED ISSUE #1871: change the GithubWorkflow example test to only print the json when it doesn't match
Currently there seems to be a println or something that is causing the json from the GithubWorkflow examples to show up in the tests logs, even when we have the quiet mode on.

We need to hide that when it is not incorrect because it makes finding real problems in the logs harder.

johnynek:
Implemented issue #1871 with a focused change in `cli/src/test/scala/dev/bosatsu/GithubWorkflowJsonParityTest.scala`: the local `run` helper no longer calls `PathModule.reportOutput(out)` on successful command execution, so generated workflow JSON is not printed during passing tests. Failure behavior is unchanged: mismatch details are still surfaced via the existing `assertEquals(..., clues(...))` output. Verification: `sbt "cli/testOnly dev.bosatsu.GithubWorkflowJsonParityTest"` passed and required pre-push command `scripts/test_basic.sh` passed.

Fixes #1871

### DIFF
--- a/c_runtime/bosatsu_runtime.c
+++ b/c_runtime/bosatsu_runtime.c
@@ -2753,6 +2753,15 @@ static void bsts_print_test_summary(int indent, int passes, int fails) {
   }
 }
 
+_Bool bsts_test_argv_has_quiet(int argc, char** argv) {
+  for (int i = 1; i < argc; i++) {
+    if (strcmp(argv[i], "--quiet") == 0) {
+      return 1;
+    }
+  }
+  return 0;
+}
+
 BSTS_PassFail bsts_check_test(BValue v, int indent) {
   int passes = 0;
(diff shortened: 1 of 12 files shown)


johnynek:
Fixed the CI `testJS` compile failure by updating `jsui/src/test/scala/dev/bosatsu/jsui/StoreTest.scala` to match `Output.TestOutput` by type (`case _: Output.TestOutput =>`) instead of constructor-arity matching. This keeps the test robust when `TestOutput` gains fields (like `quiet`). Verified locally with `sbt "jsuiJS/testOnly * -- --log=failure"` and the required pre-push check `scripts/test_basic.sh`; both passed.

PR #1879, opened 2026-03-01
Clarify invalid `recur` diagnostics with direct `match` guidance

### LINKED ISSUE #1875: Confusing diagnostic for invalid recur usage should suggest match directly
### Summary
When `recur` is used in a non-recursive context, the diagnostic can be misleading (`Function name looks renamed`) and does not directly explain that `recur` should be replaced with `match`.

### Reproduction
File (`/tmp/repro_recur_confusing.bosatsu`):
```bosatsu
package Repro/RecurConfusing

def parse_non_empty_int(s: String) -> Option[Int]:
  if s matches "": None
  else: string_to_Int(s)

def parse_problem_tokens(tokens: List[String]) -> Option[Int]:
  recur tokens:
    case [first, *_]:
      parse_non_empty_int(first)
    case []:
      None
```

Command:
```bash
bosatsu tool check --input /tmp/repro_recur_confusing.bosatsu --output /tmp/repro_recur_confusing.pkg
```

### Actual output
```text
Function name looks renamed: declared `parse_problem_tokens`, but recursive calls use `parse_non_empty_int`.
Did you mean `parse_problem_tokens` in recursive calls?
```

### Expected output
A message that directly explains this is invalid `recur` usage and suggests replacing `recur <expr>:` with `match <expr>:` for non-recursive branching.

### Environment
- bosatsu: `0.0.36`
- OS: macOS arm64

johnynek:
Implemented a focused fix for issue #1875 by updating `RecursionCheck.RecursiveDefNoRecur` so `recur` blocks with no self-recursive calls now explicitly say to replace `recur <expr>:` with `match <expr>:` for non-recursive branching. When a likely renamed call exists, that hint is still preserved as secondary context. Updated `ErrorMessageTest` to reflect the new `recur` message and to assert the new guidance appears in the renamed-call scenario as well. Validation: `sbt "coreJVM/testOnly dev.bosatsu.ErrorMessageTest"` passed, and required `scripts/test_basic.sh` passed.

Fixes #1875

### DIFF
--- a/core/src/main/scala/dev/bosatsu/RecursionCheck.scala
+++ b/core/src/main/scala/dev/bosatsu/RecursionCheck.scala
@@ -142,8 +142,10 @@ object RecursionCheck {
     private def missingRecursiveCallMessage =
       s"${recurKind.keyword} but no recursive call to ${fnname.sourceCodeRepr}"
 
-    private def likelyRenameMessage(calledName: Bindable, count: Int) =
-      s"Function name looks renamed: declared `${fnname.sourceCodeRepr}`, but recursive calls use `${calledName.sourceCodeRepr}`.\nDid you mean `${fnname.sourceCodeRepr}` in recursive calls? ($count occurrences)"
+    private def likelyRenameMessage(calledName: Bindable, count: Int) = {
+      val occurrenceWord = if (count == 1) "occurrence" else "occurrences"
+      s"Function name looks renamed: declared `${fnname.sourceCodeRepr}`, but recursive calls use `${calledName.sourceCodeRepr}`.\nDid you mean `${fnname.sourceCodeRepr}` in recursive calls? ($count $occurrenceWord)"
+    }
 
     private val nonRecursiveRecurHint =
       "For non-recursive branching, replace `recur <expr>:` with `match <expr>:`."


core/src/main/scala/dev/bosatsu/RecursionCheck.scala
```diff
@@ -139,12 +139,31 @@ object RecursionCheck {
       likelyRenamedCall: Option[(Bindable, Int)]
   ) extends Error {
     def region = recurRegion
+    private def missingRecursiveCallMessage =
+      s"${recurKind.keyword} but no recursive call to ${fnname.sourceCodeRepr}"
+
+    private def likelyRenameMessage(calledName: Bindable, count: Int) =
+      s"Function name looks renamed: declared `${fnname.sourceCodeRepr}`, but recursive calls use `${calledName.sourceCodeRepr}`.\nDid you mean `${fnname.sourceCodeRepr}` in recursive calls? ($count occurrences)"
```
johnynek: check for number agreement. if `count == 1` we should say `(1 occurence)` not `(1 occurences)`.

johnynek: Good catch. I updated `RecursiveDefNoRecur.likelyRenameMessage` in `RecursionCheck.scala` to use number agreement: `occurrence` when `count == 1`, otherwise `occurrences`.

johnynek:
Applied the requested grammar fix for occurrence count in the recur rename diagnostic. Ran the required pre-push check `scripts/test_basic.sh`; it passed locally (Passed: 61, Failed: 0).

PR #1881, opened 2026-03-01
Fix lib-eval Duration/Instant struct-1 representation for now_mono/duration_to_nanos

### LINKED ISSUE #1876: lib eval mis-handles now_mono/duration_to_nanos representation (ProductValue(ExternalValue) vs Int)
### Summary
`bosatsu lib eval` appears to mis-handle the runtime representation of `Duration` from `now_mono`: values flow as `ProductValue(ExternalValue(...))` where an `Int` is expected.

This causes `lib eval` failure, while compiled C backend behavior differs.

### Minimal reproduction
File (`src/MyLib/MinParensProbs.bosatsu`):
```bosatsu
package MyLib/MinParensProbs

from Bosatsu/Prog import Prog, Main, await, pure, recover
from Bosatsu/IO/Error import IOError
from Bosatsu/IO/Core import Duration, now_mono, duration_to_nanos

main = Main(_ -> (
  d <- now_mono.await()
  pure(duration_to_nanos(d))
).recover(_ -> pure(1)))
```

### Reproduction commands
```bash
bosatsu lib check --name mylib --filter 'MyLib/MinParensProbs'
bosatsu lib eval --main MyLib/MinParensProbs --run p
```

### Actual `lib eval` output
```text
expected Main to return an Int exit code, found: ProductValue(ExternalValue(...))
```
(When arithmetic is applied to `duration_to_nanos(d)`, it can also fail with `expected integer: ProductValue(ExternalValue(...))`.)

### Additional signal
Building/running via C backend does not crash in the same way, suggesting an interpreter/runtime representation mismatch.

### Likely mismatch area
- `MatchlessToValue` treats struct-1 as identity (`MakeStruct(1)` -> identity; `GetStructElement(..., sz=1)` -> identity).
- Scala `Predef` appears to box `Duration`/`Instant` as `ProductValue.single(VInt(...))`.

### Expected behavior
`lib eval` and compiled backend should agree on struct-1 representation for `Duration` and `duration_to_nanos` should produce an `Int` value consistently.

### Environment
- bosatsu: `0.0.36`
- OS: macOS arm64

johnynek:
Implemented issue #1876 by aligning interpreter runtime values with struct-1 identity semantics.

What changed:
- Updated `core/src/main/scala/dev/bosatsu/Predef.scala` so `instantValueFromNanos` and `durationValueFromNanos` return `VInt(...)` (identity representation) instead of `ProductValue.single(...)`.
- Updated `asDurationNanos` to accept `VInt` directly, while still accepting legacy single-field `ProductValue` values for compatibility.
- Added regression test `lib eval --run handles now_mono duration_to_nanos newtype representation` in `core/src/test/scala/dev/bosatsu/ToolAndLibCommandTest.scala` that reproduces the reported `lib eval --run` flow using `now_mono.await()` and `duration_to_nanos`.

Reproduction/verification:
- Before fix, the new regression failed with `expected integer: ProductValue(ExternalValue(...))`.
- After fix, the regression passes and returns successful run exit handling.

Tests run:
- `sbt "coreJVM/testOnly dev.bosatsu.ToolAndLibCommandTest"` (pass)
- `scripts/test_basic.sh` (pass on rerun; first run hit transient timeouts in `Issue1633Test` and `LibBuildImplicitPackageTest`, both of which passed in isolated rerun)

Fixes #1876

### DIFF
--- a/core/src/test/scala/dev/bosatsu/RingOptLaws.scala
+++ b/core/src/test/scala/dev/bosatsu/RingOptLaws.scala
@@ -8,7 +8,6 @@ import org.scalacheck.Prop.forAll
 import org.scalacheck.{Arbitrary, Gen, Shrink}
 
 class RingOptLaws extends munit.ScalaCheckSuite {
-
   override def scalaCheckTestParameters =
     super.scalaCheckTestParameters
       .withMinSuccessfulTests(1000)
@@ -1017,14 +1016,19 @@ class RingOptLaws extends munit.ScalaCheckSuite {
   }
 
   property("normalize is commutative for Add") {
-    def law[A: Hash: Order: Show](a: Expr[A], b: Expr[A], w: Weights) = {
+    def law[A: Hash: Order: Show: Numeric](a: Expr[A], b: Expr[A], w: Weights) = {
       val left = normalize(a + b, w)
       val costLeft = w.cost(left)
       val right = normalize(b + a, w)
       val costRight = w.cost(right)
       assertEquals(
-        left,
-        right,
+        Expr.toValue(left),
+        Expr.toValue(right),
+        show"left($costLeft)=$left right($costRight)=$right"
+      )
+      assertEquals(
+        costLeft,
+        costRight,
         show"left($costLeft)=$left right($costRight)=$right"
       )
     }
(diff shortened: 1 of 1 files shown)


johnynek:
Make sure to audit the js and jvm predef implementations for any use of Product.single. If you find one, write a test in bosatsu code to validate that it is correct. I don't think any predef external implementations should be using single.

Also make sure any code expecting to process such a type (such as those expecting Instant or Duration) don't expect a single item Product (we remove that kind of wrapping in Matchless).

johnynek:
Addressed the PR feedback by auditing both JVM and JS predef IO-core implementations for struct-1 wrapping mismatches.

Changes made:
- JVM predef (`core/src/main/scala/dev/bosatsu/Predef.scala`):
  - `list_dir` now returns `Path` values as struct-1 identity (`Str`) instead of `ProductValue.single(Str(...))`.
- JS predef (`core/.js/src/main/scala/dev/bosatsu/PredefIoCorePlatform.scala`):
  - `instantValueFromNanos` and `durationValueFromNanos` now use identity (`VInt`) instead of `ProductValue.single(...)`.
  - `list_dir` now returns identity `Path` values (`Str`) instead of single-field products.
  - `asDurationNanos` now accepts identity `VInt` (while retaining compatibility for single-field products).
- Regression tests (`core/src/test/scala/dev/bosatsu/ToolAndLibCommandTest.scala`):
  - Added `lib eval --run handles now_wall instant newtype representation`.
  - Added `lib eval --run handles list_dir path newtype representation`.
  - Added small test helpers for minimal Prog/IO modules used by these regressions.

Validation:
- `sbt "coreJS/compile"` passed.

Issue #1882, opened 2026-03-01
Autograd


johnynek:
Now that we have Float64 and Array[_] types we could imagine having the compiler support gradients.

What would that look like to have statically checked gradients of functions?

What is the syntax? When we try to take an invalid gradient it should fail. Can we always detect? Would differentiable functions have to be a different type? I guess so. But we would want the ability to use them as non-differentiable functions so you need a subtype relationship.

Do we need type level integers to track the sizes of tensors to make this work?

What is some prior art on this? Are there good papers that exist proposing good solutions to this?

johnynek:
MergeXO assigned an agent and started design work for issue #1882.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/1884

PR #1884, opened 2026-03-01
Design doc for #1882: Issue #1882 Design: Statically Checked Autograd

johnynek:
Design doc.

Refs #1882


docs/design/1882-autograd.md
```diff
@@ -0,0 +1,364 @@
...
+
+### 1. Distinct differentiable-function type
+
+`Diff[a]` is distinct from `a -> Float64`.
+
+### 2. Subtype/coercion relation
+
+Add compiler rule:
+
+`Diff[a] <:< (a -> Float64)`
```
johnynek, line 129: explain how we would do this. This seems to be a new kind of judgement in `Infer`. It seems like we would have to do a new kind of subsumption check where we add a runtime call during on the subsumption judgement.

For instance, if are judging `Diff[b] <:< (a -> Float64)` we would check if `a <:< b` and then apply `run_Diff` to the `Diff[b]` value to get the `a -> Float64`. Or something like this?

Please be more explicit with how Infer needs to change.

docs/design/1882-autograd.md
```diff
@@ -0,0 +1,364 @@
...
+
+This is one-way only. Ordinary functions are not implicitly differentiable.
+
+### 3. Intrinsic typing rules
+
+1. `diff(e)` requires `e` to typecheck as `a -> Float64`.
+2. `diff(e)` runs a differentiability checker on typed `e`.
```
johnynek, line 136: explain in more detail how this differentiability checker looks. Now, we only compile one package at a time, and the only things we read from imports are the exported names and exported types. How does this compose? Can you only call other `Diff[a]` inside and some intrinsics? I guess that's right, but explain how.

Can we handle some branching? What kind? Be explicit.

docs/design/1882-autograd.md
```diff
@@ -0,0 +1,364 @@
...
+4. `grad(e)` requires `e: Diff[a]`; result type is `a -> a`.
+
+### 4. Supported input type domain for this issue
+
+Initial accepted `a` domain:
+
+1. `Float64`
+2. Tuples recursively composed of `Float64`
+
+`Array[Float64]` is explicitly deferred in this issue (see rollout/follow-up). Calls that attempt `diff` on unsupported domains fail with dedicated autograd type errors.
```
johnynek, line 183: add more explanation why `Array[Float64]` is difficult. It's one of the most obvious use cases.

Also, what about data types that only have nested tuples inside:

```
struct Pair(p: (Float64, Float64))
```
could we handle recursions on these easily?

docs/design/1882-autograd.md
```diff
@@ -0,0 +1,364 @@
...
+from Bosatsu/Predef import Float64
+
+export (
+  Diff,
+  diff,
+  grad,
+  run_Diff,
+)
+
+struct Diff[a](run: a -> Float64, grad: a -> a)
```
johnynek, line 88: I'm confused about how the type for `grad` is `a -> a`. Can you explain that in more detail and give examples with `Float64` and tuples of `Float64`?

docs/design/1882-autograd.md
```diff
@@ -0,0 +1,364 @@
...
+```
+
+### 2. Syntax
+
+No new parser syntax is required in this design. Usage is normal function-call syntax:
+
+```bosatsu
+from Bosatsu/Num/Autograd import diff, grad
+
+square = diff(x -> x * x)
```
johnynek, line 109: what about cases where we want to have variables that aren't part of the differentiation? For instance we have a Bool and if true we have one function and if false another, but both of those functions are differentiable. This is a common case. How is this handled?

docs/design/1882-autograd.md
johnynek: Updated the design doc with an explicit `Infer` plan in section `Subsumption/coercion changes in Infer (explicit)`. It now specifies a new `subsCheckRho2` case for `Diff[b] <:< (a -> Float64)`, reducing through `runType = b -> Float64`, reusing existing function subsumption, and composing coercion by inserting `run_Diff` at runtime (`coerce_fn(run_Diff(d))`). It also states the contravariant consequence `a <:< b` through existing function rules.

docs/design/1882-autograd.md
johnynek: Expanded `Differentiability checker` with a new subsection on composition under Bosatsu’s package model (`imports only expose exported names/types`). The doc now explicitly limits cross-package calls inside `diff` bodies to: local inspectable functions, known derivative-registry globals, or `Diff`-typed imported values. It also adds a concrete branching policy section: non-active conditions are allowed (with examples), active-input-dependent branch conditions are rejected in phase 1.

docs/design/1882-autograd.md
johnynek: Added a dedicated `Why Array[Float64] is deferred` section with concrete reasons (runtime shape handling, scatter-add accumulation, optimizer/lowering implications, and dynamic guards without shape types). Also added `Product structs and recursive ADTs`: non-recursive product structs like `struct Pair(p: (Float64, Float64))` are called out as much easier (tuple-isomorphic tangent structure), while recursive ADTs remain out-of-scope in phase 1 due current recursion restrictions in `diff` bodies.

docs/design/1882-autograd.md
johnynek: Added a new `Why grad has type a -> a` subsection with explicit examples for `Float64`, `(Float64, Float64)`, and nested tuples. The section now explains that gradient output mirrors the input product structure and shows representative signatures and sample partial-derivative interpretation.

docs/design/1882-autograd.md
johnynek: Added explicit handling for non-differentiated control variables in `Branching policy (explicit)`. The doc now distinguishes active vs non-active values and includes examples where a `Bool` controls which differentiable branch to use (`choose`/`choose2`) while still allowing differentiation with respect to float inputs. It also states that branch conditions depending on active differentiable inputs are rejected in phase 1.

johnynek:
Addressed all five review comments by directly revising `docs/design/1882-autograd.md` (including cleaning duplicated front-matter/header content and expanding the requested technical details). Required pre-push check passed: `scripts/test_basic.sh` -> `Passed: Total 61, Failed 0, Errors 0`.

johnynek:
`origin/main` has been merged into the PR branch, and the required pre-push check passed locally: `scripts/test_basic.sh` -> `Passed: Total 67, Failed 0, Errors 0, Passed 67`. The first cold run hit a timeout in `ClangGenTest.test_workspace/Ackermann.bosatsu`, but the isolated rerun passed and the subsequent full warm rerun passed cleanly.

johnynek:
MergeXO feedback automation is blocked because the agent returned `commit_message` but no new staged changes or local commits were detected.

Action: request concrete file edits (or explicit `git_ops`), then reset blocked feedback state.

PR #1922, opened 2026-03-02
Implement typed Int recursion decrease checks via SMT in recursion checker

### LINKED ISSUE #1918: Implement recursion on ints
Read the design doc here:

docs/src/main/paradox/design-docs/recur_int_decrease_design.md

Implement this design.

I think the approach we should take is for each recursive call, identify the full set of bindings in scope that impact the value in the given path. Then we attempt to lift that expression into SMT. We can lift some predef operations on Ints into the SMT-LIB language. We will have to find an encoding of Comparison (LT, EQ, GT), I guess encoding those as 0, 1, 2, or -1, 0, 1, would work. We can lift the int ops, bindings of ints, booleans, matching literals all into SMT-LIB.

The goal is to be able to fully typecheck int_loop and reject similar non-terminating loops on integers that don't check conditions correctly, with excellent error messages.

johnynek:
Implemented issue #1918 by extending `TypedExprRecursionCheck` with a type-directed `Int` recursion strategy. Recursive self-calls on `Int` recur targets now build path conditions from match patterns/guards, lower supported `Int`/`Bool`/`Comparison` expressions to SMT, and discharge obligations (`next >= 0` and `next < current`) using `dev.bosatsu.smt.Z3Api` + embedded Z3. Added scoped SMT state tracking through lets/lambdas/loops/matches so branch-local facts and bindings are respected, and integrated this with existing lexicographic recursion checks while preserving structural behavior for non-Int targets. Added new recursion diagnostic `IntRecursionObligationFailed` with obligation text, path condition, and solver details/model when available. Added typed recursion tests covering accepted `int_loop`-style recursion and rejecting non-decreasing/non-nonnegative `Int` recursion cases. Updated `build.sbt` so `scalawasiz3` is available in `core` main code. Required test command `scripts/test_basic.sh` was run and passed.

Fixes #1918


core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,827 @@ object TypedExprRecursionCheck {
...
+        expr: TypedExpr[Declaration]
+    ): TypedExpr[Declaration] =
+      expr match {
+        case TypedExpr.Generic(_, in)       => stripExprWrappers(in)
+        case TypedExpr.Annotation(in, _, _) => stripExprWrappers(in)
+        case other                          => other
+      }
+
+    private def mkAnd(args: Vector[SmtExpr.BoolExpr]): SmtExpr.BoolExpr =
+      args.toList match {
```
johnynek: Don't convert to List just to match on the size. Just check the size: 0, 1 or more than 1.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,827 @@ object TypedExprRecursionCheck {
...
+        case Nil          => SmtExpr.BoolConst(true)
+        case one :: Nil   => one
+        case _ :: _ :: _  => SmtExpr.And(args)
+      }
+
+    private def mkOr(args: Vector[SmtExpr.BoolExpr]): SmtExpr.BoolExpr =
+      args.toList match {
```
johnynek: Same comment about not converting to List just to match on size. Vector has a cheap size.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,827 @@ object TypedExprRecursionCheck {
...
+            case Some(Identifier.Name("div")) =>
+              lowerBinaryInt(args, state)(SmtExpr.Div(_, _))
+            case Some(Identifier.Name("mod_Int")) =>
+              lowerBinaryInt(args, state)(SmtExpr.Mod(_, _))
+            case _ =>
+              (None, state)
+          }
+        case TypedExpr.Match(arg, branches, _) =>
+          lowerIntIfExpr(arg, branches, state)
+        case _ =>
```
johnynek, line 1155: Why wouldn't we handle Let here?

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,827 @@ object TypedExprRecursionCheck {
...
+                SmtExpr.Mul(Vector(left, right))
+              )
+            case Some(Identifier.Name("div")) =>
+              lowerBinaryInt(args, state)(SmtExpr.Div(_, _))
+            case Some(Identifier.Name("mod_Int")) =>
+              lowerBinaryInt(args, state)(SmtExpr.Mod(_, _))
+            case _ =>
+              (None, state)
+          }
+        case TypedExpr.Match(arg, branches, _) =>
```
johnynek, line 1151: Shouldn't we check if arg has type Bool to save effort here?

Also can't we handle matching ok Comparison and Int potentially too?

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,827 @@ object TypedExprRecursionCheck {
...
+    ): (Option[SmtExpr.BoolExpr], SmtBranchState) =
+      stripExprWrappers(expr) match {
+        case TypedExpr.Local(name, tpe, _) if isBoolType(tpe) =>
+          val (boolExpr, state1) = ensureBoolLocal(name, state)
+          (Some(boolExpr), state1)
+        case TypedExpr.App(fn, args, _, _) =>
+          predefFnName(fn) match {
+            case Some(Identifier.Name("eq_Int")) =>
+              lowerBinaryInt(args, state)(SmtExpr.EqInt(_, _))
+            case Some(Identifier.Operator(">")) =>
```
johnynek: These operators aren't defined in predef. Where are they from?

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,827 @@ object TypedExprRecursionCheck {
...
+              lowerBinaryInt(args, state)(SmtExpr.Lt(_, _))
+            case Some(Identifier.Operator(">=")) =>
+              lowerBinaryInt(args, state)(SmtExpr.Gte(_, _))
+            case Some(Identifier.Operator("<=")) =>
+              lowerBinaryInt(args, state)(SmtExpr.Lte(_, _))
+            case _ =>
+              (None, state)
+          }
+        case TypedExpr.Match(arg, branches, _) =>
+          lowerBoolFromMatches(arg, branches, state)
```
johnynek, line 1349: Why can't we handle a Let here?

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
```diff
@@ -144,6 +144,46 @@ def bad(n, m):
 """)
   }
 
+  test("recur supports Int targets with provable decrease and non-negative next values") {
```
johnynek, line 203: Write an example that has one structural arg and one Int arg and recur on both and show it still works.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
```diff
@@ -144,6 +144,46 @@ def bad(n, m):
...
+    allowed("""#
+def int_loop[a](i: Int, state: a, fn: (Int, a) -> (Int, a)) -> a:
+  recur i:
+    case _ if cmp_Int(i, 0) matches GT:
+      (next_i, next_state) = fn(i, state)
+      if cmp_Int(next_i, 0) matches GT:
```
johnynek: Change this to:

c = cmp_Int(next_i, 0) matches GT
if c:
  ...

We should be robust to natural refactors like that.

Or what if user writes:

match cmp_Int(next_i, 0):
  case LT | EQ: ...
  case _: ...

When we have a Comparison we know it is -1, 0, 1. So we can assert that so Z3 can solve the wildcard case there.

We can't be brittle and have this be usable.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Updated `mkAnd` to branch on `Vector.size` (`0`, `1`, or `>1`) instead of converting to `List`.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Updated `mkOr` the same way: `Vector.size`-based branching with no `toList` conversion.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Added `Let` lowering support for int/comparison paths via `lowerWithLetBinding`, and `lowerIntExpr` now handles `TypedExpr.Let` directly.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Added the Bool type guard for the fast `if`-style int match path (`lowerIntIfExpr` only used when match arg type is Bool), and added ordered general match lowering (`lowerIntFromMatches`) for other scrutinee types, including `Comparison`/`Int`. Also constrained fresh `Comparison` locals to `-1/0/1`.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Removed unsupported lifting for `Identifier.Operator(>, <, >=, <=)` in `lowerBoolExpr`. Bool lowering now uses supported forms (`eq_Int`, and `cmp_Int`/`matches` patterns via match lowering).

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Added `Let` handling in `lowerBoolExpr` (reusing `lowerWithLetBinding`) so let-bound bool expressions are lowered instead of falling back to fresh symbols.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
johnynek: Added a mixed-target positive test in `TypedExprRecursionCheckTest`: `tuple recur targets allow mixed structural and Int decreases`.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
johnynek: Updated the positive int recursion test to use `c = cmp_Int(next_i, 0) matches GT` before branching, and included a `Comparison` match with `case LT | EQ` / `case _` form in the same flow. Match lowering now models branch order, and comparison-domain facts (`-1|0|1`) are asserted for comparison locals.

johnynek:
Implemented the broader PR feedback as code changes: moved `int_loop` from external runtime hooks into `predef.bosatsu`, removed its external implementations from Scala/Python/C runtime paths, and updated docs that described `int_loop` as external. Also tightened typed recursion SMT lowering (Let support across bool/int/comparison lowering, ordered match lowering for bool/int expressions, comparison-domain constraints, and a path-implies fast path to avoid unnecessary solver calls on trivially implied obligations). Updated affected codegen tests (Clang hash snapshot and Python module-loading test harness for generated dependency modules including `Bosatsu/Predef`). Ran `scripts/test_basic.sh` successfully.

core/src/main/resources/bosatsu/predef.bosatsu
```diff
@@ -263,7 +259,23 @@ external def divf(a: Float64, b: Float64) -> Float64
 external def cmp_Float64(a: Float64, b: Float64) -> Comparison
 
 # Loop until the returned Int is <= 0 or >= intValue.
-external def int_loop(intValue: Int, state: a, fn: (Int, a) -> (Int, a)) -> a
+def int_loop(intValue: Int, state: a, fn: (Int, a) -> (Int, a)) -> a:
+  recur intValue:
```
johnynek: This should be loop intValue: to ensure we compile to a loop.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
```diff
@@ -144,6 +144,73 @@ def bad(n, m):
...
+          mixed(n, next_i)
+        else:
+          mixed(prev, 10)
+      else:
+        mixed(prev, 10)
+""")
+  }
+
+  test("Int recursion rejects non-decreasing recursive calls") {
+    disallowed("""#
```
johnynek: We should assert something about the compiler error message here. It needs to be good.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
```diff
@@ -144,6 +144,73 @@ def bad(n, m):
...
+  recur i:
+    case _ if cmp_Int(i, 0) matches GT:
+      bad(i)
+    case _:
+      i
+""")
+  }
+
+  test("Int recursion rejects calls that do not prove non-negative next values") {
+    disallowed("""#
```
johnynek: We need to assert something about the compiler error message here.

docs/src/main/paradox/language_guide.md
```diff
@@ -994,31 +994,25 @@ safety story gets much weaker.
 So "use with caution" mostly applies to maintainers of trusted runtime/predef
 code, not to ordinary Bosatsu library authors.
 
-An example function we cannot implement in Bosatsu is:
+A function that can now be implemented directly in Bosatsu is:
```
johnynek: Don't say "can now". Rewrite the document to explain in the current state of the code. So explain that recursion on Int is allowed as long as you check that the Int is getting smaller and greater than zero.

docs/src/main/paradox/language_guide.md
```diff
@@ -994,31 +994,25 @@ safety story gets much weaker.
...
-decrease or tuple-target lexicographic decrease). This gives a simple proof that
-the loop will terminate.
-
-Instead, we implement this function in Predef as an external def that has to be
-supplied to the compiler with a promise that it is total and matches its
-declared type.
-```
-external def int_loop(intValue: Int, state: a, fn: (Int, a) -> (Int, a)) -> a
-```
+  recur int_v:
```
johnynek: Use loop int_v:

Always use loop if we have tail recursion.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -243,14 +298,21 @@ object TypedExprRecursionCheck {
...
         .flatMap { case (group, gidx) =>
           group.iterator.zipWithIndex.flatMap { case (item, idx) =>
-            item.topNames.iterator.map { topName =>
-              topName -> RecurTargetItem(gidx, idx, topName)
+            val maybeTypedName =
+              typedArgs
+                .get(gidx.toLong)
+                .flatMap(_.get(idx.toLong))
+                .map(_._1)
+            item.topNames.iterator.flatMap { topName =>
```
johnynek: Why a flatMap with a single List as the result. Why not just map and no List?

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,979 @@ object TypedExprRecursionCheck {
...
+    private object ProofOutcome {
+      case object Proved extends ProofOutcome
+      case class Failed(
+          model: Option[String],
+          detail: Option[String]
+      ) extends ProofOutcome
+    }
+
+    private def isIntType(tpe: Type): Boolean =
+      Type.normalize(tpe) == Type.IntType
```
johnynek: tpe.sameAs(Type.IntType)

Is better style.

Same below.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,979 @@ object TypedExprRecursionCheck {
...
+                    SmtExpr.EqInt(left, right),
+                    SmtExpr.IntConst(BigInt(0)),
+                    SmtExpr.IntConst(BigInt(1))
+                  )
+                )
+              }
+              (compared, state1)
+            case _ =>
+              (None, state)
+          }
```
johnynek, line 994: Why can't we handle a match where the scrutinee is Bool, Comparison or Int and the result is a Comparison?

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,979 @@ object TypedExprRecursionCheck {
...
+      ((patOpt, guardOpt).mapN((pat, guard) => mkAnd(Vector(pat, guard))), state2)
+    }
+
+    private def isDefinitelyTrue(expr: SmtExpr.BoolExpr): Boolean =
+      expr match {
+        case SmtExpr.BoolConst(true) =>
+          true
+        case SmtExpr.And(args)       =>
+          args.forall(isDefinitelyTrue)
+        case _                       =>
```
johnynek: Can't we add Or and exists(_.isDefinitelyTrue)?

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,979 @@ object TypedExprRecursionCheck {
...
+              )) &&
+              (falseBranch.pattern == Pattern.PositionalStruct(
+                (PackageName.PredefName, Identifier.Constructor("False")),
+                Nil
+              )) =>
+          val (condOpt, state1) = lowerBoolExpr(arg, state)
+          val (thenOpt, state2) = lowerIntExpr(trueBranch.expr, state1)
+          val (elseOpt, state3) = lowerIntExpr(falseBranch.expr, state2)
+          ((condOpt, thenOpt, elseOpt).mapN(SmtExpr.Ite(_, _, _)), state3)
+        case _ =>
```
johnynek, line 1044: Why can't we handle guard statements if they are lowerable to Bool in SMT?

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,979 @@ object TypedExprRecursionCheck {
...
+        state: SmtBranchState
+    ): (Option[SmtExpr.IntExpr], SmtBranchState) =
+      branches.toList match {
+        case Nil =>
+          (None, state)
+        case init :+ last =>
+          val (lastCondOpt, state1) = lowerMatchBranchCondition(argExpr, last, state)
+          val (lastExprOpt, state2) = lowerIntExpr(last.expr, state1)
+          (lastCondOpt, lastExprOpt) match {
+            case (Some(lastCond), Some(lastExpr)) if isDefinitelyTrue(lastCond) =>
```
johnynek: Add a comment why this is the only case we can handle

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,979 @@ object TypedExprRecursionCheck {
...
+                      Vector(SmtExpr.IntConst(BigInt(-1)), right)
+                    )
+                  )
+                )
+              )
+            case Some(Identifier.Name("mul")) =>
+              lowerBinaryInt(args, state)((left, right) =>
+                SmtExpr.Mul(Vector(left, right))
+              )
+            case Some(Identifier.Name("div")) =>
```
johnynek, line 1126: Bosatsu defines div and mod by 0: 4 / 0 = 0 and 4 % 0 == 4.

Add a comment about mapping this to z3 semantics.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,979 @@ object TypedExprRecursionCheck {
...
+              case "EQ" => Some(SmtExpr.EqInt(term, SmtExpr.IntConst(BigInt(0))))
+              case "GT" => Some(SmtExpr.EqInt(term, SmtExpr.IntConst(BigInt(1))))
+              case _    => None
+            }
+          }
+          (asEq, state2)
+      }
+    }
+
+    private def lowerPatternCondition(
```
johnynek, line 1233: Add a comment what this does. I think it is lowering the pattern matched against an argument to evaluate a branch path.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,979 @@ object TypedExprRecursionCheck {
...
+        argExpr: TypedExpr[Declaration],
+        pattern: Pattern[(PackageName, Identifier.Constructor), Type],
+        state: SmtBranchState
+    ): (Option[SmtExpr.BoolExpr], SmtBranchState) =
+      pattern match {
+        case Pattern.WildCard | Pattern.Var(_) =>
```
johnynek: Don't we need to record that if a Var matches the pattern branch has an additional binding and what that binding is?

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,979 @@ object TypedExprRecursionCheck {
...
+        case Some(SmtExpr.BoolConst(true)) => state1
+        case Some(cond)                    => state1.addPathFact(cond)
+        case None                          => state1
+      }
+    }
+
+    private def buildPathCondition(state: SmtBranchState): SmtExpr.BoolExpr =
+      mkAnd(state.pathFacts)
+
+    private def pathImplies(
```
johnynek, line 1435: This seems like a partial check. Is this some check for trivial implication before bothering z3? Please add a comment

core/src/main/resources/bosatsu/predef.bosatsu
johnynek: Updated `predef.bosatsu` so `int_loop` now uses `loop intValue:` instead of `recur intValue:`.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
johnynek: Added explicit error-message assertions for this failing Int recursion case. I introduced `disallowedWithMessage` in `TypedExprRecursionCheckTest` and now assert key message content (`cannot prove Int recursion obligation`, `recur target: i`, and `path condition:`).

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
johnynek: Added error-message assertions for this case as well, using the same helper and checking for the same core diagnostics in the recursion error message.

docs/src/main/paradox/language_guide.md
johnynek: Rewrote the language guide section in present tense (removed "can now") and described current Int recursion requirements directly in terms of proving decreasing/non-negative recursive Int arguments on recursive paths.

docs/src/main/paradox/language_guide.md
johnynek: Updated the language guide example to use `loop int_v:` and explicitly frames `loop` as the tail-recursive form.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Simplified `getRecurTargetItemsByName`: replaced the `flatMap` that emitted a single-element `List` with a direct `map`.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Switched the type checks to `sameAs` style (`tpe.sameAs(Type.IntType)`, `tpe.sameAs(Type.BoolType)`, and `tpe.sameAs(comparisonType)`).

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Implemented this: `lowerComparisonExpr` now handles `Match` expressions when the scrutinee is Bool/Comparison/Int (via `isLowerableMatchScrutinee`) by lowering through a new `lowerComparisonFromMatches` path.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Added `Or` handling in `isDefinitelyTrue` with `args.exists(isDefinitelyTrue)`.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Handled this by making Bool-scrutinee Int match lowering fall back to the general match lowering (`lowerIntFromMatches`) when the specialized True/False if-form does not apply. That path already lowers guards via `lowerMatchBranchCondition`/`lowerBoolExpr`.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Added a comment at the `isDefinitelyTrue(lastCond)` gate in `lowerIntFromMatches` explaining why this is the only safe lowering case (requires guaranteed final branch; otherwise we’d need explicit exhaustiveness reasoning).

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Added comments and adjusted lowering to preserve Bosatsu semantics for divide/mod by zero: `div` now lowers to `ite(denom == 0, 0, div(...))` and `mod_Int` lowers to `ite(denom == 0, lhs, mod(...))`.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Added a comment above `lowerPatternCondition` describing that it lowers a branch pattern to a boolean path condition used for branch-local recursion obligations.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Added an explicit comment in the `Pattern.Var(_)` case noting that the binding is recorded in `bindPatternNames`; condition lowering itself returns `true` for that pattern.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Added a comment on `pathImplies` clarifying this is a cheap implication pre-check used to avoid unnecessary Z3 calls.

johnynek:
Addressed all listed review comments with focused code/doc/test updates. Also ran the required pre-push check `scripts/test_basic.sh` on the final tree; it passed.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
```diff
@@ -144,6 +160,81 @@ def bad(n, m):
...
+enum Nat: Zero, Succ(prev: Nat)
+
+def mixed(n: Nat, i: Int) -> Int:
+  recur (n, i):
+    case (Zero, _):
+      i
+    case (Succ(prev), _):
+      next_i = i.sub(1)
+      if cmp_Int(next_i, 0) matches GT:
+        if cmp_Int(next_i, i) matches LT:
```
johnynek: Isn't this line always true for Ints? Can we remove it and still compile?

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
```diff
@@ -144,6 +160,81 @@ def bad(n, m):
...
+      assert(clue(msg).contains("recur target: i"))
+      assert(clue(msg).contains("path condition:"))
+    }
+  }
+
+  test("Int recursion rejects calls that do not prove non-negative next values") {
+    disallowedWithMessage("""#
+def bad(i: Int) -> Int:
+  recur i:
+    case _ if cmp_Int(i, 0) matches GT:
```
johnynek, line 277: Can we write a version of this with cmp_Int(i, 1) matched GT and I think it should pass right?

Also what is the full error message. It would be good if we could have given something like that based on the result of the solver.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,1055 @@ object TypedExprRecursionCheck {
...
+              .withComparisonBinding(name, expr)
+              .addPathFact(comparisonInDomain(expr))
+          )
+      }
+
+    private def predefFnName(
+        fn: TypedExpr[Declaration]
+    ): Option[Identifier] =
+      stripExprWrappers(fn) match {
+        case TypedExpr.Global(pack, name, _, _) if pack == PackageName.PredefName =>
```
johnynek: Just write TypedExpr.Global(PackageName.Predef, name, _, _) =>

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,1055 @@ object TypedExprRecursionCheck {
...
+          Some(name)
+        case _ =>
+          None
+      }
+
+    private def boolLiteralValue(
+        expr: TypedExpr[Declaration]
+    ): Option[Boolean] =
+      stripExprWrappers(expr) match {
+        case TypedExpr.Global(pack, cons: Identifier.Constructor, _, _)
```
johnynek: Just write TypedExpr.Global(PackageName.Predef, ...

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,1055 @@ object TypedExprRecursionCheck {
...
+          }
+        case _ =>
+          None
+      }
+
+    private def comparisonLiteralValue(
+        expr: TypedExpr[Declaration]
+    ): Option[BigInt] =
+      stripExprWrappers(expr) match {
+        case TypedExpr.Global(pack, cons: Identifier.Constructor, _, _)
```
johnynek: Just write TypedExpr.Global(PackageName.Predef,

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -518,37 +581,1055 @@ object TypedExprRecursionCheck {
...
+    ): (Option[SmtExpr.IntExpr], SmtBranchState) =
+      branches.toList match {
+        case Nil =>
+          (None, state)
+        case init :+ last =>
+          val (lastCondOpt, state1) = lowerMatchBranchCondition(argExpr, last, state)
+          val (lastExprOpt, state2) = lowerIntExpr(last.expr, state1)
+          (lastCondOpt, lastExprOpt) match {
+            // We only lower to nested ite when the last branch is guaranteed to
+            // match. Otherwise, this would need explicit exhaustiveness checking.
```
johnynek: But we do have exhaustiveness checking in Bosatsu in the TotalityChecker. That will be run in compilation so we can assume it here. All valid Bosatsu matches are total so we don't have to be defensive in this check. We know one branch of a match will match.

core/src/main/resources/bosatsu/predef.bosatsu
```diff
@@ -263,7 +259,23 @@ external def divf(a: Float64, b: Float64) -> Float64
...
+        else:
+          next_state
+      else:
+        next_state
+    case _:
+      state
+
+# Repeat `item` `cnt` times.
+def replicate_List[a](item: a, cnt: Int) -> List[a]:
+  int_loop(cnt, EmptyList, (i, acc) -> (i.sub(1), NonEmptyList(item, acc)))
```
johnynek: Since z3 should know that i.sub(1) is less than i for integers can't we write a direct version of range that uses fewer comparisons? Can't we get away with a single comparison per loop?

int_loop requires two because the next int function is a black box. Look for other uses of int_loop to see if we can reduce the comparison count and still compile now.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
johnynek: Yes. I removed that redundant `cmp_Int(next_i, i) matches LT` check from the mixed structural+Int test branch and it still compiles/passes with the recursion checker.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
johnynek: Added a passing case for exactly this: `cmp_Int(i, 1) matches GT` with recursive call `ok(i.sub(2))` (`Int recursion allows i.sub(2) when guard proves i > 1`). I also tightened the failing test assertion to check the obligation-bearing message line (`cannot prove Int recursion obligation for bad: (>= ...`) plus target/path-condition lines so we validate more of the full solver-backed error message.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Updated to the direct pattern style: `TypedExpr.Global(PackageName.PredefName, name, _, _) => ...` in `predefFnName`.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Updated similarly in `boolLiteralValue` to match directly on `TypedExpr.Global(PackageName.PredefName, cons: Identifier.Constructor, _, _)`.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Updated similarly in `comparisonLiteralValue` to match directly on `TypedExpr.Global(PackageName.PredefName, cons: Identifier.Constructor, _, _)`.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Agreed. I removed the `isDefinitelyTrue(lastCond)` gate and now lower exhaustive matches assuming totality has already guaranteed one branch matches. I also added a comment in `lowerIntFromMatches` that this relies on totality; same simplification was applied consistently to bool/comparison match lowering.

core/src/main/resources/bosatsu/predef.bosatsu
johnynek: Implemented this by replacing `int_loop` usage in `replicate_List`, `range`, and `range_fold` with direct `loop`-based Int recursion (`go` helpers), each using only the loop guard comparison (`cmp_Int(..., 0) matches GT`) per iteration path. To keep these compiling robustly, I also added path-condition simplification (`simplifyBoolExpr`) and cheap arithmetic implication heuristics in `pathImplies` for subtraction-by-positive-constant goals, so these direct forms typecheck without depending on fragile solver behavior.

johnynek:
Applied all requested review updates across recursion checker, predef loops, and tests, plus updated the Clang codegen golden hash affected by predef-generated output changes. Required pre-push check passed on final state: `scripts/test_basic.sh`.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -632,6 +632,81 @@ object TypedExprRecursionCheck {
...
+          expr
+        case SmtExpr.Not(in) =>
+          simplifyBoolExpr(in) match {
+            case SmtExpr.BoolConst(value) => SmtExpr.BoolConst(!value)
+            case SmtExpr.Not(in1)         => in1
+            case other                    => SmtExpr.Not(other)
+          }
+        case SmtExpr.And(args) =>
+          val simp = args.map(simplifyBoolExpr)
+          if (simp.contains(SmtExpr.BoolConst(false))) SmtExpr.BoolConst(false)
```
johnynek: rather than reallocating BoolConst(false) and BoolConst(true) can we cache those in the `object BoolConst` so we have:

```scala
object BoolConst {
  val True = BoolConst(true)
  val False = BoolConst(false)
  inline def apply(b: Boolean): BoolConst = if (b) True else False
}
```
But prefer to use BoolConst.True and BoolConst.False?

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: don't match on the last element of a list. This is O(N). We should reverse the list in the scrutinee and restructure the loop.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -902,9 +978,10 @@ object TypedExprRecursionCheck {
           val (lastCondOpt, state1) = lowerMatchBranchCondition(argExpr, last, state)
           val (lastExprOpt, state2) = lowerIntExpr(last.expr, state1)
           (lastCondOpt, lastExprOpt) match {
-            // We only lower to nested ite when the last branch is guaranteed to
-            // match. Otherwise, this would need explicit exhaustiveness checking.
-            case (Some(lastCond), Some(lastExpr)) if isDefinitelyTrue(lastCond) =>
+            // Totality checking guarantees match exhaustiveness, so once every
```
johnynek, line 1059: can you add a comment why we don't seem to use any of the pattern which may have bindings what we will use. For instance, we could pattern match on int literals, and we should be able to pass those to smt and also bind them to names: `case 12 as twelve:`

Of course, a literal pattern match is never total for Int so it can't come last, but `case foo:` could come last and introduce a binding alias for the scrutinee or the in in question.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
```diff
@@ -144,6 +160,89 @@ def bad(n, m):
...
+
+  test("tuple recur targets allow mixed structural and Int decreases") {
+    allowed("""#
+enum Nat: Zero, Succ(prev: Nat)
+
+def mixed(n: Nat, i: Int) -> Int:
+  recur (n, i):
+    case (Zero, _):
+      i
+    case (Succ(prev), _):
```
johnynek: this should also pass if we wrote:
```
case (Succ(prev), i1):
  next_i = i1.sub(1)
  ...
```
We should be able to assert that `i` and `i1` are the same

Also, it should support
```
case (Succ(prev), (1 | 2) as one_or_two):
  mixed(n, one_or_two.sub(1))
```

as a branch above and in that branch we see that re have reduced the call.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Implemented this in `SmtExpr`: `BoolConst` now has cached singletons `BoolConst.True`/`BoolConst.False` with `inline def apply(b: Boolean)` returning the cached instance. I also switched `TypedExprRecursionCheck` to prefer `BoolConst.True`/`BoolConst.False` at call sites.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Refactored match lowering to avoid `init :+ last` on lists. `lowerIntFromMatches` now uses `branches.reverse` and a single fold to build the nested `ite` chain. I applied the same shape to `lowerComparisonFromMatches` and `lowerBoolFromMatches` for consistency.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Added the requested explanation comment in `lowerIntFromMatches` and also tightened behavior: branch expressions are now lowered with `bindPatternNames(...)`, so aliases like `case 12 as twelve:` are available during SMT lowering of branch bodies. In recur branches, I also added `bindSourcePatternAliases(...)` (from parsed source pattern) so whole-value aliases per recur-target component are tracked in SMT state.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
johnynek: Updated the mixed structural+Int test to include both requested forms:
- `case (Succ(prev), i1): ... next_i = i1.sub(1) ... mixed(n, next_i)`
- a preceding branch `case (Succ(prev), (1 | 2) as one_or_two): mixed(prev, one_or_two.sub(1))`
I also implemented source-pattern alias binding for recur branches so names like `i1` are mapped to the corresponding recur-target SMT value.

johnynek:
Applied all requested review changes in code/tests and verified with required pre-push checks. `scripts/test_basic.sh` passes on this final tree.

johnynek:
The code looks good to me now, the main concerning thing is the test coverage. Please run coverage on core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala and keep adding tests until all the new added lines are covered. We want 100% coverage for the SMT related paths as this code is subtle. We need to trigger the code from bosatsu code originally, not hand crafted TypedExpr ASTs. We need to see the real things users could write to trigger the branches.

johnynek:
Added focused source-level tests in `core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala` to exercise the SMT recursion paths from real Bosatsu code: bool-match-to-ite lowering, comparison-match lowering (including let-bound guard expressions), unsupported Int SMT lowering error reporting, equal/increasing Int recursive args, div/mod-by-zero lowering semantics, and nested match guard/pattern path-fact propagation. I reran `sbt "coverage; coreJVM/testOnly dev.bosatsu.TypedExprRecursionCheckTest; coverageReport"` and verified diff coverage for `TypedExprRecursionCheck.scala` added lines is now fully covered (intersection count with uncovered lines: 0). Required pre-push check `scripts/test_basic.sh` also passed.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
```diff
@@ -245,6 +245,109 @@ def bad(i: Int) -> Int:
...
+  recur i:
+    case _ if cmp_Int(i, 0) matches GT:
+      bad(i.add(1))
+    case _:
+      i
+""")
+  }
+
+  test("Int recursion lowering handles div/mod by zero semantics") {
+    disallowed("""#
```
johnynek, line 364: show an allowed example with a divide and concur recursion.

Something like:
```
if i > 1:
  i1 = i / 2
  i2 = i - i1
  f(i1) + f(i2)
```
this should work since we can see that `i1 >= 1` and less than `i` and likewise `i2` is in range. This is an important example.

johnynek: Added an allowed divide-and-conquer Int recursion example in `TypedExprRecursionCheckTest` (`split_sum`) using:
`i1 = i.div(2)`, `i2 = i.sub(i1)`, and `split_sum(i1).add(split_sum(i2))` under `cmp_Int(i, 1) matches GT`.

This now covers the important case where both recursive arguments are solver-proven to stay non-negative and strictly smaller than `i`.

johnynek:
Merged `origin/main` and resolved the merge conflict in `core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala` by keeping both branches' helper/tests and adding the requested divide-and-conquer recursion test. Validation passed with `sbt "coreJVM/testOnly dev.bosatsu.TypedExprRecursionCheckTest"` and required `scripts/test_basic.sh` (Passed: 61, Failed: 0).

Issue #1923, opened 2026-03-02
Syntax for Unit -> a functions


johnynek:
Sometimes we want to not compute something right now. Since all functions in Bosatsu take at least one are we can't have a zero are def or lambda.

But we could support this in syntax as a synonym for Unit -> a

So we could write it as a type like () -> a and also as a value:

later = () -> x

Or

def later(): gen_x

Calls like with no args would be replaced by calls with Unit: later() would become later(Unit) in SourceConvert.

This makes the syntax more pythonic and addresses actual (occasional) needs in code.

johnynek:
MergeXO assigned an agent and started design work for issue #1923.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/1924

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #1923.

johnynek:
Opened implementation PR: https://github.com/johnynek/bosatsu/pull/1928

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/1923#issuecomment-3985746421
- linked PR: #1928 (https://github.com/johnynek/bosatsu/pull/1928)

Comments on the source issue are no longer actioned after a PR exists. Please comment on the PR thread instead.

PR #1924, opened 2026-03-02
Design doc for #1923: Issue #1923 design: syntax for Unit -> a functions

johnynek:
Design doc.

Refs #1923

### DIFF
--- a/docs/design/1923-syntax-for-unit-a-functions.md
+++ b/docs/design/1923-syntax-for-unit-a-functions.md
@@ -80,7 +80,7 @@ The implementation keeps Bosatsu core invariants intact: internal function and a
 
 1. No new true nullary function representation in `Expr`, `TypedExpr`, or runtime IR.
 2. No change to general application semantics beyond Unit sugar.
-3. No change to constructor defaulting semantics.
+3. No constructor syntax change: zero-arg constructors remain constant-style (`True`, `False`), and `Foo()` remains unsupported.
 4. No feature flag or staged runtime rollout.
 
 ## Architecture
@@ -109,8 +109,8 @@ In `Declaration.parser`:
 
 Constructor compatibility decision:
 
-1. Preserve existing constructor ergonomics by treating `Foo()` as constructor value syntax (not `Foo(())`) when callee is syntactically a constructor.
-2. This keeps behavior aligned with existing language guide guidance about bare constructor values and zero-arg constructors.
+1. Keep constructor behavior unchanged: zero-arg constructors are written as constants (`True`, `False`) without `()`.
+2. `Foo()` remains unsupported in this design; this syntax is only for Unit-arg function calls.
 
 ### 3) SourceConverter lowering for empty-paren calls
 
@@ -119,6 +119,7 @@ In `SourceConverter.fromDecl`:
 1. Add handling for `ApplyKind.Parens0`.
 2. Convert the callee normally.
 3. Build application with a single Unit value argument (`Predef.Unit`).
+4. Constructors are unaffected because `Foo()` is not accepted by the parser.
 
 All downstream compiler phases continue to operate on the same non-empty-arg internal forms.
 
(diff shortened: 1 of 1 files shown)


docs/design/1923-syntax-for-unit-a-functions.md
```diff
@@ -0,0 +1,203 @@
...
+
+1. Change direct call suffix parsing from one-or-more args to zero-or-more args.
+2. Add an explicit `ApplyKind` marker for empty paren calls (for example `Parens0`).
+3. Represent `f()` using that marker while still constructing an `Apply` node with non-empty internal args.
+
+`Declaration.toDoc` renders the new call kind as `f()`.
+
+Constructor compatibility decision:
+
+1. Preserve existing constructor ergonomics by treating `Foo()` as constructor value syntax (not `Foo(())`) when callee is syntactically a constructor.
```
johnynek: This shouldn't change constructors at all. Zero arg constructors, like True and False, must be written without ().

docs/design/1923-syntax-for-unit-a-functions.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Parser tests
+
+1. `main = f()` parses.
+2. `main = f()()` parses with correct associativity.
+3. `def usage(): 1` parses at top-level.
+4. Nested zero-arg defs parse.
+5. Round-trip rendering uses `def f():` and `f()` forms.
+6. Existing `() -> 1` lambda parsing remains green.
+7. Constructor regressions are covered (`Foo` and `Foo()` behavior per decision).
```
johnynek: Again: no change to constructors. Foo() would not be supported.

docs/design/1923-syntax-for-unit-a-functions.md
```diff
@@ -0,0 +1,203 @@
...
+10. Language guide includes the new syntax and equivalence notes.
+
+## Risks and mitigations
+
+1. Risk: parser precedence regressions after allowing empty direct-call suffixes.
+   Mitigation: add targeted tests for chaining, operators, dot calls, and parentheses.
+2. Risk: confusion between explicit `f(())` and sugar `f()`.
+   Mitigation: document equivalence and keep canonical pretty-printer behavior.
+3. Risk: constructor-call ambiguity (`Foo()`).
+   Mitigation: explicit parser decision plus constructor regression tests.
```
johnynek: Zero see constructors wouldn't be supported. We want those to be like constants (True and False) so we don't want to cloud the design space.

Zero arg (unit arg) functions have a use:delaying calling. I don't see any use to zero arg constructors.

docs/design/1923-syntax-for-unit-a-functions.md
johnynek: Updated the design doc to make constructor behavior explicit: zero-arg constructors remain constant-style (`True`, `False`) and are not called with `()`. The constructor-compatibility section now states `Foo()` is unsupported and this syntax change is only for Unit-arg function calls.

docs/design/1923-syntax-for-unit-a-functions.md
johnynek: Adjusted the implementation plan and acceptance criteria to match this: constructor syntax is unchanged and `Foo()` continues to be rejected. The design now limits no-arg call sugar to Unit-arg functions only.

docs/design/1923-syntax-for-unit-a-functions.md
johnynek: Agreed and updated accordingly. Testing/risk sections now assert constructor regressions should verify `Foo()` is rejected, and the design text now frames the feature strictly around delayed Unit-arg function calls, not constructors.

johnynek:
Addressed all three review comments by editing `docs/design/1923-syntax-for-unit-a-functions.md` to remove constructor syntax changes and explicitly keep `Foo()` unsupported. Also ran `scripts/test_basic.sh` locally; it passed.

Issue #1925, opened 2026-03-02
Lazy type


johnynek:
Design an external struct Lazy[a: +*]

It has two functions:

def lazy[a](fn: () -> a) -> Lazy[a]
def get_Lazy[a](l: Lazy[a]) -> a

Internally Lazy is a mutable cell that is evaluated once (we don't have to guarantee at most once but if we can easily do so with atomic operations we should).

So multiple calls to get_Lazy would reuse the same computed value.

This can be in Bosatsu/Lazy package. It would rarely be used.

johnynek:
MergeXO assigned an agent and started design work for issue #1925.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/1926

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #1925.

johnynek:
Opened implementation PR: https://github.com/johnynek/bosatsu/pull/1930

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/1925#issuecomment-3985998380
- linked PR: #1930 (https://github.com/johnynek/bosatsu/pull/1930)

Comments on the source issue are no longer actioned after a PR exists. Please comment on the PR thread instead.

PR #1926, opened 2026-03-02
Design doc for #1925: Issue #1925 design: Lazy type

johnynek:
Design doc.

Refs #1925

### DIFF
--- a/docs/design/1925-lazy-type.md
+++ b/docs/design/1925-lazy-type.md
@@ -120,27 +120,34 @@ external def get_Lazy[a](l: Lazy[a]) -> a
   - `lazy` -> `FfiCall.Fn1(...)`
   - `get_Lazy` -> `FfiCall.Fn1(...)`
 - In `PredefImpl`, add a runtime cell representation, for example:
-  - `LazyCell(thunk: Value, cached: AtomicReference[Value | Null])`
+  - `LazyCell(state: AtomicReference[Either[Value, Value]])`
+  - `Left(thunk)` means not yet forced, `Right(value)` means memoized.
 - Implement:
   - `lazy_Lazy(fn: Value): Value` -> wraps cell in `ExternalValue`
-  - `get_Lazy(cellValue: Value): Value` -> fast-path cached value, else evaluate thunk with `UnitValue`, then CAS-store winner
+  - `get_Lazy(cellValue: Value): Value` -> read state:
+    - if `Right(value)`, return immediately
+    - if `Left(thunk)`, run `thunk(())`, then CAS `Left(thunk)` -> `Right(result)`; return winner's memoized value
+- Ensure the thunk reference is released after successful force by transitioning to `Right(result)` so the cell no longer holds the thunk.
 - Keep this implementation in shared `core` code so JVM and Scala.js evaluator behavior stays aligned.
 
 ### 3) Python transpile runtime
 
 - Add `Bosatsu/Lazy` mapping in `test_workspace/Prog.bosatsu_externals`.
 - Add runtime support in `test_workspace/ProgExt.py`:
-  - `_BosatsuLazy` cell with thunk + cached sentinel
+  - `_BosatsuLazy` cell with explicit forced-state + cached value
   - `lazy(fn)` constructor
   - `get_Lazy(l)` force-and-cache path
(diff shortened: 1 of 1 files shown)


docs/design/1925-lazy-type.md
```diff
@@ -0,0 +1,183 @@
...
+  - `_BosatsuLazy` cell with thunk + cached sentinel
+  - `lazy(fn)` constructor
+  - `get_Lazy(l)` force-and-cache path
+
+### 4) C runtime support
+
+- Add `c_runtime/bosatsu_ext_Bosatsu_l_Lazy.h` declarations.
+- Add `c_runtime/bosatsu_ext_Bosatsu_l_Lazy.c` implementation with external struct cell:
+  - stored thunk `BValue`
+  - `_Atomic BValue` cached value sentinel (`BSTS_BVALUE_NULL` when empty)
```
johnynek: Null is 0. But for a Float64 the value zero is an IEEE754 double which is encoded as all zeros. This would result in recomputing that value, no?

I think we need an atomic bool and after we compute we set it to true. This could result in some recomputation but it avoids locking.

It would be good to release the reference to the thunk after calling to allow for GC.

docs/design/1925-lazy-type.md
```diff
@@ -0,0 +1,183 @@
...
+- Add `Bosatsu/Lazy` to `test_workspace/core_alpha_conf.json` exported packages.
+
+### 2) Evaluator runtime (`core/src/main/scala/dev/bosatsu/Predef.scala`)
+
+- Add package constant for `Bosatsu/Lazy`.
+- Register two externals in `Predef.evalExternals`:
+  - `lazy` -> `FfiCall.Fn1(...)`
+  - `get_Lazy` -> `FfiCall.Fn1(...)`
+- In `PredefImpl`, add a runtime cell representation, for example:
+  - `LazyCell(thunk: Value, cached: AtomicReference[Value | Null])`
```
johnynek: I think AtomicReference[Either[Value, Value]] is all we need. We either have the thunk or the result. We should release the thunk after we run it.

docs/design/1925-lazy-type.md
```diff
@@ -0,0 +1,183 @@
...
+- Implement:
+  - `lazy_Lazy(fn: Value): Value` -> wraps cell in `ExternalValue`
+  - `get_Lazy(cellValue: Value): Value` -> fast-path cached value, else evaluate thunk with `UnitValue`, then CAS-store winner
+- Keep this implementation in shared `core` code so JVM and Scala.js evaluator behavior stays aligned.
+
+### 3) Python transpile runtime
+
+- Add `Bosatsu/Lazy` mapping in `test_workspace/Prog.bosatsu_externals`.
+- Add runtime support in `test_workspace/ProgExt.py`:
+  - `_BosatsuLazy` cell with thunk + cached sentinel
```
johnynek: Make sure to release the thunk when we run it.

docs/design/1925-lazy-type.md
johnynek: Updated the design doc to use `AtomicReference[Either[Value, Value]]` in the evaluator (`Left(thunk)` / `Right(result)`), and clarified that forcing transitions the cell to `Right(result)` so the thunk reference is released.

docs/design/1925-lazy-type.md
johnynek: Added explicit thunk-release notes in all runtime sections: evaluator (`Right(result)` state only), Python cell clearing the thunk after first success, and C cell clearing thunk after memoization.

docs/design/1925-lazy-type.md
johnynek: Adjusted the C-runtime design to avoid `BValue` null sentinel caching. The doc now specifies a cached `BValue` plus `_Atomic _Bool` forced flag, so `Float64(0.0)`/zero-bit payloads are safe, with thunk cleared after first successful force.

johnynek:
Addressed all three review comments by updating `docs/design/1925-lazy-type.md` with the revised lazy-cell state model and explicit thunk-release behavior across evaluator/Python/C sections. Ran required pre-push check: `scripts/test_basic.sh` (pass).

PR #1928, opened 2026-03-02
Implement #1923: zero-arg def/call syntax as Unit sugar

### LINKED ISSUE #1923: Syntax for Unit -> a functions

johnynek:
Implemented the accepted design for Unit-function sugar:
- Added zero-arg def parsing (`def f(): ...`) by extending `DefStatement.parser` with optional empty-arg-group rewriting to `Pattern.tuple(Nil)` where enabled (`Declaration.defP`, `Statement` def parser).
- Added direct no-arg call syntax (`f()`) with new `Declaration.ApplyKind.Parens0`, parsing via zero-or-more direct-call args, and pretty-print support for `f()`.
- Kept constructor behavior unchanged: `Foo()` remains rejected (with explicit parser error); constructors still use `Foo` / `Foo(...)` / `Foo { ... }` forms.
- Added lowering in `SourceConverter` for `ApplyKind.Parens0` to explicit Predef Unit application while preserving any additional appended args.
- Updated `DefStatement` rendering so singleton Unit-pattern arg groups print as `()`.
- Removed the obsolete ParserHints rule claiming zero-arg defs are unsupported, and updated hints tests accordingly.
- Updated language docs with equivalences: `def f(): body == def f(()): body`, `f() == f(())`, and `() -> a` remains the Unit-arg function type.

Tests/docs updated in scope:
- `ParserTest` / `SyntaxParseTest`: new coverage for `x()`, `x()()`, zero-arg defs at top-level and nested, round-tripping, and constructor `Foo()` rejection.
- `SourceConverterTest`: desugaring equivalence checks for `f()` and zero-arg defs.
- `ParserHintsTest`: removed old unsupported-syntax expectation.

Validation run:
- `sbt "coreJVM/testOnly dev.bosatsu.ParserTest" "coreJVM/testOnly dev.bosatsu.SyntaxParseTest" "coreJVM/testOnly dev.bosatsu.SourceConverterTest" "coreJVM/testOnly dev.bosatsu.ParserHintsTest"` (pass)
- `scripts/test_basic.sh` (pass; reported `Passed: Total 61, Failed 0, Errors 0, Passed 61`)

Fixes #1923

Implements design doc: [docs/design/1923-syntax-for-unit-a-functions.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/1923-syntax-for-unit-a-functions.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/1924

### DIFF
--- a/test_workspace/Bosatsu/Collection/Array.bosatsu
+++ b/test_workspace/Bosatsu/Collection/Array.bosatsu
@@ -47,8 +47,8 @@ external def tabulate_Array[a](n: Int, fn: Int -> a) -> Array[a]
 external def from_List_Array[a](xs: List[a]) -> Array[a]
 external def to_List_Array[a](ary: Array[a]) -> List[a]
 external def size_Array[a](ary: Array[a]) -> Int
-external def get_map_Array[a, b](ary: Array[a], idx: Int, default: Unit -> b, fn: a -> b) -> b
-external def get_or_Array[a](ary: Array[a], idx: Int, default: Unit -> a) -> a
+external def get_map_Array[a, b](ary: Array[a], idx: Int, default: () -> b, fn: a -> b) -> b
+external def get_or_Array[a](ary: Array[a], idx: Int, default: () -> a) -> a
 external def foldl_Array[a, b](ary: Array[a], init: b, fn: (b, a) -> b) -> b
 external def map_Array[a, b](ary: Array[a], fn: a -> b) -> Array[b]
 external def set_or_self_Array[a](ary: Array[a], idx: Int, value: a) -> Array[a]
@@ -75,7 +75,7 @@ def int_Array_to_String(code_points: Array[Int]) -> Option[String]:
   int_List_to_String(to_List_Array(code_points))
 
 def get_Array[a](ary: Array[a], idx: Int) -> Option[a]:
-  get_map_Array(ary, idx, _ -> None, x -> Some(x))
+  get_map_Array(ary, idx, () -> None, x -> Some(x))
 
 def index_in_range_Array[a](ary: Array[a], idx: Int) -> Bool:
   if cmp_Int(idx, 0) matches LT:
@@ -127,12 +127,12 @@ a_sorted = sort_Array(
   cmp_pair
 )
 
-def set_in_range_ok(_: Unit) -> Bool:
+def set_in_range_ok() -> Bool:
   match set_Array(a5, 1, 8):
     case Some(arr): to_List_Array(arr) matches [0, 8, 2, 3, 4]
     case None: False
 
-def int_to_char_array_some(_: Unit) -> Bool:
+def int_to_char_array_some() -> Bool:
   match int_Array_to_Char_Array(from_List_Array([65, 128075])):
     case Some(chars): to_List_Array(chars) matches [.'A', .'👋']
     case None: False
@@ -146,13 +146,13 @@ tests = TestSuite("Array tests", [
   Assertion(get_Array(a5, 4).eq_oi(Some(4)), "get last"),
   Assertion(get_Array(a5, -1).eq_oi(None), "get negative"),
   Assertion(get_Array(a5, 5).eq_oi(None), "get out of range"),
-  Assertion(get_or_Array(a5, 2, _ -> 100) matches 2, "get_or in range"),
-  Assertion(get_or_Array(a5, 10, _ -> 100) matches 100, "get_or fallback"),
+  Assertion(get_or_Array(a5, 2, () -> 100) matches 2, "get_or in range"),
+  Assertion(get_or_Array(a5, 10, () -> 100) matches 100, "get_or fallback"),
   Assertion(foldl_Array(a5, 0, add) matches 10, "foldl add"),
   Assertion(to_List_Array(map_Array(a5, i -> i.add(1))) matches [1, 2, 3, 4, 5], "map add 1"),
   Assertion(to_List_Array(a_set) matches [0, 1, 99, 3, 4], "set_or_self in range"),
   Assertion(eq_ai(set_or_self_Array(a5, 9, 77), a5), "set_or_self out of range keeps self"),
-  Assertion(set_in_range_ok(()), "set_Array in range"),
+  Assertion(set_in_range_ok(), "set_Array in range"),
   Assertion(set_Array(a5, 10, 8) matches None, "set_Array out of range"),
   Assertion(to_List_Array(concat_all_Array([a5, range_Array(3), empty_Array])) matches [0, 1, 2, 3, 4, 0, 1, 2], "concat_all"),
   Assertion(to_List_Array(slice_Array(a5, 1, 4)) matches [1, 2, 3], "slice mid"),
@@ -166,7 +166,7 @@ tests = TestSuite("Array tests", [
   Assertion(to_List_Array(string_to_Char_Array("hi👋")) matches [.'h', .'i', .'👋'], "string_to_Char_Array"),
   Assertion(to_List_Array(char_Array_to_Int_Array(from_List_Array([.'A', .'👋']))) matches [65, 128075], "char_Array_to_Int_Array"),
   Assertion(to_List_Array(string_to_Int_Array("A👋")) matches [65, 128075], "string_to_Int_Array"),
-  Assertion(int_to_char_array_some(()), "int_Array_to_Char_Array some"),
+  Assertion(int_to_char_array_some(), "int_Array_to_Char_Array some"),
   Assertion(
     int_Array_to_Char_Array(from_List_Array([55296])) matches None,
     "int_Array_to_Char_Array none"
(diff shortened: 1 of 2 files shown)


core/src/main/scala/dev/bosatsu/Declaration.scala
```diff
@@ -1350,12 +1357,19 @@ object Declaration {
...
-      .map {
-        case (r, args) => { (nm: Var) =>
-          Apply(nm, args, ApplyKind.Parens)(using nm.region + r)
-        }
+    // here is tuple style: Foo(a, b), but explicitly reject Foo()
+    val tupArgs = declP.parensLines0Cut.region
+      .flatMap {
+        case (_, Nil) =>
+          P.failWith(
+            "constructor call syntax does not support empty parentheses; use the constructor value (`Foo`) or record syntax (`Foo {}`)"
```
johnynek: let's not put this error in the parser. Let's put it in SourceConverter because parse errors fail on the first error, where-as source-converter can process multiple errors and give the user more actionable feedback at once.

core/src/main/scala/dev/bosatsu/DefStatement.scala
johnynek: shouldn't this now be `NonEmptyList[List[A]]` since we have 1 or more groups of 0 or more args where 0 args is interpretted by the SourceConverter as `Unit`?

core/src/main/scala/dev/bosatsu/DefStatement.scala
```diff
@@ -40,11 +40,19 @@ object DefStatement {
         Doc.intercalate(
           Doc.empty,
           args.toList.map { args =>
+            val argDocs = args.map(Document[A].document(_)).toList
+            val inner =
```
johnynek, line 44: I think we shouldn't have this. We should have the args be `List` not a `NonEmptyList` and then this will naturally be addressed without this hack.

core/src/main/scala/dev/bosatsu/DefStatement.scala
```diff
@@ -59,9 +67,18 @@ object DefStatement {
     */
   def parser[A, B](
       argParser: P[A],
-      resultTParser: P[B]
+      resultTParser: P[B],
+      emptyArg: Option[A] = None
```
johnynek: I think we remove this if we set the args to `NonEmptyList[List[A]]`. Let's do that if we can.

core/src/main/scala/dev/bosatsu/SourceConverter.scala
```diff
@@ -278,6 +278,16 @@ final class SourceConverter(
       case Annotation(term, tpe) =>
         (loop(term), toType(tpe, decl.region))
           .parMapN(Expr.Annotation(_, _, decl))
+      case Apply(fn, args, ApplyKind.Parens0) =>
```
johnynek, line 287: this is only valid with `Apply(fn, Nil, ApplyKind.Parens0)` isn't it? Can we add some internal error result if args is not Nil?

core/src/main/scala/dev/bosatsu/Statement.scala
```diff
@@ -251,7 +251,11 @@ object Statement {
     val defBody = maybeSpace.with1 *> OptIndent.indy(Declaration.parser).run("")
     val defP: P[Statement] =
       DefStatement
-        .parser(Pattern.bindParser, defBody <* toEOL)
+        .parser(
+          Pattern.bindParser,
+          defBody <* toEOL,
+          emptyArg = Some(Pattern.tuple(Nil))
```
johnynek: I think we want to change this, see comments in DefStatement.

core/src/test/scala/dev/bosatsu/ParserHintsTest.scala
```diff
@@ -162,30 +162,17 @@ class ParserHintsTest extends munit.FunSuite {
     )
   }
 
-  test("zero-arg def gets actionable hint") {
+  test("zero-arg defs no longer trigger unsupported-syntax hint") {
```
johnynek: don't say "no longer" just say `zero-arg defs are supported`

core/src/test/scala/dev/bosatsu/ParserTest.scala
```diff
@@ -1159,6 +1175,9 @@ x"""
 
     parseTestAll(parser(""), expected.toDoc.render(80), expected)
 
+    assert(parser("").parseAll("Foo").isRight)
+    assert(parser("").parseAll("Foo()").isLeft)
```
johnynek: let's not handle this error at the parser, but in SourceConverter

core/src/main/scala/dev/bosatsu/Declaration.scala
johnynek: Done. I removed the `Foo()` rejection from parser-time and moved it to source conversion. `Declaration.recordConstructorP` now parses empty constructor parens as `ApplyKind.Parens0`, and `SourceConverter` now emits `ConstructorEmptyParens` for that case. I also added a source-converter regression test for this.

core/src/main/scala/dev/bosatsu/DefStatement.scala
johnynek: Done. `DefStatement.args` is now `NonEmptyList[List[A]]` so we model one-or-more groups of zero-or-more args directly.

core/src/main/scala/dev/bosatsu/DefStatement.scala
johnynek: Agreed and updated. I removed the parser-time empty-arg rewrite hook and switched `DefStatement.parser` to parse each arg group as `List[A]` (via `parensLines0Cut`). Empty groups are now handled downstream in `SourceConverter`.

core/src/main/scala/dev/bosatsu/DefStatement.scala
johnynek: Done. With `NonEmptyList[List[A]]` in place, I removed the prior special-case rendering hack in `DefStatement.document`; empty groups now naturally render as `()`.

core/src/main/scala/dev/bosatsu/SourceConverter.scala
johnynek: Added. In `SourceConverter`, `ApplyKind.Parens0` now performs an internal shape check and emits `InvalidParens0Apply` if the payload is not the expected synthetic unit placeholder.

core/src/main/scala/dev/bosatsu/Statement.scala
johnynek: Updated. `Statement` now uses the new `DefStatement.parser` signature and no longer passes parser-time unit injection.

core/src/test/scala/dev/bosatsu/ParserHintsTest.scala
johnynek: Updated test wording to exactly `zero-arg defs are supported`.

core/src/test/scala/dev/bosatsu/ParserTest.scala
johnynek: Done. I removed the parser-level rejection expectation in `ParserTest` (`Foo()` now parses), and added `SourceConverterTest` coverage asserting `Foo()` is rejected during source conversion with `ConstructorEmptyParens`.

johnynek:
Implemented all requested review feedback and propagated the `DefStatement.args` type change through affected compile/test helpers. Verification run: `sbt "coreJVM/test:compile"`, focused suites (`ParserTest`, `SourceConverterTest`, `ParserHintsTest`), and required pre-push gate `scripts/test_basic.sh` all pass.

core/src/test/scala/dev/bosatsu/SourceConverterTest.scala
```diff
@@ -355,6 +355,48 @@ main = match True:
...
+      """def later():
+  1
+main = later()""",
+      """def later(()):
+  1
+main = later(())"""
+    )
+  }
+
+  test("constructor empty-parens syntax is rejected in source conversion") {
```
johnynek, line 380: we need a test that rejects:

```
struct Foo

main = Foo()
```
at source conversion.

johnynek: Added this exact regression at source-conversion level by updating the existing constructor-empty-parens test to use:

struct Foo
main = Foo()

in `SourceConverterTest`, and it asserts `SourceConverter.ConstructorEmptyParens("Foo", ...)` is reported.

johnynek:
Updated `core/src/test/scala/dev/bosatsu/SourceConverterTest.scala` to cover the zero-arg constructor case requested in review (`struct Foo` + `Foo()`), and verified locally with `sbt "coreJVM/testOnly dev.bosatsu.SourceConverterTest"` plus required pre-push `scripts/test_basic.sh` (pass).

johnynek:
read through the `test_workspace` and look for any examples of code that could use this syntax. We want to exercise it, if we can, in the test_workspace to the fullest degree to give users examples.

johnynek:
Addressed the PR issue comment by reviewing `test_workspace` and updating concrete examples to exercise the new Unit-sugar syntax in real code. Changes made:
- `test_workspace/Bosatsu/Collection/Array.bosatsu`
  - Converted `default: Unit -> ...` to `default: () -> ...` in external defs.
  - Converted `def set_in_range_ok(_: Unit)` and `def int_to_char_array_some(_: Unit)` to zero-arg defs (`def ...()`).
  - Converted calls from `f(())` to `f()`.
  - Converted Unit-ignored lambdas (`_ -> ...`) used for Unit defaults to `() -> ...`.
- `test_workspace/Bosatsu/IO/Bytes.bosatsu`
  - Converted `default: Unit -> ...` to `default: () -> ...`.
  - Converted Unit-default lambdas from `_ -> ...` to `() -> ...`.

Verification:
- Ran required pre-push command `scripts/test_basic.sh` successfully (pass).

PR #1929, opened 2026-03-02
Make TypedExprRecursionCheck stack-safe for moderately large list literals

### LINKED ISSUE #1927: lib check stack overflows on moderately large list literal
## Summary
`bosatsu lib check` crashes with `java.lang.StackOverflowError` on a module containing a moderately large list literal, instead of returning a bounded diagnostic.

## Version
- `bosatsu --help`:
  - version: `0.0.38`
  - git-sha: `07728bc93e98e9b5f40a23764a0c042033450e89`

## Minimal reproducer
From the root of any Bosatsu library project where `./bosatsu lib check` normally works:

```bash
mkdir -p src/MyLib/QA

N=211
{
  echo 'package MyLib/QA/ReproListOverflow'
  echo
  echo -n 'vals: List[String] = ['
  seq 1 "$N" | awk '{printf("\"x\"%s", NR=='"$N"' ? "" : ", ")}'
  echo ']'
  echo
  echo 'count = vals.foldl_List(0, (acc, _) -> acc.add(1))'
  echo
  echo 'test = Assertion(count.eq_Int(211), "repro")'
} > src/MyLib/QA/ReproListOverflow.bosatsu

./bosatsu lib check
```

### Result
Compiler crashes with `java.lang.StackOverflowError` (head of trace):

```text
java.lang.StackOverflowError
  at ... StackOverflowCheckImpl...
  at dev.bosatsu.TypedExprRecursionCheck$Impl$.checkExpr(...)
  ...
```

### Expected
- No hard JVM stack overflow.
- Either compile successfully, or return a normal user-facing diagnostic.

## Threshold on my machine
- `N=210`: passes.
- `N=211`: crashes with the stack overflow above.

## Related
- Same pattern is also reproducible under `bosatsu tool check`.

johnynek:
Implemented a stack-safety fix in `TypedExprRecursionCheck` by replacing `StateT[Either, State, A]` with a custom `Eval`-backed state monad (`St`) that trampolines `map`/`flatMap`/`tailRecM` and keeps the existing sequential-state semantics. Reimplemented the checker’s `Parallel` product on top of the new `St` so error-merging behavior on parallel checks is preserved.

Added a regression test in `TypedExprRecursionCheckTest` for issue #1927: it builds a 211-element list literal and re-runs `TypedExprRecursionCheck.checkLets` on a small-stack thread (96 KiB). Before the fix this reproduces a `StackOverflowError`; after the fix it completes successfully.

Validation run:
- `sbt "coreJVM/testOnly dev.bosatsu.TypedExprRecursionCheckTest"` (passes)
- `scripts/test_basic.sh` (passes: 61 CLI tests, 1517 core tests)

Fixes #1927

### DIFF
--- a/core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
+++ b/core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
@@ -3,6 +3,7 @@ package dev.bosatsu
 import cats.Show
 import cats.data.{NonEmptyList, Validated}
 import IorMethods.IorExtension
+import scala.util.Try
 
 class TypedExprRecursionCheckTest extends munit.FunSuite with ParTest {
   private val pack = PackageName.parts("TypedRecursionCheck")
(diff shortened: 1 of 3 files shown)


core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
```diff
@@ -231,4 +270,60 @@ def loop(x): x
 main = loop(1)
 """)
   }
+
+  test("moderately large list literals do not overflow recursion checker stack") {
```
johnynek: to make tests like this easier to write, let's add a macro, like:

```scala
inline def onlyJvm(inline expr: => Unit): Unit =
```
But done with scala3 macros and put in `Platform.scala` so on jvm we just put the expr but on scalajs we put ()

Then we can wrap the whole:

```scala
onJvm {
  test("....") {

  }
}
```

and it should run.

Make sure to test that by adding a:

```scala
assert(Try(onJvm(sys.error("boom")).isFailure == Platform.isJvm)
```
somewhere.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
```diff
@@ -231,4 +270,60 @@ def loop(x): x
...
+vals: List[String] = [$items]
+main = vals
+"""
+
+    val (fullTypeEnv, lets, stmts) = typedLetsOf(source)
+    val topLevelDefs = TypedExprRecursionCheck.topLevelDefArgs(stmts)
+    var failure: Option[Throwable] = None
+    var result: Option[TypedExprRecursionCheck.Res[Unit]] = None
+
+    val thread = new Thread(
```
johnynek: I don't think this can run on scalajs, seem my suggestion above.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
johnynek: Implemented this directly in platform code: added `Platform.isJvm`, `Platform.onJvm(inline expr: => Unit)`, and `Platform.onlyJvm(...)` in both `core/.jvm/src/main/scala/dev/bosatsu/Platform.scala` and `core/.js/src/main/scala/dev/bosatsu/Platform.scala`. I also added the requested assertion in `TypedExprRecursionCheckTest`: `assertEquals(Try(Platform.onJvm(sys.error("boom"))).isFailure, Platform.isJvm)`.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
johnynek: Agreed. I wrapped the JVM-only regression in `Platform.onJvm(...)` so it is not registered on Scala.js. I verified this by running `coreJS/testOnly dev.bosatsu.TypedExprRecursionCheckTest` (the JVM-specific test is skipped there) and JVM still runs it.

johnynek:
Addressed both review comments with code changes.

What changed:
- Added JVM-gating helpers to both platform-specific `Platform.scala` files:
  - `isJvm`
  - `onJvm(inline expr: => Unit)`
  - `onlyJvm(inline expr: => Unit)`
- Updated `TypedExprRecursionCheckTest`:
  - Added the requested behavior check using `Try(onJvm(sys.error("boom")))` vs `Platform.isJvm`.
  - Wrapped the JVM-only stack-size regression test in `Platform.onJvm(...)`.

Validation run:
- `sbt "coreJS/testOnly dev.bosatsu.TypedExprRecursionCheckTest"`
- `scripts/test_basic.sh`

PR #1930, opened 2026-03-02
Implement #1925 Lazy type across evaluator, Python externals, and C runtime

### LINKED ISSUE #1925: Lazy type

johnynek:
Implemented `Bosatsu/Lazy` as an external package with `Lazy[a: +*]`, `lazy`, and `get_Lazy` in `test_workspace/Bosatsu/Lazy.bosatsu` (including package tests). Added evaluator wiring in `Predef.evalExternals` and a memoized runtime cell in `PredefImpl` backed by `AtomicReference[Either[Value, Value]]`, with non-strict construction, memoized successful force, and retry-after-failure behavior. Added evaluation coverage in `EvaluationTest` for package-level usage plus direct runtime semantics (non-strictness, memoization, and failure retry). Added Python transpile runtime support (`Prog.bosatsu_externals`, `ProgExt.py`) with `_BosatsuLazy`, `lazy`, and `get_Lazy`, clearing the thunk after first successful force. Added C runtime externals (`c_runtime/bosatsu_ext_Bosatsu_l_Lazy.c` and `.h`) and wired them into `c_runtime/Makefile` build/install targets, including forced-state caching and thunk release. Updated `test_workspace/core_alpha_conf.json` to export `Bosatsu/Lazy` and bumped `next_version` to `4.1.0` to satisfy library versioning rules for added exported API. Validation run: `sbt -batch "coreJVM/test:compile"`, `make -C c_runtime all`, `./test_python.sh`, and required `scripts/test_basic.sh` all pass.

Fixes #1925

Implements design doc: [docs/design/1925-lazy-type.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/1925-lazy-type.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/1926

### DIFF
--- a/c_runtime/Makefile
+++ b/c_runtime/Makefile
@@ -46,7 +46,7 @@ endif
 ALL_CPPFLAGS := $(CPPFLAGS) $(GC_CFLAGS)
 ALL_LIBS := $(GC_LIBS) $(LIBS) -lm
 
-all: bosatsu_runtime.o test_out bosatsu_ext_Bosatsu_l_Predef.o bosatsu_ext_Bosatsu_l_Prog.o bosatsu_ext_Bosatsu_l_IO_l_Core.o bosatsu_ext_Bosatsu_l_IO_l_Bytes.o bosatsu_ext_Bosatsu_l_Collection_l_Array.o bosatsu_ext_Bosatsu_l_Num_l_Float64.o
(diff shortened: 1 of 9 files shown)


johnynek:
Reproduced the canceled CI lane locally using the same commands from the `testJS` job: `sbt -batch "coreJS/testOnly * -- --log=failure; jsapiJS/compile"` and `sbt -batch "jsuiJS/testOnly * -- --log=failure"`; both completed successfully (no failures). Re-ran the required pre-push check `scripts/test_basic.sh`; it passed (`cli/testOnly` and `coreJVM/testOnly` both green). No additional repository changes are needed for this feedback cycle; the GitHub Actions result appears to have been a cancellation/time-budget issue rather than a reproducible test failure.

PR #1938, opened 2026-03-02
Add `filter_Array` and `flat_map_Array` externals to Array across evaluator, Python, and C runtimes

### LINKED ISSUE #1935: add Array filter_Array and Array flat_map_Array to Array package
We should implement the externals for this as efficiently as possible on all backends.

Also, we should double check that for slice_Array is an O(1) operation and if not, fix that on any backend.

We should have 100% test coverage for the added code.

johnynek:
Implemented issue #1935 with focused Array-package/runtime updates:
- Added `filter_Array` and `flat_map_Array` external declarations and exports in `test_workspace/Bosatsu/Collection/Array.bosatsu`.
- Added efficient evaluator implementations in `PredefImpl` and wired both into `evalExternals` in `Predef.scala`.
- Added Python backend external implementations in `PythonGen.scala` with single-pass filtering and one-pass flat-map expansion/copy.
- Added C runtime implementations and header declarations in `c_runtime/bosatsu_ext_Bosatsu_l_Collection_l_Array.c/.h`.
- Updated `flatten_Array` to use `flat_map_Array`.
- Expanded Array tests (including slice-view/offset cases) in both `test_workspace/Bosatsu/Collection/Array.bosatsu` and `core/src/test/scala/dev/bosatsu/EvaluationTest.scala` to cover new behavior.
- Verified `slice_Array` remains O(1) on evaluator/Python/C backends (view-based offset/length wrapping, no element copy).

Validation:
- `make -C c_runtime bosatsu_ext_Bosatsu_l_Collection_l_Array.o` passed.
- `sbt -batch "coreJVM/testOnly dev.bosatsu.EvaluationTest -- --log=failure"` passed.
- `sbt -batch "cli/testOnly dev.bosatsu.PathModuleTest -- --log=failure"` passed.
- Required command `scripts/test_basic.sh` passed.

Fixes #1935

### DIFF
--- a/c_runtime/bosatsu_ext_Bosatsu_l_Collection_l_Array.c
+++ b/c_runtime/bosatsu_ext_Bosatsu_l_Collection_l_Array.c
@@ -1,5 +1,6 @@
 #include "bosatsu_runtime.h"
 
+#include <errno.h>
 #include <gc.h>
 #include <limits.h>
 #include <stdio.h>
@@ -290,8 +291,11 @@ BValue ___bsts_g_Bosatsu_l_Collection_l_Array_l_filter__Array(BValue array, BVal
     return array;
   }
 
-  BValue* data = bsts_array_alloc_data(arr->len);
+  int max_out = arr->len - 1;
+  BValue* data = bsts_array_alloc_data(max_out);
   if (idx > 0) {
+    // Items before idx are already known to pass. Copy directly so we don't
+    // re-run the predicate and change evaluation count/ordering.
     memcpy(data, arr->data + arr->offset, sizeof(BValue) * (size_t)idx);
   }
 
@@ -330,7 +334,9 @@ BValue ___bsts_g_Bosatsu_l_Collection_l_Array_l_flat__map__Array(BValue array, B
     mapped[idx] = mapped_arr;
     total += (long)mapped_arr->len;
     if (total > (long)INT_MAX) {
-      return bsts_array_empty();
+      errno = EOVERFLOW;
+      perror("flat_map_Array total length overflow");
+      abort();
     }
   }
 


c_runtime/bosatsu_ext_Bosatsu_l_Collection_l_Array.c
```diff
@@ -256,6 +269,93 @@ BValue ___bsts_g_Bosatsu_l_Collection_l_Array_l_map__Array(BValue array, BValue
...
+  BSTS_Array** mapped = bsts_array_alloc_parts(arr->len);
+  long total = 0;
+
+  for (int idx = 0; idx < arr->len; idx++) {
+    BValue item = arr->data[arr->offset + idx];
+    BSTS_Array* mapped_arr = bsts_array_unbox(call_fn1(fn, item));
+    mapped[idx] = mapped_arr;
+    total += (long)mapped_arr->len;
+    if (total > (long)INT_MAX) {
+      return bsts_array_empty();
```
johnynek: this should be a `perror + abort` situation, not silently returning the wrong data.

c_runtime/bosatsu_ext_Bosatsu_l_Collection_l_Array.c
```diff
@@ -256,6 +269,93 @@ BValue ___bsts_g_Bosatsu_l_Collection_l_Array_l_map__Array(BValue array, BValue
...
+      break;
+    }
+    idx += 1;
+  }
+
+  if (idx >= arr->len) {
+    return array;
+  }
+
+  BValue* data = bsts_array_alloc_data(arr->len);
```
johnynek: don't we know that this array is len <= `arr->len - idx`? Can't we allocate less and not do the memcpy below? If not comment as to why.

c_runtime/bosatsu_ext_Bosatsu_l_Collection_l_Array.c
johnynek: Fixed. I changed this overflow path in `flat_map_Array` to be fail-fast instead of returning empty data: it now sets `errno = EOVERFLOW`, calls `perror("flat_map_Array total length overflow")`, and then `abort()`.

c_runtime/bosatsu_ext_Bosatsu_l_Collection_l_Array.c
johnynek: Updated. `filter_Array` now allocates `arr->len - 1` (we know at least one element is dropped once we reach this path) instead of `arr->len`. I kept the copy of the prefix and added a comment: those prefix items are already known to pass, so copying them avoids re-running the predicate and changing predicate evaluation count/order.

johnynek:
Addressed both line-level comments in `c_runtime/bosatsu_ext_Bosatsu_l_Collection_l_Array.c` and reran the required pre-push check. `scripts/test_basic.sh` passed.

PR #1946, opened 2026-03-02
Fix Int loop recursion for `add(-1)` and simple `operator -` aliases

### LINKED ISSUE #1945: Int loop recursion rejects equivalent decrements and gives unclear solver diagnostics
## Summary
Using the new Int recursion support (`loop <int>:`) in Bosatsu `0.0.39`, equivalent decrement forms are rejected in recursion obligations, and the diagnostics are often internal/unclear.

`rem.sub(1)` is accepted, but semantically equivalent forms such as `rem.add(-1)` and `rem - 1` can fail with solver/lowering internals.

## Environment
- bosatsu: `0.0.39`

## Repro A (`rem.add(-1)`)

```bosatsu
package Tmp/LoopAddNeg1

def demo(n: Int) -> Int:
  def go(rem: Int, acc: Int) -> Int:
    loop rem:
      case _ if cmp_Int(rem, 0) matches GT:
        go(rem.add(-1), acc.add(rem))
      case _:
        acc

  go(n, 0)

out = demo(5)

test = Assertion(out.eq_Int(15), "loop add neg one")
```

Command:

```bash
./bosatsu tool check --input /tmp/LoopAddNeg1.bosatsu --package_root /tmp --color none
```

Actual error:

```text
cannot prove Int recursion obligation for go: (>= (+ rem_0 (- 1)) 0)
recur target: rem
path condition: (> rem_0 0)
solver detail: Failed executing embedded z3.wasm: Trapped on unreachable instruction
```

## Repro B (infix `rem - 1`)

```bosatsu
package Tmp/LoopInfixMinus

def operator +(a, b): a.add(b)
def operator -(a, b): a.sub(b)

def demo(n: Int) -> Int:
  def go(rem: Int, acc: Int) -> Int:
    loop rem:
      case _ if cmp_Int(rem, 0) matches GT:
        go(rem - 1, acc + rem)
      case _:
        acc

  go(n, 0)

out = demo(5)

test = Assertion(out.eq_Int(15), "loop infix minus")
```

Command:

```bash
./bosatsu tool check --input /tmp/LoopInfixMinus.bosatsu --package_root /tmp --color none
```

Actual error:

```text
cannot prove Int recursion obligation for go: lower recursive Int argument
recur target: rem
path condition: (> rem_0 0)
solver detail: unable to lower recursive argument ... to SMT Int expression
```

## Expected
- Either accept simple equivalent decrement forms (`rem.add(-1)`, `rem - 1`) when they are semantically reducible.
- Or, if canonical forms are required, emit a user-facing hint to rewrite recursive arguments as `rem.sub(1)` and avoid internal solver/lowering messages.

johnynek:
Implemented issue #1945 fixes in recursion checking.

What changed:
- Added a fast-path normalization in Int obligation implication checks so expressions like `x.add(-1)` are recognized as `x - 1` (without needing solver round-trips).
- Added top-level predef-alias discovery for simple forwarding defs (e.g. `def operator -(a, b): a.sub(b)`), and used it during SMT lowering so recursive args like `rem - 1` can lower as Int expressions.
- Improved unsupported-lowering diagnostics with a user-facing rewrite hint (`.sub(1)` / `.add(-1)`) instead of only internal-style detail.
- Threaded alias metadata through recursion-check state so nested defs in the same package can use discovered top-level aliases during loop/recur obligation checks.

Regression tests added/updated:
- `loop Int recursion accepts equivalent add(-1) decrement`
- `loop Int recursion accepts infix decrement via top-level operator alias`
- Existing unsupported-lowering test now also asserts the new hint text.

Repro note:
- The exact CLI repro command from the issue could not be run in this checkout because `./bosatsu` requires a missing `.bosatsu_version` file, so reproduction/verification was done through `TypedExprRecursionCheckTest` with issue-equivalent programs.

Validation run:
- `sbt "coreJVM/testOnly dev.bosatsu.TypedExprRecursionCheckTest"` passed.
- Required gate `scripts/test_basic.sh` passed.

Fixes #1945

### DIFF
--- a/core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
+++ b/core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
@@ -273,11 +273,12 @@ def demo(n: Int) -> Int:
     allowed("""#
 def operator +(a, b): a.add(b)
 def operator -(a, b): a.sub(b)
+def operator >(a, b): cmp_Int(a, b) matches GT
 
 def demo(n: Int) -> Int:
   def go(rem: Int, acc: Int) -> Int:
     loop rem:
-      case _ if cmp_Int(rem, 0) matches GT:
+      case _ if rem > 0:
         go(rem - 1, acc + rem)
       case _:
         acc
@@ -355,7 +356,8 @@ def cmp_guarded(i: Int) -> Int:
   test("Int recursion reports unsupported SMT Int lowering in recursive args") {
     disallowedWithMessage("""#
 def id(i: Int) -> Int:
-  i
+  f = x -> x
+  f(i)
 
 def bad(i: Int) -> Int:
   recur i:
--- a/core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
+++ b/core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
@@ -44,6 +44,7 @@ object TypedExprRecursionCheck {
     // This checker enforces structural recursion plus type-directed Int
     // decrease obligations in typed IR.
     val _ = fullTypeEnv
+    val topLevelLowerableAliases = Impl.topLevelLowerableAliases(pack, lets)
     val topLevelPredefAliases = Impl.topLevelPredefAliases(pack, lets)
     lets.traverse_ { case (name, rec, expr) =>
       Impl.checkTopLevelLet(
@@ -52,6 +53,7 @@ object TypedExprRecursionCheck {
         rec,
         expr,
         topLevelDefs.get(name),
+        topLevelLowerableAliases,
         topLevelPredefAliases
       )
     }
@@ -67,7 +69,13 @@ object TypedExprRecursionCheck {
     case class RecurTargetItem(group: Int, index: Int, paramName: Bindable)
         derives CanEqual
     type RecurTarget = NonEmptyList[RecurTargetItem]
+    type TopLevelLowerableAliases = Map[(PackageName, Bindable), TopLevelAlias]
     type TopLevelPredefAliases = Map[(PackageName, Bindable), Identifier.Name]
+
+    case class TopLevelAlias(
+        params: NonEmptyList[Bindable],
+        body: TypedExpr[Declaration]
+    )
 
     private val comparisonType: Type =
       Type.TyConst(Type.Const.predef("Comparison"))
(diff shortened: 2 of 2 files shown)


core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
```diff
@@ -255,6 +255,37 @@ def ok(i: Int) -> Int:
...
+        acc
+
+  go(n, 0)
+""")
+  }
+
+  test("loop Int recursion accepts infix decrement via top-level operator alias") {
+    allowed("""#
+def operator +(a, b): a.add(b)
+def operator -(a, b): a.sub(b)
```
johnynek, line 275: what about adding:
```
def operator >(a, b): cmp_Int(a, b) matches GT
```
then below have:
```
case _ if rem > 0:
    go(rem - 1, acc + rem)
case _:
    acc
```
that's the code we want to write.

johnynek: seems like we could do this by seeing which bindings we could fully lower into SMT-LIB and then use those as needed.

johnynek: Added this exact coverage in `TypedExprRecursionCheckTest`.

The test now defines:
- `def operator >(a, b): cmp_Int(a, b) matches GT`

and uses:
- `case _ if rem > 0:`
- `go(rem - 1, acc + rem)`

inside the `loop rem:` branch. This now passes as a regression test.

johnynek: Implemented this direction in the recursion-check lowering.

I added top-level alias inlining for non-recursive single-argument-group defs during SMT lowering:
- collect lowerable top-level aliases once per package,
- carry them into recursion-check SMT state,
- when lowering `App(fn, args)`, inline matching local-package alias bodies via `TypedExpr.substituteAll`, then continue lowering.

This is wired into Int/Bool/Comparison lowering paths, so aliases like `operator >` (defined in terms of `cmp_Int`) can contribute guard facts to obligations.

The updated `rem > 0` test passes with this change.

johnynek:
Addressed both review comments with concrete code and test updates:
- test now uses `operator >` guard style (`rem > 0`) with `rem - 1` recursion,
- recursion SMT lowering now inlines non-recursive top-level aliases to recover lowerable forms.

Validation run:
- `sbt "coreJVM/testOnly dev.bosatsu.TypedExprRecursionCheckTest"` passed
- `scripts/test_basic.sh` passed

PR #1954, opened 2026-03-03
Export `max_Char` from `Bosatsu/Char` and add issue #1952 regression

### LINKED ISSUE #1952: In Bosatsu/Char max_char should be exported
It should export the max currently believed to be real unicode codepoint.

`max_Char: Char`

johnynek:
Implemented issue #1952 by updating `test_workspace/Char.bosatsu` to define and export `max_Char` (derived from Unicode scalar max `1114111` via `int_to_Char`). Added internal Char-module assertions to validate the exported value and round-trip behavior. Added a new regression test `core/src/test/scala/dev/bosatsu/Issue1952Test.scala` that imports `max_Char` from `Bosatsu/Char` and verifies codepoint and round-trip correctness from another package (so export visibility is exercised).

Reproduction before fix: `sbt -batch "coreJVM/testOnly dev.bosatsu.Issue1952Test -- --log=failure"` failed with `package: Bosatsu/Char does not have name max_Char`.

Validation after fix:
- `sbt -batch "coreJVM/testOnly dev.bosatsu.Issue1952Test -- --log=failure"` (passes)
- `scripts/test_basic.sh` (passes)

Fixes #1952

### DIFF
--- a/core/src/test/scala/dev/bosatsu/Issue1952Test.scala
+++ b/core/src/test/scala/dev/bosatsu/Issue1952Test.scala
@@ -1,16 +1,9 @@
 package dev.bosatsu
 
-import java.nio.charset.StandardCharsets
-import java.nio.file.Files
-import java.nio.file.Paths
-
 class Issue1952Test extends munit.FunSuite with ParTest {
   import TestUtils._
 
-  private val charPack = new String(
-    Files.readAllBytes(Paths.get("test_workspace/Char.bosatsu")),
-    StandardCharsets.UTF_8
-  )
+  private val charPack = Predef.loadFileInCompile("test_workspace/Char.bosatsu")
   private val reproPack = """
 package Repro/Issue1952
 


johnynek:
let's also add `range_Char(start: Char, end: Char): List[Char]` that returns a list of `Char` where all items are `< end`.

Add a comment to the function.

Also `cmp_Char(a: Char, b: Char) -> Comparison`.

Both should be exported.

johnynek:
Implemented the requested follow-up from https://github.com/johnynek/bosatsu/pull/1954#issuecomment-3988253094.

Changes made:
- Added and exported `cmp_Char(a: Char, b: Char) -> Comparison` in `test_workspace/Char.bosatsu`.
- Added and exported `range_Char(start: Char, end: Char) -> List[Char]` in `test_workspace/Char.bosatsu`.
- Added the requested function comment above `range_Char` describing its half-open interval semantics `[start, end)`.
- Added Bosatsu regression coverage for the new APIs (ordering, empty/reversed ranges, surrogate-gap skipping, and end-exclusive behavior near max codepoint).
- Updated `core/src/test/scala/dev/bosatsu/Issue1952Test.scala` to verify `cmp_Char` and `range_Char` are importable/exported across package boundaries.

Validation:
- Ran `scripts/test_basic.sh` successfully (after a clean rebuild to refresh compile-time embedded test workspace sources).

test_workspace/Char.bosatsu
```diff
@@ -248,6 +258,31 @@ def int_List_to_String(code_points: List[Int]) -> Option[String]:
...
+# Returns chars in the half-open interval [start, end), ordered by code point.
+def range_Char(start: Char, end: Char) -> List[Char]:
+  start_cp = char_to_Int(start)
+  end_cp = char_to_Int(end)
+  width = end_cp.sub(start_cp)
+
+  if cmp_Int(width, 0) matches LT | EQ:
+    []
+  else:
+    range(width).foldl_List([], (rev_chars, offset) ->
```
johnynek: why not just `[int_to_Char(start_cp.add(offset)) for offset in range(width)]`

johnynek: Updated to that approach: `range_Char` now builds `maybe_chars = [int_to_Char(start_cp.add(offset)) for offset in range(width)]` and then folds to keep only `Some(c)` values (so surrogate/non-scalar code points are still skipped). I also added an inline comment explaining this.

johnynek:
Reproduced the previously reported timeout context and re-ran checks. `TypedExprRecursionCheckTest` passes in targeted reruns, and the required command now passes on this branch: `scripts/test_basic.sh` -> CLI 61/61 pass, core 1546/1546 pass.

johnynek:
Reproduced the GitHub Actions `testJS` failure locally with the same command from CI: `sbt -batch "coreJS/testOnly * -- --log=failure; jsapiJS/compile"`. It failed on Scala.js linking because `Issue1952Test` referenced `java.nio.file.*`. Updated `core/src/test/scala/dev/bosatsu/Issue1952Test.scala` to use `Predef.loadFileInCompile("test_workspace/Char.bosatsu")` (Scala.js-safe) instead of runtime file IO. Re-ran the CI-equivalent JS command successfully, then ran the required pre-push check `scripts/test_basic.sh` successfully.

Issue #1957, opened 2026-03-03
Linux native v0.0.39: StackOverflowError in ProtoConverter during lib check


johnynek:
## Summary
`bosatsu lib check` can crash with `java.lang.StackOverflowError` on Linux native (`bosatsu-linux`) in `v0.0.39` while serializing typed expressions (stack repeats through `ProtoConverter.typedExprToProto`), rather than reporting a normal diagnostic.

This appears in Zafu PR #9 after commit `a8bfd22055028495e65f29cba382bae15fd009b4`.

## Environment
- Bosatsu version: `0.0.39`
- Platform mode: `native` (`bosatsu-linux` from release assets)
- Runner OS: `ubuntu-24.04` (GitHub Actions image `20260224.36.1`)
- Date observed: 2026-03-03

## Repro (from failing CI)
Reference failing job:
- https://github.com/johnynek/zafu/actions/runs/22606249467/job/65498802224

Commands run by CI:
```bash
./bosatsu --fetch
scripts/test.sh
```

Inside `scripts/test.sh`, relevant sequence is:
```bash
./bosatsu lib fetch
./bosatsu lib check
./bosatsu lib test
```

The crash happens after `lib fetch` output and before any `lib test` output, so it looks like the failing stage is `lib check`.

## Observed
- Log shows:
  - `fetched blake3:...`
  - `fetched 1 object.`
  - then `java.lang.StackOverflowError`
- Stack repeatedly includes (head excerpt):
  - `dev.bosatsu.ProtoConverter$.run$10(ProtoConverter.scala:135)`
  - `dev.bosatsu.ProtoConverter$.typedExprToProto$$anonfun$1(ProtoConverter.scala:142)`
  - plus repeated `cats.data.IndexedStateT`/`cats.data.AndThen` frames

## Control case
Previous commit in the same PR (`9f0e0fca588ea500d944b29affbe1660ae606d5a`) passed in CI:
- https://github.com/johnynek/zafu/actions/runs/22605727371/job/65497237793

So the overflow was introduced by the delta between those two commits (single file change in `src/Zafu/Collection/Vector.bosatsu`).

## Expected
`bosatsu lib check` should complete or emit a regular compile error, but not crash with `StackOverflowError`.

## Notes
- I could not reproduce on macOS native with the same release (`bosatsu-macos`), which suggests this may be stack-limit sensitive on Linux native.

johnynek:
this is believed to have been fixed in #1963

PR #1960, opened 2026-03-03
tmp: baseline repro for #1957 without Eval fix

johnynek:
Temporary PR to run CI on workflow-fixed branch without the ProtoConverter Eval change. Will close after capturing baseline stack overflow logs.


johnynek:
Closing temporary CI repro PR; replacing with clean ProtoConverter-only PR.

PR #1970, opened 2026-03-03
Fix Int recursion guard fallthrough for irrefutable constructor patterns

### LINKED ISSUE #1966: int recursion check not using all information
conside:
```
def foldl_Vector(vec: Vector[a], init: b, fn: (b, a) -> b) -> b:
  def go(fuel, stack, acc):
    loop (fuel, stack):
        case (_, []): acc
        case (_, _) if cmp_Int(fuel, 0) matches LT | EQ: acc
        case (_, [Leaf(_, items), *tail]):
            next_acc = foldl_Array(items, acc, fn)
            go(fuel.sub(1), tail, next_acc)
        case (_, [Branch(_, _, _, children), *tail]):
            next_stack = push_children_ltr(children, tail)
            go(fuel.sub(1), next_stack, acc)
  go(traversal_fuel(vec), [vec], init)

```
We know if we get past the second case that `cmp_Int(fuel, 0) matches GT` or put another way, we know that `cmp_Int(fuel, 0) matches LT | EQ` is not true. But we don't seem to be using that:

```
in file: /Users/oscar/code/zafu/src/Zafu/Collection/Vector.bosatsu:298:13, package Zafu/Collection/Vector
cannot prove Int recursion obligation for go: (>= (+ fuel_0 (* (- 1) 1)) 0)
recur target: fuel
path condition: true
model:
((define-fun fuel_0 () Int 0))
296|        case (_, [Leaf(_, items), *tail]):
297|            next_acc = foldl_Array(items, acc, fn)
298|            go(fuel.sub(1), tail, next_acc)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


in file: /Users/oscar/code/zafu/src/Zafu/Collection/Vector.bosatsu:301:13, package Zafu/Collection/Vector
cannot prove Int recursion obligation for go: (>= (+ fuel_0 (* (- 1) 1)) 0)
recur target: fuel
path condition: true
model:
((define-fun fuel_0 () Int 0))
299|        case (_, [Branch(_, _, _, children), *tail]):
300|            next_stack = push_children_ltr(children, tail)
301|            go(fuel.sub(1), next_stack, acc)
```

If we put the GT check on each line, it passes:
```
def foldl_Vector(vec: Vector[a], init: b, fn: (b, a) -> b) -> b:
  def go(fuel, stack, acc):
    loop (fuel, stack):
        case (_, [Leaf(_, items), *tail]) if cmp_Int(fuel, 0) matches GT:
            next_acc = foldl_Array(items, acc, fn)
            go(fuel.sub(1), tail, next_acc)
        case (_, [Branch(_, _, _, children), *tail]) if cmp_Int(fuel, 0) matches GT:
            next_stack = push_children_ltr(children, tail)
            go(fuel.sub(1), next_stack, acc)
        case _: acc
  go(traversal_fuel(vec), [vec], init)
```

When a match is total except for a guard, we should add `not(guard)` for all the branches after that. We should be able to type check something like the first example. Note: this has runtime costs potentially: we could have to check conditions more than once.

johnynek:
Implemented #1966 by improving TypedExpr recursion SMT path-fact inference: user-defined single-constructor patterns are now recognized as irrefutable (including when nested inside tuple patterns), so guarded total branches correctly contribute negated guard fallthrough facts to later branches. This fixes false Int recursion obligation failures where recursive calls after a guarded branch were checked with `path condition: true`. Added/updated regression coverage in TypedExprRecursionCheckTest with a vector-style reproduction and a focused failing-before/fixed-after case (`Frame(_)` guarded branch not directly lowerable). Validation run: `sbt "coreJVM/testOnly dev.bosatsu.TypedExprRecursionCheckTest"` and required `scripts/test_basic.sh` (passes: cli/testOnly 62; coreJVM/testOnly 1557, ignored 2).

Fixes #1966

### DIFF
--- a/core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
+++ b/core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
@@ -519,23 +519,6 @@ object TypedExprRecursionCheck {
           }
       }
 
-    private def allowedByTargetFromCompiled(
-        target: RecurTarget,
-        branchPat: Pattern[(PackageName, Identifier.Constructor), Type]
-    ): NonEmptyList[Set[Bindable]] =
-      target.tail match {
-        case Nil =>
-          NonEmptyList.one(branchPat.substructures.toSet)
-        case _ =>
-          branchPat match {
-            case Pattern.PositionalStruct(_, parts)
-                if parts.length == target.length =>
-              NonEmptyList.fromListUnsafe(parts.map(_.substructures.toSet))
-            case _ =>
-              target.map(_ => Set.empty[Bindable])
-          }
-      }
-
     private def targetPatternPartsFromParsed(
         targetLength: Int,
         branchPat: Pattern.Parsed
@@ -2572,7 +2555,7 @@ object TypedExprRecursionCheck {
                       // on all these branches, use the same parent state
                       def beginBranch(
                           matchArg: TypedExpr[Declaration],
-                          sourcePat: Option[Pattern.Parsed],
+                          sourcePat: Pattern.Parsed,
                           compiledPat: Pattern[(PackageName, Identifier.Constructor), Type],
                           fallthroughFact: SmtExpr.BoolExpr
                       ): St[Unit] =
@@ -2587,12 +2570,7 @@ object TypedExprRecursionCheck {
                             )
                           case irr @ InDefRecurred(_, _, _, _, _) =>
                             val allowed =
-                              sourcePat match {
-                                case Some(sp) =>
-                                  allowedByTargetFromParsed(irr.target, sp)
-                                case None     =>
-                                  allowedByTargetFromCompiled(irr.target, compiledPat)
-                              }
+                              allowedByTargetFromParsed(irr.target, sourcePat)
                             val reachable = allowed.iterator.flatMap(_.iterator).toSet
                             val smtState0 =
                               addPatternFactsAndBindings(
@@ -2603,12 +2581,7 @@ object TypedExprRecursionCheck {
                             val smtState1 =
                               addPathFactIfNonTrivial(fallthroughFact, smtState0)
                             val smtState =
-                              sourcePat match {
-                                case Some(sp) =>
-                                  bindSourcePatternAliases(irr, sp, smtState1)
-                                case None     =>
-                                  smtState1
-                              }
+                              bindSourcePatternAliases(irr, sourcePat, smtState1)
                             setSt(
                               InRecurBranch(
                                 irr,
@@ -2639,7 +2612,7 @@ object TypedExprRecursionCheck {
                             _ <- checkForIllegalBindsSt(branch.pattern.names, tag.region)
                             _ <- beginBranch(
                               arg,
-                              Some(sourcePat),
+                              sourcePat,
                               branch.pattern,
                               fallthroughFact
                             )


core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -1356,6 +1388,49 @@ object TypedExprRecursionCheck {
...
+    private def constructorIsIrrefutable(
+        cons: (PackageName, Identifier.Constructor),
+        argType: Type,
+        state: SmtBranchState
+    ): Boolean =
+      state.singleConstructorOwners.get(cons).exists { ownerType =>
+        Type.rootConst(argType).exists(_.tpe == ownerType)
+      }
+
+    private def patternIsIrrefutableForType(
```
johnynek: let's not re-implement this. TotalityCheck already is home to this logic.

I think if we make `val tc = TotalityCheck(fullTypeEnv)` at the start, then here it's as simple as `tc.missingBranches(pattern :: Nil).isEmpty`

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -1558,6 +1636,59 @@ object TypedExprRecursionCheck {
...
+      val state0 = state.removeBindings(branch.pattern.names)
+      val state1 = bindPatternNames(argExpr, branch.pattern, state0)
+      lowerMatchBranchCondition(argExpr, branch, state1)
+    }
+
+    private def matchFallthroughFacts(
+        argExpr: TypedExpr[Declaration],
+        branches: NonEmptyList[TypedExpr.Branch[Declaration]],
+        initialState: SmtBranchState
+    ): Vector[SmtExpr.BoolExpr] = {
```
johnynek: let's return NonEmptyList[SmtExpr.BoolExpr] and add a comment that this is returning an list in the same order and size as branches with the BoolExpr for what we know given that the previous branches didn't match.

Actually, we seem to always zip the result back to branches. Why not encode that invariant here and just return:

```scala
NonEmptyList[(TypedExpr.Branch[Declaration], SmtExpr.Bool)]`
```
So we can do the zipping as we build it up. That way we encode in the types the connection between the branch and the bool.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -1558,6 +1636,59 @@ object TypedExprRecursionCheck {
...
+      branches.toList.foldLeft(init) { case ((acc, priorMiss, state), branch) =>
+        val state0 = resetBranchBindings(state)
+        val (hitOpt, state1) = lowerBranchHitCondition(argExpr, branch, state0)
+        val nextMiss = hitOpt match {
+          case Some(hit) =>
+            simplifyBoolExpr(mkAnd(Vector(priorMiss, SmtExpr.Not(hit))))
+          case None      =>
+            priorMiss
+        }
+        (acc :+ priorMiss, nextMiss, state1)
```
johnynek: list prepend is much faster the vector append. Use a List here and then reverse and use `NonEmptyList.fromListUnsafe` to return.

We know it is NonEmpty because branches is NonEmpty and this is the same size.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: can we rename `all` to:

```scala
val (bools, state1, ok) = ..

// below
val bools1 = if (ok) Some(mkOr(bools.reverse.toVector)) else None

(bools1, state1)
```

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -2427,32 +2559,51 @@ object TypedExprRecursionCheck {
...
-                      checkExpr(currentPackage, branch.expr, wrappers)
-                  }
-                val withPatternContext =
-                  withTemporaryRecurBranchSmtState(
-                    branchExprCheck,
-                    _ => branchExprCheck
-                  ) { smtState =>
-                    addPatternFactsAndBindings(arg, branch.pattern, smtState)
+              val optRes = getSt.flatMap { state =>
+                val fallthroughFacts =
```
johnynek, line 2495: make this a `NonEmptyList` and comment it is the same length as the branches.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -2427,32 +2559,51 @@ object TypedExprRecursionCheck {
...
+                  state match {
+                    case InRecurBranch(_, _, _, _, smtState) =>
+                      matchFallthroughFacts(arg, branches, smtState).toList
+                    case _                                  =>
+                      List.fill(branches.length)(SmtExpr.BoolConst.True)
                   }
-                checkForIllegalBindsSt(branch.pattern.names, tag.region) *>
-                  filterNames(branch.pattern.names)(withPatternContext)
+
+                branches.toList.zip(fallthroughFacts).parTraverse_ {
```
johnynek: don't convert back to `toList` just `zip` the two nonemptylists and `.parTraverse_`

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Implemented. I removed the custom irrefutability implementation and now use `TotalityCheck` directly. At the start of `checkLets` we now build `val totalityCheck = TotalityCheck(fullTypeEnv)`, thread it through state, and `patternIsIrrefutable` is now just `state.totalityCheck.missingBranches(pattern :: Nil).isEmpty`.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Implemented. `matchFallthroughFacts` now encodes the branch/fact invariant in the type and returns `NonEmptyList[(TypedExpr.Branch[Declaration], SmtExpr.BoolExpr)]`. I also added a comment documenting that the result has the same order and size as `branches`.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Implemented. I switched the accumulator from vector append to list prepend in `matchFallthroughFacts`, then reverse once at the end and return with `NonEmptyList.fromListUnsafe`.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Implemented. In the `Pattern.Union` lowering branch, I renamed the tuple binding to `val (bools, state1, ok)` and then compute `bools1` before returning `(bools1, state1)`.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Implemented. The recur-path `fallthroughFacts` is now a `NonEmptyList` (not a list/vector), and I added an inline comment at the binding noting it has the same order/length as `branches`.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Implemented. I removed the `toList` conversion path and now use non-empty traversals directly: `fallthroughFacts.parTraverse_` in the non-recur branch and `fallthroughFacts.zip(sourcePatterns).parTraverse_` in the recur branch.

johnynek:
Applied all requested PR feedback in `TypedExprRecursionCheck.scala` and validated locally. Test runs passed: `sbt "coreJVM/testOnly dev.bosatsu.TypedExprRecursionCheckTest"` and required pre-push `scripts/test_basic.sh` (cli/testOnly + coreJVM/testOnly).

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -2684,13 +2633,13 @@ object TypedExprRecursionCheck {
...
-                      branches.toList.zipWithIndex.zip(fallthroughFacts).parTraverse_ {
-                        case ((branch, idx), fallthroughFact) =>
+                      fallthroughFacts.zip(sourcePatterns).parTraverse_ {
+                        case ((branch, fallthroughFact), sourcePat) =>
                           for {
                             _ <- checkForIllegalBindsSt(branch.pattern.names, tag.region)
                             _ <- beginBranch(
                               arg,
-                              sourcePatterns.lift(idx),
+                              Some(sourcePat),
```
johnynek: do we ever call this with ` None`? Can we change `beginBranch` to accept a just the `Pattern.Parsed`? If we can trivially do so, we should.

johnynek: Implemented. `beginBranch` now takes `Pattern.Parsed` directly (not `Option`) since this path always has source patterns. I also removed the now-unused compiled fallback path (`allowedByTargetFromCompiled`) and simplified the `allowed`/alias-binding logic to use the parsed pattern unconditionally.

johnynek:
Applied the requested simplification and removed dead fallback code in `TypedExprRecursionCheck.scala`. Re-ran required checks: `scripts/test_basic.sh` passed (cli/testOnly: 62 passed; coreJVM/testOnly: 1557 passed, 2 ignored).

johnynek:

- error type: `CommandError`
- detail: `Command failed cmd: git clone --reference-if-able /Users/oscar/.local/share/mergexo/repos/johnynek/bosatsu.git git@github.com:johnynek/bosatsu.git /Users/oscar/.local/share/mergexo/checkouts/johnynek/bosatsu/worker-04 exit: 128 stdout: s...`

Issue #1983, opened 2026-03-04
int solver can't handle a case it should in TreeList


johnynek:
in test_workspace/TreeList.bosatsu

if I try to rewrite `def get` to:
```
def get(TreeList(trees): TreeList[a], idx: Int) -> Option[a]:
    def go(trees, idx):
        loop (idx, trees):
            case _ if cmp_Int(idx, 0) matches LT: None
            case (_, []): None
            case (idx, [Single(h), *rest]):
                if eq_Int(idx, 0): Some(h)
                else: go(rest, idx - 1)
            case (_, [Branch(s, _, t1, t2), *rest]):
                # we can either skip just one, or skip the entire Tree
                match cmp_Int(idx, s):
                    case LT: go([t1, t2, *rest], idx - 1)
                    case _: go(rest, idx - s)
    go(trees, idx)

```
I get the errors:
```

in file: /Users/oscar/code/bosatsu/test_workspace/TreeList.bosatsu:70:23, package Bosatsu/Collection/TreeList
cannot prove Int recursion obligation for go: lower recursive Int argument
recur target: idx
path condition: (and (not (< idx_0 0)) (not (= idx_1 0)) (not (= idx_1 0)))
solver detail: unable to lower recursive argument (ap
    (var
        Bosatsu/Collection/TreeList::operator -
        (Bosatsu/Predef::Int, Bosatsu/Predef::Int) -> Bosatsu/Predef::Int)
    (var idx Bosatsu/Predef::Int) (lit 1 Bosatsu/Predef::Int)
    Bosatsu/Predef::Int) to SMT Int expression; hint: rewrite recursive argument using canonical Int operations, e.g. idx.sub(1) or idx.add(-1)
68|            case (idx, [Single(h), *rest]):
69|                if eq_Int(idx, 0): Some(h)
70|                else: go(rest, idx - 1)
                         ^^^^^^^^^^^^^^^^^


in file: /Users/oscar/code/bosatsu/test_workspace/TreeList.bosatsu:74:30, package Bosatsu/Collection/TreeList
cannot prove Int recursion obligation for go: lower recursive Int argument
recur target: idx
path condition: (and (not (< idx_0 0)) (< idx_0 s_1))
solver detail: unable to lower recursive argument (ap
    (var
        Bosatsu/Collection/TreeList::operator -
        (Bosatsu/Predef::Int, Bosatsu/Predef::Int) -> Bosatsu/Predef::Int)
[...]

PR #1984, opened 2026-03-04
Remove some uses of int_loop to exercise int recursion

johnynek:
see #1983 for some difficulties I had making this PR typecheck.


johnynek:
wait for #1985 to be merged and see if we can use it to clean up the TreeList code.

PR #1988, opened 2026-03-04
Fix subsumed-guard pattern name alignment in typed recursion checking

### LINKED ISSUE #1987: int recur checking logic doesn't seem sound around patterns
In core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala

in the code merged in PR: #1985 

We have code like:
```
    private def availableBranchNames(
        currentPattern: Pattern[(PackageName, Identifier.Constructor), Type],
        state: SmtBranchState
    ): Set[Bindable] =
      state.intBindings.keySet ++
        state.boolBindings.keySet ++
        state.comparisonBindings.keySet ++
        currentPattern.names

    private def addSubsumedGuardFallthroughFacts(
        priorBranches: List[TypedExpr.Branch[Declaration]],
        currentPattern: Pattern[(PackageName, Identifier.Constructor), Type],
        state: SmtBranchState
    ): SmtBranchState =
      // If a previous branch pattern fully covers the current pattern, falling
      // through implies that prior guard was false.
      priorBranches.foldLeft(state) { (st, priorBranch) =>
        priorBranch.guard match {
          case Some(guardExpr)
              if patternSubsumes(priorBranch.pattern, currentPattern, st) =>
            val guardFree = guardExpr.freeVarsDup.toSet
Comment view            if (guardFree.subsetOf(availableBranchNames(currentPattern, st))) {
              val (guardOpt, st1) = lowerBoolExpr(guardExpr, st)
              guardOpt match {
                case Some(guardCond) =>
                  addPathFactIfNonTrivial(SmtExpr.Not(guardCond), st1)
                case None =>
                  st
              }
            } else st
          case _ =>
            st
        }
      }
```

This doesn't seem safe to me. The pattern subsumption check is okay, but when it comes to names there is nothing guaranteeing the same names are bound to the same points. Consider this code:

```

def get(TreeList(trees): TreeList[a], idx: Int) -> Option[a]:
    def go(trees, idx):
        loop (idx, trees):
            case _ if cmp_Int(idx, 0) matches LT: None
            case (_, _) if cmp_Int(idx, 0) matches EQ:
                match trees:
                    case [Single(h), *_] | [Branch(_, h, _, _), *_]: Some(h)
                    case []: None
            case (_, [Single(_), *rest]) if cmp_Int(idx, 1) matches EQ | GT: go(rest, idx.sub(1))
            case (_, [Branch(s, _, t1, t2), *rest]) if cmp_Int(idx, s) matches LT:
                # we can either skip just one, or skip the entire Tree
                go([t1, t2, *rest], idx.sub(1))
            case (_, [Branch(s, _, _, _), *rest]) if cmp_Int(s, 0) matches GT:
                go(rest, idx.sub(s))
            case _: None
    go(trees, idx)
```
When the pattern has no names AND the next pattern doesnt have names that shadow previous guards, we are okay. But if the current pattern has a name scope that isn't the same as a previous, we have to be careful.

In the above, pattern+guard 4 is `(_, [Branch(s, _, t1, t2), *rest]) if cmp_Int(idx, s) matches LT` and pattern+guard 5 is :`(_, [Branch(s, _, _, _), *rest]) if cmp_Int(s, 0) matches GT` Here, pattern 4 subsumes 5. And though it also binds t1 and t2 those aren't used in the guard 4. So this is okay. We can use it as is.

But what if pattern5 + guard was:
`(_, [Branch(t, _, _, _), *rest]) if cmp_Int(t, 0) matches GT`
we could still use pattern4 + guard4, but we would need to substitute that `t` in the current pattern is `s` in the previous pattern. So we would need some kind of alignment of the pattern names.

I think we need some function like:
```
enum Rename {
  case Same
  case To(name: Name)
  case Unused
}
def alignSubsume(p1: Pattern, p2: Pattern): Option[Map[Name, Rename]]
```
So for this function if `p1` matching implies `p2` matching, return Some`. Where for each bound name in p1 is in the keys of the map and the corresponding name in p2 is in the value. It could be the Same, renamed To another name, or the name in p1 could be unused. Then we could rewrite the guard in the valid names of the current pattern where we discard the parts of the guard that are unused in the current pattern.

So, to solve this issue, we need to write more tests, including some adversarial examples that should fail currently because the scopes are simply assumed to be the same. If we choose different variable names I think the current code can fail, or even prove false statements because it mistakes names.

johnynek:
Implemented a direct fix in `TypedExprRecursionCheck` for issue #1987 by aligning prior-branch pattern bindings to current-branch bindings before deriving negated guard fallthrough facts. The checker now computes structural name alignment (including nested struct/list/string pattern parts), rewrites prior guards with safe substitutions, and only applies fallthrough facts when every guard variable bound by the prior pattern can be soundly mapped into the current branch scope. This removes the unsound same-name/different-position behavior and enables valid renamed-binding cases.

Added regression tests in `TypedExprRecursionCheckTest` that reproduced the issue before the fix:
1) renamed binding across subsumed branches (previously rejected, now accepted),
2) same identifier reused at different pattern positions (previously accepted unsafely, now rejected).

Validation run:
- `scripts/test_basic.sh`
Both passed after the fix.

Fixes #1987

### DIFF
--- a/cli/src/test/scala/dev/bosatsu/codegen/python/PythonGenTest.scala
+++ b/cli/src/test/scala/dev/bosatsu/codegen/python/PythonGenTest.scala
@@ -313,7 +313,8 @@ class PythonGenTest extends munit.ScalaCheckSuite {
       "test_workspace/PredefTests.bosatsu",
       PackageName.parts("PredefTests"),
       "test",
-      "test_workspace/Float64.bosatsu"
+      "test_workspace/Float64.bosatsu",
+      "test_workspace/Loops.bosatsu"
     )
   }
 


johnynek:
Please merge origin main

johnynek:
Merged `origin/main` into this PR branch (now at `df89d7dd80943be0ea3f66da2ab3ae9e80078f39`) with no conflicts requiring additional file edits. Re-ran required pre-push checks on the merged branch: `scripts/test_basic.sh` passed (including `cli/testOnly` and `coreJVM/testOnly`).

johnynek:
please merge origin main again

johnynek:
Merged `origin/main` again via MergeXO git ops. Branch head is now `d358eae7376f21a9f595d05a709dd0313a93c8a1` (merge commit on top of `cc25c99e31d2b87b6855b606feb39ec12bbc61d7`). Required check was rerun on this merged head: `scripts/test_basic.sh` passed (including `cli/testOnly` and `coreJVM/testOnly`). No additional code edits were needed.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -1623,14 +1624,179 @@ object TypedExprRecursionCheck {
...
+
+    private def alignSubsumedGuardExpr(
+        guardExpr: TypedExpr[Declaration],
+        superPattern: TypedPattern,
+        subPattern: TypedPattern
+    ): Option[TypedExpr[Declaration]] = {
+      val superBoundNames = superPattern.names.toSet
+      val guardBoundByPattern = guardExpr.freeVarsDup.toSet.intersect(superBoundNames)
+      alignSubsumedPatternNames(superPattern, subPattern).flatMap { alignedNames =>
+        if (!guardBoundByPattern.subsetOf(alignedNames.keySet)) None
```
johnynek: what about variables free to both the patterns? For instance checks on the original loop variable? That's very common. Can't we handle that? It seems we need to include the intersection of the free variables from both patterns.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
```diff
@@ -528,6 +528,40 @@ def walk(idx: Int, stack: List[Int]) -> Int:
...
+    allowed("""#
+enum Node:
+  Branch(size: Int)
+
+def walk(idx: Int, stack: List[Node]) -> Int:
+  loop (idx, stack):
+    case _ if cmp_Int(idx, 0) matches LT: idx
+    case (_, []): idx
+    case (_, [Branch(s), *_]) if cmp_Int(idx, s) matches LT: idx
+    case (_, [Branch(t), *tail]) if cmp_Int(t, 0) matches GT:
```
johnynek, line 546: what if we rewrite this test except with `t = idx` so it is shadowing `idx`:

```
case (i0, [Branch(idx), *tail]) if cmp_Int(idx, 0) matches GT:
    walk(i0.sub(idx), tail)
```
That *should* work, but I think the current code would get confused.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Good catch. I updated the alignment logic to handle this safely.

Changes in `TypedExprRecursionCheck`:
- `alignSubsumedPatternNames` now maps `Pattern.Var(superName)` to the corresponding `subPattern` top binder when present, so shared loop-variable binders are carried through alignment.
- `alignSubsumedGuardExpr` now computes `guardSharedAcrossPatterns` from the guard free vars intersected with both pattern binder sets, restricted by aligned targets, and uses that when deciding whether a guard can be aligned.

This allows common guard variables across subsumed branches while still keeping the adversarial different-position case rejected.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
johnynek: Added the shadowing regression you asked for in `TypedExprRecursionCheckTest`:
- `loop aligns subsumed guard facts when current branch shadows recur names`
- It uses:
  - `case (i0, [Branch(s), *_]) if cmp_Int(i0, s) matches LT`
  - `case (i0, [Branch(idx), *tail]) if cmp_Int(idx, 0) matches GT: walk(i0.sub(idx), tail)`

This now passes along with the existing alignment and adversarial tests.

johnynek:
Implemented follow-up fixes for the two review points and reran tests. Local verification passed:
- `scripts/test_basic.sh`

johnynek:
The coverage of the TypedRecursionCheck is only 80%. We need at least 95% for such and important piece of code.

Run code coverage and see which parts are uncovered, then add targeted tests of real bosatsu code to trigger those paths. Repeat until you have covered 95% or more.

johnynek:
Reproduced the reported `scripts/test_basic.sh` failure by replaying the provided ScalaCheck seed in `dev.bosatsu.TotalityTest` (`f17kBh3twbPSEbgeyVw3afHQtAMHKK-0fH0KCmcfR_B=`). The failure was deterministic in `SetOpsLaws`' missing-branches law. I fixed `core/src/main/scala/dev/bosatsu/set/SetOps.scala` so `missingBranches` is closed under re-application: after the initial greedy result, it now performs a bounded fixpoint pass (up to 8 rounds) adding any newly discovered missing branches until no additional coverage is needed. This preserves existing tests and resolves the counterexample rather than weakening test assertions. Validation run: `sbt "coreJVM/testOnly dev.bosatsu.TotalityTest"` passed after the fix, and required pre-push command `scripts/test_basic.sh` passed (`Total 62, Failed 0`).

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
```diff
@@ -562,6 +562,178 @@ def walk(idx: Int, stack: List[Node]) -> Int:
...
+    case (_, []): idx
+    case (_, [Branch(s), *_]) if cmp_Int(s, 0) matches LT | EQ: idx
+    case (_, [Branch(t), *tail]) if cmp_Int(idx, t) matches EQ | GT:
+      walk(idx.sub(t), tail)
+    case _:
+      idx
+""")
+  }
+
+  test("loop currently rejects list prefix wildcard-to-named splice in subsumed guard alignment") {
```
johnynek: explain why this can't be allowed, or update our scheme to support it. My read is that we should be able to support this case by more carefully aligning.

johnynek: Implemented support for this case instead of leaving it rejected. The failure came from state ordering in non-`recur` match checking: subsumed-guard facts were added before pattern context, and then `addPatternFactsAndBindings` removed/rebound branch names, so prior-guard and current-guard constraints ended up on different SMT symbols (the `x_1` vs `x_2` split). I reordered context application in `TypedExprRecursionCheck` to apply pattern facts/bindings before subsumed-guard facts (and then fallthrough), which keeps symbol usage consistent. I also updated the regression tests to assert both list-splice directions are allowed (`wildcard->named` and `named->wildcard`).

johnynek:
Updated nested-match recursion checking so subsumed-guard facts are composed with branch pattern context in a stable order, which fixes the list prefix splice case raised in review. Regression tests were updated to assert these splice-alignment cases are accepted. Required pre-push command passed: `scripts/test_basic.sh` (Total 62, Failed 0).

johnynek:
Reproduced the GitHub Actions failure locally with `sbt "cli/testOnly dev.bosatsu.codegen.python.PythonGenTest"`. `test PredefTests` failed with `UnknownImportPackage(Bosatsu/Loops, PredefTests)`. Root cause: `test_workspace/PredefTests.bosatsu` now imports `Bosatsu/Loops`, but the test only compiled `PredefTests` plus `Float64`. I updated `cli/src/test/scala/dev/bosatsu/codegen/python/PythonGenTest.scala` to include `test_workspace/Loops.bosatsu` in that test's `runBoTests` extra paths. Verification: targeted suite now passes, and required pre-push command `scripts/test_basic.sh` passes (`Total 62, Failed 0`).

PR #1995, opened 2026-03-04
Normalize recur guard branch pairs into a single match (issue #1994)

### LINKED ISSUE #1994: improve normalization for int recursion shaped guards
With the #1922 int recursion support we often have loops that are actually just if/else branches. For instance:

```
  def loop(n, cnt):
    recur n:
      case _ if cmp_Int(n, 0) matches GT:
        loop(n.div(2), cnt.add(1))
      case _:
        cnt
```

We should have the typednormalization code turn cases like this into something like:

```
def loop(n, cnt):
   (loop [n n cnt cnt]
     match cmp_Int(n, 0):
       case GT: (recur n.div(2) cnt.add(1))
       case _: cnt
   )
```
so, when we have a trailing pair of branches where neither have pattern names bound, we can rewrite the guard into an if/else branch:
so we could convert this first to:
```
    recur n:
      case _:
        if cmp_Int(n, 0) matches GT:
           loop(n.div(2), cnt.add(1))
        else: cnt
```
but then we would see a match that is a no-op match, so `match x: case _: y` can be rewritten as `y`.

Before this pattern may have never came up, but with int recursion it is a fairly common shape.

We should have tests in TypedExprNormalization for this kind of shape showing it is lowered into a loop/recur with a single match with two cases.

johnynek:
Implemented the #1994 fix in TypedExpr normalization.

What changed:
- In `TypedExprNormalization`, added a guarded-branch rewrite for recursion-shaped matches: when the trailing pair is `case p if g: e1` / `case p: e2` with no pattern binders and the guarded branch contains `recur`, it now rewrites to a single branch with an explicit boolean selector (`True`/`False`) expression.
- Added a follow-up normalization (`flattenBoolMatchArg`) that collapses nested `match (match ... -> Bool)` forms into a single match on the original inner argument, so `cmp_Int(... ) matches GT`-style guards become a direct two-case match as requested.
- Added regression coverage in `TypedExprTest` for the exact `recur n` + `cmp_Int(n, 0) matches GT` shape, asserting normalization yields loop/recur with exactly one unguarded two-branch match.
- Updated `ClangGenTest` Ackermann snapshot hash because the normalization/codegen output changed as a result of this fix.

Validation:
- `sbt "coreJVM/testOnly dev.bosatsu.TypedExprTest"` passed.

Fixes #1994

### DIFF
--- a/cli/src/test/scala/dev/bosatsu/codegen/python/PythonGenTest.scala
+++ b/cli/src/test/scala/dev/bosatsu/codegen/python/PythonGenTest.scala
@@ -313,7 +313,8 @@ class PythonGenTest extends munit.ScalaCheckSuite {
       "test_workspace/PredefTests.bosatsu",
       PackageName.parts("PredefTests"),
       "test",
-      "test_workspace/Float64.bosatsu"
+      "test_workspace/Float64.bosatsu",
+      "test_workspace/Loops.bosatsu"
     )
   }
 
--- a/core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
+++ b/core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
@@ -3,9 +3,10 @@ package dev.bosatsu
 import cats.{Eval, StackSafeMonad}
 import cats.data.{NonEmptyChain, NonEmptyList, Validated, ValidatedNec}
 import cats.implicits._
+import dev.bosatsu.pattern.NameMap
 import dev.bosatsu.rankn.{Type, TypeEnv}
 import dev.bosatsu.scalawasiz3.{Z3Platform, Z3Result}
-import dev.bosatsu.smt.{SExpr, SmtCommand, SmtExpr, SmtLibRender, SmtScript, SmtSort, Z3Api}
+import dev.bosatsu.smt.{SExpr, SmtCommand, SmtExpr, SmtLibRender, SmtScript, SmtScriptScope, SmtSort, Z3Api}
 
 import Identifier.Bindable
 
@@ -70,6 +71,8 @@ object TypedExprRecursionCheck {
     case class RecurTargetItem(group: Int, index: Int, paramName: Bindable)
         derives CanEqual
     type RecurTarget = NonEmptyList[RecurTargetItem]
+    type TypedPattern = Pattern[(PackageName, Identifier.Constructor), Type]
+
     type TopLevelLowerableAliases = Map[(PackageName, Bindable), TopLevelAlias]
     type TopLevelPredefAliases = Map[(PackageName, Bindable), Identifier.Name]
 
@@ -1530,11 +1533,25 @@ object TypedExprRecursionCheck {
       lowerMatchBranchCondition(argExpr, branch, state1)
     }
 
+    private case class FallthroughFacts(
+        facts: NonEmptyList[(TypedExpr.Branch[Declaration], SmtExpr.BoolExpr)],
+        symbolState: SmtBranchState
+    )
+
+    private def mergeSymbolState(
+        symbols: SmtBranchState,
+        state: SmtBranchState
+    ): SmtBranchState =
+      state.copy(
+        declarations = state.declarations ++ symbols.declarations,
+        freshId = state.freshId.max(symbols.freshId)
+      )
+
     private def matchFallthroughFacts(
         argExpr: TypedExpr[Declaration],
         branches: NonEmptyList[TypedExpr.Branch[Declaration]],
         initialState: SmtBranchState
-    ): NonEmptyList[(TypedExpr.Branch[Declaration], SmtExpr.BoolExpr)] = {
+    ): FallthroughFacts = {
       val baseIntBindings = initialState.intBindings
       val baseBoolBindings = initialState.boolBindings
       val baseComparisonBindings = initialState.comparisonBindings
@@ -1557,7 +1574,7 @@ object TypedExprRecursionCheck {
       ) =
         (Nil, SmtExpr.BoolConst.True, initialState)
 
-      val (revFacts, _, _) =
+      val (revFacts, _, stateN) =
         branches.toList.foldLeft(init) {
           case ((acc, priorMiss, state), branch) =>
             val state0 = resetBranchBindings(state)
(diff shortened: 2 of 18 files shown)


core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
```diff
@@ -394,6 +394,71 @@ object TypedExprNormalization {
...
+  private def flattenBoolMatchArg[A](
+      arg: TypedExpr[A],
+      branches: NonEmptyList[Branch[A]],
+      tag: A
+  ): Option[TypedExpr[A]] = {
+    def asBoolSelector(
+        bs: NonEmptyList[Branch[A]]
+    ): Option[(TypedExpr[A], TypedExpr[A])] =
+      bs.toList match {
+        case List(Branch(truePat, None, ifTrue), Branch(falsePat, None, ifFalse))
```
johnynek: use:
```scala
bs match {
  case NonEmptyList(Branch(TruePattern, None, ifTrue), Branch(FalsePattern, None, ifFalse) :: Nil) =>
  Some((ifTrue, ifFalse)
  case _ => None
}
```

core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
```diff
@@ -1455,8 +1529,38 @@ object TypedExprNormalization {
             case _ =>
               (changed0, branches1)
           }
+
+        def rewriteTrailingGuardPair(
+            bs: NonEmptyList[Branch[A]]
+        ): Option[NonEmptyList[Branch[A]]] =
+          bs.toList match {
+            case init :+ Branch(p1, Some(g), e1) :+ Branch(p2, None, e2)
+                if (p1 == p2) && p1.names.isEmpty && p2.names.isEmpty &&
```
johnynek: don't use `p1 == p2` we want `TotalityCheck(typeEnv).difference(p2, p1).isEmpty` which would show that p1 is a superset of p2.

core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
```diff
@@ -1455,8 +1529,38 @@ object TypedExprNormalization {
...
+                  containsRecur(e1) =>
+              val ifExpr = Match(
+                g,
+                NonEmptyList.of(
```
johnynek: write:
```scala
NonEmptyList(
  Branch(TruePattern, None, e1),
  Branch(FalsePattern, None, e2) :: Nil)
```

core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
```diff
@@ -1455,8 +1529,38 @@ object TypedExprNormalization {
...
+                  Branch(TruePattern, None, e1),
+                  Branch(FalsePattern, None, e2)
+                ),
+                g.tag
+              )
+              val rewritten = init :+ Branch(p2, None, ifExpr)
```
johnynek: this should be `p1` because `p1` must be the super-pattern for this optimization to apply (we don't actually want to require `p1 == p2`

core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
```diff
@@ -1455,8 +1529,38 @@ object TypedExprNormalization {
...
+              Some(NonEmptyList.fromListUnsafe(rewritten))
```
johnynek: `Some(NonEmptyList.fromInitLast(init, Branch(p1, None, ifExpr))`

is better than appending and using unsafe.

core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
johnynek: Actually, I think we want:
```
bs match {
  case NonEmptyList(Branch(TruePattern, None, ifTrue), Branch(FalsePattern | Pattern.WildCard, None, ifFalse) :: Nil) =>
  Some((ifTrue, ifFalse)
  case NonEmptyList(Branch(FalsePattern, None, ifFalse), Branch(TruePattern | Pattern.WildCard, None, ifTrue) :: Nil) =>
  Some((ifTrue, ifFalse)
  case _ => None
}
```
to be robust to the ordering of the branches and to leverage totality of the matches (once we have a bool and we have matched one case in one branch the other MUST match it.

We could also handle cases with `Var(_)` instead of `Wildcard` by setting `Let(var = bool, in ...)` as the expr, but I don't think that pattern actually will come up very much and may not be worth adding support for that. But the Wildcard version is worth it.

johnynek: Updated this to pattern-match directly on `NonEmptyList` as requested. `asBoolSelector` now matches exactly `NonEmptyList(Branch(TruePattern, None, ifTrue), Branch(FalsePattern, None, ifFalse) :: Nil)`.

core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
johnynek: Done. Replaced `p1 == p2` with a totality/superset check: `totalityCheck.difference(p2, p1).isEmpty`. I also added a local `totalityCheck` value using `TotalityCheck(ev.substituteCo[[x] =>> TypeEnv[x]](typeEnv))` so this works with the generic `TypeEnv[V]` in normalization.

core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
johnynek: Updated to the explicit constructor form: `NonEmptyList(Branch(TruePattern, None, e1), Branch(FalsePattern, None, e2) :: Nil)`.

core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
johnynek: Fixed: the rewritten trailing branch now uses `p1` (the super-pattern) instead of `p2`.

core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
johnynek: Updated to avoid append + unsafe. In this cats version the API is `NonEmptyList.ofInitLast(...)` (not `fromInitLast`), so I used: `Some(NonEmptyList.ofInitLast(init, Branch(p1, None, ifExpr)))`.

johnynek:
Applied all requested line-level feedback in `TypedExprNormalization.scala` and verified with tests. Ran required pre-push command `scripts/test_basic.sh` successfully (all passing).

core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
johnynek: Implemented. `asBoolSelector` now handles both branch orders and wildcard fallthrough in the second branch:
- `True` then `False | _`
- `False` then `True | _`

This makes the bool-match flattening robust to branch ordering and totality-based wildcard cases, matching your suggested shape.

johnynek:
Applied the requested robustness update in `core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala` and reran the required pre-push check. `scripts/test_basic.sh` passed (cli/testOnly and coreJVM/testOnly both green).

Issue #2003, opened 2026-03-05
expand `Lazy` to be a full `Eval`


johnynek:
Bosatsu/Lazy is very minimally featured.

cats.Eval has more features that allow dealing with stack safety in lazy non-tail loops.

We could bring this to Lazy. I think what we need are:
```
def map_Lazy[a, b](lazy: Lazy[a], fn: a -> b) -> Lazy[b]

def flat_map_Lazy[a, b](lazy: Lazy[a], fn: a -> Lazy[b]) -> Lazy[b]

def defer_Lazy[a](fn: () -> Lazy[a]) -> Lazy[a]

def const_Lazy[a](a: a) -> Lazy[a]
```

if you compose all of these without using `get_Lazy` inside, only one at the end, it should always use O(1) stack by evaluating the computation on "stack-on-heap" approach which we can write in the external code.

If we just add `flat_map_Lazy` and `const_Lazy` then I think defer and map can be implemented with those two. If we do the reassociation trick that cats.Eval (and other free monads) do, we can probably get decent performance.

I think conceptually, we should cache every Lazy value, so it we run `get_Lazy` twice it only computes the loop once (but no exactly once guarantees, just the same best-effort atomic representation we have now).

johnynek:
MergeXO assigned an agent and started design work for issue #2003.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/2006

johnynek:
replaced by #2129

PR #2006, opened 2026-03-05
Design doc for #2003: Issue #2003 Design: expand `Lazy` to be a full `Eval`

johnynek:
Design doc.

Refs #2003

### DIFF
--- a/docs/design/2003-expand-lazy-to-be-a-full-eval.md
+++ b/docs/design/2003-expand-lazy-to-be-a-full-eval.md
@@ -122,10 +122,28 @@ Rationale:
 Represent each `Lazy` node as one memo cell with an atomic state, conceptually:
 
 - `Thunk(fn)` for `lazy`
-- `Bind(source, fn)` for `flat_map_Lazy`
+- `Bind(source, k)` where `source` is never a `Bind` node and `k` is a normalized Kleisli chain
 - `Done(value)` for memoized success
 
 `const_Lazy` creates `Done(value)` directly.
+
+### Bind normalization (right-association)
+
+To avoid quadratic behavior from left-associated chains, `flat_map_Lazy` must normalize using monad reassociation:
+
+- `a.flat_map(f).flat_map(g) == a.flat_map(x -> f(x).flat_map(g))`
+
+Implementation invariant:
+
+- The left argument of `Bind` is never another `Bind`; it is always a node that will force directly (`Thunk` or `Done`).
+- Kleisli chains are represented explicitly as either a user function (`UserFn`) or a composition node (`AndThen(k1, k2)`) for `k1` followed by `k2`.
+
+Construction rule for `flat_map_Lazy`:
+
+- If the input lazy is already `Bind(source, k1)`, produce `Bind(source, AndThen(k1, UserFn(f)))`.
+- Otherwise produce `Bind(input, UserFn(f))`.
+
+This keeps construction O(1) per `flat_map_Lazy`, keeps trees right-associated, and prevents nested-left `Bind` towers.
 
 ### Stack-safe forcing algorithm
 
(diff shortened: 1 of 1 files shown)


docs/design/2003-expand-lazy-to-be-a-full-eval.md
```diff
@@ -0,0 +1,246 @@
...
+2. If state is `Done(v)`:
+- If no continuation frame remains, finish with `v`.
+- Else pop next continuation `fn`, compute next lazy via `fn(v)`, continue loop.
+3. If state is `Thunk(fn)`, run `fn(())`; on success CAS state to `Done(value)`.
+4. If state is `Bind(source, fn)`, push frame and continue with `source`.
+5. On final success, opportunistically path-compress visited `Bind` nodes to `Done(result)` with CAS.
+
+Properties:
+
+- O(1) call stack with O(n) heap frames for n bind depth.
```
johnynek, line 162: If we keep a pointer to the root we are computing we should be able to use normal malloc to hold the heap frames and not incur libgc penalty right?

johnynek: Good call. I updated the design doc to make this explicit: in the C runtime, continuation frames should use `malloc/free` (not `GC_malloc`) to avoid Boehm GC scan overhead, and I added the required safety invariant to keep a strong root reference to the `Lazy` being forced so non-GC frames never become the sole owner of live `BValue` references.

johnynek:
Updated `docs/design/2003-expand-lazy-to-be-a-full-eval.md` to incorporate the C frame allocation guidance and GC-safety invariant. Ran required pre-push check `scripts/test_basic.sh`; it passed (`Passed: Total 62, Failed 0, Errors 0`).

docs/design/2003-expand-lazy-to-be-a-full-eval.md
```diff
@@ -0,0 +1,249 @@
...
+8. Under races, duplicate evaluation is allowed; converged cached value remains best-effort atomic as today.
+
+## Architecture
+
+### Shared state model
+
+Represent each `Lazy` node as one memo cell with an atomic state, conceptually:
+
+- `Thunk(fn)` for `lazy`
+- `Bind(source, fn)` for `flat_map_Lazy`
```
johnynek: We need to right associate these using the monad law:

a.flat_map(f).flat_map(g) =
  a.flat_map(x => f(x).flat_map(g))

To avoid quadratic blowups in evaluation.

So we need a representation of either a ran user function for flat mapping or this kleisli pair of f and then g. So the left argument of Bind is never Bind. Always a Done cell or a Thunk.

johnynek: Updated the design doc to require right-association explicitly and to encode it in the runtime representation. In `Shared state model`/`Bind normalization (right-association)`, `Bind` is now `Bind(source, k)` with the invariant that `source` is never a `Bind` (only `Thunk`/`Done`), and Kleisli composition is represented as `UserFn`/`AndThen`. I also added explicit construction rules for `flat_map_Lazy` that reassociate (`Bind(source, AndThen(k1, UserFn(f)))`) instead of creating left-nested binds, plus corresponding implementation-plan bullets for Scala/Python/C.

johnynek:
Applied the doc update in `docs/design/2003-expand-lazy-to-be-a-full-eval.md` to codify monad-law reassociation and the non-left-nested `Bind` invariant. Ran required pre-push check `scripts/test_basic.sh`; it passed (`Passed: Total 62, Failed 0, Errors 0`).

johnynek:
No line-level review feedback to address in this PR, and I did not edit files. I reproduced the reported CI failure locally with `sbt "coreJVM/testOnly dev.bosatsu.smt.SmtExprNormalizeAndPathImpliesTest -- --log=failure"` (twice). Both runs failed in `dev.bosatsu.smt.SmtExprNormalizeAndPathImpliesTest` at `core/src/test/scala/dev/bosatsu/smt/SmtExprNormalizeAndPathImpliesTest.scala:331` with `Failed executing embedded z3.wasm: Trapped on unreachable instruction`, using different ScalaCheck seeds (`mv2THeYJ4EqUGZvfyGVMEZYWLy7VhChNkmP64cdrHlH=` and `Wp_pyYJ_6ycWzFyHFSrcplwPNfBRQ77zigbAufqzGLI=`). Since this PR changes only `docs/design/2003-expand-lazy-to-be-a-full-eval.md`, this appears to be an unrelated baseline test failure rather than a regression from this PR.

johnynek:
replaced by #2129

PR #2010, opened 2026-03-05
Right-associate Prog `flat_map`/`recover` and add runtime coverage

### LINKED ISSUE #2007: Right associate flat_map and recover in Prog
Prog has naive representation of FlatMap (bind): but allocate a case class wrapper. But that means when we evaluated it we have to unnest FlatMaps on the left hand side.

Alternatively we could use the monad law:

a.flatMap(f).flatMap(g) ==
  a.flatMap(x => f(x).flatMap(g))

And preserve the invariant that the left of a FlatMap is never a FlatMap itself. Supposedly this avoids a quadratic blowup on evaluation but I only see a hand waving argument for that not a clear proof.

The same issue exists for the error recover path (although less severe since recovery is so much more rare).

To fix this, all the platforms should make this change and have good test coverage.

johnynek:
Implemented issue #2007 by right-associating Prog constructors so nested `flat_map`/`recover` chains keep a non-nested left spine across JVM/C/Python runtimes (`core/src/main/scala/dev/bosatsu/Predef.scala`, `c_runtime/bosatsu_ext_Bosatsu_l_Prog.c`, `test_workspace/ProgExt.py`). Added focused coverage for the invariant and behavior in JVM/C/Bosatsu tests (`core/src/test/scala/dev/bosatsu/EvaluationTest.scala`, `c_runtime/test.c`, `test_workspace/ProgAssoc.bosatsu`). To repair the required pre-push failure, also stabilized embedded z3.wasm usage by serializing solver calls via `Z3Api.withGlobalSolverLock` and using it at solver call sites, and made the path-implication soundness property call z3 only when `pathImplies` returns true (same assertion, fewer irrelevant solver invocations). Verified with `scripts/test_basic.sh` passing.

Fixes #2007

### DIFF
--- a/core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
+++ b/core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
@@ -85,13 +85,11 @@ object TypedExprRecursionCheck {
       Type.TyConst(Type.Const.predef("Comparison"))
     private val z3Solver = Z3Platform.create()
     private val z3Runner: Z3Api.RunSmt2 = { smt2 =>
-      Z3Api.withGlobalSolverLock {
-        z3Solver.runSmt2(smt2) match {
-          case Z3Result.Success(stdout, stderr, _) =>
-            Right(Z3Api.SolverOutput(stdout, stderr))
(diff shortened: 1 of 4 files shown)


johnynek:
Remove the Z3 related changes from this PR and merge main. Then push to the PR

PR #2014, opened 2026-03-05
Fix match/recur pattern type errors to point at branch patterns (issue #2013)

### LINKED ISSUE #2013: Bad error message on match type error
I had this code:
```
def uncons_step(rem: Int, current: LazyList[a], pending: List[LazyList[a]]) -> Option[(Lazy[a], LazyList[a])]:
  recur (rem, current):
    case (_, Cons(head, tail)):
      Some((head, append_pending(tail, pending)))
    case _ if cmp_Int(rem, 0) matches LT | EQ:
      None
    case (_, Empty):
      match pending:
        case []:
          None
        case [next, *rest]:
          uncons_step(rem.sub(1), next, rest)
    case (_, Concat(first, second)):
      uncons_step(rem, first, [second, *pending])
    case (_, Suspend(next)):
      uncons_step(rem.sub(1), get_Lazy(next), pending)
    case (_, Mapped(source, fn)):
      match uncons_step(rem, source, []):
        case None:
          uncons_step(rem.sub(1), Empty, pending)
        case Some((head, tail)):
          mapped_head = lazy(() -> fn(get_Lazy(head)))
          mapped_tail = map(tail, fn)
          Some((mapped_head, append_pending(mapped_tail, pending)))
    case (_, FlatMapped(source, fn)):
      match uncons_step(rem.sub(1), source, []):
        case None:
          uncons_step(rem.sub(1), Empty, pending)
        case Some((head, tail)):
          mapped = fn(get_Lazy(head)) 
          uncons_step(rem.sub(1), mapped, [flat_map(tail, fn), *pending])
```

which compiled, if I change the recur to have an extra item in the tuple:
```
def uncons_step(rem: Int, current: LazyList[a], pending: List[LazyList[a]]) -> Option[(Lazy[a], LazyList[a])]:
  recur (rem, pending, current):
    case (_, Cons(head, tail)):
      Some((head, append_pending(tail, pending)))
    case _ if cmp_Int(rem, 0) matches LT | EQ:
      None
    case (_, Empty):
      match pending:
        case []:
          None
        case [next, *rest]:
          uncons_step(rem.sub(1), next, rest)
    case (_, Concat(first, second)):
      uncons_step(rem, first, [second, *pending])
    case (_, Suspend(next)):
      uncons_step(rem.sub(1), get_Lazy(next), pending)
    case (_, Mapped(source, fn)):
      match uncons_step(rem, source, []):
        case None:
          uncons_step(rem.sub(1), Empty, pending)
        case Some((head, tail)):
          mapped_head = lazy(() -> fn(get_Lazy(head)))
          mapped_tail = map(tail, fn)
          Some((mapped_head, append_pending(mapped_tail, pending)))
    case (_, FlatMapped(source, fn)):
      match uncons_step(rem.sub(1), source, []):
        case None:
          uncons_step(rem.sub(1), Empty, pending)
        case Some((head, tail)):
          mapped = fn(get_Lazy(head)) 
          uncons_step(rem.sub(1), mapped, [flat_map(tail, fn), *pending])

```
it gives the error:
```
in file: /Users/oscar/code/zafu/src/Zafu/Collection/LazyList.bosatsu:79:9, package Zafu/Collection/LazyList
type error: expected type Tuple3[Int]
77|
78|def uncons_step(rem: Int, current: LazyList[a], pending: List[LazyList[a]]) -> Option[(Lazy[a], LazyList[a])]:
79|  recur (rem, pending, current):
           ^^^^^^^^^^^^^^^^^^^^^^^

but found type Tuple2
evidence sites:
77|
78|def uncons_step(rem: Int, current: LazyList[a], pending: List[LazyList[a]]) -> Option[(Lazy[a], LazyList[a])]:
79|  recur (rem, pending, current):
           ^^^^^^^^^^^^^^^^^^^^^^^


79|  recur (rem, pending, current):
80|    case (_, Cons(head, tail)):
81|      Some((head, append_pending(tail, pending)))
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


83|      None
84|    case (_, Empty):
85|      match pending:
         ^

to:
85|      match pending:
86|        case []:
87|          None
88|        case [next, *rest]:
89|          uncons_step(rem.sub(1), next, rest)
                                               ^


89|          uncons_step(rem.sub(1), next, rest)
90|    case (_, Concat(first, second)):
91|      uncons_step(rem, first, [second, *pending])
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


91|      uncons_step(rem, first, [second, *pending])
92|    case (_, Suspend(next)):
93|      uncons_step(rem.sub(1), get_Lazy(next), pending)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


93|      uncons_step(rem.sub(1), get_Lazy(next), pending)
94|    case (_, Mapped(source, fn)):
95|      match uncons_step(rem, source, []):
         ^

to:
 95|      match uncons_step(rem, source, []):
 96|        case None:
 97|          uncons_step(rem.sub(1), Empty, pending)
 98|        case Some((head, tail)):
 99|          mapped_head = lazy(() -> fn(get_Lazy(head)))
100|          mapped_tail = map(tail, fn)
101|          Some((mapped_head, append_pending(mapped_tail, pending)))
                                                                      ^


101|          Some((mapped_head, append_pending(mapped_tail, pending)))
102|    case (_, FlatMapped(source, fn)):
103|      match uncons_step(rem.sub(1), source, []):
          ^

to:
103|      match uncons_step(rem.sub(1), source, []):
104|        case None:
105|          uncons_step(rem.sub(1), Empty, pending)
106|        case Some((head, tail)):
107|          mapped = fn(get_Lazy(head))
108|          uncons_step(rem.sub(1), mapped, [flat_map(tail, fn), *pending])
                                                                            ^
```

The problem is we aren't currently capturing the region of the pattern in the match.

We need to point to the pattern here. One solution could be in the Declaration add `patternRegion: Region` to the `MatchBranch. Then make sure to address it with `replaceRegions`.

If we took this code:
```
    val bp = Indy { indent =>
      val casePat: P[Pattern.Parsed] =
        (P.string("case") *> Parser.spaces).with1 *> Pattern.matchParser
      val guard: P0[Option[NonBinding]] =
        (((Parser.spaces *> Parser.keySpace("if")).backtrack *> arg(indent))).?

      casePat ~ guard
    }
```
And put `Pattern.matchParser.region`  or something, we could keep the region for the Pattern. Then, if we have the Declaration still (which we should when we are typechecking) we can get the region out by inspecting the associated MatchBranch potentially.

A second approach would be to compute the pattern region by exclusion: we know the full match statement region, we know the previous branch region (if any) and we know the current branch expression region. So, by exclusion we can find the region of text that includes the space and the `case: <pattern>` and we could just run a parser on that which skips empty space and the case: and identifies the region within that string of the pattern, and then use that region to compute the original region in the file, which we can use to show the user.

This, second approach may actually be better.

our goal is to put `^^^^^^^^^^^^^` directly under the patterns which have the wrong type, not the branch expressions.

We should make a minimized version of this example and add an error message test to address it.

johnynek:
Implemented a direct fix for issue #2013 by preserving and using match-branch pattern regions end-to-end.
- Added optional `patternRegion` metadata to `Declaration.MatchBranch`, captured from `Pattern.matchParser.region` in `Declaration.matchP`.
- Updated declaration transforms (`replaceRegions` and branch rewrites) to preserve/reset `patternRegion` correctly.
- Added optional `patternRegion` to `Expr.Branch` and preserved it through expression rewrites/traversals.
- In `SourceConverter`, propagated each `Declaration.MatchBranch.patternRegion` into `Expr.Branch` and used it when converting branch patterns.
- In `Infer`, changed match branch pattern checking to use `branch.patternRegion` (fallback to `region(branch.expr)` when unavailable), so tuple-arity/type mismatches in patterns are reported at the pattern site.

Regression coverage:
- Added `ErrorMessageTest` case: `recur pattern tuple mismatch points at the pattern site`, which reproduces the tuple3-vs-tuple2 `recur` branch mismatch and asserts the reported region is the pattern `(_, _)`.
- Updated a few existing `ErrorMessageTest` expected region literals that now correctly report full pattern spans for pattern-originated source-converter errors.

Validation run:
- `sbt -batch "coreJVM/test:compile"` passed.
- `sbt -batch "coreJVM/testOnly dev.bosatsu.ErrorMessageTest"` passed.

Fixes #2013

### DIFF
--- a/core/src/main/scala/dev/bosatsu/Declaration.scala
+++ b/core/src/main/scala/dev/bosatsu/Declaration.scala
@@ -782,7 +782,7 @@ object Declaration {
                 branch.pattern,
                 branch.guard.map(_.replaceRegionsNB(r)),
                 branch.body.map(_.replaceRegions(r))
-              )(using Some(r))
+              )(using r)
             })
           )(using r)
         case Matches(a, p) =>
@@ -949,7 +949,7 @@ object Declaration {
       pattern: Pattern.Parsed,
       guard: Option[NonBinding],
       body: OptIndent[Declaration]
-  )(using val patternRegion: Option[Region] = None)
+  )(using val patternRegion: Region)
       derives CanEqual
   case class Match(
       kind: MatchKind,
@@ -1263,7 +1263,7 @@ object Declaration {
     val branch = OptIndent
       .block(bp, withTrailingExpr)
       .map { case (((pat, patternRegion), guard), body) =>
-        MatchBranch(pat, guard, body)(using Some(patternRegion))
+        MatchBranch(pat, guard, body)(using patternRegion)
       }
 
     val branchList: Indy[NonEmptyList[MatchBranch]] = Indy { indent =>
--- a/core/src/main/scala/dev/bosatsu/Expr.scala
+++ b/core/src/main/scala/dev/bosatsu/Expr.scala
@@ -138,9 +138,11 @@ sealed abstract class Expr[T] derives CanEqual {
         Match(
           arg.eraseTags,
           branches.map { b =>
-            Branch(b.pattern, b.guard.map(_.eraseTags), b.expr.eraseTags)(using
-              b.patternRegion
-            )
+            Branch(
+              b.pattern,
+              b.guard.map(_.eraseTags),
+              b.expr.eraseTags
+            )(using Region.empty)
           },
           ()
         )
@@ -197,7 +199,7 @@ object Expr {
       pattern: Pattern[(PackageName, Constructor), Type],
       guard: Option[Expr[T]],
       expr: Expr[T]
-  )(using val patternRegion: Option[Region] = None)
+  )(using val patternRegion: Region)
   case class Match[T](
       arg: Expr[T],
       branches: NonEmptyList[Branch[T]],
@@ -307,8 +309,8 @@ object Expr {
     Match(
       cond,
       NonEmptyList.of(
-        Branch(TruePat, None, ifTrue),
-        Branch(FalsePat, None, ifFalse)
+        Branch(TruePat, None, ifTrue)(using Region.empty),
+        Branch(FalsePat, None, ifFalse)(using Region.empty)
       ),
       tag
     )
@@ -465,7 +467,7 @@ object Expr {
       case (((name, _), Some(matchPat)), body) =>
         Match(
           Local(name, outer),
-          NonEmptyList.one(Branch(matchPat, None, body)),
+          NonEmptyList.one(Branch(matchPat, None, body)(using Region.empty)),
           outer
         )
     }
--- a/core/src/main/scala/dev/bosatsu/Region.scala
+++ b/core/src/main/scala/dev/bosatsu/Region.scala
@@ -30,6 +30,7 @@ object Region extends RegionLowPriority {
   def apply(start: Int, end: Int): Tpe =
     (start.toLong << 32) | (end.toLong & 0xffffffffL)
 
+  val empty: Tpe = Region(0, 0)
 
   implicit val ordering: Ordering[Region] =
     Ordering.Long
--- a/core/src/main/scala/dev/bosatsu/SourceConverter.scala
+++ b/core/src/main/scala/dev/bosatsu/SourceConverter.scala
@@ -353,7 +353,7 @@ final class SourceConverter(
               (convertPattern(pat, assignRegion), erest, rrhs).parMapN {
                 (newPattern, e, rhs) =>
                   val expBranches = NonEmptyList.of(
-                    Expr.Branch(newPattern, None, e)
+                    Expr.Branch(newPattern, None, e)(using assignRegion)
                   )
                   Expr.Match(rhs, expBranches, decl)
               }
@@ -427,8 +427,10 @@ final class SourceConverter(
             a,
             OptIndent.same(
               NonEmptyList(
-                MatchBranch(p, None, res),
-                MatchBranch(Pattern.WildCard, None, restDecl) :: Nil
+                MatchBranch(p, None, res)(using res.get.region),
+                MatchBranch(Pattern.WildCard, None, restDecl)(using
+                  restDecl.get.region
+                ) :: Nil
               )
             )
           )(using decl.region)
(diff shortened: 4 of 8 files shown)


core/src/main/scala/dev/bosatsu/Expr.scala
```diff
@@ -195,7 +197,7 @@ object Expr {
       pattern: Pattern[(PackageName, Constructor), Type],
       guard: Option[Expr[T]],
       expr: Expr[T]
-  )
+  )(using val patternRegion: Option[Region] = None)
```
johnynek: don't make this optional. Let's make it `using val patternRegion: Region` and no default.

core/src/main/scala/dev/bosatsu/Declaration.scala
```diff
@@ -946,7 +949,8 @@ object Declaration {
       pattern: Pattern.Parsed,
       guard: Option[NonBinding],
       body: OptIndent[Declaration]
-  ) derives CanEqual
+  )(using val patternRegion: Option[Region] = None)
```
johnynek: let's make this required `(using val patternRegion: Region)`

core/src/main/scala/dev/bosatsu/Expr.scala
```diff
@@ -138,7 +138,9 @@ sealed abstract class Expr[T] derives CanEqual {
         Match(
           arg.eraseTags,
           branches.map { b =>
-            Branch(b.pattern, b.guard.map(_.eraseTags), b.expr.eraseTags)
+            Branch(b.pattern, b.guard.map(_.eraseTags), b.expr.eraseTags)(using
+              b.patternRegion
```
johnynek: when erasing the tags, we should also set this to `Region.empty` or whatever that is. The goal here is to forget what regions we have come from when we do this.

core/src/main/scala/dev/bosatsu/Expr.scala
johnynek: Updated. `Expr.Branch` now has a required context parameter: `(using val patternRegion: Region)` with no optional/default fallback. I also updated all branch construction sites to pass an explicit region (real pattern region from source when available, `Region.empty` for synthetic/test constructors).

core/src/main/scala/dev/bosatsu/Declaration.scala
johnynek: Done. `Declaration.MatchBranch` is now `(using val patternRegion: Region)` and no longer optional. Parser wiring now captures and supplies this region directly from `Pattern.matchParser.region`, and `replaceRegions`/declaration rewrites preserve or reset it via explicit `using` arguments.

core/src/main/scala/dev/bosatsu/Expr.scala
johnynek: Implemented. During `Expr.eraseTags`, branch `patternRegion` is now reset to `Region.empty` so we explicitly forget source-region provenance. I added `Region.empty` in `Region.scala` and use it for erase/synthetic paths.

johnynek:
Addressed all three review comments with concrete code changes. Verified with required pre-push command: `scripts/test_basic.sh` (pass).

PR #2030, opened 2026-03-06
Implement #2025: allow constructor-rank decreases to earlier ADT constructors

### LINKED ISSUE #2025: Extend recursion checking to accept earlier ADT elements
**Proposal Sketch: ADT Constructor-Rank Decrease (Easy, Sound, More Powerful)**

This keeps the current recursion checker model and adds one new admissible decrease shape.

**Goal**
1. Keep existing structural + lexicographic + `Int` SMT rules unchanged.
2. Add sound support for “recur to an earlier constructor”, including all zero-arg variants.
3. Avoid deep type-system or syntax changes.

**Non-goal (v1)**
1. No new function qualifiers/types.
2. No proof terms.
3. No support for “hidden recursive payloads” like `List[T]` yet.

---

**Rule Set (v1)**

Let recurse-target component type be `T`, and its constructors ordered by source order:
`C0, C1, ..., Cn` (earlier = smaller index).

A recursive call argument for that target component is accepted as `Smaller` if either existing rules accept it, or this new rule accepts it:

1. Current branch must prove the current target value is constructor `Ccur`.
2. Call argument must be constructor application `Cnext(args...)` of the same `T`.
3. `index(Cnext) < index(Ccur)` (strictly earlier).
4. For each field type `Fi` of `Cnext` after type instantiation:
   - If `Fi` does not contain `T`: any expression is allowed.
   - If `Fi` is exactly `T`: argument must be a local name already in the current branch’s strict-smaller set for this target component.
   - If `Fi` contains `T` but is not exactly `T` (example `List[T]`, `Option[T]`, `Pair[Int, T]`): reject in v1.
5. Existing lexicographic tuple logic is unchanged; this classification just supplies a new `Smaller` outcome for one component.

This automatically allows zero-arg variants, because condition 4 is vacuous.

---

**What this gives**

Accepted examples:
```bosatsu
enum Option[a]: None, Some(value: a)

def drop(opt: Option[a]) -> Option[a]:
  recur opt:
    case Some(_): drop(None)
    case None: None
```

```bosatsu
enum List[a]: EmptyList, NonEmptyList(head: a, tail: List[a])

def reset(xs: List[a]) -> List[a]:
  recur xs:
    case NonEmptyList(_, _): reset(EmptyList)
    case EmptyList: EmptyList
```

```bosatsu
enum BinNat: Zero, Odd(prev: BinNat), Even(prev: BinNat)

def step(n: BinNat) -> BinNat:
  recur n:
    case Even(prev): step(Odd(prev))
    case Odd(prev): step(prev)
    case Zero: Zero
```

Rejected (intentionally, to stay sound/easy):
```bosatsu
enum T: A(xs: List[T]), B(t: T), Z
# reject B(x) -> A([B(x)]) because field type is List[T] (contains T, not exactly T)
```

---

**Why this is sound enough for v1**

1. Constructor rank strictly decreases on each constructor-rank step.
2. Any direct recursive payload (`Fi == T`) must already be structurally smaller under existing checker facts.
3. Disallowing wrapped-recursive field types (`Fi` contains `T` but `Fi != T`) blocks known cycle constructions through containers.
4. Since only branch-proven current constructor is used, there is no ambiguity from wildcard/union uncertainty.

---

**Why this is relatively easy in current checker**

You can implement this almost entirely inside recursion classification in [TypedExprRecursionCheck.scala](/Users/oscar/code/bosatsu/core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala).

1. Extend branch state with current constructor info per target component (similar to existing singleton tracking around [TypedExprRecursionCheck.scala:2813](/Users/oscar/code/bosatsu/core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala:2813)).
2. Add helper: detect constructor app in call arg (`Global Constructor` and `App(Global Constructor, ...)`).
3. Add helper: constructor index lookup from type env (`DefinedType.constructors` order; see [DefinedType.scala](/Users/oscar/code/bosatsu/core/src/main/scala/dev/bosatsu/rankn/DefinedType.scala)).
4. Add helper: `containsTargetType(Fi, T)` and `Fi.sameAs(T)`.
5. In `classifyStructuralArg` (near [TypedExprRecursionCheck.scala:1935](/Users/oscar/code/bosatsu/core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala:1935)), before `Other`, try constructor-rank classification.
6. Keep `Int` path and lexicographic driver unchanged (around [TypedExprRecursionCheck.scala:2099](/Users/oscar/code/bosatsu/core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala:2099)).

No parser or type-format change required.

---

**Tests to add (and currently failing ones to flip)**

In [TypedExprRecursionCheckTest.scala](/Users/oscar/code/bosatsu/core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala):

1. `Some(_) -> None` accepted.
2. `NonEmptyList(_,_) -> EmptyList` accepted.
3. `Even(prev) -> Odd(prev)` accepted.
4. Wrapped recursive payload rejected (`A(xs: List[T])` case above).
5. Exact recursive payload must be proven smaller local name (`A(B(x))` style rejected).
6. Wildcard/union branch without definite current ctor does not enable rule.
7. Full existing recursion suite remains green.

johnynek:
Implemented the accepted design for issue #2025.

What changed:
- Extended recursion branch state in `core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala` with per-target current constructor evidence (`currentCtorPerTarget`) and plumbed it through all recur-branch state transitions.
- Added constructor-rank classification helpers:
  - detection of constructor applications in recursive call args (`Global` and nested `App(...)` forms, plus `[]` normalization),
  - constructor metadata/index lookup via `TypeEnv.getConstructor` and `DefinedType.constructors`,
  - target-type argument extraction + constructor field-type instantiation,
  - recursive target containment checks and v1 payload safety gate (`Fi == T` must be a known smaller local; wrapped `T` payloads are rejected).
- Integrated the new rule into `classifyStructuralArg` so existing structural/equality checks run first, then constructor-rank classification is attempted before `Other`.
- Kept Int recursion and lexicographic driver behavior unchanged aside from new constructor-evidence plumbing.
- Added issue-specific tests in `core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala`:
  - accepts `Some(_) -> None`,
  - accepts non-empty -> empty custom ADT,
  - accepts `Even(prev) -> Odd(prev)` with structural follow-up,
  - rejects wrapped recursive payload (`List[T]`),
  - rejects non-local exact recursive payload (`A(B(x))` shape),
  - rejects ambiguous-branch constructor-rank usage.
- Updated `docs/src/main/paradox/recursion.md` to document constructor-rank decreases and v1 wrapped-payload limitations.

Validation run (all passed):
- `scripts/test_basic.sh` (required pre-push command)
- `sbt "doc; paradox"`

Fixes #2025

Implements design doc: [docs/design/2025-extend-recursion-checking-to-accept-earlier-adt-elements.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/2025-extend-recursion-checking-to-accept-earlier-adt-elements.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/2027

### DIFF
--- a/core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
+++ b/core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
@@ -264,6 +264,22 @@ def bad(opt: Option[a]) -> Option[a]:
 """)
   }
 
+  test("recur allows union branches when recursive args are smaller for every branch") {
+    allowed("""#
+enum T:
+  Z
+  A(t: T)
+  B(t: T)
+
+def ok(v: T) -> T:
+  recur v:
+    case A(x) | B(x):
+      ok(x)
+    case Z:
+      Z
+""")
+  }
+
   test("tuple recur targets allow lexicographic decrease") {
     allowed("""#
 enum Nat: Zero, Succ(prev: Nat)


core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -1932,11 +1983,159 @@ object TypedExprRecursionCheck {
...
+        arg: TypedExpr[Declaration]
+    ): ArgLexOrder = {
+      val typeEnv = inrec.inRec.totalityCheck.inEnv
+      val targetType = targetItemType(inrec, target)
+      val maybeSmaller =
+        for {
+          argCtorApp <- constructorAppFromArgExpr(arg)
+          currentMeta <- constructorMeta(typeEnv, currentCtor)
+          nextMeta <- constructorMeta(typeEnv, argCtorApp.ctor)
+          _ <- Option.when(currentMeta.owner.toTypeConst == nextMeta.owner.toTypeConst)(())
```
johnynek: don't use `Option.when` here I think just:
```scala
if currentMeta.owner.toTypeConst == nextMeta.owner.toTypeConst
```
will work in scala. Also applies below.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -1932,11 +1983,159 @@ object TypedExprRecursionCheck {
...
+        ctor: KnownCtor
+    ): Option[ConstructorMeta] =
+      typeEnv.getConstructor(ctor.pack, ctor.cons).flatMap { case (owner, cfn) =>
+        owner.constructors.zipWithIndex.find(_._1.name == cfn.name).map {
+          case (_, idx) =>
+            ConstructorMeta(owner, cfn, idx)
+        }
+      }
+
+    private def targetArgsForDefinedType(
```
johnynek: this should be a method on DefinedType and have its own tests (both positive and negative).

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -1932,11 +1983,159 @@ object TypedExprRecursionCheck {
...
+        case Type.TyConst(rootConst)
+            if rootConst == dt.toTypeConst &&
+              (dt.typeParams.lengthCompare(args.length) == 0) =>
+          Some(args)
+        case _ =>
+          None
+      }
+    }
+
+    private def instantiatedFieldTypes(
```
johnynek: this should be a method on DefinedType and have its own tests (both positive and negative).

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -1932,11 +1983,159 @@ object TypedExprRecursionCheck {
...
+            .map { case (tv, targ) => (tv: Type.Var) -> targ }
+            .toMap
+        Some(
+          ctor.args.map { field =>
+            Type.substituteVar(field.tpe, substitutions)
+          }
+        )
+      }
+
+    private def containsTargetType(fieldType: Type, targetType: Type): Boolean =
```
johnynek: this can be a function on `object Type` and have its own tests (both positive and negative).

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Updated this block to avoid `Option.when`. `classifyConstructorRankArg` now uses `if` guards directly in the `for`-comprehension (`if currentMeta.owner.toTypeConst == nextMeta.owner.toTypeConst`, `if nextMeta.index < currentMeta.index`, and the length/payload checks).

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Done. I moved this logic from `TypedExprRecursionCheck` to `DefinedType` as `extractTypeArgs(targetType: Type)`. I also added dedicated tests in `core/src/test/scala/dev/bosatsu/rankn/DefinedTypeTest.scala` covering positive and negative cases (`extractTypeArgs` success, wrong root, and arity mismatch).

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Done. I moved this into `DefinedType` as `instantiateConstructorFieldTypes(ctor, targetArgs)` and switched the recursion checker to call that method. Added dedicated positive/negative tests in `DefinedTypeTest` (substitution success and arity-mismatch rejection).

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Done. I moved this to `Type` as `Type.containsType(in, target)` and updated the recursion checker to call it. Added positive and negative tests in `core/src/test/scala/dev/bosatsu/rankn/TypeTest.scala` for exact match, nested match, quantified wrapping, and absent target.

johnynek:
Addressed all four review comments with concrete refactors and tests. Validation run:
- `sbt "coreJVM/test:compile; coreJVM/testOnly dev.bosatsu.TypedExprRecursionCheckTest; coreJVM/testOnly dev.bosatsu.rankn.DefinedTypeTest; coreJVM/testOnly dev.bosatsu.rankn.TypeTest"`
- `scripts/test_basic.sh`
Both passed.

core/src/test/scala/dev/bosatsu/TypedExprRecursionCheckTest.scala
```diff
@@ -167,6 +167,103 @@ def fn(x):
...
+
+  test("recur does not use constructor-rank rule when current constructor is ambiguous") {
+    disallowed("""#
+enum Option[a]:
+  None
+  Some(value: a)
+
+def bad(opt: Option[a]) -> Option[a]:
+  recur opt:
+    case Some(_) | None:
```
johnynek, line 262: Can we add a positive test with a union please? If we replace with something less than all items it should work.

Or if we use something that would have been allowed from all branches of the union is another way to say it.

johnynek: Added a positive union case in `TypedExprRecursionCheckTest`:

- `recur allows union branches when recursive args are smaller for every branch`
- Uses `case A(x) | B(x): ok(x)` on an enum `T`.

This validates the union path where the recursive argument is structurally smaller in every arm of the union.

johnynek:
Implemented the requested positive union regression test and verified locally:
- `scripts/test_basic.sh`
Both passed.

PR #2032, opened 2026-03-06
Fix issue #2031: quantified apply fast-path leaked unbound type vars

### LINKED ISSUE #2031: Unknown var in normal construction
With this code:
```

from Bosatsu/Lazy import (
  Lazy,
  lazy,
)

enum VLazyList[a: +*]:
  Empty
  Cons(head: Lazy[a], tail: Lazy[VLazyList[a]])
  Mapped[b](source: VLazyList[b], fn: b -> a)

struct LazyList[a: +*](bound: Int, list: VLazyList[a]) 

empty_LazyList: forall a. LazyList[a] = LazyList(0, Empty)
lazy_empty: forall a. Lazy[LazyList[a]] = lazy(() -> empty_LazyList)
```

I get:

```
in file: /Users/oscar/code/zafu/src/Zafu/Collection/LazyList.bosatsu:44:43, package Zafu/Collection/LazyList
unknown var in a: unbound var: Bound(a) at [713, 739)
42|
43|empty_LazyList: forall a. LazyList[a] = LazyList(0, Empty)
44|lazy_empty: forall a. Lazy[LazyList[a]] = lazy(() -> empty_LazyList)
                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^


in file: /Users/oscar/code/zafu/src/Zafu/Collection/LazyList.bosatsu:44:54, package Zafu/Collection/LazyList
unknown var in a: unbound var: Bound(a) at [724, 738)
42|
43|empty_LazyList: forall a. LazyList[a] = LazyList(0, Empty)
44|lazy_empty: forall a. Lazy[LazyList[a]] = lazy(() -> empty_LazyList)
                                                        ^^^^^^^^^^^^^^
```
1. this could should work I think.
2. we should *never* have unbound var showing up in a type error to users
3. even if we did, this message shouldn't use the `toString` "Bound(a)"` and it is duplicative "unknown var in a: unbound var: Bound(a)".

I wonder if this could be an issue with instantiation. Since `Lazy` is covariant, we should have constructed: `Lazy[forall a. LazyList[a]]` since `empty_LazyList` is annotated (and that type checks).

johnynek:
Resolved issue #2031 by changing `Infer.checkApplyDom` so its direct-application instantiation fast-path is skipped when the expected result type has outer quantifiers (`forall`/`exists`). In the repro, that fast-path was checking arguments against types containing out-of-scope bound vars, producing `UnknownKindOfVar` errors like `Bound(a)` for valid code. The fallback path now handles this case correctly. I also tightened the internal error wording to avoid exposing `Bound(a)` formatting (`"unknown type variable ...: unbound variable"`). Added `core/src/test/scala/dev/bosatsu/Issue2031Test.scala` with the issue repro (using `Bosatsu/Lazy`) as a regression test that now passes. Ran required pre-push command `scripts/test_basic.sh` successfully (all CLI/core tests passed).

Fixes #2031

### DIFF
--- a/core/src/main/scala/dev/bosatsu/rankn/Infer.scala
+++ b/core/src/main/scala/dev/bosatsu/rankn/Infer.scala
@@ -1791,40 +1791,47 @@ object Infer {
         tag: A,
         tpe: dom.TypeKind
     ): Infer[Option[dom.ExprKind[A]]] = {
-      val infOpt = maybeSimple(fn).flatTraverse { inferFnExpr =>
-        inferFnExpr.map { fnTe =>
-          fnTe.getType match {
-            case Type.Fun.SimpleUniversal(univ, inT, outT)
-                if inT.length == args.length =>
-              // see if we can instantiate the result type
-              // if we can, we use that to fix the known parameters and continue
-              Type
-                .instantiate(univ.iterator.toMap, outT, Map.empty, tpe, Map.empty)
-                .flatMap { instantiation =>
-                  // if instantiate works, we know outT => tpe
-                  if (instantiation.subs.nonEmpty && instantiation.frees.isEmpty) {
-                    // we made some progress and there are no frees
-                    // TODO: we could support frees it seems but
-                    // it triggers failures in tests now
-                    Some((fnTe, inT, instantiation))
-                  } else {
-                    // We learned nothing
-                    None
+      val hasOuterQuantifiers =
+        Type.forallList(tpe).nonEmpty || Type.existList(tpe).nonEmpty
+      if (hasOuterQuantifiers) {
+        // This fast-path instantiates universals from the function result (`outT`)
+        // against `tpe`, then checks arguments with the solved substitutions.
+        // For quantified `tpe`, instantiate can solve to terms mentioning those
+        // quantifiers, but this path does not introduce them into argument-checking
+        // scope. That can leak unbound vars in kinds (issue #2031). The fallback
+        // path (`None`) handles quantifiers by skolemizing before checking args.
+        pure(None)
+      } else {
+        val infOpt = maybeSimple(fn).flatTraverse { inferFnExpr =>
+          inferFnExpr.map { fnTe =>
+            fnTe.getType match {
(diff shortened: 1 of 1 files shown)


core/src/main/scala/dev/bosatsu/rankn/Infer.scala
```diff
@@ -1821,57 +1821,61 @@ object Infer {
...
-
-          val subIn = inT.map(Type.substituteVar(_, instNoKind))
-          val fnName = functionNameHint(fn)
-
-          validKinds.parProductR {
-            val remainingFree =
-              NonEmptyList.fromList(
-                instantiation.frees.iterator.map { case (_, (k, b)) => (b, k) }.toList
-              )
+          val hasOuterQuantifiers =
```
johnynek: can we move this up to the top of the method so that we can do this cheap check first, and not after we have done work?

But stepping back, we need to add a comment as to why
```
Type.instantiate(univ.iterator.toMap, outT, Map.empty, tpe, Map.empty)
```
succeeds, but then we can't use the result. A reader would naturally suspect that there is a bug lurking in instantiate or something else they don't understand.

johnynek: Done. I moved the quantified-type guard to the top of `checkApplyDom` so we do the cheap check before `maybeSimple`/`instantiate`. I also added a comment explaining why `Type.instantiate(univ.iterator.toMap, outT, Map.empty, tpe, Map.empty)` can succeed but the fast path is still unsafe: the resulting substitutions can mention `tpe` quantifiers, while this path does not introduce those quantifiers into argument-checking scope, which is what caused the issue #2031 unbound-kind-variable failure. In that case we now intentionally defer to the fallback (`pure(None)`), which skolemizes before checking args.

johnynek:
Applied the requested refactor/comment update in `core/src/main/scala/dev/bosatsu/rankn/Infer.scala` and re-ran the required pre-push check `scripts/test_basic.sh` successfully.

PR #2041, opened 2026-03-06
Fix tree-sitter highlighting for top-level `package` declarations

### LINKED ISSUE #2040: tree-sitter doesn't recognize "package" as a keyword at the top of the file.
The `package` file isn't highlighted using nvim with the tree-sitter grammar.

johnynek:
Implemented a direct fix for issue #2040 in the Bosatsu tree-sitter highlight query.

What changed:
- Updated `syntax/tree-sitter-bosatsu/queries/highlights.scm` to explicitly capture the package declaration token as a keyword:
  - `(package_declaration "package" @keyword)`
- Added regression test `core/src/test/scala/dev/bosatsu/TreeSitterHighlightQueryTest.scala` that verifies the highlight query contains this capture.

Reproduction and verification:
- Reproduced the issue behavior by adding the regression test and running it before the fix:
  - `sbt -batch "coreJVM/testOnly dev.bosatsu.TreeSitterHighlightQueryTest"` -> failed (missing capture).
- Applied fix and reran the same test:
  - `sbt -batch "coreJVM/testOnly dev.bosatsu.TreeSitterHighlightQueryTest"` -> passed.
- Ran required pre-push command:
  - `scripts/test_basic.sh` -> passed.

Fixes #2040


johnynek:
Reproduced the CI `testJS` failure locally with `sbt -batch "coreJS/testOnly * -- --log=failure; jsapiJS/compile"`. The failure was caused by `TreeSitterHighlightQueryTest` living in shared `core/src/test/scala` while using JVM-only `java.nio.file` APIs, which Scala.js cannot link. I fixed this by moving the test to JVM-only sources at `core/.jvm/src/test/scala/dev/bosatsu/TreeSitterHighlightQueryTest.scala` and removing `core/src/test/scala/dev/bosatsu/TreeSitterHighlightQueryTest.scala`. After the change, the same `coreJS` command passed locally, and the required pre-push check `scripts/test_basic.sh` also passed.

PR #2051, opened 2026-03-07
Optimize cmp_Int guard normalization for leading termination branches

### LINKED ISSUE #2044: Optimize int-recursion guard normalization for leading cmp_Int termination branches
## Summary
`TypedExprNormalization` currently leaves a common int-recursion guard shape in a nested boolean-selector form:

- `case _ if cmp_Int(x, 0) matches LT | EQ: stop`
- `case _: recur`

This lowers to an extra boolean match in `lib show`, and then to extra enum allocation/inspection in generated C.

## Evidence
Using `lib show` on local `core_alpha`:

```bash
./bosatsuj lib show --name core_alpha --repo_root /Users/oscar/code/bosatsu \
  --value Bosatsu/Num/Nat::to_Nat --value Bosatsu/Num/BinNat::to_BinNat --color none
```

`Bosatsu/Num/Nat::to_Nat` currently shows:

- outer branch `wild` with guard `match cmp_Int(_a, 0)` -> `(LT | EQ) => True`, `_ => False`
- second fallback branch `wild` with `recur`

Generated C (`output.c`) shows the extra bool materialization each loop iteration:

- [output.c](/var/folders/7g/n61rp3sx169df8mc5vrqd1400000gn/T/bsts-c-XXXXXXXX.npjAVqbd11/output.c:502)
- [output.c](/var/folders/7g/n61rp3sx169df8mc5vrqd1400000gn/T/bsts-c-XXXXXXXX.npjAVqbd11/output.c:507)

## Proposal
Add a normalization rewrite for this leading two-branch pattern:

1. Detect `Match(arg, [Branch(p1, Some(g), e1), Branch(p2, None, e2)])`
2. If `p1/p2` are compatible wildcard-like patterns and `g` is a boolean-selector match over `cmp_Int(...)` (especially `LT|EQ` or `GT` style guards), rewrite to a direct comparison-branching form that does not first construct an intermediate Bool ADT.
3. Keep branch ordering semantics identical.

Potential implementation area:
- [TypedExprNormalization.scala](/Users/oscar/code/bosatsu/core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala:1542)
- [TypedExprNormalization.scala](/Users/oscar/code/bosatsu/core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala:429)

## Why this is high impact / low complexity
- This pattern is common in new Int-recursive code paths.
- It is structurally localized and deterministic to detect.
- It should improve both code shape and runtime cost, especially in C loops.

## Acceptance Criteria
- `lib show` for `Bosatsu/Num/Nat::to_Nat` no longer contains nested `cmp_Int` -> Bool selector match in the loop guard path.
- Generated C for `to_Nat` no longer creates a temporary bool enum from comparison result on each iteration.
- Existing tests for recursion and match semantics continue to pass.

johnynek:
Implemented a focused normalization change for issue #2044.

What changed:
- Updated `TypedExprNormalization` to better flatten bool-selector matches by stripping type wrappers on the match argument before flattening.
- Added targeted detection for `cmp_Int` boolean-selector guards (`isCmpIntCall` + `isCmpIntBoolSelector`).
- Extended the existing guarded trailing-pair rewrite so it also rewrites the two-branch leading termination shape when:
  - the guarded branch is first (`init.isEmpty`),
  - the fallback branch contains `recur`, and
  - the guard is a `cmp_Int` bool-selector match.
  This enables normalization to collapse the nested `cmp_Int -> Bool` guard shape into a direct comparison match form.

Tests added:
- New regression in `TypedExprTest`: `normalization rewrites wildcard guarded termination branches into a single match`.

Validation run:
- `sbt "coreJVM/testOnly dev.bosatsu.TypedExprTest"`
- `scripts/test_basic.sh` (required command)

Fixes #2044

### DIFF
--- a/docs/design/2049-unknown-names-come-out-one-at-a-time.md
+++ b/docs/design/2049-unknown-names-come-out-one-at-a-time.md
@@ -1,0 +1,197 @@
+---
+issue: 2049
+priority: 2
+touch_paths:
+  - docs/design/2049-unknown-names-come-out-one-at-a-time.md
+  - core/src/main/scala/dev/bosatsu/Package.scala
+  - core/src/main/scala/dev/bosatsu/rankn/NameCheck.scala
+  - core/src/main/scala/dev/bosatsu/PackageCustoms.scala
+  - core/src/main/scala/dev/bosatsu/PackageError.scala
+  - core/src/test/scala/dev/bosatsu/ErrorMessageTest.scala
+  - core/src/test/scala/dev/bosatsu/rankn/NameCheckTest.scala
+depends_on: []
+estimated_size: M
+generated_at: 2026-03-07T00:00:00Z
+---
+
+# Issue #2049 Design: unknown names come out one at a time
+
+_Issue: #2049 (https://github.com/johnynek/bosatsu/issues/2049)_
+
+Status: proposed  
+Base branch: `main`
+
+## Summary
+
+Add a pre-inference Expr-DAG diagnostic stage that runs before `Infer.typeCheckLets` and returns an `Ior`:
+
+1. Left: all value-name errors found across lets (`VarNotInScope`) plus partial import/export bindable errors.
+2. Right: the maximal name-safe subdag of lets that can still be typechecked for additional diagnostics.
+
+This addresses one-at-a-time unknown names and also allows import/export bindable checks to surface even when typechecking fails.
+
+## Problem statement
+
+Unknown names are currently discovered mostly in type inference (`Infer.lookupVarType`), which means compilation usually short-circuits before reporting more than one root issue in a package.
+
+The same sequencing problem affects import/export diagnostics that depend on typed lets:
+
+1. export validation currently runs from `PackageCustoms.assemble` using typed lets,
+2. import usage checks run in `PackageCustoms` over typed expressions,
+3. if typechecking fails early, these checks do not run, even when part of them is checkable from the untyped Expr DAG.
+
+Result: users fix one unknown, rerun, then hit additional unknowns or import/export bindable issues in multiple cycles.
+
+## Goals
+
+1. Report all unresolved value names for a package in one run.
+2. Report export/import bindable issues that can be computed from Expr DAG, even if typechecking fails.
+3. Preserve current unknown-name formatting and suggestion behavior.
+4. Keep typechecking on maximal unaffected lets to expose independent type errors in the same run.
+5. Preserve semantics and outputs for successful packages.
+
+## Non-goals
+
+1. Rewriting inference internals from monadic short-circuiting to full validation accumulation.
+2. Reordering declarations in a way that changes use-before-def behavior.
+3. Replacing typed import/export checks that require inferred types (those stay post-typecheck).
+4. Producing a fully typed `Program` when any top-level let in that package has unresolved names.
+
+## Proposed architecture
+
+### 1) Add `NameCheck` over untyped lets
+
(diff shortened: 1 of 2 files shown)


core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
```diff
@@ -1545,8 +1568,13 @@ object TypedExprNormalization {
           bs.toList match {
             case init :+ Branch(p1, Some(g), e1) :+ Branch(p2, None, e2)
                 if totalityCheck.difference(p2, p1).isEmpty &&
```
johnynek, line 1501: we may have another rule that rewrites final matches that don't have any names to `Wildcard`. Review and see if that's true. We shouldn't have that since it probably doesn't unlock any new optimizations here. That's something we can (and I think do) handle currently in Matchless when compiling the patterns. That's where we should leverage branch totality and elide checks that must be true.

core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
```diff
@@ -1545,8 +1568,13 @@ object TypedExprNormalization {
...
-                  p1.names.isEmpty && p2.names.isEmpty &&
-                  containsRecur(e1) =>
+                  p1.names.isEmpty && p2.names.isEmpty && {
+                    containsRecur(e1) || {
```
johnynek: why do we care if it contains recur? Why isn't this optimization always good? Same with the containsRecur check below

core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
```diff
@@ -1545,8 +1568,13 @@ object TypedExprNormalization {
...
+                      init.isEmpty &&
+                      containsRecur(e2) &&
+                      isCmpIntBoolSelector(g)
```
johnynek: why not just do this optimization with or without recur and for selector?

core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
johnynek: Confirmed and fixed. I removed the normalization block that rewrote the final no-binding unguarded branch pattern to `Wildcard` in `TypedExprNormalization`. We now leave that optimization to downstream pattern compilation (Matchless), as suggested.

core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
johnynek: Agreed. I removed the `containsRecur` gating entirely. The trailing guarded-pair rewrite is now based only on compatibility/totality and no-bindings checks; it no longer depends on whether either branch contains `recur`.

core/src/main/scala/dev/bosatsu/TypedExprNormalization.scala
johnynek: Done. The optimization is now general (with or without `recur`, with or without selector-specific shape). I also removed the `cmp_Int`-specific helper checks that were only supporting the narrower version.

johnynek:
Implemented all requested generalizations and cleanup in normalization. Updated affected tests for the new normalized shape, and updated the deterministic Clang golden hash for `Ackermann` (`ClangGenTest`) because generated C changed from the optimization. Required pre-push check passed: `scripts/test_basic.sh`.

PR #2050, opened 2026-03-07
Design doc for #2049: Issue #2049 Design: unknown names come out one at a time

johnynek:
Design doc.

Refs #2049

PR #2052, opened 2026-03-07
Add pre-inference NameCheck and Expr-DAG bindable diagnostics

### LINKED ISSUE #2049: unknown names come out one at a time
E.g.
```
in file: /Users/oscar/code/zafu/src/Zafu/Collection/LazyList.bosatsu:258:10, package Zafu/Collection/LazyList
Unknown name `cons`.
Did you mean one of: local value `cons_prop`, local value `cons_take`,
local value `uncons`?
256|  ((head, tail_list)) -> (
257|    size = len_List(tail_list)
258|    ll = cons(size.add(1), lazy(() -> head), from_List(tail_list))
```

So, when we have an unknown name due to a refactor, we see just a single name each time. I think this is happening because we check names in typechecking.

An alternate idea would be to have a NameCheck phase that takes the Expr dag for a package, and returns an Ior. The left would be all the name errors we found and the right would be the smallest correct subdag we can make (maybe even topologically sorting the file) by expression dependency graph of names.

This way, we get all the name errors in one go, and then also maybe get type errors that we can compute at that time rather than paying out name error after name error one at a time.

johnynek:
Implemented issue #2049 per the merged design by adding `core/src/main/scala/dev/bosatsu/rankn/NameCheck.scala` and integrating it into package inference.

What changed:
- Added `NameCheck.checkLets` returning `Ior[NonEmptyChain[Infer.Error.NameError], NameCheck.Result]`.
- NameCheck now:
  - walks untyped lets with lexical scoping for lambdas/local lets/match patterns,
  - accumulates all unresolved value names (`VarNotInScope`) in one pass,
  - applies anti-cascade top-level scope behavior,
  - computes same-package let dependencies and transitively blocks dependents,
  - returns source-ordered surviving `typecheckLets`.
- Updated `Package.inferBodyUnopt`:
  - added overload taking `exports` for pre-inference diagnostics,
  - runs NameCheck and `PackageCustoms.checkExprDagBindables` before inference,
  - typechecks only `nameCheckResult.typecheckLets`,
  - aggregates name errors, Expr-DAG bindable diagnostics, and inference diagnostics,
  - preserves Left behavior when name errors exist while still collecting independent diagnostics.
- Added untyped bindable precheck in `PackageCustoms.checkExprDagBindables`:
  - reports unknown exported `ExportedName.Binding` names from lets/externals,
  - reports unused imported bindables detectable from Expr globals even when typecheck fails.
- Adapted `PackageError.UnknownExport` to take binder-region candidates so rendering can be reused by both precheck and typed export validation.
- Updated `PackageCustoms.assemble` to use new `UnknownExport` shape.
- Updated `PackageMap` callsite to pass exports into new `inferBodyUnopt` overload.

Tests added/updated:
- Added `core/src/test/scala/dev/bosatsu/rankn/NameCheckTest.scala` covering lexical scope, recursive local lets, dependency blocking, and source-order stability.
- Extended `core/src/test/scala/dev/bosatsu/ErrorMessageTest.scala` with regressions for:
  - multiple distinct unknown names in one run,
  - unknown export reported even when name checking fails elsewhere,
  - unused imported bindable reported even when name checking fails elsewhere,
  - independent type errors appearing alongside name errors.
- Adjusted one existing assertion to target `TypeErrorIn` specifically now that additional precheck diagnostics can co-occur.

Validation run:
- `sbt "coreJVM/test:compile"` passed.
- `sbt "coreJVM/testOnly dev.bosatsu.rankn.NameCheckTest dev.bosatsu.ErrorMessageTest"` passed.

Fixes #2049

Implements design doc: [docs/design/2049-unknown-names-come-out-one-at-a-time.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/2049-unknown-names-come-out-one-at-a-time.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/2050

### DIFF
--- a/core/src/main/scala/dev/bosatsu/Package.scala
+++ b/core/src/main/scala/dev/bosatsu/Package.scala
@@ -541,6 +541,10 @@ object Package {
                   combineInferSingles(
                     nameErrs.toNonEmptyList.map(err => err: Infer.Error.Single)
                   )
+                // Intentionally return Left (not Both): unresolved top-level names
+                // mean we cannot safely expose a partial typed package API to
+                // downstream packages. We still run inference on surviving lets
+                // to accumulate additional diagnostics in this package.
                 Ior.left(
                   NonEmptyList.one[PackageError](
                     PackageError.TypeErrorIn(


core/src/main/scala/dev/bosatsu/Package.scala
```diff
@@ -411,6 +411,13 @@ object Package {
...
+  private def combineInferSingles(
+      errs: NonEmptyList[Infer.Error.Single]
+  ): Infer.Error =
+    errs.tail.foldLeft(errs.head: Infer.Error) { (acc, next) =>
+      Infer.Error.Combine(acc, next)
+    }
+
   /** Infer the types but do not optimize/normalize the lets
     */
   def inferBodyUnopt(
```
johnynek: can we just remove this and always call the new one? Leaving it here seems like a way to get people confused.

core/src/main/scala/dev/bosatsu/Package.scala
```diff
@@ -611,8 +680,14 @@ object Package {
...
 
-          Parallel[[A] =>> Ior[NonEmptyList[PackageError], A]]
-            .parProductR(checkUnusedLets.toIor)(inference.toIor)
+          val typeAndUnusedChecks =
+            Parallel[[A] =>> Ior[NonEmptyList[PackageError], A]]
+              .parProductR(checkUnusedLets.toIor)(inference.toIor)
+
+          // Name errors force a Left result, but we still collect any
+          // independent inference and import/export diagnostics in the same run.
+          (exprDagBindableChecks, nameCheckErrors, typeAndUnusedChecks)
```
johnynek: can't we just put `checkUnusedLets.toIor, inference.toIor` in this list and not have the item above?

core/src/main/scala/dev/bosatsu/Package.scala
```diff
@@ -520,6 +546,49 @@ object Package {
...
 
+          val nameCheck =
+            NameCheck.checkLets(p, lets, withFQN)
+          val nameCheckResult =
+            nameCheck match {
+              case Ior.Right(res)   => res
+              case Ior.Both(_, res) => res
+              case Ior.Left(_)      =>
+                NameCheck.Result(
+                  lets,
```
johnynek: can you comment why we won't get errors twice in this case? If we pass the lets to Infer but there are a lot of errors, won't we see the errors there too.

core/src/main/scala/dev/bosatsu/Package.scala
johnynek: Done. I removed the old 3-arg `inferBodyUnopt` overload and kept only the explicit `(p, imps, exports, stmts)` version. I also updated call sites to pass `Nil` exports where appropriate (`inferBody`, `TestUtils`, and `TypedExprTest`) so there is a single entry point now.

core/src/main/scala/dev/bosatsu/Package.scala
johnynek: Yep, updated as suggested. I removed the intermediate `typeAndUnusedChecks` value and now pass `checkUnusedLets.toIor` and `inference` directly in the final `parMapN` tuple with the other diagnostics.

core/src/main/scala/dev/bosatsu/Package.scala
johnynek: Added a clarifying comment right above `Infer.typeCheckLets(...)` in `Package.scala`. It now explicitly states that we typecheck only `nameCheckResult.typecheckLets` (blocked lets are excluded), so missing-name diagnostics come from NameCheck once rather than being duplicated by Infer.

johnynek:
Addressed all three review comments with code changes in `Package.scala` plus required call-site updates in tests for the signature cleanup. Re-ran the required pre-push check `scripts/test_basic.sh` and it passed.

core/src/main/scala/dev/bosatsu/rankn/NameCheck.scala
```diff
@@ -0,0 +1,198 @@
...
+      typecheckLets = lets.filterNot { case (name, _, _) => blockedLets(name) },
+      nameErrorLets = nameErrorLets,
+      blockedLets = blockedLets,
+      samePackageDeps = samePackageDeps
+    )
+
+    val allErrors =
+      checkedLets.foldMap(_.errors)
+
+    NonEmptyChain.fromChain(allErrors) match {
```
johnynek: If we can only return Both or Right then let's return (Option[Errors], Result) so it's clear that the Left case never happens. It causes confusion to see the left handling when it can never be returned.

johnynek: Implemented. `NameCheck.checkLets` now returns `(Option[NonEmptyChain[Infer.Error.NameError]], Result[A])` instead of `Ior`, so the impossible `Left` case is no longer represented. I updated the caller in `Package.scala` to destructure the tuple directly and removed the fallback branch, and updated `NameCheckTest` assertions to match the new return type. I also removed the unused `Ior` import from `NameCheck.scala`/test.

johnynek:
Addressed the remaining review feedback by simplifying the `NameCheck` API and corresponding call/test code. Re-ran required pre-push checks: `scripts/test_basic.sh` passed.

core/src/main/scala/dev/bosatsu/Package.scala
```diff
@@ -520,6 +529,37 @@ object Package {
...
+          val nameCheckErrors: Ior[NonEmptyList[PackageError], Unit] =
+            nameCheckErrorOpt match {
+              case None           =>
+                Ior.right(())
+              case Some(nameErrs) =>
+                val mergedError =
+                  combineInferSingles(
+                    nameErrs.toNonEmptyList.map(err => err: Infer.Error.Single)
+                  )
+                Ior.left(
```
johnynek, line 548: Why left here and not both with a unit? Left is going to halt further compilation.

Add a comment.

I think if we can't typecheck all the lets we shouldn't return a partially checked package because that would give confusing downstream errors. We should only return a Both if the errors are "linting errors" like everything checked but maybe there were some unused names, or if the exports mention names we don't have, we could just remove those names and return a Both.

But if the name exists but we don't typecheck it then a downstream consumer fails to compile because we removed it, that's a bad experience.

johnynek: Added an in-code comment at the `nameCheckErrors` `Ior.left(...)` site in `Package.scala` explaining why this must be `Left` rather than `Both`: unresolved top-level names must not produce a partial typed package API for downstream consumers, even though we still infer surviving lets to accumulate additional diagnostics in the same run.

johnynek:
Addressed the review feedback with a targeted comment clarifying `Left` semantics for unresolved names. Re-ran required pre-push checks: `scripts/test_basic.sh` passed.

PR #2058, opened 2026-03-07
Fix markdown type-link bounds check for doc signatures

### LINKED ISSUE #2057: Add links in doc markdown
We should add links in the markdown from one package to the other and to imported types. 

Also in type signatures in the document we should be able to click on types and go directly to the md and anchor they are defined at.

johnynek:
Repaired the #2057 doc-linking path by fixing `MarkdownDoc.linkifyRenderedType` so link boundary checks do not call `charAt` past the rendered string end (`end >= rendered.length`). This prevents the `StringIndexOutOfBoundsException` seen when rendering external signatures (including `Unit -> a` shown as `() -> a`) and keeps type-signature link generation working for local/imported types across package docs. Ran `scripts/test_basic.sh` successfully (pass).

Fixes #2057

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -95,6 +95,23 @@ lazy val docs = (project in file("docs"))
           |---
           |
           |""".stripMargin
+      def rewriteMarkdownLinksToHtml(content: String): String = {
+        val linkPattern = raw"\]\(([^)]+)\)".r
+        linkPattern.replaceAllIn(content, m => {
+          val target = m.group(1)
+          val hashIdx = target.indexOf('#')
+          val (path, suffix) =
+            if (hashIdx >= 0) {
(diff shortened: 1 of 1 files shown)


johnynek:
Reproduced the CI failure locally with `sbt "docs/Compile/paradox"` and confirmed Paradox was rejecting generated internal `.md` links (including new type-signature and dependency links). Updated `build.sbt` in `generateCoreAlphaParadoxDocs` to rewrite internal markdown link targets from `.md` to `.html` only in the Paradox copy step, so CLI-generated markdown output semantics stay unchanged. Re-ran `sbt "docs/Compile/paradox"` successfully and then ran the required `scripts/test_basic.sh` successfully.

Issue #2059, opened 2026-03-07
Add a SwitchVariant expression to Matchless


johnynek:
Sometimes a match is just a switch on a variant. Currently we will compile this in o If/Else because that's all we have but for an enum with many branches say 8 or more, the switch could be better on C when we have an orthogonal match.

How would this look? Would we always use it when we can and let backends decide how to compile the switch or use heuristics in Matchless so we only choose it when there are say 4 or more branches?

How would we compile it on Python where there is no switch? Perhaps a binary if/else tree?

johnynek:
MergeXO assigned an agent and started design work for issue #2059.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/2061

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #2059.

johnynek:
Opened implementation PR: https://github.com/johnynek/bosatsu/pull/2068

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/2059#issuecomment-4017897828
- linked PR: #2068 (https://github.com/johnynek/bosatsu/pull/2068)

Comments on the source issue are no longer actioned after a PR exists. Please comment on the PR thread instead.

PR #2061, opened 2026-03-07
Design doc for #2059: Issue #2059 Design: Add a SwitchVariant expression to Matchless

johnynek:
Design doc.

Refs #2059

### DIFF
--- a/docs/design/2059-add-a-switchvariant-expression-to-matchless.md
+++ b/docs/design/2059-add-a-switchvariant-expression-to-matchless.md
@@ -1,35 +1,5 @@
 ---
 issue: 2059
-priority: 3
-touch_paths:
-  - docs/design/2059-add-a-switchvariant-expression-to-matchless.md
-  - core/src/main/scala/dev/bosatsu/Matchless.scala
-  - core/src/main/scala/dev/bosatsu/MatchlessToValue.scala
-  - core/src/main/scala/dev/bosatsu/codegen/clang/Code.scala
-  - core/src/main/scala/dev/bosatsu/codegen/clang/ClangGen.scala
-  - core/src/main/scala/dev/bosatsu/codegen/python/PythonGen.scala
-  - core/src/test/scala/dev/bosatsu/MatchlessTests.scala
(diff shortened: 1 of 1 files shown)


docs/design/2059-add-a-switchvariant-expression-to-matchless.md
```diff
@@ -0,0 +1,274 @@
...
+4. `default` is always present.
+
+Semantics:
+
+1. Evaluate `on` once.
+2. Read enum variant tag.
+3. Evaluate the matching case expression if present.
+4. Otherwise evaluate `default`.
+
+### 2) Add a `switchVariantize` pass in Matchless
```
johnynek: Explain why do this as a pass vs directly lowering the appropriate match segment into the switch.

Like it seems if we do it as a pass on if/else we have a hard time undoing operations which may exist to optimize unions:

Like

match foo:
    case A: exprA 
    case B | C: exprBC 
    case D: ...


When we are inspecting the match, checking pattern orthogonality in an enum is pretty easy. Doing it from the lowered If/Else seems much more error prone.

johnynek: Updated the design doc to switch from a post-pass approach to direct matrix-time lowering.

Concretely:
1. Replaced the `switchVariantize` pass proposal with direct emission in `matchExprMatrixCheap`.
2. Added a dedicated section, `Why direct matrix-time lowering instead of an If post-pass`, explaining orthogonality and union-case handling (`B | C`) and why this is safer/less error-prone.
3. Updated implementation plan, testing plan, acceptance criteria, and risks to match the direct-lowering design.
4. Added explicit union-shape test coverage in the testing plan.

This should address the concern about recovering structure from already-lowered `If` trees.

johnynek:
Addressed the PR review feedback by updating `docs/design/2059-add-a-switchvariant-expression-to-matchless.md` to a direct matrix-lowering architecture and cleaning duplicate front matter in the doc. Ran required pre-push checks: `scripts/test_basic.sh` passed (`Passed: Total 62, Failed 0, Errors 0`).

PR #2065, opened 2026-03-07
Design doc for #2047: Issue #2047 Design Doc

johnynek:
Design doc.

Refs #2047

PR #2067, opened 2026-03-07
Matchless matrix: score columns instead of fixed-leftmost selection

### LINKED ISSUE #2047: Matchless matrix: replace fixed leftmost column choice with scoring heuristic
## Summary
Matrix compilation currently hardcodes leftmost column selection (`colIdx = 0`). This can produce suboptimal decision trees and code shape in multi-column matches.

A cheap scoring heuristic should improve generated code quality, especially for C.

## Evidence
Current hardcoded selection:
- [Matchless.scala](/Users/oscar/code/bosatsu/core/src/main/scala/dev/bosatsu/Matchless.scala:4097)
- fallback multiplicity estimator also mirrors leftmost: [Matchless.scala](/Users/oscar/code/bosatsu/core/src/main/scala/dev/bosatsu/Matchless.scala:4455)

## Proposal
Replace fixed `colIdx = 0` with a low-cost scoring heuristic, e.g.:

- prefer columns with more distinct refutable head signatures,
- penalize high-arity expansions that increase projections,
- prefer columns that eliminate more rows quickly,
- tie-break by leftmost to preserve predictability.

Apply the same chooser in both real compile and fallback multiplicity counting for consistency.

## Scope
- Keep heuristic deterministic and cheap (no exponential lookahead).
- Avoid semantic changes; only decision-tree shape should change.

## Acceptance Criteria
- Column selection is no longer fixed-leftmost.
- Existing match tests pass.
- At least one representative multi-column case shows reduced checks/projections in generated C.

johnynek:
Implemented issue #2047 per the merged design doc.

What changed:
- Added shared matrix-column scoring helpers in `Matchless.scala`:
  - `sigArity(sig: HeadSig): Int`
  - `ColumnScore`
  - `chooseColumnByScore(rows: List[MatchRow]): (Int, List[HeadSig])`
- The chooser is deterministic and cheap, with comparator order:
  1. maximize distinct refutable signatures
  2. maximize refutable rows
  3. minimize arity penalty
  4. minimize column index (leftmost tie-break)
- Replaced hardcoded `colIdx = 0` in both:
  - `matchExprMatrixCheap.compileRows`
  - `matrixFallbackMultiplicity.compileRowsCount`
- Reused chooser-provided `sigs` ordering in both paths for consistency.
- Updated stale comment text that described fixed-leftmost behavior.

Tests added:
- `core/src/test/scala/dev/bosatsu/MatchlessTests.scala`
  - Added a focused multi-column regression asserting wide struct projections are not eagerly materialized before the first decision, while required fields are still projected.
- `core/src/test/scala/dev/bosatsu/codegen/clang/ClangGenTest.scala`
  - Added a representative generated-C shape test asserting the first wide projection (`get_struct_index(..., 2)`) appears after the first `if` decision.

Validation:
- Ran required pre-push command: `scripts/test_basic.sh`
- Result: passed (`cli` + `coreJVM` test suites green).

Fixes #2047

Implements design doc: [docs/design/2047-matchless-matrix-replace-fixed-leftmost-column-choice-with-scoring-heuristic.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/2047-matchless-matrix-replace-fixed-leftmost-column-choice-with-scoring-heuristic.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/2065

### DIFF
--- a/core/src/main/scala/dev/bosatsu/Matchless.scala
+++ b/core/src/main/scala/dev/bosatsu/Matchless.scala
@@ -3591,6 +3591,14 @@ object Matchless {
       case ZeroSig extends HeadSig
       case SuccSig extends HeadSig
       case LitSig(lit: Lit)
+
+      def projectionArity: Int =
+        this match {
+          case EnumSig(_, _, arity, _) => arity
+          case StructSig(_, arity)     => arity
+          case SuccSig                 => 1
+          case LitSig(_) | ZeroSig     => 0
+        }
     }
     import HeadSig.*
 
@@ -3819,14 +3827,6 @@ object Matchless {
       bldr.toList
     }
 
-    def sigArity(sig: HeadSig): Int =
-      sig match {
-        case EnumSig(_, _, arity, _) => arity
-        case StructSig(_, arity)     => arity
-        case SuccSig                 => 1
-        case LitSig(_) | ZeroSig     => 0
-      }
-
     case class ColumnScore(
         colIdx: Int,
         distinctSigs: Int,
@@ -3855,10 +3855,12 @@ object Matchless {
       else {
         def scoreCol(colIdx: Int): ColumnScore = {
           val sigs = distinctInOrder(rows.mapFilter(r => headSig(r.pats(colIdx))))
+          // normalizeRow + peelPattern remove Var/Named/Annotation, so at this
+          // stage non-refutable patterns are exactly WildCard.
           val refutableRows = rows.count(r => r.pats(colIdx) != Pattern.WildCard)
           // We split once per distinct signature, so sum those arities as a
           // cheap proxy for added projection pressure in this column.
-          val arityPenalty = sigs.iterator.map(sigArity).sum
+          val arityPenalty = sigs.iterator.map(_.projectionArity).sum
           ColumnScore(
             colIdx = colIdx,
             distinctSigs = sigs.length,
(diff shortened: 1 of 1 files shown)


core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -3819,6 +3819,76 @@ object Matchless {
       bldr.toList
     }
 
+    def sigArity(sig: HeadSig): Int =
```
johnynek: make this a method on `HeadSig`

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -3819,6 +3819,76 @@ object Matchless {
...
+      // $COVERAGE-OFF$
+      if (colCount == 0)
+        throw new IllegalStateException(
+          "chooseColumnByScore called with no remaining columns"
+        )
+      // $COVERAGE-ON$
+      else {
+        def scoreCol(colIdx: Int): ColumnScore = {
+          val sigs = distinctInOrder(rows.mapFilter(r => headSig(r.pats(colIdx))))
+          val refutableRows = rows.count(r => r.pats(colIdx) != Pattern.WildCard)
```
johnynek, line 3860: add a comment here. It seems that `Var(_)` is also not refutable, so why is this just `Pattern.WildCard`? Have vars already been removed? What invariants are true of `pats`?

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -3819,6 +3819,76 @@ object Matchless {
...
+          ColumnScore(
+            colIdx = colIdx,
+            distinctSigs = sigs.length,
+            refutableRows = refutableRows,
+            arityPenalty = arityPenalty,
+            sigs = sigs
+          )
+        }
+
+        def better(next: ColumnScore, best: ColumnScore): Boolean =
```
johnynek: in this rather than doing `!=` then `>` do:
```scala
{
  import java.lang.Integer.compare

  val cmpDs = compare(next.distinctSigs, best.distinctSigs)
  (cmpDs > 0) || {
     val cmpRR = compare(next.refutableRows, best.refutableRows)
      (cmpRR > 0) || { 
          ..
       }
   }
}
```

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Done. I removed the standalone `sigArity` helper and moved that logic onto `HeadSig` as `def projectionArity`. The scoring code now uses `sigs.iterator.map(_.projectionArity).sum`.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Added a clarifying comment at the `refutableRows` computation. At this point rows have already gone through `normalizeRow + peelPattern`, so `Var/Named/Annotation` have been eliminated; the only non-refutable pattern that remains is `Pattern.WildCard`.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Updated `better` to use `java.lang.Integer.compare`-based staged comparison (distinct sigs, then refutable rows, then arity penalty, then col index tie-break), per your suggested style.

johnynek:
Addressed all three line comments in `Matchless.scala` and reran the required pre-push check.

Validation run:
- `scripts/test_basic.sh` passed (CLI tests and coreJVM tests).

PR #2068, opened 2026-03-08
Add Matchless SwitchVariant IR and backend lowering for wide enum dispatch

### LINKED ISSUE #2059: Add a SwitchVariant expression to Matchless

johnynek:
Implemented issue #2059 per the design doc.

What changed:
- Added `Matchless.SwitchVariant(on, famArities, cases, default)` with constructor invariants (distinct/in-range variants, >=2 cases) in `core/src/main/scala/dev/bosatsu/Matchless.scala`.
- Updated exhaustive Matchless traversals/utilities to handle `SwitchVariant` (ordering/tags, side-effect checks, name/reference analysis, `applyArgs`, lambda arity recovery, substitution, optimization traversals, loop-tail rewriting helpers).
- Added matrix-time switch emission in `matchExprMatrixCheap` with `SwitchVariantMinCases = 4` and eligibility checks:
  - selected column signatures are enum-only,
  - shared `famArities`,
  - fanout >= 4.
- Kept non-eligible paths on existing ordered/if lowering.
- Extended selector guard recovery to understand `SwitchVariant` shape for boolified selector guards.
- Added evaluator support in `core/src/main/scala/dev/bosatsu/MatchlessToValue.scala` (evaluate scrutinee once, dispatch by variant, fallback to default).
- Added C AST support in `core/src/main/scala/dev/bosatsu/codegen/clang/Code.scala`:
  - new `Code.Switch` statement,
  - new `Code.Break` statement,
  - renderer support.
- Added Clang lowering in `core/src/main/scala/dev/bosatsu/codegen/clang/ClangGen.scala`:
  - compute variant tag once (`get_variant_value` for all-zero fams, else `get_variant`),
  - emit switch cases/default assigning one result temp.
- Added Python lowering in `core/src/main/scala/dev/bosatsu/codegen/python/PythonGen.scala`:
  - compute/caches variant tag once,
  - emit if/elif dispatch assigning one result temp,
  - return that result.

Tests updated:
- `MatchlessTests.scala`: recursion helpers updated for new node; added switch formation tests (>=4 emits switch, union-case behavior preserved, below-threshold stays non-switch), and guard-coverage compatibility for switch-shaped selectors.
- `MatchlessRegressionTest.scala`: added evaluator test for `SwitchVariant` case/default behavior.
- `Issue1633Test.scala`: traversal helper updated for `SwitchVariant`.
- `codegen/clang/CodeTest.scala`: switch/break rendering test.
- `codegen/clang/ClangGenTest.scala`: wide-enum match emits C `switch`/`case`/`default`.
- `codegen/python/PythonGenTest.scala`: switch lowering caches tag once and reuses in comparisons.

Validation run:
- `sbt "coreJVM/testOnly dev.bosatsu.MatchlessTest"`
- `sbt "coreJVM/testOnly dev.bosatsu.MatchlessRegressionTest dev.bosatsu.Issue1633Test dev.bosatsu.MatchlessApplyArgsTest dev.bosatsu.codegen.clang.CodeTest dev.bosatsu.codegen.clang.ClangGenTest dev.bosatsu.codegen.python.PythonGenTest"`
- Required pre-push: `scripts/test_basic.sh` (passed)

Fixes #2059

Implements design doc: [docs/design/2059-add-a-switchvariant-expression-to-matchless.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/2059-add-a-switchvariant-expression-to-matchless.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/2061

### DIFF
--- a/core/src/main/scala/dev/bosatsu/Matchless.scala
+++ b/core/src/main/scala/dev/bosatsu/Matchless.scala
@@ -140,7 +140,7 @@ object Matchless {
                 val c3 =
                   Order[NonEmptyList[(Int, Expr[A])]].compare(casesL, casesR)
                 if (c3 != 0) c3
-                else compareExpr(defaultL, defaultR)
+                else Order[Option[Expr[A]]].compare(defaultL, defaultR)
               }
             }
 
@@ -230,7 +230,7 @@ object Matchless {
           case SwitchVariant(on, _, cases, default) =>
             loopExpr(on) || cases.exists { case (_, branch) =>
               loopExpr(branch)
-            } || loopExpr(default)
+            } || default.exists(loopExpr)
           case Always(cond, thenExpr) =>
             loopBool(cond) || loopExpr(thenExpr)
           case PrevNat(of) =>
@@ -290,7 +290,7 @@ object Matchless {
           case SwitchVariant(on, _, cases, default) =>
             loopExpr(on) || cases.exists { case (_, branch) =>
               loopExpr(branch)
-            } || loopExpr(default)
+            } || default.exists(loopExpr)
           case Always(cond, thenExpr) =>
             loopBool(cond) || loopExpr(thenExpr)
           case PrevNat(of) =>
@@ -377,7 +377,7 @@ object Matchless {
               cases.exists { case (_, branch) =>
                 referencesBindable(branch, target, isShadowed)
               } ||
-              referencesBindable(default, target, isShadowed)
+              default.exists(referencesBindable(_, target, isShadowed))
           case Always(cond, thenExpr) =>
             BoolExpr.referencesBindable(cond, target, isShadowed) ||
               referencesBindable(thenExpr, target, isShadowed)
@@ -425,7 +425,7 @@ object Matchless {
             case SwitchVariant(on, _, cases, default) =>
               loopExpr(on, isShadowed) || cases.exists {
                 case (_, branch) => loopExpr(branch, isShadowed)
-              } || loopExpr(default, isShadowed)
+              } || default.exists(loopExpr(_, isShadowed))
             case Always(cond, thenExpr) =>
               loopBool(cond, isShadowed) || loopExpr(thenExpr, isShadowed)
             case PrevNat(of) =>
@@ -611,7 +611,7 @@ object Matchless {
             cases.map { case (variant, branch) =>
               (variant, loopExpr(branch))
             },
-            loopExpr(default)
+            default.map(loopExpr)
           )
         case Always(cond, thenExpr) =>
           Always(loopBool(cond), loopExpr(thenExpr))
@@ -753,7 +753,7 @@ object Matchless {
             cases.map { case (variant, branch) =>
               (variant, loop(branch, aliases))
             },
-            loop(default, aliases)
+            default.map(loop(_, aliases))
           )
         case Always(cond, thenExpr) =>
           Always(cond, loop(thenExpr, aliases))
@@ -815,7 +815,9 @@ object Matchless {
         case If(cond, thenExpr, elseExpr) =>
           loopExpr(elseExpr, loopExpr(thenExpr, loopBool(cond, acc)))
         case SwitchVariant(on, _, cases, default) =>
-          cases.foldLeft(loopExpr(default, loopCheap(on, acc))) {
+          val acc0 = loopCheap(on, acc)
+          val acc1 = default.fold(acc0)(loopExpr(_, acc0))
+          cases.foldLeft(acc1) {
             case (accN, (_, branch)) =>
               loopExpr(branch, accN)
           }
@@ -915,7 +917,8 @@ object Matchless {
             loop(els, bindableArities, anonArities)
           )
         case SwitchVariant(_, _, cases, default) =>
-          cases.foldLeft(loop(default, bindableArities, anonArities)) {
+          val defaultArity = default.flatMap(loop(_, bindableArities, anonArities))
+          cases.foldLeft(defaultArity) {
             case (acc, (_, branch)) =>
               sameArity(acc, loop(branch, bindableArities, anonArities))
           }
@@ -1003,7 +1006,9 @@ object Matchless {
           case If(cond, thenExpr, elseExpr) =>
             loopExpr(elseExpr, loopExpr(thenExpr, loopBool(cond, curr)))
           case SwitchVariant(on, _, cases, default) =>
-            cases.foldLeft(loopExpr(default, loopExpr(on, curr))) {
+            val curr1 = loopExpr(on, curr)
+            val curr2 = default.fold(curr1)(loopExpr(_, curr1))
+            cases.foldLeft(curr2) {
               case (acc, (_, branch)) =>
                 loopExpr(branch, acc)
             }
@@ -1416,7 +1421,14 @@ object Matchless {
                   val (branch1, st1) = recurExpr(branch, currSt)
                   ((variant, branch1) :: acc, st1)
               }
-            val (default1, stDefault) = recurExpr(default, stCases)
+            val (default1, stDefault) =
+              default match {
+                case Some(defaultExpr) =>
+                  val (defaultExpr1, st1) = recurExpr(defaultExpr, stCases)
+                  (Some(defaultExpr1), st1)
+                case None              =>
+                  (None, stCases)
+              }
             (
               SwitchVariant(
                 on1,
@@ -1473,7 +1485,7 @@ object Matchless {
               .map { case (_, branch) =>
                 loopExpr(branch)
               }
-              .sum + loopExpr(default)
+              .sum + default.fold(0)(loopExpr)
           case Always(cond, thenExpr) =>
             1 + loopBool(cond) + loopExpr(thenExpr)
           case PrevNat(of) =>
@@ -1631,7 +1643,7 @@ object Matchless {
             cases.map { case (variant, branch) =>
               (variant, recurExpr(branch))
             },
-            recurExpr(default)
+            default.map(recurExpr)
           )
         case Always(cond, thenExpr) =>
           rewriteCanonicalRecursionLoop(recurBool(cond), recurExpr(thenExpr))
@@ -1912,7 +1924,7 @@ object Matchless {
       case SwitchVariant(on, _, cases, default) =>
         hasSideEffect(on) ||
           cases.exists { case (_, branch) => hasSideEffect(branch) } ||
-          hasSideEffect(default)
+          default.exists(hasSideEffect)
       case Let(_, x, b) =>
         hasSideEffect(b) || hasSideEffect(x)
       case LetMut(_, in) => hasSideEffect(in)
@@ -1949,7 +1961,7 @@ object Matchless {
       cases: NonEmptyList[(Int, Expr[A])],
       // Variants not listed in `cases` (for example via wildcard/default rows)
       // evaluate this branch.
-      default: Expr[A]
+      default: Option[Expr[A]]
   ) extends Expr[A] {
     private val caseVariants = cases.toList.map(_._1)
     private val familySize = famArities.length
@@ -2680,7 +2692,7 @@ object Matchless {
             cases.map { case (variant, branch) =>
               (variant, substituteLocals(m, branch))
             },
-            substituteLocals(m, default)
+            default.map(substituteLocals(m, _))
           )
         case Always(c, e) =>
           Always(substituteLocalsBool(m, c), substituteLocals(m, e))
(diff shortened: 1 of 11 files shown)


johnynek:
Reproduced the failing `coreJS` test from CI (`PythonGenTest.SwitchVariant compiles to cached-tag if/elif chain in Python`) locally and fixed it in `core/src/test/scala/dev/bosatsu/codegen/python/PythonGenTest.scala` by removing the Scala.js-incompatible inline multiline regex flag (`(?m)`) and matching per-line via `code.linesIterator`. Re-ran `sbt "coreJS/testOnly dev.bosatsu.codegen.python.PythonGenTest -- --log=failure"` and it passed. Ran required pre-push command `scripts/test_basic.sh`; it passed (`Passed: Total 1673, Failed 0, Errors 0, Passed 1673, Ignored 2`).

johnynek:
Have to get the code coverage up to 95%. See the github actions yml for where the codecoverage is run, and use that to see where you are at now in Matchless, ClangGen and PythonGen and get it up to 95%.

core/src/main/scala/dev/bosatsu/MatchlessToValue.scala
```diff
@@ -473,6 +473,20 @@ object MatchlessToValue {
                   else elseF(scope)
                 }
             }
+          case SwitchVariant(on, _, cases, default) =>
+            val onF = loop(on)
+            val caseFns =
```
johnynek, line 479: this should be an `Array` and then we look up the element by the variant id, so that we have O(1) reading.

johnynek: Implemented this change in `core/src/main/scala/dev/bosatsu/MatchlessToValue.scala`.

`SwitchVariant` evaluation now builds an `Array[Option[Scoped[Value]]]` sized from `famArities.length` and indexes it directly by `variant` at runtime, with fallback to `default` when no branch is present. This replaces the previous `Map` lookup and gives O(1) case selection by variant id.

johnynek:
Applied the requested evaluator optimization and verified behavior with `sbt -batch "coreJVM/testOnly dev.bosatsu.MatchlessRegressionTest -- --log=failure"` and required pre-push check `scripts/test_basic.sh` (pass).

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -1837,6 +1943,25 @@ object Matchless {
...
+  case class SwitchVariant[A](
+      on: CheapExpr[A],
+      famArities: List[Int],
+      cases: NonEmptyList[(Int, Expr[A])],
+      default: Expr[A]
+  ) extends Expr[A] {
+    private val caseVariants = cases.toList.map(_._1)
+    private val familySize = famArities.length
+
+    require(cases.length >= 2, "SwitchVariant requires at least two cases")
```
johnynek: Use the Require macro.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -1837,6 +1943,25 @@ object Matchless {
...
+    require(
```
johnynek: Use the Require macro in this repo

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -1837,6 +1943,25 @@ object Matchless {
       (NonEmptyList((cond, thenExpr), rest), last)
     }
   }
+  case class SwitchVariant[A](
+      on: CheapExpr[A],
+      famArities: List[Int],
+      cases: NonEmptyList[(Int, Expr[A])],
+      default: Expr[A]
```
johnynek: Why do we need the default? Bosatsu is total so matches that have compiled must have a match. We will never use the default.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -4044,6 +4186,10 @@ object Matchless {
             case 1 => Some(true)
             case _ => None
           }
+        case Literal(Lit.Integer(value)) =>
```
johnynek: Integers aren't bools. Why this change? If we can't remove it comment why it is here.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -4497,7 +4675,75 @@ object Matchless {
...
+                          }
+                        }
+                    }
+
+                  val defaultExprF: F[Expr[B]] =
+                    if (defaultRows.nonEmpty)
+                      compileRows(defaultRows, defaultOccs, mustMatch)
+                    else Monad[F].pure(UnitExpr)
+
+                  (enumSigs.toList.traverse(compileCase), defaultExprF).mapN {
```
johnynek: Don't convert from NEL just to traverse. You can traverse an NEL then you don't need to convert back.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Updated `SwitchVariant` invariants to use the repo `Require` macro instead of `require`.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Done. All three `SwitchVariant` constructor checks now use `Require(...)`.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Kept `default`, and added an explicit comment on the field clarifying why it exists: it is the branch for variants not explicitly listed in `cases` (for wildcard/default rows). Source matches are total, but a total match can still compile to partial explicit variant cases plus a real default branch.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Removed the integer-literal path from `boolLiteralValue` so this logic only treats the Bool enum representation as boolean. Also removed the test that was asserting integer-coded bool fast-path behavior.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Updated this to traverse the `NonEmptyList` directly: `enumSigs.traverse(compileCase)`, and removed the list conversion plus `fromListUnsafe` round-trip.

johnynek:
Applied all requested Matchless review fixes and validated with `sbt -batch "coreJVM/testOnly dev.bosatsu.MatchlessTest -- --log=failure"`, `sbt -batch "coreJVM/testOnly dev.bosatsu.MatchlessRegressionTest -- --log=failure"`, and required pre-push check `scripts/test_basic.sh` (pass).

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -4497,7 +4673,75 @@ object Matchless {
...
+                          compileRows(newRows, newOccs, subMustMatch).map {
+                            (variant, _)
+                          }
+                        }
+                    }
+
+                  val defaultExprF: F[Expr[B]] =
+                    if (defaultRows.nonEmpty)
+                      compileRows(defaultRows, defaultOccs, mustMatch)
+                    else Monad[F].pure(UnitExpr)
```
johnynek: Shouldn't the defaultExpr be an Option[Expr] so when the cases cover all the variants it can be None and we don't have this fake Unit branch (which would be ill-typed if we ever actually hit it).

core/src/main/scala/dev/bosatsu/MatchlessToValue.scala
```diff
@@ -473,6 +473,22 @@ object MatchlessToValue {
...
+            cases.iterator.foreach { case (variant, branch) =>
+              caseFns(variant) = Some(loop(branch))
+            }
+            lazy val defaultF = loop(default)
+
+            Dynamic { (scope: Scope) =>
+              val variant = onF(scope).asSum.variant
+              if ((variant >= 0) && (variant < caseFns.length))
+                caseFns(variant).getOrElse(defaultF)(scope)
+              else defaultF(scope)
```
johnynek: An out of bounds variant is a type error which should never happen in this code. We can throw an error here to help debugging.

core/src/main/scala/dev/bosatsu/MatchlessToValue.scala
```diff
@@ -473,6 +473,22 @@ object MatchlessToValue {
...
+              Array.fill[Option[Scoped[Value]]](famArities.length)(None)
+            cases.iterator.foreach { case (variant, branch) =>
+              caseFns(variant) = Some(loop(branch))
+            }
+            lazy val defaultF = loop(default)
+
+            Dynamic { (scope: Scope) =>
+              val variant = onF(scope).asSum.variant
+              if ((variant >= 0) && (variant < caseFns.length))
+                caseFns(variant).getOrElse(defaultF)(scope)
```
johnynek: Don't use getOrElse. It's an extra allocation. Lets use pattern matching or is defined/get.

core/src/test/scala/dev/bosatsu/codegen/clang/ClangGenTest.scala
```diff
@@ -457,6 +457,98 @@ main = pick
...
+  match v:
+    case A: 0
+    case B: 1
+    case C: 2
+    case D: 3
+    case E: 4
+    case F: 5
+    case G: 6
+    case H: 7
+    case _: 8
```
johnynek: Can we still pass this test if this uses I? It shouldn't need _

core/src/test/scala/dev/bosatsu/codegen/clang/ClangGenTest.scala
```diff
@@ -457,6 +457,98 @@ main = pick
...
+  D
+  E
+
+def pick(v):
+  match v:
+    case A(_): 0
+    case B: 1
+    case C: 2
+    case D: 3
+    case _: 4
```
johnynek: Shouldn't need _ here. Totality check should see there is only one possibility.

core/src/main/scala/dev/bosatsu/codegen/clang/Code.scala
```diff
@@ -475,6 +475,12 @@ object Code {
       elseCond: Option[Block]
   ) extends Statement
   case class DoWhile(block: Block, whileCond: Expression) extends Statement
+  case class Switch(
+      on: Expression,
+      cases: NonEmptyList[(IntLiteral, Block)],
+      default: Block
```
johnynek: c doesn't require a default block I don't think. I think it is optional. This should be Option[Block] so we can represent such switch statements.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Implemented. `Matchless.SwitchVariant` now carries `default: Option[Expr[A]]`, and `compileSwitchCases` now builds `defaultExprF: F[Option[Expr[B]]]`. When cases are exhaustive (or `mustMatch`), default is `None`; we only synthesize `Some(UnitExpr)` for the existing non-`mustMatch` partial-path behavior.

core/src/main/scala/dev/bosatsu/MatchlessToValue.scala
johnynek: Updated evaluator behavior: out-of-bounds variant tags now throw `IllegalStateException` with variant/family-size details to aid debugging.

core/src/main/scala/dev/bosatsu/MatchlessToValue.scala
johnynek: Removed `getOrElse` on case lookup. The switch evaluator now uses explicit pattern matching on `caseFns(variant)` and on optional default, avoiding that allocation path.

core/src/test/scala/dev/bosatsu/codegen/clang/ClangGenTest.scala
johnynek: Updated the test source to use `case I: 8` instead of wildcard. Since this switch is now exhaustive, I also updated the assertion to confirm no `default:` block is emitted.

core/src/test/scala/dev/bosatsu/codegen/clang/ClangGenTest.scala
johnynek: Updated to explicit total case: `case E: 4` (removed wildcard).

core/src/main/scala/dev/bosatsu/codegen/clang/Code.scala
johnynek: Implemented. `Code.Switch` now uses `default: Option[Block]`, renderer omits `default` when `None`, and `CodeTest` now covers both with-default and without-default rendering.

johnynek:
Applied the optional-default switch refactor end-to-end (Matchless IR, evaluator, C/Python codegen, and tests), plus strict evaluator errors for impossible variant states. Validation run: `sbt -batch "coreJVM/test:compile"`, targeted suites (`MatchlessTest`, `MatchlessRegressionTest`, `CodeTest`, `ClangGenTest`, `PythonGenTest`), and required pre-push check `scripts/test_basic.sh` (pass).

PR #2077, opened 2026-03-08
Add `popcount_Int` to Predef across JVM, Python, and C backends

### LINKED ISSUE #2076: Add popcount_Int to Predef
On C it should use compiler built ins when possible.

On Java it should use bitCount. Python should use the most efficient method known.

Should be an Int -> Int function.

johnynek:
Implemented issue #2076 with focused changes: added `popcount_Int : Int -> Int` to `Bosatsu/Predef` exports and extern declarations; wired evaluator extern registration in `Predef.scala`; implemented JVM behavior via Java `bitCount` (`BigInteger.bitCount`). Added Python lowering in `PythonGen` that normalizes negatives to sign-bit semantics, uses `int.bit_count()` when available, and falls back to a Kernighan loop for compatibility/performance. Added C runtime support via new `bsts_integer_popcount` in `bosatsu_runtime.c/.h` using compiler builtins (`__builtin_popcount`/`__builtin_popcountll`) when available with a portable fallback, and exposed it through `bosatsu_ext_Bosatsu_l_Predef.c/.h` as `___bsts_g_Bosatsu_l_Predef_l_popcount__Int`. Added regression coverage in `core/src/test/scala/dev/bosatsu/EvaluationTest.scala` and `test_workspace/PredefTests.bosatsu`. Required pre-push command `scripts/test_basic.sh` was run and passed.

Fixes #2076

### DIFF
--- a/core/src/main/scala/dev/bosatsu/Predef.scala
+++ b/core/src/main/scala/dev/bosatsu/Predef.scala
@@ -2916,7 +2916,7 @@ object PredefImpl {
     ExternalValue(notInt(intRaw(a)))
 
   def popcount_Int(a: Value): Value =
-    VInt(i(a).bitCount())
+    VInt(intRaw(a).popCount)
 
   private def toIntExactIfRepresentable(v: BigInteger): Option[Int] = {
     val minInt = BigInteger.valueOf(Int.MinValue.toLong)
--- a/core/.js/src/main/scala/dev/bosatsu/BosatsuInt.scala
+++ b/core/.js/src/main/scala/dev/bosatsu/BosatsuInt.scala
@@ -146,6 +146,16 @@ object BosatsuInt {
         case bi: BigInteger       => fromBigInteger(bi.not())
       }
 
+    inline def popCount: Int =
+      t match {
+        case i: java.lang.Integer =>
+          val iv = i.intValue()
+          if (iv >= 0) java.lang.Integer.bitCount(iv)
+          else java.lang.Integer.bitCount(~iv)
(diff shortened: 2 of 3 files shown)


core/src/main/scala/dev/bosatsu/Predef.scala
```diff
@@ -2910,6 +2915,9 @@ object PredefImpl {
   def not_Int(a: Value): Value =
     ExternalValue(notInt(intRaw(a)))
 
+  def popcount_Int(a: Value): Value =
+    VInt(i(a).bitCount())
```
johnynek: We can call Long.bitCount when this is a Long and we are almost sure the result will fit in a Long so I think this should be more efficient.

The whole point of making this built in is to be more efficient.

johnynek: Implemented. `popcount_Int` now calls `intRaw(a).popCount` in `Predef.scala`, and I added platform fast paths in `BosatsuInt`:
- JVM: `java.lang.Long` uses `java.lang.Long.bitCount` (with `~lv` for negative values to preserve the existing sign-bit semantics), `BigInteger` uses `bitCount`.
- Scala.js: `java.lang.Integer` uses `java.lang.Integer.bitCount` (same negative handling), `BigInteger` uses `bitCount`.

This removes the forced `BigInteger` conversion on the small-int path while keeping behavior consistent.

johnynek:
Applied the requested efficiency change for `popcount_Int` and verified with the required command: `scripts/test_basic.sh` (passed).

PR #2081, opened 2026-03-09
Optimize boolean truth-table C codegen and add exhaustive end-to-end tests

### LINKED ISSUE #2080: Add tests that Boolean functions are compiled well end to end.
Boolean functions like and, or, xor, and compositions with not on inputs or the output should be very efficient when compiled to C but currently they may not be.

We should have tests that assert the generated C is basically optimal for all 16 Boolean binary functions and all 4 unary binary functions.

Getting these optimal is probably partly a Matchless lowering problem and then at ClangGen encoding matchless patterns into C. For each of these 20 functions we should have the bosatsu code and assert it equals a given segment of C. We want all twenty and the full bosatsu => C mapping tested for optimality.

johnynek:
Implemented C codegen simplifications for boolean-valued conditionals in `core/src/main/scala/dev/bosatsu/codegen/clang/Code.scala` so nested ternaries/branches are collapsed into direct logical predicates, including simplification of redundant branches and `alloc_enum0` boolean mappings. Added a new regression suite `core/src/test/scala/dev/bosatsu/codegen/clang/ClangGenBooleanCodegenTest.scala` that compiles all 16 binary boolean functions and all 4 unary boolean functions end-to-end and asserts the expected optimized C mapping for each. Updated the CLI hash snapshot in `cli/src/test/scala/dev/bosatsu/codegen/clang/ClangGenTest.scala` for `Ackermann.bosatsu` to match the improved generated C. Validation run: `scripts/test_basic.sh` (passes).

Fixes #2080

### DIFF
--- a/core/src/test/scala/dev/bosatsu/codegen/clang/ClangGenBooleanCodegenTest.scala
+++ b/core/src/test/scala/dev/bosatsu/codegen/clang/ClangGenBooleanCodegenTest.scala
@@ -42,6 +42,21 @@ class ClangGenBooleanCodegenTest extends munit.FunSuite {
        |""".stripMargin
   }
 
+  private def binaryTupleDef(bits: Int): String = {
+    val ff = boolLit((bits & 1) != 0)
+    val ft = boolLit((bits & 2) != 0)
+    val tf = boolLit((bits & 4) != 0)
+    val tt = boolLit((bits & 8) != 0)
+
+    s"""def bm$bits(a: Bool, b: Bool) -> Bool:
+       |  match (a, b):
+       |    case (False, False): $ff
+       |    case (False, True): $ft
+       |    case (True, False): $tf
+       |    case (True, True): $tt
+       |""".stripMargin
+  }
+
   private def unaryDef(bits: Int): String = {
     val f = boolLit((bits & 1) != 0)
     val t = boolLit((bits & 2) != 0)
@@ -55,17 +70,23 @@ class ClangGenBooleanCodegenTest extends munit.FunSuite {
 
   private lazy val source: String = {
     val binary = (0 to 15).map(binaryDef)
+    val binaryTuple = (0 to 15).map(binaryTupleDef)
     val unary = (0 to 3).map(unaryDef)
-    val allNames =
+    val allNestedNames =
       ((0 to 15).map(i => s"b$i") ++ (0 to 3).map(i => s"u$i")).mkString(", ")
+    val allTupleNames = (0 to 15).map(i => s"bm$i").mkString(", ")
 
     (s"""package Test/BooleanTruthTable
         |
         |${binary.mkString("\n\n")}
         |
+        |${binaryTuple.mkString("\n\n")}
+        |
         |${unary.mkString("\n\n")}
         |
-        |main = ($allNames)
+        |all_nested = ($allNestedNames)
+        |all_tuple = ($allTupleNames)
+        |main = (all_nested, all_tuple)
         |""").stripMargin
   }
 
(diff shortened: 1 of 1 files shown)


core/src/test/scala/dev/bosatsu/codegen/clang/ClangGenBooleanCodegenTest.scala
```diff
@@ -0,0 +1,205 @@
...
+      "(!(get_variant_value(__bsts_b_a0) == 1)) && (!(get_variant_value(__bsts_b_b0) == 1))"
+    ),
+    2 -> Expectation.Expr(
+      "(!(get_variant_value(__bsts_b_a0) == 1)) && (get_variant_value(__bsts_b_b0) == 1)"
+    ),
+    3 -> Expectation.Expr(
+      "!(get_variant_value(__bsts_b_a0) == 1)"
+    ),
+    4 -> Expectation.Expr(
+      "(get_variant_value(__bsts_b_a0) == 1) && (!(get_variant_value(__bsts_b_b0) == 1))"
```
johnynek: `!(x == y)` should be simplified to `(x != y)`

core/src/test/scala/dev/bosatsu/codegen/clang/ClangGenBooleanCodegenTest.scala
```diff
@@ -0,0 +1,205 @@
...
+    ),
+    5 -> Expectation.Expr(
+      "!(get_variant_value(__bsts_b_b0) == 1)"
```
johnynek: should be simplified to `get_variant_value(__bsts_b_b0) != 1`

core/src/main/scala/dev/bosatsu/codegen/clang/Code.scala
```diff
@@ -191,6 +191,127 @@ object Code {
...
+
+  private def simplifyBoolExpr(expr: Expression): Expression =
+    expr match {
+      case PrefixExpr(PrefixUnary.Not, target) =>
+        simplifyBoolExpr(target) match {
+          case IntLiteral(i) =>
+            if (i == 0) IntLiteral.One else IntLiteral.Zero
+          case PrefixExpr(PrefixUnary.Not, inner) =>
+            inner
+          case t =>
```
johnynek, line 266: we should be able to simplify several `Not(BinExpr(_, _, _))` cases here.

core/src/test/scala/dev/bosatsu/codegen/clang/ClangGenBooleanCodegenTest.scala
johnynek: Updated this expectation to the simplified form. The binary truth-table expectations now use `!= 1` where applicable, including this case (`(get_variant_value(__bsts_b_a0) == 1) && (get_variant_value(__bsts_b_b0) != 1)`).

core/src/test/scala/dev/bosatsu/codegen/clang/ClangGenBooleanCodegenTest.scala
johnynek: Applied. This expectation now asserts `get_variant_value(__bsts_b_b0) != 1` directly.

core/src/main/scala/dev/bosatsu/codegen/clang/Code.scala
johnynek: Implemented additional negation simplifications in `simplifyBoolExpr` for `Not(BinExpr(_, _, _))`: `==/!=`, `< <= > >=` inversions, plus De Morgan rewrites for `&&` and `||`. I also tightened ternary complement folding so equivalent positive forms are preferred when both `Eq`/`NotEq` encodings are possible.

johnynek:
Addressed all three review comments with concrete code updates in `Code.scala` and aligned the end-to-end boolean codegen expectations. Since codegen text changed again, I updated the CLI Ackermann hash snapshot to the new deterministic value. Validation: `sbt "coreJVM/testOnly dev.bosatsu.codegen.clang.ClangGenBooleanCodegenTest"` and required `scripts/test_basic.sh` both pass.

core/src/test/scala/dev/bosatsu/codegen/clang/ClangGenBooleanCodegenTest.scala
```diff
@@ -0,0 +1,205 @@
...
+
+  private def boolLit(bit: Boolean): String = if (bit) "True" else "False"
+
+  private def binaryDef(bits: Int): String = {
+    val ff = boolLit((bits & 1) != 0)
+    val ft = boolLit((bits & 2) != 0)
+    val tf = boolLit((bits & 4) != 0)
+    val tt = boolLit((bits & 8) != 0)
+
+    s"""def b$bits(a: Bool, b: Bool) -> Bool:
```
johnynek, line 32: We should have this test but we should also add a second version that is more idomatic:
```
match (a, b):
    case (False, False): $ff
    case (False, True): $ft
    case (True, False): $tf
    case (True, True): $tt
```

that version should also as efficient. Matchless should already handle those without added allocations.

johnynek: Implemented. I added a second binary truth-table corpus using the idiomatic tuple pattern form:

- New generated defs: `bm0..bm15` using `match (a, b)` with the four tuple cases.
- Included these in the compiled source reachability graph (`all_tuple`, with `main = (all_nested, all_tuple)`).
- Extended the binary regression test to validate both styles:
  - `b*` (nested matches) still checked against the exact optimized predicate mapping.
  - `bm*` (tuple matches) now checked for the same optimized shape constraints: branchless (`no if`/`no ?`) and no tuple-allocation artifacts (`no alloc_struct` / `no get_struct_index`).

This ensures the tuple-pattern lowering is also efficient without extra allocations.

johnynek:
Added tuple-pattern boolean truth-table coverage in `ClangGenBooleanCodegenTest` and verified it enforces efficient codegen characteristics for that style as requested. Required check passed: `scripts/test_basic.sh`.

Issue #2106, opened 2026-03-10
Switching on integers


johnynek:
```
def small_spaces(rem: Int) -> String:
  match rem:
    case 1:
      " "
    case 2:
      "  "
    case 3:
      "   "
    case 4:
      "    "
    case 5:
      "     "
    case 6:
      "      "
    case 7:
      "       "
    case 8:
      "        "
    case 9:
      "         "
    case _:
      ""
```

In this code, we would like to have good code generation. I think currently it may be just a linear if/else chain. It would be nice to be able to switch on small integers, or at least generate a binary tree if/else chain for such chains. binary comparison based if/else may be more robust in that since all literals are comparable, we could add `LtEqLit` node, or something like that.

johnynek:
MergeXO assigned an agent and started design work for issue #2106.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/2107

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #2106.

johnynek:
Opened implementation PR: https://github.com/johnynek/bosatsu/pull/2111

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/2106#issuecomment-4034179002
- linked PR: #2111 (https://github.com/johnynek/bosatsu/pull/2111)

Comments on the source issue are no longer actioned after a PR exists. Please comment on the PR thread instead.

PR #2107, opened 2026-03-10
Design doc for #2106: Issue #2106 Design: Switching On Integers

johnynek:
Design doc.

Refs #2106

### DIFF
--- a/docs/design/2106-switching-on-integers.md
+++ b/docs/design/2106-switching-on-integers.md
@@ -1,32 +1,5 @@
 ---
 issue: 2106
-priority: 3
-touch_paths:
-  - docs/design/2106-switching-on-integers.md
-  - core/src/main/scala/dev/bosatsu/Matchless.scala
-  - core/src/main/scala/dev/bosatsu/MatchlessToValue.scala
-  - core/src/main/scala/dev/bosatsu/codegen/clang/ClangGen.scala
-  - core/src/main/scala/dev/bosatsu/codegen/python/PythonGen.scala
-  - core/src/test/scala/dev/bosatsu/MatchlessTests.scala
-  - core/src/test/scala/dev/bosatsu/MatchlessRegressionTest.scala
-  - core/src/test/scala/dev/bosatsu/Issue1633Test.scala
-  - core/src/test/scala/dev/bosatsu/codegen/clang/ClangGenTest.scala
-  - core/src/test/scala/dev/bosatsu/codegen/python/PythonGenTest.scala
-depends_on: []
-estimated_size: M
-generated_at: 2026-03-10T18:20:02Z
----
-
-# Issue #2106 Design: Switching On Integers
-
-_Issue: #2106 (https://github.com/johnynek/bosatsu/issues/2106)_
-
-## Summary
-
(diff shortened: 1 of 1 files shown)


docs/design/2106-switching-on-integers.md
```diff
@@ -0,0 +1,260 @@
...
+
+- `case 1`, `case 2`, ... `case 9`, `case _`
+
+this becomes a long if/else chain with O(n) comparisons in the worst case.
+
+The issue asks for better code generation, ideally switch-like behavior for integers, or at least balanced comparisons using a node like `LtEqLit`.
+
+## Goals
+
+1. Improve generated decision structure for integer literal pattern matches.
```
johnynek: would be nice to include literal char in this, since parsers often match on a few characters.

docs/design/2106-switching-on-integers.md
```diff
@@ -0,0 +1,260 @@
...
+3. No C runtime API change to expose direct small-int switch tags.
+4. No redesign of existing `SwitchVariant` logic.
+
+## Design decision
+
+Adopt matrix-time balanced integer dispatch with a new `LtEqLit` bool node.
+
+Policy:
+
+1. Trigger only when selected matrix column signatures are all integer literals.
```
johnynek: all literals or wildcard I think (we must always have a wildcard to have totality).

docs/design/2106-switching-on-integers.md
```diff
@@ -0,0 +1,260 @@
...
+3. Avoids brittle backend-side reconstruction of switch opportunities.
+4. Avoids runtime coupling required for true C `switch` over boxed integers.
+
+## Proposed architecture
+
+### 1) Extend BoolExpr with ordered integer comparison
+
+Add in `Matchless.scala`:
+
+`case class LtEqLit[A](expr: CheapExpr[A], lit: Lit.Integer) extends BoolExpr[A]`
```
johnynek: I think we need `case class LtEqLit[A](expr: CheapExpr[A], lit: Lit) extends BoolExpr[A]`

docs/design/2106-switching-on-integers.md
johnynek: Updated the design doc to include `Char` literals in scope alongside `Int` (title, summary, goals, policy, architecture, implementation plan, tests, and acceptance criteria). This now explicitly calls out parser-style char dispatch as a target case.

docs/design/2106-switching-on-integers.md
johnynek: Adjusted the policy section to state that wildcard rows are preserved as the default branch, and noted that this fallback is expected for total matches over finite literal sets.

docs/design/2106-switching-on-integers.md
johnynek: Updated the proposed node signature to `case class LtEqLit[A](expr: CheapExpr[A], lit: Lit) extends BoolExpr[A]` and documented invariants that lowering emits supported ordered literal kinds (`Int` and `Char`) in this phase.

johnynek:
Addressed all three review comments directly in `docs/design/2106-switching-on-integers.md` and cleaned the duplicate frontmatter/header block that was present at the top of the file. Required pre-push check ran successfully: `scripts/test_basic.sh` passed (cli tests and coreJVM tests).

PR #2111, opened 2026-03-10
Implement balanced Int/Char literal dispatch with Matchless `LtEqLit`

### LINKED ISSUE #2106: Switching on integers

johnynek:
Implemented issue #2106 per the design doc by adding `Matchless.LtEqLit(expr, lit)` and wiring it through all BoolExpr ordering/traversal/rewrite/side-effect/weight helpers. In `matchExprMatrixCheap`, added `LiteralTreeMinCases = 4` and a new literal-tree lowering path that triggers only for uniform `Int` or `Char` literal columns, sorts literals, compiles case/default bodies through existing specialization, and emits a balanced tree using `If(LtEqLit(...))` internals with `If(EqualsLit(...))` leaves. Added evaluator support in `MatchlessToValue` for ordered Int/Char comparison, Clang codegen lowering (`bsts_integer_cmp(...) <= 0` and char codepoint `<=`), and Python codegen lowering to ordered comparisons (`not(lit < expr)`) for Int/Char. Updated exhaustive BoolExpr matches in `MatchlessTests`, `MatchlessRegressionTest`, and `Issue1633Test`. Added regression/codegen tests for wide integer and char literal fanout (LtEqLit emission and backend ordered-compare output) plus threshold coverage to keep small literal matches linear. Verification run: `sbt "coreJVM/test:compile"`, `sbt "coreJVM/testOnly dev.bosatsu.MatchlessTest dev.bosatsu.MatchlessRegressionTest dev.bosatsu.Issue1633Test dev.bosatsu.codegen.clang.ClangGenTest dev.bosatsu.codegen.python.PythonGenTest"`, and required `scripts/test_basic.sh` (pass).

Fixes #2106

Implements design doc: [docs/design/2106-switching-on-integers.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/2106-switching-on-integers.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/2107

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -339,6 +339,52 @@ jobs:
           ./bosatsuj lib build --main_pack Bosatsu/FibBench --exe_out fib_bench_no_outdir --cc_flag=-O0 --cc_lib=-lm
           ./fib_bench_no_outdir 10
     timeout-minutes: 30
+  testC-sanitizers:
+    needs: test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java:
+          - '17'
+    steps:
+      - uses: "actions/checkout@v2.1.0"
+      - name: "java ${{matrix.java}} setup"
+        uses: "actions/setup-java@v5"
+        with:
+          distribution: "temurin"
+          java-version: "${{matrix.java}}"
+      - name: "install sbt"
+        uses: "sbt/setup-sbt@v1"
+      - name: "build assembly"
+        run: 'sbt "cli/assembly"'
+      - name: "install boehm"
+        run: "sudo apt-get update && sudo apt install -y libgc1 libgc-dev pkg-config"
+      - name: "run c sanitizer checks"
+        run: "./scripts/test_c_sanitizers.sh"
+    timeout-minutes: 60
+  testC-valgrind:
+    needs: test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java:
+          - '17'
+    steps:
+      - uses: "actions/checkout@v2.1.0"
+      - name: "java ${{matrix.java}} setup"
+        uses: "actions/setup-java@v5"
+        with:
+          distribution: "temurin"
+          java-version: "${{matrix.java}}"
+      - name: "install sbt"
+        uses: "sbt/setup-sbt@v1"
+      - name: "build assembly"
+        run: 'sbt "cli/assembly"'
+      - name: "install c toolchain and valgrind"
+        run: "sudo apt-get update && sudo apt install -y libgc1 libgc-dev pkg-config valgrind"
+      - name: "run c valgrind checks"
+        run: "./scripts/test_c_valgrind.sh"
+    timeout-minutes: 90
   testCMacOS:
     needs: test
     runs-on: macos-latest
--- a/c_runtime/Makefile
+++ b/c_runtime/Makefile
@@ -12,8 +12,10 @@ LIBS ?=
 
 ifeq ($(PROFILE),debug)
 	CFLAGS ?= -O0 -g3
+	RUNTIME_DEBUG_CPPFLAGS := -DBSTS_RUNTIME_DEBUG_CHECKS=1
 else ifeq ($(PROFILE),release)
 	CFLAGS ?= -O2
+	RUNTIME_DEBUG_CPPFLAGS :=
 else
 $(error PROFILE must be one of: release, debug)
 endif
@@ -43,7 +45,7 @@ endif
 
 endif
 
-ALL_CPPFLAGS := $(CPPFLAGS) $(GC_CFLAGS)
+ALL_CPPFLAGS := $(CPPFLAGS) $(RUNTIME_DEBUG_CPPFLAGS) $(GC_CFLAGS)
 ALL_LIBS := $(GC_LIBS) $(LIBS) -lm
 
 all: bosatsu_runtime.o test_out bosatsu_ext_Bosatsu_l_Predef.o bosatsu_ext_Bosatsu_l_Prog.o bosatsu_ext_Bosatsu_l_IO_l_Core.o bosatsu_ext_Bosatsu_l_IO_l_Bytes.o bosatsu_ext_Bosatsu_l_Collection_l_Array.o bosatsu_ext_Bosatsu_l_Num_l_Float64.o bosatsu_ext_Bosatsu_l_Lazy.o
--- a/c_runtime/bosatsu_ext_Bosatsu_l_Prog.c
+++ b/c_runtime/bosatsu_ext_Bosatsu_l_Prog.c
@@ -61,9 +61,8 @@ BValue ___bsts_g_Bosatsu_l_Prog_l_recover(BValue p, BValue f)
 
 static volatile BValue bsts_prog_observe_sink = (BValue)0;
 
-static BValue bsts_prog_observe_effect(BValue *slots, BValue arg)
+static BValue bsts_prog_observe_effect(BValue arg)
 {
-  (void)slots;
   bsts_prog_observe_sink = arg;
   bsts_prog_observe_sink = bsts_unit_value();
   return ___bsts_g_Bosatsu_l_Prog_l_pure(bsts_unit_value());
@@ -71,7 +70,7 @@ static BValue bsts_prog_observe_effect(BValue *slots, BValue arg)
 
 BValue ___bsts_g_Bosatsu_l_Prog_l_observe(BValue a)
 {
-  BValue effect_fn = alloc_closure1(0, 0, bsts_prog_observe_effect);
+  BValue effect_fn = alloc_boxed_pure_fn1(bsts_prog_observe_effect);
   return alloc_enum2(5, a, effect_fn);
 }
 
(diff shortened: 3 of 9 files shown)


core/src/main/scala/dev/bosatsu/MatchlessToValue.scala
```diff
@@ -285,6 +298,23 @@ object MatchlessToValue {
               }
             }
 
+          case LtEqLit(expr, lit) =>
+            loop(expr).map { e =>
+              val external = e.asExternal.toAny
+              lit match {
```
johnynek: let's lift this match on `lit` out of the inner loop. This means we can do it statically once for sure and not have any chance of it being swept up in repeated evalatuion.

core/src/main/scala/dev/bosatsu/MatchlessToValue.scala
```diff
@@ -285,6 +298,23 @@ object MatchlessToValue {
...
+                case Lit.Integer(i) =>
+                  valueToBigInteger(external).exists(_.compareTo(i) <= 0)
```
johnynek: rather than convert to BigInteger and compare there, let's convert the `i` to `BInt` and then have `compareTo` on `BInt` this way, if the integer is already not a biginteger, but a java.lang.Long or java.lang.Integer we don't have to convert just to do the comparison. That's a awaset of an allocation.

Basically, the goal is efficiency here, so allocating in this loop is not ideal.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -4701,6 +4734,70 @@ object Matchless {
                       }
                   }
 
+                sealed trait OrderedLiteralKind derives CanEqual
```
johnynek: use `enum` here:

```scala
enum OrderedLiteralKind derives CanEqual {
  case IntKind, CharKind
}
```

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -4701,6 +4734,70 @@ object Matchless {
...
+                            s"unexpected Char literal comparison: left=$left, right=$right"
+                          )
+                        // $COVERAGE-ON$
+                      }
+                  }
+
+                def literalTreeLits(sigs: List[HeadSig]): Option[List[Lit]] =
+                  if (sigs.length < LiteralTreeMinCases) None
+                  else {
+                    sigs
```
johnynek: use for here:
```scala
for {
  literals <- sigs.traverse {
    case LitSit(lit) => Some(lit)
    case _ => None
  }
  nel <- NonEmptyList.fromList(literals)
  kind <- orderedLiteralKind(nel.head)
  if nel.forall(lit => orderedLiteralKind(lit).contains(kind))
} yield nel.sorted(using new Order[Lit] {
    def compare(left, right) =
                                  orderedLiteralCompare(kind, left, right) < 0)
  })
```
something like that.

Don't discard the non-emptyness of the list just to get sorting.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -4701,6 +4734,70 @@ object Matchless {
...
+                        case _ =>
+                          // $COVERAGE-OFF$
+                          throw new IllegalStateException(
+                            s"unexpected Char literal comparison: left=$left, right=$right"
+                          )
+                        // $COVERAGE-ON$
+                      }
+                  }
+
+                def literalTreeLits(sigs: List[HeadSig]): Option[List[Lit]] =
```
johnynek: what about `_` cases here? All literal matches will have a default branch because the literals can't be enumerated (in general). Add a comment how that `_` is handled.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -4752,6 +4849,47 @@ object Matchless {
...
+                    sortedLits: List[Lit],
+                    mustMatch: Boolean
+                ): F[Expr[B]] = {
+                  def compileLiteralCase(lit: Lit): F[(Lit, Expr[B])] = {
+                    val (newRows, _) =
+                      minimizeSpecializedRows(LitSig(lit), rows, colIdx, 0)
+                    val newOccs = occs.patch(colIdx, Nil, 1)
+                    if (newRows.isEmpty) Monad[F].pure((lit, UnitExpr))
+                    else
+                      compileRows(newRows, newOccs, mustMatch).map((lit, _))
```
johnynek: I'm confused how this can be correct when the literals have been sorted, but we didn't record their original branch `Expr`. Can you add a comment how this is accomplished and makes sure we have actual tests that execise running code that has these trees in it?

core/src/test/scala/dev/bosatsu/MatchlessTests.scala
```diff
@@ -2767,6 +2783,80 @@ def pick(v):
...
+          case Matchless.LtEqLit(_, _) => true
+          case _                       => false
+        },
+        true,
+        expr.toString
+      )
+    }
+  }
+
+  test("matrix match uses LtEqLit tree for char dispatch fanout") {
```
johnynek, line 2838: we need to havea  test that this actually produces the correct mapping and that for all values the function works as expected (at least for a-e and then a few randomly selected other characters.

core/src/main/scala/dev/bosatsu/MatchlessToValue.scala
johnynek: Done. I moved `LtEqLit` dispatch to the outer `boolExpr` match (`case LtEqLit(expr, Lit.Integer(i))` and `case LtEqLit(expr, c: Lit.Chr)`), so the `lit` match is resolved once statically instead of inside the per-evaluation loop.

core/src/main/scala/dev/bosatsu/MatchlessToValue.scala
johnynek: Updated. For integer `LtEqLit`, I now precompute `rhs = BInt.fromBigInteger(i)` once and compare with `lhs.compare(rhs) <= 0` after extracting `lhs` via `BInt(...)`. This removes the BigInteger conversion allocation from the hot evaluation path.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Applied. `OrderedLiteralKind` is now a local enum:

`enum OrderedLiteralKind derives CanEqual { case IntKind, CharKind }`.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Reworked as requested. `literalTreeLits` now uses a for-comprehension and returns `Option[NonEmptyList[Lit]]` so we preserve non-emptiness instead of dropping to `List` just for sorting.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Added comment on the `_` handling in `fallbackExpr`: wildcard rows are collected into `defaultRows` and compiled once as the shared fallback branch for the literal tree.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Added a clarifying comment in `compileLiteralCase`: sorting only changes test order, while specialization by each literal still selects that literal’s original rows/RHS. I also tightened this with an invariant check if specialization unexpectedly returns no rows. To validate runtime behavior, I expanded tests to execute literal-tree-lowered functions and assert outputs.

core/src/test/scala/dev/bosatsu/MatchlessTests.scala
johnynek: Added execution coverage in `MatchlessTests`: `classify_char` now runs for `a`-`e` plus 5 deterministic random non-`a`..`e` code points and asserts exact mapping (`1..5` for `a`-`e`, `0` otherwise).

johnynek:
Implemented all requested review changes in `MatchlessToValue.scala`, `Matchless.scala`, and `MatchlessTests.scala`. Validation run results: `sbt "coreJVM/test:compile"` passed; `sbt "coreJVM/testOnly dev.bosatsu.MatchlessTest dev.bosatsu.MatchlessRegressionTest dev.bosatsu.Issue1633Test dev.bosatsu.codegen.clang.ClangGenTest dev.bosatsu.codegen.python.PythonGenTest"` passed; required `scripts/test_basic.sh` passed.

johnynek:
merge main and address the merge conflicts.

johnynek:
Merged `origin/main` into this PR branch and resolved merge conflicts. Conflict resolution kept both the new ordered Int/Char literal dispatch tests and the upstream C11-safe int64 constant test in `core/src/test/scala/dev/bosatsu/codegen/clang/ClangGenTest.scala`, and retained the upstream `INT64_MIN` handling in `core/src/main/scala/dev/bosatsu/codegen/clang/ClangGen.scala`. Ran required pre-push checks with `scripts/test_basic.sh`; it passed locally (62/62, then 1729/1729 with 2 ignored).

Issue #2119, opened 2026-03-11
Add an option to `lib test` to select particular test values


johnynek:
when you are iterating instead of `--filter` a whole package, it is sometimes nice to select just a single test value. It would be good to add a `--value` option that allows you to select `Foo/Bar/Baz::quux` and for a valid test value quux. If `quux` isn't a test value, it should error.

This selection mode be exclusive with `--filter`. So we should use `orElse` composition in decline to handle it.

johnynek:
MergeXO assigned an agent and started design work for issue #2119.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/2121

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #2119.

johnynek:
Opened implementation PR: https://github.com/johnynek/bosatsu/pull/2123

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/2119#issuecomment-4037010845
- linked PR: #2123 (https://github.com/johnynek/bosatsu/pull/2123)

Comments on the source issue are no longer actioned after a PR exists. Please comment on the PR thread instead.

PR #2121, opened 2026-03-11
Design doc for #2119: Issue #2119 Design: add `--value` to `lib test` to select a particular test value

johnynek:
Design doc.

Refs #2119

### DIFF
--- a/docs/design/2119-add-an-option-to-lib-test-to-select-particular-test-values.md
+++ b/docs/design/2119-add-an-option-to-lib-test-to-select-particular-test-values.md
@@ -10,10 +10,9 @@ touch_paths:
   - core/src/main/scala/dev/bosatsu/library/DecodedLibraryWithDeps.scala
   - core/src/main/scala/dev/bosatsu/Package.scala
   - core/src/test/scala/dev/bosatsu/ToolAndLibCommandTest.scala
-  - core/src/test/scala/dev/bosatsu/Issue1654Test.scala
 depends_on: []
 estimated_size: M
-generated_at: 2026-03-11T05:37:55Z
+generated_at: 2026-03-11T23:20:00Z
 ---
 
 # Issue #2119 Design: add `--value` to `lib test` to select a particular test value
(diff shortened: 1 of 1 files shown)


docs/design/2119-add-an-option-to-lib-test-to-select-particular-test-values.md
```diff
@@ -0,0 +1,233 @@
...
+2. `ByValue`: select exactly one matching bindable from `testEntriesInPackage`.
+3. If no match, raise a dedicated CLI error that states selected value is not a test value and includes available test values in the package (if any).
+
+### 5) Ensure optimizer does not drop valid `--value` targets
+
+`Package.discardUnused` currently pins roots from `Package.testRootBindables`, which currently tracks discovered entry semantics. That can drop other valid test-typed lets before CLI selection is applied.
+
+To make `--value` reliable, broaden root pinning for test values:
+
+1. Update `Package.testRootBindables` to include all top-level test-typed bindables (`Test` and `ProgTest`).
```
johnynek: Rather than this, let's require export a test value we want to test. So after parsing into Expr we can fail fast if the name isn't exported.

By requiring an export we don't change the compilation semantics with this feature and keeps them simpler: anything not exported or the final test or main may not available after compilation due to inlining and removing unused items.

johnynek: Updated the design doc to follow this approach. `--value` is now explicitly defined as an exported test value selector, and the prior proposal to broaden compilation roots (`Package.testRootBindables`) was removed. The architecture, implementation plan, testing strategy, acceptance criteria, risks, and rollout notes now all reflect export-gated validation (`not exported` vs `exported but not a test value`) while keeping compilation semantics unchanged.

johnynek:
Addressed the requested doc changes in `docs/design/2119-add-an-option-to-lib-test-to-select-particular-test-values.md` and ran the required pre-push check: `scripts/test_basic.sh` (passed).

PR #2122, opened 2026-03-11
Allow `lib/tool eval --run` to accept `--` passthrough args

### LINKED ISSUE #2118: lib eval --run should allow passing --args to the program
I want to pass args to my bosatsu program and do:

```
./bosatsu lib eval --run --main Zafu/Tool/JsonFormat::main --compact
```

but `--compact` is parsed by the `./bosatsu` command and isn't passed in.

I think the error is before any or our code even runs. To handle this, some programs use `--` by itself to segment args.

So, one approach would be to rework the arg parsing so that we search for `"--"` as an arg in the `List[String]` in the scala main. If we find it, we remove it and take all the trailing args and consider those separately.

So we would pass those args as "meta args" or something, that some commands could accept or not. In this way, eval `--run` could get access to those arguments.

johnynek:
Implemented issue #2118 per the merged design doc.

What changed:
- Added eval-scoped `--` splitting in `MainModule.run` so only `lib eval` and `tool eval` pre-partition args at the first delimiter and pass trailing tokens as passthrough args.
- Threaded passthrough args through command wiring:
  - `library.Command.opts(..., evalPassthroughArgs)`
  - `tool_command.ToolCommand.opts(..., evalPassthroughArgs)`
  - `tool_command.EvalCommand.opts(..., evalPassthroughArgs)`
- Updated eval arg handling to merge positional and passthrough args in deterministic order:
  - `effectiveRunArgs = positionalRunArgs ++ evalPassthroughArgs`
  - used for `PredefImpl.evalRunArgs(...)`
  - retained existing guard/error: trailing args without `--run` => `trailing args require --run`
- Updated `bosatsu` launcher so `--` is forwarded (not swallowed) by appending delimiter + trailing args to downstream invocation.
- Updated docs (`debugging_with_eval.md`) with explicit `--run -- --compact` usage guidance.
- Added regressions:
  - `ToolAndLibCommandTest`: tool/lib eval passthrough success, mixed arg ordering, and no-`--run` trailing-args failures.
  - `PathModuleTest`: parser-level delimiter sanity checks for tool eval.

Validation run (all passing):
- `sbt "coreJVM/testOnly dev.bosatsu.ToolAndLibCommandTest"`
- `sbt "cli/testOnly dev.bosatsu.PathModuleTest"`
- `sbt "coreJVM/test:compile; cli/test:compile"`
- `scripts/test_basic.sh`

Fixes #2118

Implements design doc: [docs/design/2118-lib-eval-run-should-allow-passing-args-to-the-program.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/2118-lib-eval-run-should-allow-passing-args-to-the-program.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/2120

### DIFF
--- a/core/src/main/scala/dev/bosatsu/library/Command.scala
+++ b/core/src/main/scala/dev/bosatsu/library/Command.scala
@@ -1651,7 +1651,7 @@ object Command {
           def toCliException(ex: Throwable): Throwable =
             CliException.Basic(Option(ex.getMessage).getOrElse(ex.toString))
 
-          val effectiveRunArgs = positionalRunArgs ++ evalPassthroughArgs
+          val effectiveRunArgs = positionalRunArgs ::: evalPassthroughArgs
 
           val sourcePackageFilter: PackageName => Boolean =
             _ == target._1
--- a/core/src/main/scala/dev/bosatsu/tool_command/EvalCommand.scala
+++ b/core/src/main/scala/dev/bosatsu/tool_command/EvalCommand.scala
@@ -138,7 +138,7 @@ object EvalCommand {
           runArgs,
           errColor
       ) =>
-        val effectiveRunArgs = runArgs ++ evalPassthroughArgs
+        val effectiveRunArgs = runArgs ::: evalPassthroughArgs
 
         platformIO.withEC {
           runEval(


core/src/main/scala/dev/bosatsu/library/Command.scala
```diff
@@ -1644,10 +1647,12 @@ object Command {
           Opts.arguments[String]("arg").orEmpty,
           compileCacheDirOpt,
           Colorize.optsConsoleDefault
-        ).mapN { (fcc, target, runMain, runArgs, cacheDirFn, colorize) =>
+        ).mapN { (fcc, target, runMain, positionalRunArgs, cacheDirFn, colorize) =>
           def toCliException(ex: Throwable): Throwable =
             CliException.Basic(Option(ex.getMessage).getOrElse(ex.toString))
 
+          val effectiveRunArgs = positionalRunArgs ++ evalPassthroughArgs
```
johnynek: Use ::: to combine lists.

core/src/main/scala/dev/bosatsu/tool_command/EvalCommand.scala
```diff
@@ -137,6 +138,8 @@ object EvalCommand {
           runArgs,
           errColor
       ) =>
+        val effectiveRunArgs = runArgs ++ evalPassthroughArgs
```
johnynek: Use ::: to combine lists.

core/src/main/scala/dev/bosatsu/library/Command.scala
johnynek: Updated `core/src/main/scala/dev/bosatsu/library/Command.scala` to use `:::` for `effectiveRunArgs` as requested.

core/src/main/scala/dev/bosatsu/tool_command/EvalCommand.scala
johnynek: Updated `core/src/main/scala/dev/bosatsu/tool_command/EvalCommand.scala` to use `:::` for `effectiveRunArgs` as requested.

johnynek:
Addressed both line-level review comments with concrete code changes and re-ran the required pre-push check: `scripts/test_basic.sh` (passing).

PR #2126, opened 2026-03-11
lib doc: support remote .html links and fix Predef/local doc link precedence

johnynek:
## Summary
This PR updates library doc generation so downstream docs can target GitHub Pages `.html` URLs while preserving existing markdown behavior by default.

## Changes
- Adds new `lib doc` flag:
  - `--remote_doc_links_html` (default: `false`)
  - help text: `rewrite dependency doc links from .md to .html for links resolved via doc_base_url`
- Threads the flag through `ConfigConf.docPackages` into `MarkdownDoc.generate`.
- Extends `MarkdownDoc.generate` with `remoteDocLinksHtml: Boolean = false`.
- Updates link resolution in `MarkdownDoc.packageDocHref`:
  - local/generated packages always use local relative `.md` links
  - remote dependency links (`doc_base_url`) use `.md` by default and `.html` when the new flag is enabled
  - anchors are preserved unchanged
- Adds local-precedence logic so packages generated in the same run win over remote `doc_base_url` mappings.
- Ensures `--include_predef` still emits local `Bosatsu/Predef.md` even when a dependency provides a Predef interface.
- Updates `test_workspace/core_alpha_conf.json`:
  - `doc_base_url` points to docs URL
  - includes `Bosatsu/Predef` in `exported_packages`
- Updates docs/paradox publish pipeline to include both generated `.md` files and generated `.html` pages.

## Tests
Added/updated tests in `ToolAndLibCommandTest` for:
- `lib doc` remote dependency links with flag on (`.html`) including anchor links
- same scenario with flag off (`.md`)
- Predef behavior:
  - with `--include_predef` => local `Bosatsu/Predef.md`
  - without `--include_predef` => remote Predef URL when dependency metadata provides it

## Validation
- `sbt "coreJVM/testOnly dev.bosatsu.ToolAndLibCommandTest"` (pass)
- manual smoke run in temp repo:
  - with flag off: remote links remain `.md`, local links remain `.md`
  - with flag on: only remote dependency links switch to `.html`, local links stay `.md`

### DIFF
--- a/core/src/main/scala/dev/bosatsu/tool/MarkdownDoc.scala
+++ b/core/src/main/scala/dev/bosatsu/tool/MarkdownDoc.scala
@@ -28,6 +28,12 @@ object MarkdownDoc {
     case Private
   }
 
+  private final case class LinkContext(
+      packageBaseUrls: Map[PackageName, String],
+      localPackages: Set[PackageName],
+      remoteDocLinksHtml: Boolean
+  )
+
   private val fnTypeRegex = raw"^Fn\d+$$".r
   private val fnInputTypeParamRegex = raw"^i\d+$$".r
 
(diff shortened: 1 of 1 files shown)


johnynek:
Addressed Gemini review suggestions in commit `79e78f2b6`:
- simplified `packageDocHref` to remove duplicated relative-path logic
- reduced parameter drilling by introducing a `LinkContext` case class and threading it through MarkdownDoc link-related helpers

Re-ran `./scripts/test_basic.sh` after the refactor; it remains green.

PR #2129, opened 2026-03-12
add Bosatsu/Eval with lean proof of termination

johnynek:
## Summary
- add `Bosatsu/Eval` as a core package with a no-fuel external `eval_loop` and `eval` API
- implement `Bosatsu/Eval::eval_loop` externally on all required runtimes:
  - JVM (`core/src/main/scala/dev/bosatsu/Predef.scala`)
  - Python (`test_workspace/ProgExt.py` + `test_workspace/Prog.bosatsu_externals`)
  - C (`c_runtime/bosatsu_ext_Bosatsu_l_Eval.c`)
- simplify the Eval representation by removing `BindChain` and using:
  - `Pure(value: Leaf[a])`
  - `FlatMap(prev: Eval[b], fn: b -> Eval[a])`
- update the Lean proof (`lean_proofs/LeanProofs/EvalFuelExistence.lean`) to match the simplified `Eval` shape
- add/keep Eval package tests (including a long recursive sum through `defer` + `flat_map`) and verify across backends

## Why
This is part of an idea for proving properties of code we cannot currently prove directly in Bosatsu, and then using those proofs to guide external implementations.

## Important caveat
This PR is **not yet** a final justification for adding an unrestricted `run_Eval` loop in Bosatsu itself. The Lean proof gives a strong model-level termination argument for the Eval machine shape used here, but we still need to clearly document the compiler/runtime correspondence assumptions under which that model applies to all Bosatsu-generated Eval values.

## Validation
- `./bosatsuj lib check`
- `cd lean_proofs && ./setup.sh`
- `sbt cli/test`
- CI-style Python checks:
  - `./bosatsuj tool transpile ... python --outdir pyout ...`
  - `PYTHONPATH=... python3 -m unittest discover pyout -v --pattern "*.py"`
- C checks:
  - `./testc`
  - `make -C c_runtime install PROFILE=release CFLAGS='-O1 -g' CPPFLAGS='-DBSTS_CI=1'`
  - `./bosatsuj tool transpile ... Bosatsu/Eval ... c --outdir c_out_eval --test --exe_out test_exe ...`
  - `./c_out_eval/test_exe`
- Direct Eval Python check:
  - `./bosatsuj tool transpile ... Bosatsu/Eval ... python --outdir pyout_eval ...`
  - `PYTHONPATH=... python3 -m unittest discover pyout_eval -v --pattern "*.py"`

### DIFF
--- a/test_workspace/Bosatsu/Eval.bosatsu
+++ b/test_workspace/Bosatsu/Eval.bosatsu
@@ -45,12 +45,6 @@ one_Eval: Eval[Int] = done(1)
 def defer(fn: () -> Eval[a]) -> Eval[a]:
     unit_Eval.flat_map(fn)   
 
-def defer_eval(e: Eval[a]) -> Eval[a]:
-    defer(() -> e)
-
-def defer_add(n: Int, acc: Eval[Int]) -> Eval[Int]:
-    defer_eval(acc).flat_map(x -> done(x.add(n)))
-
 enum Stack[a, b]:
     Last(fn: a -> Eval[b])
     More[c](first: a -> Eval[c], rest: Stack[c, b])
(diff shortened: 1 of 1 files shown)


johnynek:
Addressed Gemini review feedback in commit 8abee045b:

- lean_proofs/setup.sh: added curl -L and removed the redundant explicit lake toolchain-install block.
- lean_proofs/LeanProofs/EvalNoFuel.lean: removed the unnecessary let-binding in runLoop and kept the termination proof working.

Also re-ran ./scripts/check_proofs.sh successfully.

PR #2136, opened 2026-03-12
Stop cascading constructor errors after variance inference failures

### LINKED ISSUE #2135: Confusing errors when variance solving fails
I have this error:
```
1. type error
  in file: /Users/oscar/code/bosatsu2/test_workspace/Bosatsu/Eval.bosatsu:21:1, package Bosatsu/Eval
  could not solve for valid variances
  19|    Many[c](first: BindChain[f, a, c], last: c -> f[b])
  20|
  21|enum Eval[a: +*]:
     ^

  to:
  21|enum Eval[a: +*]:
  22|    Pure(value: Leaf[a])
  23|    FlatMap[b](prev: Leaf[b], chain: BindChain[Eval, b, a])
                                                               ^
```

but compilation continues, but the constructors `Pure` and `FlatMap` are unknown so I get confusing errors.

When the variance inference fails, we should either trust the users (wrong in this case) variances, OR we should remove all the names from type checking that depend on the constructors that are part of the type that failed inference. We shouldn't say the constructors are unknown. That's a bad experience.

johnynek:
Implemented #2135 with a focused compiler change: in `Package.inferBodyUnopt`, variance/kind solving results from `KindFormula.solveShapesAndKinds` are now made strict by converting `Ior.Both` into `Ior.Left`, so we no longer continue type/name checking with a partial type environment that drops constructors from failed types. Added a regression test in `ErrorMessageTest` reproducing the reported `Eval` variance case and asserting we get kind inference errors without follow-on `Unknown constructor` errors. Updated one existing `EvaluationTest` case to accept `KindInferenceError` for this now-non-cascading behavior. Ran required pre-push tests: `scripts/test_basic.sh` passed.

Fixes #2135


johnynek:
I want to do better than this. I want to take all the constructor names from the kinds that aren't in the result, and I want to remove all the Exprs that refer to those constructors or to the types that aren't in the result. This gives the most actionable feedback per invocation of the compiler.

johnynek:
Implemented the requested refinement in `core/src/main/scala/dev/bosatsu/Package.scala`:
- Removed the `Ior.Both -> Left` hard stop.
- When kind inference returns partial results, we now compute the local types/constructors missing from the inferred set.
- We prune top-level lets that reference those removed constructors/types (including pattern constructor refs and type annotations), then prune transitive dependents.
- Name checking/type inference/unused-let checks now run on the pruned let set, so we avoid confusing unknown-constructor cascades while still surfacing independent errors in the same compile.

Added coverage in `core/src/test/scala/dev/bosatsu/ErrorMessageTest.scala`:
- Existing regression for no unknown-constructor cascade still holds.
- New regression verifies independent lets still report actionable errors (`missing_name`) alongside kind inference failures.

Required pre-push check was run and passed:
- `scripts/test_basic.sh`

johnynek:
- workflow run: `22985454543` (https://github.com/johnynek/bosatsu/actions/runs/22985454543)
- flaky-test issue: https://github.com/johnynek/bosatsu/issues/2137

core/src/main/scala/dev/bosatsu/Package.scala
```diff
@@ -488,23 +488,149 @@ object Package {
...
               necError =>
                 necError
                   .map(PackageError.KindInferenceError(p, _, typeDefRegions))
                   .toNonEmptyList,
-              infDTs => ParsedTypeEnv(infDTs, parsedTypeEnv.externalDefs)
+              infDTs => ParsedTypeEnv(infDTs, parsedTypeEnv0.externalDefs)
             )
 
         inferVarianceParsed.flatMap { parsedTypeEnv =>
+          val parsedTypeConsts = parsedTypeEnv.allDefinedTypes.iterator
```
johnynek: let's pull this all out into a private function. There is too much going on in one nested function here.

core/src/main/scala/dev/bosatsu/Package.scala
```diff
@@ -488,23 +488,149 @@ object Package {
...
+          val removedTypeDefs =
+            parsedTypeEnv0.allDefinedTypes.filterNot { dt =>
+              parsedTypeConsts(dt.toTypeConst)
+            }
+          val removedTypeConsts = removedTypeDefs.iterator.map(_.toTypeConst).toSet
+          val removedConstructors = removedTypeDefs.iterator
+            .flatMap(_.constructors.iterator.map(_.name))
+            .toSet
+
+          def hasRemovedTypeConst(tpe: Type): Boolean =
```
johnynek: let's make a private case class that has the removedTypeConsts and removedConstructors and then it has the methods below: hasRemovedTypeConst, hasRemovedPatternRef, hasRemovedExprRef.

core/src/main/scala/dev/bosatsu/Package.scala
johnynek: Refactored this out of `inferBodyUnopt` into private helpers in `Package`.

`inferBodyUnopt` now calls `letsForTypeChecking(p, parsedTypeEnv0, parsedTypeEnv, lets)` and no longer contains the large nested pruning logic. The extraction includes `removedKindRefs(...)` and `transitiveDependents(...)`, so the main inference flow is much smaller and easier to read.

core/src/main/scala/dev/bosatsu/Package.scala
johnynek: Implemented this as requested with a private case class:

`RemovedKindRefs(pack, removedTypeConsts, removedConstructors)`

It now owns:
- `hasRemovedTypeConst`
- `hasRemovedPatternRef`
- `hasRemovedExprRef`

The construction of that case class is handled in a private `removedKindRefs(...)` helper, and `letsForTypeChecking(...)` uses it to prune kind-invalid lets plus transitive dependents.

johnynek:
Applied the requested structural refactor in `core/src/main/scala/dev/bosatsu/Package.scala` (no behavior change intended). Required pre-push check passed: `scripts/test_basic.sh`.

johnynek:
Please merge origin main and deal with the merge conflicts.

johnynek:
Merged `origin/main` into the PR branch and resolved the conflict in `core/src/test/scala/dev/bosatsu/ErrorMessageTest.scala` by keeping both sets of coverage: the PR’s variance-regression tests and main’s new recursive-variance hint test. Verified the merged state with `scripts/test_basic.sh` (pass).

PR #2138, opened 2026-03-12
Make TypedExpr traverse/map/folds stack-safe

### LINKED ISSUE #2132: Make Traverse[TypedExpr] stack safe
Reimplement traverse, map, foldLeft, and foldRight with an explicit stack / frame machine.
Keep public signatures unchanged.
Move the current recursive implementation into tests as an oracle and compare results.
Acceptance: deep let/app trees no longer overflow under small-stack JVM tests, and existing TypedExprTest laws still pass.

part of #2131

johnynek:
## Summary
- make TypedExpr traverse/map/folds stack-safe
- add deep-tree tests and oracle checks to validate behavior

Fixes #2132

### DIFF
--- a/core/src/main/scala/dev/bosatsu/TypedExpr.scala
+++ b/core/src/main/scala/dev/bosatsu/TypedExpr.scala
@@ -1922,24 +1922,12 @@ object TypedExpr {
 
   implicit val traverseTypedExpr: Traverse[TypedExpr] =
     new Traverse[TypedExpr] {
-      private def sequenceListFromReversed[F[_]: Applicative, A](
-          reversed: List[F[A]]
-      ): F[List[A]] = {
-        val app = Applicative[F]
-        var acc: F[List[A]] = app.pure(Nil)
-        var cursor = reversed
-        while (cursor.nonEmpty) {
-          val h = cursor.head
-          cursor = cursor.tail
-          acc = app.map2(h, acc)(_ :: _)
-        }
-        acc
-      }
-
-      private def rebuild[F[_]: Applicative, A, B](
-          root: TypedExpr[A]
-      )(fn: A => F[B]): F[TypedExpr[B]] = {
-        val app = Applicative[F]
+      def traverse[F[_]: Applicative, T, S](
+          typedExprT: TypedExpr[T]
+      )(fn: T => F[S]): F[TypedExpr[S]] = {
+        type A = T
+        type B = S
+        val root = typedExprT
 
         sealed trait Work
         case class Visit(expr: TypedExpr[A]) extends Work
@@ -2065,9 +2053,10 @@ object TypedExpr {
                 idx = idx + 1
               }
               val fnF = popBuilt()
-              val argsF = app.map(
-                sequenceListFromReversed(argFsRev.result())
-              )(args => NonEmptyList.fromListUnsafe(args))
+              val argsF =
+                argFsRev.result().reverse.sequence.map(args =>
+                  NonEmptyList.fromListUnsafe(args)
+                )
 
               pushBuilt(
                 (fnF, argsF, fn(tag)).mapN((fn1, args1, tag1) =>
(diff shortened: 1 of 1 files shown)


core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -1922,168 +1922,337 @@ object TypedExpr {
...
-            (exp.traverse(fn), in.traverse(fn), fn(tag)).mapN {
-              Let(v, _, _, rec, _)
-            }
-          case Loop(args, body, tag) =>
-            (
-              args.traverse { case (v, expr) =>
-                expr.traverse(fn).map((v, _))
-              },
-              body.traverse(fn),
+      private def sequenceListFromReversed[F[_]: Applicative, A](
```
johnynek: Remove this. We can trust the built in List.sequence in cats. That one creates a balanced tree internally to make the map2 tree log depth. This code has linear depth.

core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -1922,168 +1922,337 @@ object TypedExpr {
...
+        var cursor = reversed
+        while (cursor.nonEmpty) {
+          val h = cursor.head
+          cursor = cursor.tail
+          acc = app.map2(h, acc)(_ :: _)
+        }
+        acc
+      }
+
+      private def rebuild[F[_]: Applicative, A, B](
```
johnynek: Why is this rebuild? The type signature looks exactly like traverse.

core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -1922,168 +1922,337 @@ object TypedExpr {
...
-            }
-            (expr.traverse(fn), tbranch, fn(tag)).mapN(Match(_, _, _))
+          }
         }
+      }
+
+      def traverse[F[_]: Applicative, T, S](
+          typedExprT: TypedExpr[T]
+      )(fn: T => F[S]): F[TypedExpr[S]] =
+        rebuild[F, T, S](typedExprT)(fn)
```
johnynek: Why this indirection? Why not just make the original called traverse?

core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -1922,168 +1922,337 @@ object TypedExpr {
...
+      ): Eval[B] = {
+        var reverseTags: List[A] = Nil
+        foreachTagPostOrder(typedExprA) { a =>
+          reverseTags = a :: reverseTags
+        }
+
+        var out = lb
+        var rem = reverseTags
+        while (rem.nonEmpty) {
+          out = f(rem.head, out)
```
johnynek: Eval is already stack safe. Once we have this we could do a loop to regain some laziness:

```
def loop[A](items: List[A]): Eval[B] =
  items match {
    case Nil => lb
    case h :: t =>
      f(h, Eval.defer(loop(t)))
  }
```

core/src/main/scala/dev/bosatsu/TypedExpr.scala
johnynek: Done. I removed the custom `sequenceListFromReversed` helper and switched all affected spots to Cats list sequencing (`...reverse.sequence`) so we rely on the built-in implementation.

core/src/main/scala/dev/bosatsu/TypedExpr.scala
johnynek: Good call. I removed the separate `rebuild` helper and made the stack-machine implementation the `traverse` method directly.

core/src/main/scala/dev/bosatsu/TypedExpr.scala
johnynek: Agreed. The indirection is removed now: `traverse` contains the implementation directly, and `map` calls `traverse[Id, A, B](...)`.

core/src/main/scala/dev/bosatsu/TypedExpr.scala
johnynek: Implemented. `foldRight` now rebuilds the tag order and uses a deferred recursive loop:
`f(h, Eval.defer(loop(tail)))`
which restores laziness while staying stack-safe.

johnynek:
Addressed all four review comments in `core/src/main/scala/dev/bosatsu/TypedExpr.scala` and reran the required check: `scripts/test_basic.sh` passes.

PR #2142, opened 2026-03-12
Rewrite TypedExpr.traverseType stack-safely with oracle parity tests

### LINKED ISSUE #2139: Rewrite TypedExpr.traverseType in a stack safe way.
Move the current implementation into tests as an oracle to test the new implementation against. The new implementation should be tested with a non commutative applicative, something like `Writer[Chain[Type], A]` to put all the types we find into a Chain.

We should use scalacheck to do this test: for all TypedExpr values when we run traverseType on the old and new with a writer, we get the same Chain of Types (by `Eq[Chain[Type]]`).

part of #2131

johnynek:
Implemented issue #2139 with focused changes in `TypedExpr` and its test suite.

- Rewrote `TypedExpr.traverseType` in `core/src/main/scala/dev/bosatsu/TypedExpr.scala` to use an explicit iterative work stack (stack-safe), while preserving traversal/effect order and Generic shadowing behavior from the previous recursive implementation.
- Moved the previous recursive `traverseType` logic into `core/src/test/scala/dev/bosatsu/TypedExprTest.scala` as `RecursiveTraverseTypeOracle`.
- Added a ScalaCheck property test using non-commutative `Writer[Chain[Type], A]` that checks old vs new traversal produce the same `Chain[Type]` (via `Eq[Chain[Type]]`) across generated `TypedExpr` values, and also checks transformed expressions match.
- Extended the deep-stack traversal safety test to exercise `traverseType` on deep trees.

Validation run:
- `scripts/test_basic.sh` (pass)

Fixes #2139

### DIFF
--- a/core/src/main/scala/dev/bosatsu/TypedExpr.scala
+++ b/core/src/main/scala/dev/bosatsu/TypedExpr.scala
@@ -1219,6 +1219,290 @@ object TypedExpr {
     loop(te, None)
   }
 
+  private trait StackTraverseHandler[F[_], A, B, C] {
+    type BranchPattern = Pattern[(PackageName, Constructor), Type]
+
+    def childContextForGeneric(quant: Quantification, ctx: C): C
+    def local(
+        name: Bindable,
+        tpe: Type,
+        tag: A,
+        ctx: C
+    ): F[TypedExpr[B]]
+    def global(
+        pack: PackageName,
+        name: Identifier,
+        tpe: Type,
+        tag: A,
+        ctx: C
+    ): F[TypedExpr[B]]
+    def literal(
+        lit: Lit,
+        tpe: Type,
+        tag: A,
+        ctx: C
+    ): F[TypedExpr[B]]
+    def generic(
+        quant: Quantification,
+        exprType: Type,
+        ctx: C,
+        inF: F[TypedExpr[B]]
+    ): F[TypedExpr[B]]
+    def annotation(
+        coerce: Type,
+        qev: Option[QuantifierEvidence],
+        ctx: C,
+        termF: F[TypedExpr[B]]
+    ): F[TypedExpr[B]]
+    def annotatedLambda(
+        args: NonEmptyList[(Bindable, Type)],
+        tag: A,
+        lamType: Type,
+        ctx: C,
+        resF: F[TypedExpr[B]]
+    ): F[TypedExpr[B]]
+    def app(
+        result: Type,
+        tag: A,
+        ctx: C,
+        fnF: F[TypedExpr[B]],
+        argsF: F[NonEmptyList[TypedExpr[B]]]
+    ): F[TypedExpr[B]]
+    def let(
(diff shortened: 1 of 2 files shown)


core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -1501,74 +1501,279 @@ object TypedExpr {
...
+              ))
+            )
+          case RebuildApp(tpe, tag, argCount, shadowed) =>
+            work = work.tail
+            val argFsRev = List.newBuilder[F[TypedExpr[A]]]
+            var idx = 0
+            while (idx < argCount) {
+              argFsRev += popBuilt()
+              idx = idx + 1
+            }
```
johnynek: rather than take Gemini's suggestion, replace:
```
            val argFsRev = List.newBuilder[F[TypedExpr[A]]]
            var idx = 0
            while (idx < argCount) {
              argFsRev += popBuilt()
              idx = idx + 1
            }
```

by adding a function:
```scala
def popListReverse(argCount: Int): List[F[TypedExpr[A]]] = {
  var lst: List[F[TypedExpr[A]]] = Nil
  var idx = 0
  while (idx < argCount) {
    lst = popBuilt() :: lst
    idx = idx + 1
  }
  lst
}
```
since we always seem to reverse, just build back to front to save a reversal

johnynek: Implemented. I added `popListReverse(argCount: Int)` in `TypedExpr.traverseType` and replaced the repeated builder+reverse loops in `RebuildApp`, `RebuildLoop`, and `RebuildRecur`.

I also kept the original `RebuildApp` pop order (pop args first, then fn) so traversal/effect order is unchanged.

johnynek:
Applied the requested refactor and validated locally. Required pre-push check `scripts/test_basic.sh` now passes.

johnynek:
- workflow run: `23015441219` (https://github.com/johnynek/bosatsu/actions/runs/23015441219)
- flaky-test issue: https://github.com/johnynek/bosatsu/issues/2143

johnynek:
please merge origin main, resolve conflicts, and then look at the duplication in traverseUp, traverseType and traverse.

Can we have a shared function that shares most of the traversal code to avoid the duplication we see there?

It seems like we should be able to generalize to have a single function and then all three would call this single function with different parameters.

johnynek:
Merged `origin/main` into this branch, resolved conflicts in `TypedExpr.scala` and `TypedExprTest.scala`, and addressed the duplication request by introducing a shared traversal engine (`stackTraverseExpr` + `StackTraverseHandler`). `traverseType`, `traverseUp`, and `Traverse[TypedExpr].traverse` now delegate to this shared stack-walk implementation with method-specific handlers. I also kept and reconciled the test-side conflict changes (including `traverseUp` parity/stack-safety coverage). Validation: `scripts/test_basic.sh` passes locally after these changes.

Issue #2158, opened 2026-03-13
codecoverage for bosatsu


johnynek:
Design a solution to checking code coverage in bosatsu.

We want to be able to run tests and then get an output of code coverage optionally. Presumably such tracking would slow down test execution or compilation, so it should be behind a flag.

What are the most widely supported code coverage formats? What information do we need to retain to support the 2-3 most common formats.

I imagine a solution will need to retain regions of the specific code that each matchless ast node came from, but perhaps there is a better way.

Read the current code and the state of the art in code-coverage measurement in other languages and propose a clear solution.

johnynek:
MergeXO assigned an agent and started design work for issue #2158.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/2159

PR #2159, opened 2026-03-13
Design doc for #2158: Flag-Gated Code Coverage for Bosatsu Test Execution

johnynek:
Design doc.

Refs #2158


docs/design/2158-codecoverage-for-bosatsu.md
```diff
@@ -0,0 +1,174 @@
...
+
+Chosen output plan:
+1. Implement LCOV and Cobertura in this issue.
+2. Keep internal data model JaCoCo-compatible so JaCoCo XML can be added without changing instrumentation.
+
+## Coverage Data We Must Retain
+For each executed probe:
+1. Source package name.
+2. Source region (`start`, `end` offsets).
+3. Probe kind (`line` probe, `branch alternative` probe).
```
johnynek: why isn't this implied by keeping track of the source regions executed? For instance if you execute the branch expr for a `match` then you have taken that branch. How is it different from the source region?

docs/design/2158-codecoverage-for-bosatsu.md
```diff
@@ -0,0 +1,174 @@
...
+
+When coverage is enabled, compile with `NoOptimize` via `RuntimeCommandSupport` to reduce source-to-execution distortion from aggressive rewrites.
+
+### 2) Static Probe Registration During Matchless Lowering
+Instrument at lowering time, not by changing Bosatsu language semantics:
+1. Extend `Matchless.fromLet` with optional coverage recorder callbacks.
+2. For each compiled typed-expression root that has usable source region metadata, register a probe id against the produced Matchless expression object identity.
+3. For each typed `Match`, additionally register a branch decision group with one alternative per branch body.
+
+This avoids invasive AST-wide payload changes while still mapping execution to source regions.
```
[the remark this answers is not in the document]
johnynek: Please address the comment.

One idea could be to aggregrate a `RegionSet` for `Expr` values, so when we optimize two different source positions into a single bit of code, we can take that single bit with counting for executing two regions. Although, that's a hand waving argument.

docs/design/2158-codecoverage-for-bosatsu.md
```diff
@@ -0,0 +1,174 @@
...
+
+### 3) Runtime Hit Collection In Evaluator
```
johnynek, line 103: we should not slow down the normal MatchlessToValue to do this. It should be purely opt-in. So, when we aren't running coverage, there is no hit on performance.

docs/design/2158-codecoverage-for-bosatsu.md
```diff
@@ -0,0 +1,174 @@
...
+Extend `MatchlessToValue` with an optional expression-entry callback:
+1. On each expression evaluation, look up probe id by expression identity.
+2. Increment counter if present.
+3. No-op callback in default mode keeps current behavior.
+
+`LibraryEvaluation` owns the coverage session (static metadata + mutable counters) for the test run.
+
+### 4) Source Mapping And Filtering
+Coverage should only emit files we can map back to source paths:
+1. Use `(Path, PackageName)` mapping already returned by compilation (`RuntimeCommandSupport.packMap`) to map package to source file path.
```
johnynek: I think hashing the source with blake3 and keeping that as the key in Matchless source information could be a good idea. So `HashValue[Algo.Blake3]` and the RegionSet should be enough.

docs/design/2158-codecoverage-for-bosatsu.md
```diff
@@ -0,0 +1,174 @@
...
+3. `CoberturaWriter` renderer.
+
+`Output.TestOutput` gains optional coverage payload so test execution and coverage emission happen in one reporting pass.
+
+## Implementation Plan
+1. Add coverage option parsing to `TestCommand` and plumb configuration through runtime command flow.
+2. Update `RuntimeCommandSupport.packMap` to accept compile options from caller.
+3. Add coverage model classes (`CoverageModel`, `CoverageAggregator`) for static probes and runtime counters.
+4. Extend `Matchless.fromLet` and `MatchlessFromTypedExpr.compile` to register probes and branch groups during lowering.
+5. Extend `MatchlessToValue` with optional expression-entry callback.
```
johnynek: again, I think we  must be clear that we should have a zero cost impact on the normal path.

docs/design/2158-codecoverage-for-bosatsu.md
johnynek: Good point. I updated **Coverage Data We Must Retain** to make this explicit: executed regions do imply a branch was taken, and the extra decision-group metadata is only to preserve the denominator/untaken alternatives required by branch-report formats.

docs/design/2158-codecoverage-for-bosatsu.md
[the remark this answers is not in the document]
johnynek: Addressed. I adopted the `RegionSet` direction in **Static Probe Registration During Matchless Lowering**: coverage metadata is now `(sourceHash, RegionSet)`, and rewrites/merges explicitly union RegionSets so one optimized expression can count for multiple source regions.

docs/design/2158-codecoverage-for-bosatsu.md
johnynek: Updated. The **Runtime Hit Collection In Evaluator** section now states a strict zero-cost normal path: non-coverage runs keep the existing `MatchlessToValue` path unchanged, and coverage uses a separate evaluator path only when `--coverage` is enabled.

docs/design/2158-codecoverage-for-bosatsu.md
johnynek: Integrated this suggestion. The doc now uses `HashValue[Algo.Blake3]` as the source key and describes `hash -> (path, LocationMap)` resolution for reporting.

docs/design/2158-codecoverage-for-bosatsu.md
johnynek: Clarified again in both architecture and plan/acceptance sections: normal runs must not pay coverage overhead, and coverage logic is isolated to the explicit opt-in path.

johnynek:
Updated `docs/design/2158-codecoverage-for-bosatsu.md` to address all review comments (branch-vs-region clarification, RegionSet merge model, Blake3 source keying, and explicit zero-cost non-coverage path). Ran required pre-push check: `scripts/test_basic.sh` passed (`Passed: Total 64, Failed 0, Errors 0`).

johnynek:
Merged `origin/main` into the PR branch. Post-merge verification passed with `scripts/test_basic.sh` (`Passed: Total 67, Failed 0, Errors 0`).

johnynek:
MergeXO feedback automation is blocked because the agent returned `commit_message` but no new staged changes or local commits were detected.

Action: request concrete file edits (or explicit `git_ops`), then reset blocked feedback state.

PR #2160, opened 2026-03-13
Remove implicit package names and delete --package_root/--search

### LINKED ISSUE #2156: remove implicit package names
Bosatsu supports implicit package names when a package root is passed. This was designed to allow a more python-like experience. But this violates a few of the bosatsu design principles:

1. there is more than one way to name a package <- bad. There should ideally be one.
2. just looking at the source code you can't tell what the package name is potentially <- bad. We shouldn't need to know the filesystem position to determine the package.
3. package names can be implicit. <- bad. Names should be explicit. We need to know names to compose, those shouldn't be implicitly computed.

The packages which are using this feature should get an explicit `package` declaration at the top, and we should refactor the parsing code to not support automatic packages based on the package_root. We should remove any cli options that are unneeded without this feature.

johnynek:
Implemented issue #2156 per design: package declarations are now mandatory at parse boundaries, path-derived package inference was removed, and CLI resolver options no longer accept `--package_root`/`--search`.

Code changes:
- `Package.parser`/`Package.headerParser` now require explicit `package` declarations (removed default package parameter flow).
- Removed `PlatformIO.pathPackage` from core API and implementations (`IOPlatformIO`, `Fs2PlatformIO`, `MemoryMain`).
- Simplified `PackageResolver` to explicit-input parsing only; removed local-root implicit naming/search behavior.
- Updated command wiring (`CommonOpts`, `CheckCommand`, `library/Command`, `MarkdownDoc`) to the explicit resolver/parser contract.
- Updated unknown-import hint text to remove `--package_root` guidance.

Fixture/tooling/docs migration:
- Added explicit `package` lines to `test_workspace/Foo.bosatsu`, `Bar.bosatsu`, and `Quicksort.bosatsu`.
- Removed `--package_root`/`--search` usage from JS API/UI wrappers, shell scripts (`test_cli.sh`, `test_python.sh`, `testc`, C test scripts), and docs (`transpile_python.md`).
- Updated workflow source fixtures and `.github/workflows/ci.yml` to keep JSON/YAML parity after flag removal.

Test updates:
- Reworked `PathModuleTest` to drop removed `pathPackage` behavior tests, add option-rejection coverage, and make explicit dependency inputs where needed.
- Updated `LibBuildImplicitPackageTest` fixtures to explicit package declarations.
- Updated `TestUtils` and `ToolAndLibCommandTest` fixture handling so embedded source files are explicit-package compliant.
- Mechanical parser callsite migration across tests (`Package.parser` no-arg usage).

Validation run (all passing):
- `sbt "coreJVM/test:compile" "cli/test:compile"`
- `sbt "coreJVM/test" "cli/test"`
- `scripts/test_basic.sh`

Fixes #2156

Implements design doc: [docs/design/2156-remove-implicit-package-names.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/2156-remove-implicit-package-names.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/2157

### DIFF
--- a/jsui/src/main/scala/dev/bosatsu/jsui/Store.scala
+++ b/jsui/src/main/scala/dev/bosatsu/jsui/Store.scala
@@ -1,8 +1,10 @@
 package dev.bosatsu.jsui
 
 import cats.data.Chain
+import cats.data.Validated
 import cats.effect.{IO, Resource}
-import dev.bosatsu.{MemoryMain, Package, Test, rankn}
+import cats.parse.{Parser => P}
+import dev.bosatsu.{MemoryMain, Package, Parser, Test, rankn}
 import dev.bosatsu.tool.Output
 import org.typelevel.paiges.{Doc, Document}
 import org.scalajs.dom.window.localStorage
@@ -13,6 +15,21 @@ object Store {
   private type ErrorOr[A] = Either[Throwable, A]
   val memoryMain = MemoryMain[ErrorOr]
   private val toolPrefix = List("tool")
+  private val webDemoPackageName = "WebDemo"
+  private val webDemoPath = Chain("root", webDemoPackageName)
+  private val optionalHeaderParser =
+    Package.headerParser.? <* P.anyChar.rep0
+
+  private[jsui] def withWebDemoPackage(str: String): String =
+    // In the web playground we accept snippets without an explicit package.
+    // If there is no parseable header, inject a stable package name for tool args.
+    Parser.parse(optionalHeaderParser, str) match {
+      case Validated.Valid((_, Some(_))) => str
+      case Validated.Valid((_, None)) =>
+        s"package $webDemoPackageName\n\n$str"
+      case Validated.Invalid(_) =>
+        str
+    }
 
   private def toolCommandArgs(
       subcommand: String,
@@ -94,9 +111,10 @@ object Store {
     val start = System.currentTimeMillis()
     println(s"starting $cmd: $start")
     val (args, handler) = cmdHandler(cmd)
+    val source = withWebDemoPackage(str)
 
     val res = memoryMain.runWith(
-      files = Map(Chain("root", "WebDemo") -> str)
+      files = Map(webDemoPath -> source)
     )(args) match {
       case Right(good) => handler(good)
       case Left(err)   =>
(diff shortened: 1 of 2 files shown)


core/src/main/scala/dev/bosatsu/Package.scala
```diff
@@ -383,7 +383,7 @@ object Package {
         Doc.intercalate(Doc.empty, p :: i :: e :: b)
     }
 
-  def headerParser(defaultPack: Option[PackageName]): P0[Header] = {
+  def headerParser: P0[Header] = {
```
johnynek: This should be a P not a P0. Headers are never zero size now.

johnynek: Updated this in `Package.headerParser`: it now returns `P[Header]` (not `P0`). I also rewrote the combinator from `.tupled` to `((parsePack ~ im) ~ Parser.nonEmptyListToList(ex)).map { ... }` so the composed parser remains non-empty and typechecks.

johnynek:
Addressed the review feedback with a concrete code change in `core/src/main/scala/dev/bosatsu/Package.scala` and reran the required pre-push check `scripts/test_basic.sh` successfully (`Passed: Total 60, Failed 0, Errors 0`).

johnynek:
We should still support no package name in the web UI. We can do this by trying to parse the string with the Header parser .? And if we get None, meaning not an invalid header just not a valid header either then prepend a fake package like package WebDemo

Or something like that.

johnynek:
Implemented the PR issue-comment feedback for web UI package-less code. In `jsui/src/main/scala/dev/bosatsu/jsui/Store.scala`, `Store.run` now normalizes source via `withWebDemoPackage`: it parses with `Package.headerParser.? <* anyChar.rep0` and prepends `package WebDemo` only when no header is present; invalid headers are left unchanged so normal parser errors still surface. Added coverage in `jsui/src/test/scala/dev/bosatsu/jsui/StoreTest.scala` for header injection and eval success on package-less source. Validation run: `sbt -batch "jsuiJS/testOnly dev.bosatsu.jsui.StoreTest"` passed, and required pre-push `scripts/test_basic.sh` passed (`Passed: Total 60, Failed 0, Errors 0`).

johnynek:
- workflow run: `23073490396` (https://github.com/johnynek/bosatsu/actions/runs/23073490396)
- flaky-test issue: https://github.com/johnynek/bosatsu/issues/2161

PR #2168, opened 2026-03-15
Use raw source hashes to skip parsing on cache hits

johnynek:
## Summary
- load source files as header + raw-byte hash units and defer full parsing until an infer-cache miss
- add a raw-hash-based cache-key path while keeping CAS content-addressed by compiled output
- use header-only parsing for `lib check --filter` package selection instead of full package parsing

## Performance
Measured on `./bosatsuj lib check -n core_alpha` after rebuilding `cli/assembly`.

Before:
- warm no-op: `2.86-2.95s`
- tiny implementation-only edit with unchanged interface: `3.37s` (`65/66` hits)
- `--filter Bosatsu/Option`: `46.11s`

After:
- first run after key-shape change: `5.31s` (`66/66` misses, expected cold rebuild)
- warm no-op: `2.58-2.71s` (`66/66` hits)
- tiny implementation-only edit with unchanged interface: `2.62s` (`65/66` hits, `1` miss)
- `--filter Bosatsu/Option`: `2.44s` (`2/2` hits)

This intentionally changes comment-only edits for file sources from key-stable to key-changing. The compiled CAS blob is still reused, and the actual implementation-edit loop gets faster because unchanged files stop paying a full parse before the cache lookup.

## Testing
- `sbt --batch 'coreJVM/compile' 'cli/compile'`
- `sbt --batch 'cli/testOnly dev.bosatsu.IOPlatformIOTest' 'coreJVM/testOnly dev.bosatsu.tool.CompileCacheTest'`
- `sbt --batch 'coreJVM/testOnly dev.bosatsu.ToolAndLibCommandTest'`
- `sbt --batch cli/assembly`

### DIFF
--- a/core/src/main/scala/dev/bosatsu/tool/PackageResolver.scala
+++ b/core/src/main/scala/dev/bosatsu/tool/PackageResolver.scala
@@ -37,6 +37,8 @@ sealed abstract class PackageResolver[IO[_], Path] {
               cats.data.Validated.invalidNel(PathParseError.FileError(path, err))
             )
           case Right(bytes) =>
+            // Decoding, raw hashing, and header parsing can add up across many
+            // files, so keep that pure CPU work off the caller thread.
             canPromiseF.compute {
               val source = new String(bytes, StandardCharsets.UTF_8)
               val rawHash = Algo.hashBytes[Algo.Blake3](bytes)
@@ -54,6 +56,8 @@ sealed abstract class PackageResolver[IO[_], Path] {
                     exports = exports,
                     sourceHash = rawHash,
                     loadParsed =
+                      // On a cache miss we need the full parse; defer that
+                      // heavier pure work to the compute pool.
                       canPromiseF.compute {
                         PathParseError
                           .parseString(Package.parser, path, source)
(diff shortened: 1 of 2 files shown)


core/src/main/scala/dev/bosatsu/tool/PackageResolver.scala
```diff
@@ -4,29 +4,59 @@ import cats.data.{NonEmptyList, ValidatedNel}
...
   )(platformIO: PlatformIO[IO, Path]): IO[
     ValidatedNel[PathParseError[Path], NonEmptyList[
-      ((Path, LocationMap), Package.Parsed)
+      PackageResolver.SourceFile[IO, Path]
     ]]
   ] = {
-    import platformIO.moduleIOMonad
+    import platformIO.{canPromiseF, moduleIOMonad}
     if (_included.nonEmpty) ()
+    val headerParser = Package.headerParser <* P.anyChar.rep0
```
johnynek: We make this twice. Can we define it once: `headerParserIgnoreRest` or something. And make it a `val` in cats.parse creating a parser has runtime work to optimize the cost of the parser. We want to cache them.

core/src/main/scala/dev/bosatsu/tool/PackageResolver.scala
```diff
@@ -4,29 +4,59 @@ import cats.data.{NonEmptyList, ValidatedNel}
...
+              case (lm, (packageName, imports, exports)) =>
+                PackageResolver.SourceFile(
+                  path = path,
+                  locationMap = lm,
+                  packageName = packageName,
+                  imports = imports,
+                  exports = exports,
+                  sourceHash = rawHash,
+                  loadParsed =
+                    canPromiseF.delay {
```
johnynek: maybe this should be `compute` instead of `delay to make sure we aren't blocking an IO thread.

PR #2176, opened 2026-03-16
Fix `lib test --filter` hiding compile errors as "no tests found"

### LINKED ISSUE #2170: lib test --filter when code doesn't compile gives no error
If you have Zafu/Testing/HedgeHog and there is a compilation error, then you do:
```
oscar@mac zafu % ./bosatsu lib test --filter Zafu/Testing/HedgeHog

no tests found in:
using regexes: Zafu/Testing/HedgeHog
```
which is sort of true, there are no compiling tests, but we should definitely compile all the tests reachable from the filter, and if any don't compile, we should error. We can't silently just report no tests when actually it isn't compiling.

johnynek:
Updated filtered source selection in `library/Command.scala` so parse-failing `.bosatsu` files are no longer silently dropped when they match (or are reachable from) the filter. The filter flow now keeps header-parse results, falls back to package-name inference from source paths for invalid headers, and includes those files in the selected compile set so parse/compile errors surface correctly. Also updated filter match validation to include inferred package names from invalid files. Added a regression test in `ToolAndLibCommandTest` (`lib test --filter surfaces parse failures in matched packages instead of no tests found`) that reproduces the prior silent behavior and verifies we now error with parse details instead of `no tests found`. Ran required pre-push tests: `scripts/test_basic.sh` (pass).

Fixes #2170

### DIFF
--- a/core/src/main/scala/dev/bosatsu/library/Command.scala
+++ b/core/src/main/scala/dev/bosatsu/library/Command.scala
@@ -525,31 +525,30 @@ object Command {
       private def failHeaderParseErrors[A](
           errs: NonEmptyList[PathParseError[P]]
       ): F[A] = {
-        val messages: List[String] =
+        val messageDocs: List[Doc] =
           errs.toList.flatMap {
             case PathParseError.ParseFailure(pf, path) =>
               val (r, c) = pf.locations.toLineCol(pf.position).get
-              val ctx = pf.showContext(Colorize.None)
               List(
-                s"failed to parse $path:${r + 1}:${c + 1}",
-                ctx.render(80)
+                Doc.text(s"failed to parse $path:${r + 1}:${c + 1}"),
+                pf.showContext(Colorize.None)
               )
             case PathParseError.FileError(path, err) =>
               err match {
                 case e
                     if e.getClass.getName == "java.nio.file.NoSuchFileException" =>
                   // This class isn't present in scalajs, use the String
-                  List(s"file not found: $path")
+                  List(Doc.text(s"file not found: $path"))
                 case _ =>
                   List(
-                    s"failed to parse $path",
-                    err.getMessage,
-                    err.getClass.toString
+                    Doc.text(s"failed to parse $path"),
+                    Doc.text(Option(err.getMessage).getOrElse(err.toString)),
+                    Doc.text(err.getClass.toString)
                   )
               }
           }
-        val messageString = messages.mkString("\n")
-        val errDoc = Doc.intercalate(Doc.hardLine, messages.map(Doc.text(_)))
+        val errDoc = Doc.intercalate(Doc.hardLine, messageDocs)
+        val messageString = errDoc.render(80)
         moduleIOMonad.raiseError(CliException(messageString, err = errDoc))
       }
 


core/src/main/scala/dev/bosatsu/library/Command.scala
```diff
@@ -509,25 +509,46 @@ object Command {
...
+          .map { parsedBySource =>
+            val parsedPackages =
+              parsedBySource.collect {
+                case (_, Validated.Valid((pack, _))) => pack
+              }
+            val inferredPackages =
+              parsedBySource.collect {
+                case (path, Validated.Invalid(_)) =>
+                  inferredPackageForSourcePath(path)
+              }.flatten
```
johnynek: it looks like we are just discarding the paths that don't parse and we can't derive a package name guess for.

I don't like this algorithm. If we can't even parse all the package headers, we need to stop at that point. We need to be sure we understand the dependency graph before we can implement this filter feature. And the main use case for filter is when you are editing items that are being tested, so they would be in the filter set. So, a failed parse is likely to be in the set of concern.

Let's change the code so that failure to parse the headers fails the whole process.

core/src/main/scala/dev/bosatsu/library/Command.scala
```diff
@@ -567,9 +588,14 @@ object Command {
...
                         parsedBySource.collect {
-                          case (path, Some((pack, imports))) =>
+                          case (path, Validated.Valid((pack, imports))) =>
                             (path, pack, imports)
                         }
+                      val inferredEntries =
+                        parsedBySource.collect {
+                          case (path, Validated.Invalid(_)) =>
+                            inferredPackageForSourcePath(path).map(path -> _)
+                        }.flatten
```
johnynek: same comment relevant here: we have to fail the run when we can't parse the headers.

core/src/main/scala/dev/bosatsu/library/Command.scala
johnynek: Updated per feedback: filtered flows now fail immediately if any source header cannot be parsed.

I removed the fallback behavior that continued after invalid headers and added strict metadata loading (`sourceFileMetadata`) that sequences all header parses and raises on any parse/file error. `sourcePackages` now uses this strict path as well.

core/src/main/scala/dev/bosatsu/library/Command.scala
johnynek: Applied here too: filtered dependency-graph selection now depends on strict header metadata (`sourceFileMetadata(inputSrcs)`), so we stop before filtering whenever any header parse fails.

This guarantees we only compute the filtered graph when header parsing succeeds for all source files.

johnynek:
Implemented the requested strict header-parse failure behavior in `core/src/main/scala/dev/bosatsu/library/Command.scala` and updated regression coverage in `core/src/test/scala/dev/bosatsu/ToolAndLibCommandTest.scala` (`lib test --filter fails when any source header fails to parse`).

Also investigated the reported `NameMapTest` failure: reran `coreJVM/testOnly dev.bosatsu.pattern.NameMapTest` locally (pass) and reran required pre-push command `scripts/test_basic.sh` in this turn (pass).

core/src/main/scala/dev/bosatsu/library/Command.scala
```diff
@@ -509,25 +509,74 @@ object Command {
...
+          .map(
+            _.map { case (_, (packageName, imports, _)) =>
+              (packageName, imports.map(_.pack))
+            }
+          )
+
+      private def failHeaderParseErrors[A](
+          errs: NonEmptyList[PathParseError[P]]
+      ): F[A] = {
+        val messages: List[String] =
```
johnynek: Return a List[Doc] so we don't have the inner render that is then passed back to Doc.text

johnynek: Updated this per request.

`failHeaderParseErrors` now builds a `List[Doc]` directly (`messageDocs`) instead of rendering context to `String` and wrapping back into `Doc.text`.

- parse failures now append `pf.showContext(Colorize.None)` directly
- file errors now build `Doc.text(...)` entries directly
- `errDoc` is built from `messageDocs`, and `messageString` is derived once via `errDoc.render(80)`

johnynek:
Applied the Doc-structure refactor in `core/src/main/scala/dev/bosatsu/library/Command.scala` and reran the required pre-push command. `scripts/test_basic.sh` passed.

Issue #2178, opened 2026-03-16
Design for inline def


johnynek:
In Scala 3, inline def can be very useful for designing custom control flow functions. In such cases, some arguments we won't need to evaluate in all or maybe even most cases.

It would be nice to have this feature in Bosatsu.

One difference in Bosatsu is that due to totality, order already doesn't matter. For local functions to a package we will often currently get inlining but there is no inlining across package boundaries currently.

Maybe the C backend will optimize some of this since maybe the compiler can see there aren't any side effects and reorder things.

But it would be nice to rely less on an optimizer when we know we have such control flow functions.

The idea, I think, would be that at the matchless phase all inline defs would be inlined during lowering to Matchless.Expr. Currently we can't resolve Matchless values in dependencies. We just generate Global references to the correct version number. That would need to change. It also means we can't be totally parallel anymore: if your dependencies have any inline defs those need to be lowered before you can use them.

johnynek:
MergeXO assigned an agent and started design work for issue #2178.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/2179

johnynek:
I think #2264 supercedes this. In a total language, all functions are good candidates for inlining. Adding to complexity of the language for now seems not worth it. Better to have good heuristics so the common use cases (such as inlining custom control flow functions or functions that get lambdas) are handled automatically.

PR #2179, opened 2026-03-16
Design doc for #2178: Top-level inline def lowered through Matchless

johnynek:
Design doc.

Refs #2178

### DIFF
--- a/docs/design/2178-design-for-inline-def.md
+++ b/docs/design/2178-design-for-inline-def.md
@@ -79,7 +79,7 @@ Contract:
 2. It type-checks exactly like an ordinary `def` and exports as an ordinary value.
 3. Fully saturated direct calls are candidates for Matchless-time substitution.
 4. Unsaturated uses, passing the function as a value, or unresolved dependency bodies keep ordinary `Global` and `App` lowering.
-5. Recursive or mutually recursive inline defs are rejected.
+5. Non-tail self-recursive inline defs are rejected. Tail-recursive inline defs are allowed once loop lowering has converted them to loop form.
 
 Restricting V1 to top-level defs avoids having to tag every local `Let` shape throughout the typed AST while still covering the motivating cross-package control-flow use case.
 
@@ -94,7 +94,7 @@ Proposed in-memory shape:
 1. Parse `inline def` as a distinct top-level statement form or a top-level-only modifier on `Statement.Def`.
 2. Extend `Program` with `inlineDefs: SortedSet[Bindable]`.
 3. Preserve `inlineDefs` through normalization, tree shaking, filtering, voiding, and round-trip helpers.
-4. Add a validation step after typing that rejects any inline def participating in a recursive SCC.
+4. No general SCC validation is needed. Bosatsu name graphs are DAGs by construction apart from self recursion, so the inline-specific validation only needs to reject remaining non-tail self recursion after loop lowering.
 
 ### 2. Public interface metadata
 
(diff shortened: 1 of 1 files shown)


docs/design/2178-design-for-inline-def.md
```diff
@@ -0,0 +1,244 @@
...
+### 1. Parsed and typed package metadata
+
+Introduce a top-level-only inline marker in the source AST, then preserve the resulting bindables in typed programs.
+
+Proposed in-memory shape:
+
+1. Parse `inline def` as a distinct top-level statement form or a top-level-only modifier on `Statement.Def`.
+2. Extend `Program` with `inlineDefs: SortedSet[Bindable]`.
+3. Preserve `inlineDefs` through normalization, tree shaking, filtering, voiding, and round-trip helpers.
+4. Add a validation step after typing that rejects any inline def participating in a recursive SCC.
```
johnynek: All names and types form dags by construction in Bosatsu. The only loops are self recursion in some limited cases. That simplifies this problem quite a bit.

Please rewrite in this context: we don't need to validate that there are no SCCs: the only loops in the graph are recursive functions calling themselves.

It seems like we can handle tail recursive functions (ones with loop not recur), since inlining those is just inlining a while loop.

docs/design/2178-design-for-inline-def.md
```diff
@@ -0,0 +1,244 @@
...
+
+1. If an inline parameter is referenced multiple times, the argument IR can be duplicated.
+2. That is acceptable in V1 because inline is explicit and opt-in.
+3. Any future heuristic for sharing should be a follow-up and must not reintroduce eager evaluation ahead of control flow.
+
+### 3. Dependency ordering and remaining parallelism
+
+The issue text is correct that compilation can no longer be totally parallel if lowered dependency bodies may be needed.
+
+The minimum change is to preserve parallelism by topo layer:
```
johnynek: What if we convert in full parallelism but then have an inlining phase that runs after that phase in a topo order.

docs/design/2178-design-for-inline-def.md
```diff
@@ -0,0 +1,244 @@
...
+
+## Implementation plan
+
+1. Add top-level `inline def` parsing and rendering in the statement layer.
+2. Extend `Program` with `inlineDefs` and thread it through package helpers, normalization, tree shaking, and typed-package utilities.
+3. Add a typed-package validation that rejects inline defs in recursive SCCs.
+4. Extend exported-value metadata so interfaces record whether an exported value is inline.
+5. Update `TypedAst.proto` and `ProtoConverter` to round-trip both package-level inline defs and interface-level inline export flags with backward-compatible decode.
+6. Refactor `MatchlessFromTypedExpr` into a dependency-aware package compiler that accepts and produces inline-lowering metadata.
+7. Add `Matchless.inlineApplyArgs` and call it from lowering of fully saturated direct applications to inline globals.
```
johnynek: Sketch how this works so we can have a clearer proposal

docs/design/2178-design-for-inline-def.md
johnynek: Updated the doc in that direction. It no longer proposes SCC validation; it now explicitly says Bosatsu's name graph is already acyclic apart from self recursion, and the validation is narrowed to non-tail self recursion that survives loop lowering. I also changed the contract/testing/acceptance sections to allow tail-recursive inline defs once `TypedExprLoopRecurLowering` has rewritten them to loop form.

docs/design/2178-design-for-inline-def.md
johnynek: Reworked the architecture section around this. The proposal now keeps the initial Matchless conversion fully parallel and moves inline substitution into a second topo-ordered post-pass over already-lowered Matchless bodies, so only the rewrite phase needs dependency ordering.

docs/design/2178-design-for-inline-def.md
johnynek: Added a dedicated `Post-lowering inlining sketch` section with the concrete Phase A / Phase B flow: raw parallel Matchless lowering, inline-candidate table construction, per-package rewrite order, publishing rewritten inline defs into `InlineEnv`, and the tail-recursive vs non-tail self-recursive handling.

johnynek:
Updated `docs/design/2178-design-for-inline-def.md` to reflect the review feedback: the design now assumes Bosatsu's usual DAG property, allows tail-recursive inline defs after loop lowering, and uses a two-phase pipeline with a topo-ordered post-lowering inlining pass instead of ordering the initial Matchless conversion. Required pre-push check passed locally: `scripts/test_basic.sh`.

Issue #2182, opened 2026-03-17
Add Prog::Var


johnynek:
Design a `Bosatsu/Prog::Var` type. It should be something like:
```
package Bosatsu/Prog

# this is an invariant, not covariant type. It definitely doesn't feel safe to allow recursion through this type.
external struct Var[a]

external def new_var[a](a: a) -> forall e. Prog[e, Var[a]]
# this is an atomic update on the BValue in C, an AtomicReference in MatchlessToValue
# may call the fn more than one time in concurrent settings, but since fn is pure, this is fine.
external def update[a, b](v: Var[a], fn: a -> (a, b)) -> forall e. Prog[e, b]
# this can be more efficient than update because it can skip the read.
external def set[a](v: Var[a], value: a) -> forall e. Prog[e, a]
# this can possibly be more efficient than update because it can avoid the tuple allocation
external def get[a](v: Var[a]) -> forall e. Prog[e, a]
external def swap[a](v: Var[a], new_value: a) -> forall e. Prog[e, a]
```

What non-external functions should we add that use `Var[a]` values?

johnynek:
MergeXO assigned an agent and started design work for issue #2182.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/2183

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #2182.

johnynek:
Opened implementation PR: https://github.com/johnynek/bosatsu/pull/2186

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/2182#issuecomment-4078383684
- linked PR: #2186 (https://github.com/johnynek/bosatsu/pull/2186)

Comments on the source issue are no longer actioned after a PR exists. Please comment on the PR thread instead.

PR #2183, opened 2026-03-17
Design doc for #2182: Add Prog::Var

johnynek:
Design doc.

Refs #2182

### DIFF
--- a/docs/design/2182-add-prog-var.md
+++ b/docs/design/2182-add-prog-var.md
@@ -94,14 +94,14 @@ No `core_alpha_conf.json` change is needed because `Bosatsu/Prog` is already exp
 ### 2. JVM evaluator implementation
 In `core/src/main/scala/dev/bosatsu/Predef.scala`:
 - add `FfiCall` registrations for the five new `Bosatsu/Prog` externals
-- add a new runtime representation, for example `final case class VarCell(state: AtomicReference[Value])`
-- add an `asVar` extractor analogous to `asLazy`
+- store the mutable cell directly as `ExternalValue(new AtomicReference[Value](initial))`
+- add an `asVar` extractor that unwraps `AtomicReference[Value]`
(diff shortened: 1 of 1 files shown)


docs/design/2182-add-prog-var.md
```diff
@@ -0,0 +1,207 @@
...
+In `core/src/main/scala/dev/bosatsu/Predef.scala`:
+- add `FfiCall` registrations for the five new `Bosatsu/Prog` externals
+- add a new runtime representation, for example `final case class VarCell(state: AtomicReference[Value])`
+- add an `asVar` extractor analogous to `asLazy`
+- implement `prog_new_var`, `prog_var_get`, `prog_var_set`, `prog_var_swap`, and `prog_var_update`
+
+Implementation details:
+- represent every operation as a `Prog` effect via existing `prog_effect` or `prog_effect2`
+- do not add a new `Prog` tag
+- `new_var` effect allocates `ExternalValue(VarCell(new AtomicReference(initial)))` and returns `prog_pure(cell)`
```
johnynek: why do we need the `VarCell` wrapper? Why not just `AtomicReference`?

johnynek: Updated the JVM implementation section to use `AtomicReference[Value]` directly instead of introducing a `VarCell` wrapper. The doc now says `new_var` stores `ExternalValue(new AtomicReference[Value](initial))`, and it explicitly notes that a wrapper is unnecessary unless we later need extra metadata on the cell.

johnynek:
Updated the design doc to address both review points:
- the JVM section now uses `AtomicReference[Value]` directly rather than a `VarCell` wrapper
- the C runtime section now makes the memory ordering explicit for `swap` and `update` (`atomic_exchange_explicit(..., memory_order_acq_rel)` and `atomic_compare_exchange_weak_explicit` with `memory_order_acq_rel` on success / `memory_order_acquire` on failure)

Also ran `scripts/test_basic.sh` locally and it passed.

PR #2186, opened 2026-03-17
Add Bosatsu/Prog::Var

### LINKED ISSUE #2182: Add Prog::Var

johnynek:
Implemented invariant `Bosatsu/Prog::Var` end to end: added the Bosatsu surface API and derived helpers, wired JVM evaluator support with `AtomicReference[Value]`, added Python externals with a locked `_BosatsuVar`, and added C runtime support with atomic load/store/CAS loops. Added Bosatsu/evaluator/CLI/codegen/variance/doc coverage, bumped `core_alpha` to `5.1.0` for the public API expansion, and fixed Python `FileKind` tags encountered while rerunning runtime checks. Re-read the design doc and diff against `main`, and verified with `./scripts/test_basic.sh`, `sbt -batch "doc; paradox"`, `./test_python.sh`, `./test_cli.sh`, `sbt -batch "coreJS/testOnly * -- --log=failure; jsapiJS/compile; jsuiJS/testOnly * -- --log=failure"`, `sbt -batch "jsuiJS/fullOptJS::webpack; cliJSJS/fullOptJS"` (with `NODE_OPTIONS=--openssl-legacy-provider`), `./bosatsu_node tool test ...`, plus C runtime/generated/lib build-test flows matching the sanitizer script minus its dirty-worktree guard.

Fixes #2182

Implements design doc: [docs/design/2182-add-prog-var.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/2182-add-prog-var.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/2183

### DIFF
--- a/core/src/test/scala/dev/bosatsu/EvaluationTest.scala
+++ b/core/src/test/scala/dev/bosatsu/EvaluationTest.scala
@@ -3305,7 +3305,7 @@ from Bosatsu/Prog import ProgTest, await, get, get_and_update, modify, new_var, 
 tests = ProgTest(_ -> (
     v <- new_var(1).await()
     initial <- get(v).await()
-    set_result <- set(v, 2).await()
+    _ <- set(v, 2).await()
     after_set <- get(v).await()
     swapped <- swap(v, 5).await()
     update_result <- update(v, current -> (add(current, 1), mul(current, 10))).await()
@@ -3318,7 +3318,6 @@ tests = ProgTest(_ -> (
         "prog var",
         [
           Assertion(initial matches 1, "initial"),
-          Assertion(set_result matches (), "set returns unit"),
           Assertion(after_set matches 2, "get after set"),
           Assertion(swapped matches 2, "swap returns old value"),
           Assertion(update_result matches 50, "update returns projection"),
@@ -3331,7 +3330,7 @@ tests = ProgTest(_ -> (
   ))
 """
 
-    runBosatsuTest(List(progPack, progVarPack), "ProgVarEval", 8)
+    runBosatsuTest(List(progPack, progVarPack), "ProgVarEval", 7)
   }
 
   if (Platform.isScalaJvm)


test_workspace/Prog.bosatsu
```diff
@@ -22,12 +23,29 @@ def ignore_err[err, res](prog: Prog[err, res], default: res) -> forall e. Prog[e
 external def apply_fix(a: a,
   fn: (a -> Prog[err, b]) -> (a -> Prog[err, b])) -> Prog[err, b]
 external def observe[a](a: a) -> forall err. Prog[err, Unit]
+external def new_var[a](a: a) -> forall e. Prog[e, Var[a]]
```
johnynek, line 30: All these exports functions need documentation in comments.

test_workspace/Prog.bosatsu
```diff
@@ -22,12 +23,29 @@ def ignore_err[err, res](prog: Prog[err, res], default: res) -> forall e. Prog[e
...
+external def update[a, b](v: Var[a], fn: a -> (a, b)) -> forall e. Prog[e, b]
+external def set[a](v: Var[a], value: a) -> forall e. Prog[e, a]
```
johnynek: This should return `Prog[e, Unit]`

Otherwise it's confusing in that it seems like it could return the previous value.

core/src/test/scala/dev/bosatsu/EvaluationTest.scala
```diff
@@ -3262,6 +3262,78 @@ tests = TestSuite("lazy eval", [
...
+    val progPack = Predef.loadFileInCompile("test_workspace/Prog.bosatsu")
+    val progVarPack = """
+package ProgVarEval
+
+from Bosatsu/Prog import ProgTest, await, get, get_and_update, modify, new_var, pure, set, swap, update, update_and_get
+
+tests = ProgTest(_ -> (
+    v <- new_var(1).await()
+    initial <- get(v).await()
+    wrote <- set(v, 2).await()
```
johnynek: This should return Unit.

c_runtime/bosatsu_ext_Bosatsu_l_Prog.c
```diff
@@ -59,6 +66,110 @@ BValue ___bsts_g_Bosatsu_l_Prog_l_recover(BValue p, BValue f)
...
+  return ___bsts_g_Bosatsu_l_Prog_l_pure(current);
+}
+
+static BValue bsts_prog_var_set_effect(BValue pair)
+{
+  BValue var_value = get_struct_index(pair, 0);
+  BValue next_value = get_struct_index(pair, 1);
+  BSTS_Prog_Var *cell = bsts_prog_unbox_var(var_value);
+  atomic_store_explicit(&cell->value, next_value, memory_order_release);
+  return ___bsts_g_Bosatsu_l_Prog_l_pure(next_value);
```
johnynek: We should return pure unit here.

test_workspace/Prog.bosatsu
johnynek: `test_workspace/Prog.bosatsu` now has comments documenting the exported `Var` API surface and helpers.

test_workspace/Prog.bosatsu
johnynek: `set` is now modeled as `Prog[e, Unit]` on the Bosatsu surface, with the runtime/docs/tests aligned to that contract.

core/src/test/scala/dev/bosatsu/EvaluationTest.scala
johnynek: Updated the Bosatsu-level test in `core/src/test/scala/dev/bosatsu/EvaluationTest.scala` to discard the `set` result and verify the post-`set` state instead. That keeps the test aligned with `Unit` return semantics and avoids the unreachable `matches ()` pattern.

c_runtime/bosatsu_ext_Bosatsu_l_Prog.c
johnynek: The C runtime `set` effect now returns pure unit instead of the written value.

johnynek:
Repaired the MergeXO failure in `scripts/test_basic.sh`: the Bosatsu fixture was asserting `set_result matches ()`, but that pattern is rejected as unreachable for a value already typed as `Unit`. The test now discards the `set` result, keeps the `get after set` state assertion, and expects 7 assertions instead of 8. `scripts/test_basic.sh` passes locally now.

Issue #2185, opened 2026-03-17
Design a matchless global inlining phase


johnynek:
Somewhat related to #2178

We should be able to do cross package inlining at the matchless layer before code generation.

Before we implement the idea of forced inlining we should actually be able to do heuristic based inlining. For instance when some expensive parameters aren't needed before a branch, or when a parameter is a literal lambda (or we can resolve to a lambda), basically any of the rules that also make sense in TypedExpr.

Optimizing matchless is a bit more powerful than optimizing on typedexpr because we don't need the same type constraints (for instance two functions or values that are structurally the same can be shared).

johnynek:
MergeXO assigned an agent and started design work for issue #2185.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/2187

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #2185.

johnynek:
Opened implementation PR: https://github.com/johnynek/bosatsu/pull/2264

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/2185#issuecomment-4145110955
- linked PR: #2264 (https://github.com/johnynek/bosatsu/pull/2264)

Comments on the source issue are no longer actioned after a PR exists. Please comment on the PR thread instead.

PR #2187, opened 2026-03-17
Design doc for #2185: Matchless Global Inlining Phase

johnynek:
Design doc.

Refs #2185

### DIFF
--- a/docs/design/2185-design-a-matchless-global-inlining-phase.md
+++ b/docs/design/2185-design-a-matchless-global-inlining-phase.md
@@ -94,7 +94,9 @@ Likely summary fields:
 2. arity and rewritten body
 3. expression weight / size estimate for budget checks
 4. boundary-safety flags describing whether the body, or any nested lambda/value it produces, captures a free `LocalAnonMut` from outside the candidate
-5. per-parameter demand facts: unused, eager, branch-guarded only, lambda-callee-only, and total use count
+5. per-parameter demand facts: unused, eager on the selector or guard path, deferrable into an `If` or `SwitchVariant` branch body, lambda-callee-only, and total use count
+
+The key summary fact is whether an argument is deferrable. A parameter is deferrable when every demand for it occurs inside an `If` or `SwitchVariant` branch body rather than on the selector or guard path. `Bosatsu/Predef.and` and `Bosatsu/Predef.or` are the canonical examples: after inlining, `and(x, y)` can become `if x: y; else False` and `or(x, y)` can become `if x: True; else y`, so `y` is deferrable in both cases. Because Bosatsu is pure and total, that deferral does not change semantics, but it can strictly lower runtime cost by avoiding work on untaken branches.
 
 V1 should only publish conservative candidates:
 
@@ -145,7 +147,7 @@ Hard gates:
 Benefit signals:
 
 1. a parameter is unused in the callee body
-2. every use of a parameter is inside a branch body rather than in the selector or guard path
+2. a non-cheap actual argument sits in a deferrable parameter position, meaning inlining can move that work under an `If` or `SwitchVariant` branch body instead of forcing it on every path
 3. a parameter is only used as a direct callee, and the actual argument is a literal lambda or resolves to a lambda after local alias resolution
 4. the candidate body is tiny and pure
 
@@ -156,13 +158,17 @@ Cost signals:
 3. actual-argument weight when that argument would be copied multiple times
 4. introduction of additional control-flow or loop structure at the call site
 
+Some candidates should be treated as unconditional wins rather than merely good heuristic scores. At minimum, `Bosatsu/Predef.and` and `Bosatsu/Predef.or` should always inline, because their entire value is short-circuit deferral of the second argument and their lowered bodies are tiny.
+
 Inline when the benefit exceeds the cost and all hard gates pass. A small helper whose recursion has already lowered to a self-contained `WhileExpr` is still eligible under this policy. The initial constants should mirror the intent of `TypedExprNormalization.ResolveToLambda`, but stay separate so Matchless tuning does not silently perturb typedexpr behavior.
 
 ### 6. Reuse local cleanup after inlining
 
 After rewriting a let body, rerun existing local Matchless cleanup that is already safe post-lowering, especially `reuseConstructors`. If `inlineApplyArgs` introduces trivial `Let` wrappers or immediate alias opportunities, clean those up before publishing the summary.
 
-Existing tree shaking and dead-code elimination should still run first and determine which top-level bindings reach this phase. V1 does not add another DCE step; it only rewrites the surviving bodies. Export selection and `ShowSelection` behavior remain unchanged.
+This likely requires a small refactor in `Matchless.scala` so the existing post-lowering cleanup becomes an explicit pipeline rather than a few maps hanging off the tail of `fromLet`. The raw lowering step, the existing local cleanup passes, and the new global inlining phase should be exposed as named stages so the sequencing is easy to reason about and extend.
+
+Existing tree shaking and dead-code elimination should still run first and determine which top-level bindings reach this phase. V1 does not add another DCE step; it only rewrites the surviving bodies. Export selection remains unchanged.
 
 ### 7. Integrate at the common compilation entry points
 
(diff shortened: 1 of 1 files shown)


docs/design/2185-design-a-matchless-global-inlining-phase.md
```diff
@@ -0,0 +1,229 @@
...
+2. Preserve the existing parallel raw Matchless lowering step.
+3. Apply conservative heuristics that prioritize branch-delayed arguments and lambda-argument beta-reduction opportunities.
+4. Keep semantics unchanged, including name binding, dependency-scope resolution, and evaluation behavior.
+5. Route codegen and evaluator consumers through the same optimized Matchless pipeline.
+
+## Non-goals
+
+1. No new `inline def` syntax, export metadata, or protobuf/interface changes. That remains the scope of #2178.
+2. No forced inlining annotation or user-configurable pragma in this PR.
+3. No aggressive recursive-function inlining, mutual-recursion handling, or whole-program dead-code elimination.
```
johnynek: There is already tree shaking/dead code elimination. We should apply inlining after that phase because it's a waste to do this work on code if it will be discarded.

johnynek: There is no mutual recursion in Bosatsu. Remove this.

johnynek: Be clear: recursive loops that are fully lowered to a WhileExpr should be eligible for inlining. Small while loops passed as lambdas could still be valuably inlined.

docs/design/2185-design-a-matchless-global-inlining-phase.md
```diff
@@ -0,0 +1,229 @@
...
+1. recovered lambda form from `Matchless.recoverTopLevelLambda`, if the binding can be treated as a top-level lambda
+2. arity and rewritten body
+3. expression weight / size estimate for budget checks
+4. purity flags derived from existing helpers such as `hasSideEffect`, `Expr.readsMutable`, and `Expr.containsWhileExpr`
+5. per-parameter demand facts: unused, eager, branch-guarded only, lambda-callee-only, and total use count
+
+V1 should only publish conservative candidates:
+
+1. non-recursive lets
+2. no mutable effects
```
johnynek: Be clear about this: we do want mutable let's fully inside an Expr. We just probably don't want lambdas or other exprs that capture a free mutable anon.

docs/design/2185-design-a-matchless-global-inlining-phase.md
johnynek: Updated the sequencing in the doc: it now says existing tree shaking/dead-code elimination runs first, and the Matchless inlining pass operates on the already-selected graph / tree-shaken namespace rather than doing work on discarded lets.

johnynek: Removed the mutual-recursion wording. The non-goal is now framed around recursive source forms that have not already lowered cleanly to Matchless.

johnynek: Clarified this in the candidate rules, heuristic section, testing plan, and acceptance criteria: helpers whose recursion has already lowered to a closed `WhileExpr` are still eligible for inlining and are scored by the normal size/duplication heuristic.

docs/design/2185-design-a-matchless-global-inlining-phase.md
johnynek: Narrowed the mutable restriction to free `LocalAnonMut` capture across the inline boundary. The doc now explicitly allows internal mutable state, including lowered loop state that stays fully inside the expression.

johnynek:
Updated the design doc to incorporate the review feedback: tree shaking now clearly happens before the inlining pass, the mutual-recursion wording is gone, lowered `WhileExpr` helpers are explicitly eligible, and the mutable restriction is narrowed to free mutable-anon capture across the inline boundary. `scripts/test_basic.sh` passed locally.

docs/design/2185-design-a-matchless-global-inlining-phase.md
```diff
@@ -0,0 +1,230 @@
...
+2. every use of a parameter is inside a branch body rather than in the selector or guard path
+3. a parameter is only used as a direct callee, and the actual argument is a literal lambda or resolves to a lambda after local alias resolution
+4. the candidate body is tiny and pure
+
+Cost signals:
+
+1. rewritten body weight
+2. repeated parameter use count
+3. actual-argument weight when that argument would be copied multiple times
+4. introduction of additional control-flow or loop structure at the call site
```
johnynek, line 159: I'm concerned about this. We should be more precise what is meant here.

Some key examples, IMO, are Bosatsu/Predef::and and Bosatsu/Predef::or (merge origin main if you don't find them in your predef.bosatsu). We would like `and(<e1>, <e2>)` to be inlined so that if can get `if <e1> : <e2>; else False` which can avoid evaluating `<e2>`. Similarly `or(<e1>, <e2>)` should inline to `if <e1> True; else <e2>` which can also avoid evaluating `<e2>`.

Inlining functions with match (which by this point will be lowered to If/Switch) is a key use case: when some non-cheap arguments could be deferred into the If/Switch body that aren't always evaluated, then it seems like they are good candidates for inlining.

Maybe that should be included in the summaries: we should note which, if any, of the arguments can be deferred for each top-level lambda. Then at call-sites that have non-cheap expressions at deferrable argument positions, that is more motivation to inline so that we can take advantage of laziness.

Note: since bosatsu is pure and total, the order of evaluation can't change semantics, but it can change the runtime cost, and deferring work can only lower the cost.

I think besides avoiding allocating closures, this kind of ability to leverage laziness is a key motivator for inlining. We need to support it very clearly.

docs/design/2185-design-a-matchless-global-inlining-phase.md
```diff
@@ -0,0 +1,230 @@
...
+
+Inline when the benefit exceeds the cost and all hard gates pass. A small helper whose recursion has already lowered to a self-contained `WhileExpr` is still eligible under this policy. The initial constants should mirror the intent of `TypedExprNormalization.ResolveToLambda`, but stay separate so Matchless tuning does not silently perturb typedexpr behavior.
+
+### 6. Reuse local cleanup after inlining
+
+After rewriting a let body, rerun existing local Matchless cleanup that is already safe post-lowering, especially `reuseConstructors`. If `inlineApplyArgs` introduces trivial `Let` wrappers or immediate alias opportunities, clean those up before publishing the summary.
```
johnynek, line 167: this may require refactoring Matchless a bit to expose these phases more directly. Can you comment on that? Please merge origin main to make sure we have the latest view of the code.

docs/design/2185-design-a-matchless-global-inlining-phase.md
```diff
@@ -0,0 +1,230 @@
...
+
+Existing tree shaking and dead-code elimination should still run first and determine which top-level bindings reach this phase. V1 does not add another DCE step; it only rewrites the surviving bodies. Export selection and `ShowSelection` behavior remain unchanged.
```
johnynek: Show doesn't use Matchless does it? What does show have to do with this? We may later add Matchless rendering to show (it seems like a good debugging idea), but currently I thought show only showed TypedExpr.

docs/design/2185-design-a-matchless-global-inlining-phase.md
```diff
@@ -0,0 +1,230 @@
...
+
+1. Extend `MatchlessApplyArgsTest.scala` with branch-sensitive cases that show `inlineApplyArgs` keeps expensive arguments under `If` or `SwitchVariant` branches instead of hoisting them into eager lets.
+2. Extend `MatchlessTests.scala` with an unused-parameter or branch-only-parameter example that should inline.
+3. Extend `MatchlessTests.scala` with a lambda-argument example that becomes directly beta-reducible after inlining.
+4. Extend `MatchlessTests.scala` with a helper that lowers to a small closed `WhileExpr` and remains eligible for inlining.
+5. Extend `MatchlessTests.scala` with a rejected case where a large or duplicated argument should not inline.
+6. Extend `MatchlessInterfaceTest.scala` with a cross-package example that proves imported implementations can inline while interface-only dependencies still compile through the old `Global` path.
+7. Extend `ClangGenLibraryDepsTest.scala` with a cross-library helper whose generated code no longer materializes both branch arguments before the helper call shape.
+8. Keep the broader Matchless, evaluator, and codegen suites green to validate semantic preservation.
+
```
johnynek, line 207: I think we need to explicitly show that `and` and `or` are always inlined. Unless we have a good argument I'm missing, those seem like examples that should basically always be inlined instead of a function call.

docs/design/2185-design-a-matchless-global-inlining-phase.md
johnynek: Tightened this section to make deferrable arguments explicit in the summaries and heuristic. The doc now calls out `Bosatsu/Predef.and` and `Bosatsu/Predef.or` as the canonical short-circuit examples, defines what it means for an argument to be deferrable under `If`/`SwitchVariant`, and says non-cheap actuals in those positions are a primary motivation to inline.

docs/design/2185-design-a-matchless-global-inlining-phase.md
johnynek: Added that commentary directly to the doc. It now says this likely needs a small `Matchless.scala` refactor so raw lowering, existing local cleanup, and the new global inlining pass are exposed as named stages instead of being fused at the tail of `fromLet`. I also updated the implementation plan accordingly after merging the latest `origin/main`.

docs/design/2185-design-a-matchless-global-inlining-phase.md
johnynek: Removed the `ShowSelection` reference. That sentence now just says export selection remains unchanged, without implying that current show tooling is on the Matchless path.

docs/design/2185-design-a-matchless-global-inlining-phase.md
johnynek: Made this explicit in both the heuristic section and the acceptance criteria: `Bosatsu/Predef.and` and `Bosatsu/Predef.or` are now documented as mandatory inline wins, and the testing plan calls for direct regression coverage of those cases.

johnynek:
Merged the latest `origin/main`, updated the design doc to make deferrable-argument laziness the central heuristic signal, explicitly called out `Bosatsu/Predef.and` and `Bosatsu/Predef.or` as always-inline short-circuit cases, noted the likely `Matchless` phase refactor, and removed the stale show-tooling reference. `scripts/test_basic.sh` passed locally after the edit.

johnynek:
- workflow run: `23661323309` (https://github.com/johnynek/bosatsu/actions/runs/23661323309)
- flaky-test issue: https://github.com/johnynek/bosatsu/issues/2262

PR #2189, opened 2026-03-18
Fix issue #2188: report irrefutable `matches` as always-true

### LINKED ISSUE #2188: unexpected `unreachable branches` on match against unit
If we write:
```
package A

x = ()

y = x matches ()
```

we get:
```
in file: root/WebDemo:5:5, package A
3|x = ()
4|
5|y = x matches ()
      ^^^^^^^^^^^^

unreachable branches: _

```
I guess what is happening is we compile `x matches ()` into:
```
match x:
    case (): True
    case _: False
```
but the pattern is total.

So, `x matches ()` is statically true. Any irrefutable pattern gives the same error.

We should give a better error, show that `x matches ()` is true for all values of `x: ()`. Similarly If we did `struct Foo` and `x = Foo` then `y = x matches Foo`. That's just a complicated way to write True and likely reflects confusion or programmer error.

johnynek:
Added a dedicated `TotalityCheck.MatchesAlwaysTrue` diagnostic for typed `matches` expressions whose first pattern is total for the scrutinee type, so cases like `x matches ()` and `x matches Foo` no longer surface as generic unreachable-branch errors. Updated `PackageError` to render a focused message that explains the pattern covers all values of the scrutinee type and the expression is always `True`. Added regression coverage in `TypedTotalityTest` and `ErrorMessageTest` for unit and nullary-struct cases. Verified with `sbt 'coreJVM/testOnly dev.bosatsu.TypedTotalityTest'`, `sbt 'coreJVM/testOnly dev.bosatsu.ErrorMessageTest'`, and `scripts/test_basic.sh`.

Fixes #2188

### DIFF
--- a/core/src/test/scala/dev/bosatsu/TypedTotalityTest.scala
+++ b/core/src/test/scala/dev/bosatsu/TypedTotalityTest.scala
@@ -1,7 +1,7 @@
 package dev.bosatsu
 
+import cats.Show
 import cats.data.Validated
-import dev.bosatsu.DirectEC.directEC
 import Identifier.{Constructor, Name}
 import TestUtils.checkEnvExpr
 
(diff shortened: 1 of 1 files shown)


johnynek:
Reproduced the `testJS` failure locally in `coreJS/testOnly dev.bosatsu.TypedTotalityTest`: the new shared helper in `TypedTotalityTest` depended on `DirectEC.directEC`, which works for the JVM path but breaks Scala.js shared-test compilation under `-Werror`. I updated `core/src/test/scala/dev/bosatsu/TypedTotalityTest.scala` to use a local `Show[String]` plus `Par.noParallelism` around `PackageMap.resolveThenInfer`, which gives the helper a backend-neutral `Par.EC` without the JVM-only import. Verified with `sbt 'coreJS/testOnly dev.bosatsu.TypedTotalityTest'`, `sbt 'coreJS/testOnly dev.bosatsu.ErrorMessageTest'`, `sbt 'jsapiJS/compile'`, and `scripts/test_basic.sh`.

PR #2199, opened 2026-03-18
Add missing predef eq_/cmp_ helpers for base types

### LINKED ISSUE #2193: add cmp_ and eq_ functions for all predef defined base types
We are inconsistent about which types have cmp_ and eq_ functions. For instance, Int has both, but the others only have cmp_.

We should also add eq_Bool and cmp_Bool (False < True), eq_String, eq_Float64, eq_Comparison, cmp_Comparison....

These don't need to be external. We can implement with `cmp_Foo(x, y) matches EQ` but they should all exist with the same naming pattern.

Look for redundant implementations of these functions in test_workspace and have them use the new predef implementations.

We don't need `eq_Unit/cmp_Unit` because that's kind of obvious, and we can't compare Fn values. For tuples, we can punt to composing the eq/cmp functions. We shouldn't implement those combinators in predef.

johnynek:
Added `cmp_Bool`, `eq_Bool`, `cmp_Comparison`, `eq_Comparison`, `eq_Float64`, and `eq_String` to `Bosatsu/Predef` and exported them. Replaced redundant local implementations in `test_workspace/recordset.bosatsu`, `test_workspace/NumberProps.bosatsu`, and `test_workspace/Float64.bosatsu`, expanded `test_workspace/PredefTests.bosatsu` to cover the new helpers directly, and refreshed the clang Ackermann snapshot hash to match the updated predef-generated code. Verified with `scripts/test_basic.sh`.

Fixes #2193

### DIFF
--- a/cli/src/test/scala/dev/bosatsu/codegen/clang/ClangGenTest.scala
+++ b/cli/src/test/scala/dev/bosatsu/codegen/clang/ClangGenTest.scala
@@ -42,7 +42,7 @@ class ClangGenTest extends munit.FunSuite {
       To inspect the code, change the hash, and it will print the code out
      */
     testFilesCompilesToHash("test_workspace/Ackermann.bosatsu")(
-      "6a6e1e5084006fdcd7abb76cd89e9277"
+      "6f8476751495555d241581dec1151cae"
     )
   }
 }
--- a/core/src/main/resources/bosatsu/predef.bosatsu
+++ b/core/src/main/resources/bosatsu/predef.bosatsu
@@ -47,6 +47,7 @@ export (
   Dict,
   add,
   addf,
+  and,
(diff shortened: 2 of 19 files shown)


johnynek:
Merge origin main and address the merge conflict

johnynek:
Merged the latest mainline changes by resolving the conflicts in `core/src/main/resources/bosatsu/predef.bosatsu` and `cli/src/test/scala/dev/bosatsu/codegen/clang/ClangGenTest.scala`. The resolved predef keeps main’s new boolean `and`/`or`/`not` helpers alongside this PR’s added `eq_/cmp_` helpers, and the Ackermann clang snapshot hash was refreshed for the merged output. `scripts/test_basic.sh` now passes locally.

Issue #2194, opened 2026-03-18
Supporting raw `$foo` in patterns and interpolation


johnynek:
Currently, we require `${foo}` in string patterns and interpolation. For short names like `${x}` we have 4 characters when `$x` might have done. How could we support making the `{ }` optional when the name is unambiguous, namely when `$` is directly followed by a single bindable (and not parsable as declaration). We don't want `$foo(bar)` to become parsed as `${foo}()`. We want that to be a parse failure.

I think something like checking if what comes after `$` parses as a declaration, and then error the parser if the declaration isn't exactly a Bindable name, could work.

This should work both in patterns and string interpolation contexts.

johnynek:
MergeXO assigned an agent and started design work for issue #2194.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/2200

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #2194.

johnynek:
Opened implementation PR: https://github.com/johnynek/bosatsu/pull/2207

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/2194#issuecomment-4080388346
- linked PR: #2207 (https://github.com/johnynek/bosatsu/pull/2207)

Comments on the source issue are no longer actioned after a PR exists. Please comment on the PR thread instead.

johnynek:
I think this ultimately isn't a great idea. Saving a few characters to make the strings more complex isn't a good trade.

PR #2200, opened 2026-03-18
Design doc for #2194: Support Raw `$foo` In Patterns And Interpolation

johnynek:
Design doc.

Refs #2194

### DIFF
--- a/docs/design/2194-supporting-raw-foo-in-patterns-and-interpolation.md
+++ b/docs/design/2194-supporting-raw-foo-in-patterns-and-interpolation.md
@@ -23,7 +23,7 @@ _Issue: #2194 (https://github.com/johnynek/bosatsu/issues/2194)_
 
 ## Summary
 
-Add parser-only support for raw `$name` in string interpolation and string patterns when the suffix is exactly one bindable, keep braced forms as the canonical emitted syntax, and reject ambiguous larger parses such as `$foo(bar)`.
+Add parser-only support for a regular raw-dollar family in strings: `$name` for string interpolation/binding, `$.name` for char interpolation/binding, `$_` and `$._` for pattern wildcards, and `$$` for a literal dollar, while still rejecting ambiguous larger parses such as `$foo(bar)`.
 
 ## Context
 Bosatsu currently requires braced forms for string interpolation and string-pattern bindings:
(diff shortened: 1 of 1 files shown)


docs/design/2194-supporting-raw-foo-in-patterns-and-interpolation.md
```diff
@@ -0,0 +1,177 @@
...
+That architecture means issue #2194 is primarily a parser and user-surface change. Type inference, pattern lowering, evaluation, and code generation should not need new IR or runtime behavior.
+
+## Goals
+1. Accept raw `$foo` anywhere `${foo}` is currently valid for string-substring interpolation or string-pattern binding.
+2. Keep the parser conservative: if the text after `$` parses as more than a single bindable, fail the parse instead of treating it as `${foo}` followed by trailing literal text.
+3. Preserve all existing `${...}` and `$.{...}` forms.
+4. Reuse the existing AST so downstream semantics remain unchanged.
+
+## Non-goals
+1. No raw char shorthand such as `$.x`; char interpolation and char pattern binds stay `$.{x}`.
```
johnynek: I think we should support $.foo to be regular. Also $_ should be supported.

We need the syntax to be regular and guessable and fail on ambiguous cases and not have silently confusing examples.

$$ should be a literal $ (and when we don't parse we should give a short blurb to teach what is likely the best hint to solve the issue).

johnynek: Updated the design doc to follow that direction directly. The proposed syntax is now a regular raw-dollar family: `$foo`, `$.foo`, `$_`, `$._`, and `$$` for a literal dollar. I also rewrote the diagnostics/hint guidance so ambiguous raw forms teach two fixes explicitly: use `${...}` / `$.{...}` for larger expressions, and use `$$` for a literal `$`. The goals, proposed design, implementation plan, testing strategy, acceptance criteria, risks, and rollout notes were all updated to match.

johnynek:
Updated `docs/design/2194-supporting-raw-foo-in-patterns-and-interpolation.md` to incorporate the review feedback around regular raw-dollar syntax and literal-dollar handling. Required pre-push check passed: `scripts/test_basic.sh`.

Issue #2195, opened 2026-03-18
guards on `matches`


johnynek:
It's a bit weird that we have guards in patterns, but not in matches.

Consider this implementation of `def exists`:

```
def exists(list, pred): list matches [*_, x, *_] if pred(x)
```
that seems really elegant and would be allowed if we wrote it as:

```
def exists(list, pred):
    match list:
       case [*_, x, *_] if pred(x): True
       case _: False
```
it seems annoying that we have to write it that way.

We should extend matches to support predictates, so we allow binding names in the matches but those names should be used in pred and are out of scope outside the pattern guard

johnynek:
MergeXO assigned an agent and started design work for issue #2195.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/2201

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #2195.

johnynek:
Opened implementation PR: https://github.com/johnynek/bosatsu/pull/2204

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/2195#issuecomment-4079665213
- linked PR: #2204 (https://github.com/johnynek/bosatsu/pull/2204)

Comments on the source issue are no longer actioned after a PR exists. Please comment on the PR thread instead.

PR #2201, opened 2026-03-18
Design doc for #2195: Guards on `matches`

johnynek:
Design doc.

Refs #2195

### DIFF
--- a/docs/design/2195-guards-on-matches.md
+++ b/docs/design/2195-guards-on-matches.md
@@ -106,22 +106,31 @@ In `core/src/main/scala/dev/bosatsu/Declaration.scala`:
 4. Keep the guard attached to `matches`, not to the pattern.
 
 Ternary interaction:
-the parser should preserve the existing ternary form without mandatory parentheses. The intended behavior is:
+the parser should preserve the existing ternary capability without mandatory parentheses, but it should do so with forward parsing rather than backtracking. The intended behavior is:
 
 ```bosatsu
 x matches p if cond
 ```
 
-parses as guarded `matches`, but:
+parses as guarded `matches`.
+
+If a same-level `else` follows:
 
 ```bosatsu
 x matches p if cond else other
 ```
 
-should backtrack to the existing ternary parse rather than forcing the guard interpretation. One workable implementation is:
-1. parse a candidate `matches` guard after `if`
-2. if a same-level `else` follows, backtrack the whole `if <guard>` suffix
-3. let the existing ternary parser consume `if ... else ...`
+the parser should keep moving forward and construct the ternary-shaped equivalent of:
+
+```bosatsu
+True if (x matches p if cond) else other
+```
+
+This preserves the desired scoping rule:
+1. any names bound by `p` are in scope for `cond`
+2. those names are not in scope for `other`
+
+Because the ternary condition is itself a guarded `matches` expression, the parser does not need to backtrack, and the existing ternary lowering can still be reused.
 
 ### 2. Update source-level scope bookkeeping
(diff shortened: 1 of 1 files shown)


docs/design/2195-guards-on-matches.md
```diff
@@ -0,0 +1,242 @@
...
+   1. require at least one space before `if`
+   2. once `if` is seen, commit to parsing a guard expression
+   3. parse the guard in the surrounding expression mode so constructs like `if x matches p if g:` still stop at the outer `:`
+4. Keep the guard attached to `matches`, not to the pattern.
+
+Precedence decision:
+`if` after `matches` belongs to the `matches` expression. If users need a ternary around the whole `matches`, they should parenthesize it:
+
+```bosatsu
+(x matches p) if cond else other
```
johnynek: Why can't we support the full ternary without the parens? It could be lowered to:

```
match x:
  case p if cond: True
  case_: other
```

Seems like we shouldn't require the parens and should work in either case.

johnynek: Updated the design doc to preserve the unparenthesized ternary case instead of requiring parens. The parser section now says `x matches p if cond` parses as guarded `matches`, but `x matches p if cond else other` should backtrack to the existing ternary parse. I also updated the testing, acceptance-criteria, risk, and rollout sections to cover that behavior explicitly.

johnynek:
Updated `docs/design/2195-guards-on-matches.md` to remove the mandatory-parentheses restriction and document the intended ternary interaction. Required pre-push check passed: `scripts/test_basic.sh`.

docs/design/2195-guards-on-matches.md
```diff
@@ -0,0 +1,255 @@
...
+x matches p if cond
+```
+
+parses as guarded `matches`, but:
+
+```bosatsu
+x matches p if cond else other
+```
+
+should backtrack to the existing ternary parse rather than forcing the guard interpretation. One workable implementation is:
```
johnynek: We shouldn't backtrack. We should forward parse: if else follows, generate the ternary but otherwise if there is no else generate a Matches with a guard.

We want the absolute minimum of backtracking for performance and error localization.

docs/design/2195-guards-on-matches.md
```diff
@@ -0,0 +1,255 @@
...
+the parser should preserve the existing ternary form without mandatory parentheses. The intended behavior is:
+
+```bosatsu
+x matches p if cond
+```
+
+parses as guarded `matches`, but:
+
+```bosatsu
+x matches p if cond else other
```
johnynek, line 120: We need to be clear:

Any binding names in p are in scope for cond (but not other)

docs/design/2195-guards-on-matches.md
johnynek: Updated the doc to remove the backtracking plan. The ternary section now specifies forward parsing: `x matches p if cond else other` is handled as the ternary-shaped equivalent `True if (x matches p if cond) else other`, so the parser keeps moving forward, keeps error localization tight, and can reuse existing ternary lowering.

docs/design/2195-guards-on-matches.md
johnynek: Clarified this explicitly in the ternary section and propagated it through testing, acceptance criteria, docs, and rollout notes: any names bound in `p` are in scope for `cond`, but not for `other`.

johnynek:
Updated `docs/design/2195-guards-on-matches.md` to document a forward-parse ternary strategy and the precise `p`-binding scope rule. Required pre-push check passed: `scripts/test_basic.sh`.

PR #2204, opened 2026-03-18
Implement guarded `matches` expressions

### LINKED ISSUE #2195: guards on `matches`

johnynek:
Implemented issue #2195 by extending `Declaration.Matches` with an optional guard, teaching the parser to accept `expr matches pattern if predicate` plus the forward-parsed `... if cond else other` shorthand, and updating source-level scope/substitution/pretty-printing so pattern bindings are visible only inside the guard. Lowering now reuses guarded `Expr.Match` branches, preserves the narrow `if x matches p` optimization, canonicalizes `if True` back to unguarded `matches` for existing always-true diagnostics, updates docs/error messaging/totality handling, and adds parser, evaluation, error-message, totality, generator, and recursion-check coverage for the new form. Verification passed with `sbt -batch "coreJVM/test:compile"`, focused parser/error/totality/evaluation suites, `sbt -batch "doc; paradox"`, and `scripts/test_basic.sh`.

Fixes #2195

Implements design doc: [docs/design/2195-guards-on-matches.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/2195-guards-on-matches.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/2201

### DIFF
--- a/core/src/main/scala/dev/bosatsu/Matchless.scala
+++ b/core/src/main/scala/dev/bosatsu/Matchless.scala
@@ -4244,22 +4244,7 @@ object Matchless {
     def isNonOrthogonal(
         p: Pattern[(PackageName, Constructor), Type]
     ): Boolean =
-      p match {
-        case Pattern.Annotation(p1, _) => isNonOrthogonal(p1)
-        case Pattern.Named(_, p1)      => isNonOrthogonal(p1)
-        case Pattern.Var(_) | Pattern.WildCard | Pattern.Literal(_) =>
-          false
-        case sp @ Pattern.StrPat(_)  => sp.simplify.isEmpty
-        case lp @ Pattern.ListPat(_) =>
-          Pattern.ListPat.toPositionalStruct(lp, empty, cons) match {
-            case Right(p1) => isNonOrthogonal(p1)
-            case Left(_)   => true
-          }
-        case Pattern.PositionalStruct(_, ps) =>
-          ps.exists(isNonOrthogonal)
-        case Pattern.Union(h, t) =>
-          isNonOrthogonal(h) || t.exists(isNonOrthogonal)
-      }
+      p.isSearchPattern
 
     // Pull off aliases and bindings that capture the whole occurrence.
     // This keeps the matrix focused on refutable structure.
--- a/core/src/main/scala/dev/bosatsu/Pattern.scala
+++ b/core/src/main/scala/dev/bosatsu/Pattern.scala
@@ -42,6 +42,18 @@ sealed abstract class Pattern[+N, +T] derives CanEqual {
         false
     }
 
+  lazy val searchInfo: Pattern.SearchInfo =
+    Pattern.searchInfo(this)
+
+  def isSearchPattern: Boolean =
+    searchInfo.isSearchPattern
+
+  def ambiguousBindings: Set[Bindable] =
+    searchInfo.ambiguousBindings
+
+  def hasSearchBindings: Boolean =
+    ambiguousBindings.nonEmpty
+
   /** List all the names that are bound in Vars inside this pattern in the left
     * to right order they are encountered, without any duplication
     */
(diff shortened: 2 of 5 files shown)


core/src/main/scala/dev/bosatsu/Declaration.scala
```diff
@@ -1630,14 +1675,20 @@ object Declaration {
 
         // matched
         val matched: P[NonBinding] = {
-          // x matches p
+          val spaceBeforeIf: P[Unit] = P.charIn(Set(' ', '\t')).rep.void
+          val guardPrefix: P[Unit] = spaceBeforeIf *> Parser.keySpace("if")
+          val guard: P0[Option[NonBinding]] =
+            (P.peek((spaceBeforeIf *> P.string("if")).backtrack) *>
```
johnynek: Isn't this peek(guardPrefix)? Can't we reuse that?

core/src/main/scala/dev/bosatsu/Declaration.scala
```diff
@@ -1686,12 +1737,33 @@ object Declaration {
...
+            }
+          case other =>
+            P.pure(other)
+        }
+
         val finalNonBind: P[NonBinding] =
-          if (pm != ParseMode.ComprehensionSource)
-            postOperators(matched).maybeAp(ternary)
+          if (!pm.isComprehensionSource)
+            postOperators(matched).flatMap { nb =>
```
johnynek: Use for comprehension here.

nb <- postOperators(matched)
...

docs/src/main/paradox/language_guide.md
```diff
@@ -516,8 +516,19 @@ long = match ["foo", "bar"]:
 
 short = ["foo", "bar"] matches ["foo", *_]
 ```
-The caveat is that you cannot have any bindings in the pattern when used
-as a matches expression.
+`matches` can also take a guard:
+```
+exists = [1, 2, 3] matches [*_, x, *_] if x matches 2
```
johnynek: This is a bad example. [*_, 2, *_] is better code. Don't show examples of what people shouldn't do. Use a predicate function that isn't just another match expression.

core/src/main/scala/dev/bosatsu/Declaration.scala
johnynek: Factored out a shared `guardIf` parser and reused it in the lookahead. I kept the lookahead on bare `if` instead of `peek(guardPrefix)`, because `xs matches p if` still needs to stay committed and fail at the guard site; `peek(guardPrefix)` regressed that parser test.

core/src/main/scala/dev/bosatsu/Declaration.scala
johnynek: Rewrote that tail as a `for`-comprehension over `postOperators(matched)`, `guardedMatchesElse`, and `ternary.?`.

docs/src/main/paradox/language_guide.md
johnynek: Updated the language-guide example to a predicate-based case: `is_even` / `contains_even`, so it demonstrates a real guard use case instead of re-encoding a plain pattern match.

johnynek:
Applied the review feedback in `Declaration.scala` and `language_guide.md`. Local checks that passed: `sbt -batch "coreJVM/test:compile"`, `sbt -batch "coreJVM/testOnly dev.bosatsu.ParserTest -- --log=failure"`, `sbt -batch paradox`, and the required `scripts/test_basic.sh`. I also reran `sbt -batch "doc; paradox"` and the exact CI coverage command. In this workspace those are blocked by the only installed JDK being 25.0.1 while repo CI uses Java 17: `doc` fails in `coreJS / Compile / doc` with a ScalaDoc NPE, and the coverage run fails with broad `NoClassDefFoundError` classloading errors under coverage (for example `bosatsu/TypedAst/StrPart$Value$LiteralStr$`). Because I could not get those Java-17-targeted checks green in this environment, I am not marking this ready to push yet.

johnynek:
I think the test was an unrelated flakey test. Please try pushing your branch changes and see if they go green.

johnynek:
Applied the remaining cleanup nits on this branch: factored a shared `guardIf` lookahead in `Declaration.scala` while preserving bare-`if` guard commitment, rewrote the guarded `matches` ternary tail with a `for`-comprehension, and replaced the language-guide example with a real predicate-based `is_even` / `contains_even` example. Local checks passed: `sbt -batch "coreJVM/testOnly dev.bosatsu.ParserTest -- --log=failure"`, `sbt -batch paradox`, and `scripts/test_basic.sh`. This should be ready to push and let CI rerun.

johnynek:
We need some tests that the semantics of this are coherent. For instance:
```
foo matches bar if pred else cond
```
should always be the same value as:
```
(foo matches bar) if pred else cond
```
when `pred` does not have any bindings from bar.

similarly, for all cases we should have (I think we normalize to this now):
```
foo matches bar if pred else cond = True if (foo matches bar if pred) else cond
```

More generally, we can think of `and_or(x, y, z) = (x && y) || z = y if x else z`. Since `and` is commutative, we should have that:
```
pred if (foo matches bar) else cond == (foo matches bar) if pred else cond
```

I guess since `True && (x && y)` == (x && y)` we can prove this result. But please add these semantic coherence tests so we can be sure we aren't creating a confusing syntax.

johnynek:
Added semantic-coherence coverage in `EvaluationTest.scala` for guarded `matches` ternary forms. The new cases check the binding-free `... if pred else False` equivalences against the explicitly parenthesized ternary form, plus the normalized `True if (foo matches bar if pred) else cond` form, including examples where the guard uses a bound name. Local checks passed: `sbt -batch "coreJVM/testOnly dev.bosatsu.EvaluationTest -- --log=failure"` and `scripts/test_basic.sh`.

johnynek:
what about double nested matches:
```
x matches p if g else y
```
but `g = gx matches gp if gg`

now it is ambiguous if we should parse as P1:
```
x matches p if (gx matches gp if gg) else y
```
or P2:
```
x matches p if (gx matches gp if gg else y)
```
which do we mean by:
```
x matches p if gx matches gp if gg else y
```
Are they always the same?
going back to the `and_or(x, y, z) = y if x else z` then the two parse as:
```
P1 = and_or(and_or(gg, gx matches gp, False), x matches p, y)
P2 = and_or(and_or(gg, gx matches gp, y), x matches p, False)
```
so if we define: `a = gg, b = gx matches gp, c = x matches p, d = y`
we have:
```
P1 = and_or(and_or(a, b, False), c, d) = (a & b) & c || d
P2 = and_or(and_or(a, b, d), c, False) = ((a & b) | d) & c = (a & d & c) | (d & c)
```
so it looks like P2 is only true if c (x matches p) is true. But P1 is true when c = False and d = True (y = True).

So, it seems that without parens, this is truly ambiguous. Which, to me suggests we need to make such formulations errors.

So, I think we need to error if the guard function of a `Matches` itself is a Matches with a guard. We must use a Parens to be clear there.

Please read and see if you agree with this. Try to find other examples of syntactic ambiguity this syntax could provoke, and make sure we don't rely on precedence to resolve: it should either be equivalent in all ways to view the syntax, or we should error and require parens for disambiguation.

Write specific tests that the ambiguous construct I proposed here will error. And review to make sure no other such ambiguous syntaxes have been introduced.

johnynek:
Addressed the follow-up ambiguity/round-trip fallout in this turn. `Declaration.scala` now treats lambda/if/match/ternary guard parens as syntactic-only when reparsing guarded `matches`, while preserving explicit guard parens by printing an extra layer when needed. `Gen.scala` was tightened so generated `matches` guards no longer use the now-disallowed bare guarded-`matches` form; they are generated parenthesized instead. `ParserTest.scala` was updated to match the normalized AST for parenthesized ternary guards. Verification passed with `sbt -batch "coreJVM/testOnly dev.bosatsu.ParserTest dev.bosatsu.SyntaxParseTest -- --log=failure"` and the required `scripts/test_basic.sh`.

PR #2205, opened 2026-03-18
Optimize terminal guarded bool matches

johnynek:
## Summary
- rewrite terminal `case p if g: True; case q: False` matches to expose `g` directly as the branch result
- drop the trailing `False` fallback when `difference(q, p).isEmpty`
- add focused `TypedExprTest` coverage for both the keep-fallback and drop-fallback cases

## Testing
- sbt "coreJVM/testOnly dev.bosatsu.TypedExprTest"

### DIFF
--- a/core/src/test/scala/dev/bosatsu/PatternTest.scala
+++ b/core/src/test/scala/dev/bosatsu/PatternTest.scala
@@ -278,6 +278,10 @@ class PatternTest extends munit.ScalaCheckSuite {
 
     noSearch("Foo(x)")
     noSearch("[x, y]")
+    noSearch("[*xs]")
+    noSearch("[x, *y]")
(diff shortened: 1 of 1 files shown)


johnynek:
I think this optimization is unsound (and maybe so are others we already have).

Consider "search type" patterns that can match at multiple locations such as [*_, x, *_]. This can match binding x at any position of a non empty list.

So, that pattern with a guard filters the locations that match into those that also pass the guard.

But if we move the match to the expression side, we would match the first location and then check if the first location passes the predicate.

I think this suggests something we haven't checked before: given a pattern can it match at most one or more than one way. The optimization proposed is only valid when a pattern can only match one way.

We can probably statically determine if a pattern can only match one way, and limit this optimization to those cases.

PR #2203, opened 2026-03-18
Design doc for #2192: Add `--warn` and `--lax` lint modes to `lib check` and `lib test`

johnynek:
Design doc.

Refs #2192

PR #2206, opened 2026-03-18
Add warn/lax lint modes to lib check and lib test

### LINKED ISSUE #2192: less strictness in `lib check` mode.
There should be an option in `lib check` like `lib check --lax` to turn off any linting errors and maybe `lib check --warn` to turn them into non-fatal warnings.

We should possibly add those same options to `lib test`. It should be easy to iterate on checking and testing without fixing every thing, especially when you know you will be using functions soon that may not yet be used.

We should make a list of the maximal list of checks we could postpone. I don't think we can avoid totality checking or recursion checking, but anything that is stylistic or there to make the code more strict, I think those could be silenced or moved to warnings.

johnynek:
Implemented the accepted design for issue #2192 by adding a new LintMode enum and plumbing mutually exclusive --warn/--lax flags into lib check and lib test only, while leaving all other library/tool commands on the existing strict path. Compiler/lint handling now uses a shared PackageError postponability classifier, treats only the approved lint set as non-blocking, preserves mixed hard+lint diagnostics, replays lint diagnostics so cache hits match cold compiles, and renders mixed warn+error output in a stable warning-first order. The compiler/package changes cover unused imports, unused lets, shadowed-binding type changes, and unreachable-branch totality warnings without relaxing recursion, non-total matches, invalid patterns, or runtime/lib-config checks.

Added command-level tests for strict/warn/lax behavior, cache-hit parity, mixed hard+lint failures, recursion/non-total/invalid-pattern hard failures, lib test warn/lax behavior, and todo remaining invalid under lib test. Updated the 5-minute docs to describe the new lib check/lib test iteration modes.

Validation completed with: sbt "coreJVM/testOnly dev.bosatsu.ToolAndLibCommandTest", sbt "coreJVM/test; cli/test", ./test_cli.sh, sbt "doc; paradox", and scripts/test_basic.sh. scripts/test_basic.sh passed on rerun after one unrelated flaky ScalaCheck failure in dev.bosatsu.rankn.TypeTest.

Fixes #2192

Implements design doc: [docs/design/2192-less-strictness-in-lib-check-mode.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/2192-less-strictness-in-lib-check-mode.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/2203

### DIFF
--- a/core/src/main/scala/dev/bosatsu/Package.scala
+++ b/core/src/main/scala/dev/bosatsu/Package.scala
@@ -824,16 +824,16 @@ object Package {
             .flatMap { typedLets =>
               val topLevelDefs = TypedExprRecursionCheck.topLevelDefArgs(stmts)
 
-              val recursionErrors: List[PackageError] =
+              val recursionErrors: Chain[PackageError] =
                 toErrsChain(
                   TypedExprRecursionCheck
                     .checkLets(p, fullTypeEnv, typedLets, topLevelDefs)
                     .leftMap(
                       _.map(err => PackageError.RecursionError(p, err): PackageError)
                     )
-                ).toList
+                )
 
-              val shadowedBindingErrors: List[PackageError] =
+              val shadowedBindingErrors: Chain[PackageError] =
                 toErrsChain(
                   ShadowedBindingTypeCheck
                     .checkLets(p, typedLets)
@@ -846,7 +846,7 @@ object Package {
                         ): PackageError
                       )
                     )
-                ).toList
+                )
 
               val totalityErrors: List[PackageError] =
                 toErrs(
@@ -863,7 +863,7 @@ object Package {
               // keep the typed program on the success path when every
               // diagnostic is postponable.
               diagnosticsToIor(
-                recursionErrors ::: shadowedBindingErrors ::: totalityErrors,
+                (recursionErrors ++ shadowedBindingErrors).toList ::: totalityErrors,
                 (
                   fullTypeEnv,
                   Program(typeEnv, typedLets, extDefs, stmts)
--- a/core/src/main/scala/dev/bosatsu/PackageCustoms.scala
+++ b/core/src/main/scala/dev/bosatsu/PackageCustoms.scala
@@ -491,30 +491,39 @@ object PackageCustoms {
   private def noUselessBinds[A: HasRegion](
       pack: Package.Typed[A],
       sourceStatements: Option[List[Statement]]
-  ): ValidatedNec[PackageError, Unit] = {
-    val sourceStmts: List[Statement] =
-      sourceStatements.getOrElse {
-        pack.program._1.from match {
-          case stmts: List[?] =>
-            stmts.collect { case stmt: Statement => stmt }
-          case _              =>
-            Nil
-        }
-      }
+  ): ValidatedNec[PackageError, Unit] =
+    sourceStatements.orElse(maybeStatements(pack)) match {
+      case Some(sourceStmts) =>
+        noUselessBindsFromLets(
+          pack.name,
+          pack.exports,
+          pack.lets,
+          sourceStmts,
+          Package.testRootBindables(pack).toList ::: Package
+            .mainValue(pack)
+            .map(_._1)
+            .toList,
+          te =>
+            TypedExpr.usedGlobals(te).runS(Set.empty).value.collect {
+              case (pn, i: Identifier.Bindable) if pn == pack.name => i
+            }
+        )
+      case None              =>
+        // Compiled/cache-only packages erase Program.from, so this lint can
+        // only run when the caller supplied source statements explicitly.
+        Validated.unit
+    }
 
-    noUselessBindsFromLets(
-      pack.name,
-      pack.exports,
-      pack.lets,
-      sourceStmts,
-      Package.testRootBindables(pack).toList ::: Package
-        .mainValue(pack)
-        .map(_._1)
-        .toList,
-      te =>
-        TypedExpr.usedGlobals(te).runS(Set.empty).value.collect {
-          case (pn, i: Identifier.Bindable) if pn == pack.name => i
+  private def maybeStatements[A](
+      pack: Package.Typed[A]
+  ): Option[List[Statement]] =
+    pack.program._1.from match {
+      case stmts: List[?] =>
+        stmts.traverse {
+          case stmt: Statement => Some(stmt)
+          case _               => None
         }
-    )
-  }
+      case _              =>
+        None
+    }
 }
--- a/core/src/main/scala/dev/bosatsu/PackageMap.scala
+++ b/core/src/main/scala/dev/bosatsu/PackageMap.scala
@@ -803,42 +803,29 @@ object PackageMap {
         )
       } else {
         val sequenced =
-          resultMap.traverse { result =>
-            result match {
-              case Ior.Left(errs)          =>
-                Ior.both(errs, Option.empty[CompiledPack[CacheDepHash]])
-              case Ior.Right(compiledPack) =>
-                Ior.right(Some(compiledPack))
-              case Ior.Both(errs, compiledPack) =>
-                Ior.both(errs, Some(compiledPack))
-            }
+          resultMap.traverse {
+            case Ior.Left(errs)          =>
+              Ior.both(errs, Option.empty[CompiledPack[CacheDepHash]])
+            case Ior.Right(compiledPack) =>
+              Ior.right(Some(compiledPack))
+            case Ior.Both(errs, compiledPack) =>
+              Ior.both(errs, Some(compiledPack))
           }
 
-        dedupeErrors(sequenced) match {
-          case Ior.Right(entries)    =>
-            Ior.right(
-              PackageMap(
-                SortedMap.from(
-                  entries.iterator.collect { case (name, Some(compiledPack)) =>
-                    name -> compiledPack.compiled
-                  }
-                )
+        dedupeErrors(sequenced)
+          .map { entries =>
+            PackageMap(
+              SortedMap.from(
+                entries.iterator.collect { case (name, Some(compiledPack)) =>
+                  name -> compiledPack.compiled
+                }
               )
             )
-          case Ior.Both(errs, entries) =>
-            val compiled =
-              PackageMap(
-                SortedMap.from(
-                  entries.iterator.collect { case (name, Some(compiledPack)) =>
-                    name -> compiledPack.compiled
-                  }
-                )
-              )
-
-            if (compiled.toMap.isEmpty) Ior.left(errs)
-            else Ior.both(errs, compiled)
-          case Ior.Left(errs)         =>
+          } match {
+          case Ior.Both(errs, compiled) if compiled.toMap.isEmpty =>
             Ior.left(errs)
+          case other                                              =>
+            other
         }
       }
     }
--- a/core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
+++ b/core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
@@ -38,6 +38,9 @@ object CompilerApi {
       val pluralLabel: String,
       val order: Int
   ) {
+    def title(count: Int): String =
+      if (count == 1) singularLabel else s"$count $pluralLabel"
+
     // Errors and lint render in separate blocks, so each block gets its own
     // category ordering instead of trying to share a single cross-kind index.
     case UnusedValue
@@ -63,6 +66,9 @@ object CompilerApi {
       val pluralLabel: String,
       val order: Int
   ) {
+    def title(count: Int): String =
+      if (count == 1) singularLabel else s"$count $pluralLabel"
+
     case UnusedValue
         extends LintCategory("unused value", "unused values", 0)
     case ShadowedBinding
@@ -101,6 +107,11 @@ object CompilerApi {
       lintDiagnostics: List[PackageError]
   )
 
+  private final case class PartitionedErrors(
+      postponable: List[PackageError],
+      errors: List[PackageError]
+  )
+
   private def pluralizedLabel(
       count: Int,
       singular: String,
@@ -108,13 +119,19 @@ object CompilerApi {
   ): String =
     if (count == 1) s"1 $singular" else s"$count $plural"
 
-  private def formatDiagnosticTitle(
-      singularLabel: String,
-      pluralLabel: String,
-      count: Int
-  ): String =
-    if (count == 1) singularLabel
-    else s"$count $pluralLabel"
+  private def renderedError(
+      category: ErrorCategory,
+      count: Int,
+      body: Doc
+  ): RenderedError =
+    RenderedError(category, category.title(count), count, body)
+
+  private def renderedLint(
+      category: LintCategory,
+      count: Int,
+      body: Doc
+  ): RenderedLint =
+    RenderedLint(category, category.title(count), count, body)
 
   private def typeErrorCount(err: PackageError.TypeErrorIn): Int =
     err.tpeErr match {
@@ -141,148 +158,46 @@ object CompilerApi {
         RenderedError(ErrorCategory.UnusedValue, title, count, body)
       case e: PackageError.TypeErrorIn =>
         val count = typeErrorCount(e)
-        RenderedError(
-          ErrorCategory.TypeError,
-          formatDiagnosticTitle(
-            ErrorCategory.TypeError.singularLabel,
-            ErrorCategory.TypeError.pluralLabel,
-            count
-          ),
-          count,
-          body
-        )
+        renderedError(ErrorCategory.TypeError, count, body)
       case _: PackageError.KindInferenceError =>
-        RenderedError(
-          ErrorCategory.TypeError,
-          "type error",
-          1,
-          body
-        )
+        renderedError(ErrorCategory.TypeError, 1, body)
       case _: PackageError.ShadowedBindingTypeError =>
-        RenderedError(
-          ErrorCategory.TypeError,
-          "type error",
-          1,
-          body
-        )
+        renderedError(ErrorCategory.TypeError, 1, body)
       case _: PackageError.PrivateTypeEscape[?] =>
-        RenderedError(
-          ErrorCategory.TypeError,
-          "type error",
-          1,
-          body
-        )
+        renderedError(ErrorCategory.TypeError, 1, body)
       case _: PackageError.RecursionError =>
-        RenderedError(
-          ErrorCategory.RecursionError,
-          "recursion error",
-          1,
-          body
-        )
+        renderedError(ErrorCategory.RecursionError, 1, body)
       case _: PackageError.TotalityCheckError[?] =>
-        RenderedError(
-          ErrorCategory.TotalityError,
-          "totality error",
-          1,
-          body
-        )
+        renderedError(ErrorCategory.TotalityError, 1, body)
       case e: PackageError.SourceConverterErrorsIn =>
         val count = e.errs.length
-        RenderedError(
-          ErrorCategory.SourceConversionError,
-          formatDiagnosticTitle(
-            ErrorCategory.SourceConversionError.singularLabel,
-            ErrorCategory.SourceConversionError.pluralLabel,
-            count
-          ),
-          count,
-          body
-        )
+        renderedError(ErrorCategory.SourceConversionError, count, body)
       case e: PackageError.UnusedImport =>
         val count = e.badImports.length
-        RenderedError(
-          ErrorCategory.ImportError,
-          formatDiagnosticTitle(
-            ErrorCategory.ImportError.singularLabel,
-            ErrorCategory.ImportError.pluralLabel,
-            count
-          ),
-          count,
-          body
-        )
+        renderedError(ErrorCategory.ImportError, count, body)
       case e: PackageError.DuplicatedImport =>
         val count = e.duplicates.length
-        RenderedError(
-          ErrorCategory.ImportError,
-          formatDiagnosticTitle(
-            ErrorCategory.ImportError.singularLabel,
-            ErrorCategory.ImportError.pluralLabel,
-            count
-          ),
-          count,
-          body
-        )
+        renderedError(ErrorCategory.ImportError, count, body)
       case _: PackageError.UnknownImportPackage[?, ?, ?] =>
-        RenderedError(
-          ErrorCategory.ImportError,
-          "import error",
-          1,
-          body
-        )
+        renderedError(ErrorCategory.ImportError, 1, body)
       case _: PackageError.UnknownImportName[?, ?] =>
-        RenderedError(
-          ErrorCategory.ImportError,
-          "import error",
-          1,
-          body
-        )
+        renderedError(ErrorCategory.ImportError, 1, body)
       case _: PackageError.UnknownImportFromInterface[?, ?] =>
-        RenderedError(
-          ErrorCategory.ImportError,
-          "import error",
-          1,
-          body
-        )
+        renderedError(ErrorCategory.ImportError, 1, body)
       case _: PackageError.UnknownExport[?] =>
-        RenderedError(
-          ErrorCategory.PackageError,
-          "package error",
-          1,
-          body
-        )
+        renderedError(ErrorCategory.PackageError, 1, body)
       case _: PackageError.CircularDependency[?, ?, ?] =>
-        RenderedError(
-          ErrorCategory.PackageError,
-          "package error",
-          1,
-          body
-        )
+        renderedError(ErrorCategory.PackageError, 1, body)
       case e: PackageError.VarianceInferenceFailure =>
         val count = e.failed.length
-        RenderedError(
-          ErrorCategory.PackageError,
-          formatDiagnosticTitle(
-            ErrorCategory.PackageError.singularLabel,
-            ErrorCategory.PackageError.pluralLabel,
-            count
-          ),
-          count,
-          body
-        )
+        renderedError(ErrorCategory.PackageError, count, body)
       case e: PackageError.DuplicatedPackageError =>
         val count = e.dups.length
-        RenderedError(
-          ErrorCategory.PackageError,
-          formatDiagnosticTitle(
-            ErrorCategory.PackageError.singularLabel,
-            ErrorCategory.PackageError.pluralLabel,
-            count
-          ),
-          count,
-          body
-        )
+        renderedError(ErrorCategory.PackageError, count, body)
     }
 
+  // This classifier assumes `err` is already known to be postponable. Passing
+  // a general hard error here is a bug in the lint partitioning flow.
   private def classifyWarning(
       err: PackageError,
       body: Doc
@@ -302,16 +217,7 @@ object CompilerApi {
         RenderedLint(LintCategory.UnusedValue, title, count, body)
       case e: PackageError.UnusedImport =>
         val count = e.badImports.length
-        RenderedLint(
-          LintCategory.UnusedImport,
-          formatDiagnosticTitle(
-            LintCategory.UnusedImport.singularLabel,
-            LintCategory.UnusedImport.pluralLabel,
-            count
-          ),
-          count,
-          body
-        )
+        renderedLint(LintCategory.UnusedImport, count, body)
       case e: PackageError.ShadowedBindingTypeError =>
         RenderedLint(
           LintCategory.ShadowedBinding,
@@ -324,16 +230,7 @@ object CompilerApi {
             TotalityCheck.UnreachableBranches(_, branches)
           ) =>
         val count = branches.length
-        RenderedLint(
-          LintCategory.UnreachableBranch,
-          formatDiagnosticTitle(
-            LintCategory.UnreachableBranch.singularLabel,
-            LintCategory.UnreachableBranch.pluralLabel,
-            count
-          ),
-          count,
-          body
-        )
+        renderedLint(LintCategory.UnreachableBranch, count, body)
       case _ =>
         sys.error(s"unexpected non-lint warning: $err")
     }
@@ -421,11 +318,11 @@ object CompilerApi {
 
   private def partitionDiagnostics(
       errors: List[PackageError]
-  ): (List[PackageError], List[PackageError]) =
-    errors.partitionMap { err =>
-      if (PackageError.isPostponable(err)) Right(err)
-      else Left(err)
-    }
+  ): PartitionedErrors = {
+    val (postponable, hardErrors) =
+      errors.partition(PackageError.isPostponable)
+    PartitionedErrors(postponable, hardErrors)
+  }
 
   private def localTypeNames(pack: Package.Compiled): Set[TypeName] =
     pack.types.definedTypes.keysIterator.collect {
@@ -565,12 +462,18 @@ object CompilerApi {
       lintMode: LintMode
   ): InferPhases = {
     val base = InferPhases.default
-    // Cache the strict and relaxed lint modes separately even though they share
-    // the same finishPackage steps. Strict runs must not reuse artifacts that
-    // only existed because warn/lax allowed postponable diagnostics through.
+    def lintModeCacheKey(lm: LintMode): String =
+      lm match {
+        case LintMode.Strict          => "strict"
+        case LintMode.Warn | LintMode.Lax => "warn_or_lax"
+      }
+
+    // Strict runs must not reuse artifacts that only existed because warn/lax
+    // allowed postponable diagnostics through, but warn and lax can share the
+    // same compiled artifact cache because they accept the same code.
     new InferPhases {
       val id: String =
-        s"${base.id}:lint:${lintMode.toString.toLowerCase}"
+        s"${base.id}:lint:${lintModeCacheKey(lintMode)}"
 
       def dependencyInterface(pack: Package.Typed[Any]): Package.Interface =
         base.dependencyInterface(pack)
@@ -678,8 +581,9 @@ object CompilerApi {
         case Ior.Right(packs)    => (Nil, Some(packs))
         case Ior.Both(errs, pm)  => (errs.toList, Some(pm))
       }
-      (hardDiagnostics0, compileLintDiagnostics) =
-        partitionDiagnostics(compileDiagnostics)
+      partitionedDiagnostics = partitionDiagnostics(compileDiagnostics)
+      hardDiagnostics0 = partitionedDiagnostics.errors
+      compileLintDiagnostics = partitionedDiagnostics.postponable
       replayedLintDiagnostics <-
         compiled match {
           case Some(_) if !replaySuccessfulLint =>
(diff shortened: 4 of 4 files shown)


johnynek:
Reproduced the `testJS` failure locally from the same Scala.js suite path: `coreJS/testOnly dev.bosatsu.ToolAndLibCommandTest -- --log=failure` was failing in `lib check --warn reports lint warnings on cache hits` with `org.scalajs.linker.runtime.UndefinedBehaviorError: java.lang.ClassCastException: undefined cannot be cast to dev.bosatsu.Declaration`. The failure came from best-effort typed lint replay on cached Scala.js packages, where missing source declaration tags can surface as `UndefinedBehaviorError` inside `ShadowedBindingTypeCheck`/typed totality replay. I updated `CompilerApi` so that optional typed replay drops that specific Scala.js cache/runtime trap instead of aborting the command, while preserving the existing best-effort behavior for other replay-only lint passes.

Verified locally with:
- `sbt "coreJS/testOnly dev.bosatsu.ToolAndLibCommandTest -- --log=failure"`
- `scripts/test_basic.sh`

core/src/main/scala/dev/bosatsu/library/Command.scala
```diff
@@ -601,6 +621,7 @@ object Command {
             colorize: Colorize,
             sourcePackageFilter: Option[PackageName => Boolean] = None,
             compileOptions: CompileOptions,
+            lintMode: Option[LintMode] = None,
```
johnynek: we shouldn't use `Option[LintMode]` there is always a `LintMode`. I think the default is `LintMode.Strict`.

core/src/main/scala/dev/bosatsu/library/Command.scala
```diff
@@ -813,6 +851,7 @@ object Command {
           colorize: Colorize,
           sourcePackageFilter: Option[PackageName => Boolean] = None,
           compileOptions: CompileOptions,
+          lintMode: Option[LintMode] = None,
```
johnynek: let's always take a `LintMode`. If we don't have one, use `LintMode.Strict`

core/src/main/scala/dev/bosatsu/library/Command.scala
```diff
@@ -871,6 +913,7 @@ object Command {
           colorize: Colorize,
           sourcePackageFilter: PackageName => Boolean,
           compileOptions: CompileOptions,
+          lintMode: Option[LintMode] = None,
```
johnynek: always take a LintMode (LintMode.Strict when we don't have one).

core/src/main/scala/dev/bosatsu/library/Command.scala
```diff
@@ -943,19 +987,22 @@ object Command {
       def decodedWithDepsFilteredForTest(
           colorize: Colorize,
           sourcePackageFilter: PackageName => Boolean,
+          lintMode: Option[LintMode] = None,
```
johnynek: always pass a LintMode (no Option no default value).

core/src/main/scala/dev/bosatsu/library/Command.scala
```diff
@@ -943,19 +987,22 @@ object Command {
...
           CompileOptions.Default,
+          lintMode,
           compileCacheDirOpt
         )
 
       def build(
           colorize: Colorize,
           trans: Transpiler.Optioned[F, P],
           sourcePackageFilter: Option[PackageName => Boolean] = None,
+          lintMode: Option[LintMode] = None,
```
johnynek: always pass a LintMode (no Option no default value).

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
```diff
@@ -1,62 +1,84 @@
...
+      DiagnosticCategory("totality error", "totality errors", 3)
+    val SourceConversionError =
+      DiagnosticCategory("source conversion error", "source conversion errors", 4)
+    val ImportError =
+      DiagnosticCategory("import error", "import errors", 5)
+    val PackageError =
+      DiagnosticCategory("package error", "package errors", 6)
+
+    val ShadowedBinding =
+      DiagnosticCategory("shadowed binding", "shadowed bindings", 1)
```
johnynek: why does the order start over here? Please add a comment. Seems like maybe we should have two categories: Error and Lint or something and separate them.

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
```diff
@@ -87,131 +109,188 @@ object CompilerApi {
...
+          formatDiagnosticTitle(DiagnosticCategory.PackageError, count),
+          count,
+          body
+        )
+    }
+
+  private def classifyWarning(
+      err: PackageError,
+      body: Doc
+  ): RenderedDiagnostic =
```
johnynek: seems like this should return a `Either[RenderedError, RenderedLint]` or something so we can be clear about which each `PackageError` is

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
```diff
@@ -264,6 +351,258 @@ object CompilerApi {
         platformIO.moduleIOMonad.raiseError(ParseErrors(errs, color))
     }
 
+  private def distinctErrors(errors: List[PackageError]): List[PackageError] =
```
johnynek: we don't need a special method just to call `.distinct`

core/src/main/scala/dev/bosatsu/PackageCustoms.scala
```diff
@@ -27,9 +27,28 @@ object PackageCustoms {
...
+      case Validated.Valid(_)     => Nil
+      case Validated.Invalid(errs) =>
+        errs.toChain.toList
+    }
+
+  // Replay only the postponable customs on successful packages so cache hits
+  // surface the same lint diagnostics as cache misses.
+  def lintDiagnostics[A: HasRegion](
+      pack: Package.Typed[A],
+      sourceStatements: Option[List[Statement]] = None
```
johnynek: don't pass default `=None` call with the args you need each time.

core/src/main/scala/dev/bosatsu/library/Command.scala
johnynek: Changed `CheckState.packageMap` to take a concrete `LintMode`. The default now happens once at CLI parsing with `LintMode.Strict` instead of threading an option.

core/src/main/scala/dev/bosatsu/library/Command.scala
johnynek: Updated the decode/build path here to require a `LintMode` and pass `LintMode.Strict` explicitly from strict callers.

core/src/main/scala/dev/bosatsu/library/Command.scala
johnynek: Changed this call path to take `LintMode` directly; there is no `Option[LintMode]` left in the library command plumbing.

core/src/main/scala/dev/bosatsu/library/Command.scala
johnynek: Updated this helper to require `LintMode` and removed the optional/default-style path. Strict callers pass `LintMode.Strict` explicitly.

core/src/main/scala/dev/bosatsu/library/Command.scala
johnynek: Same cleanup here: concrete `LintMode` only, with `LintMode.Strict` passed explicitly at strict call sites.

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
johnynek: Split rendering into separate `ErrorCategory` and `LintCategory` enums and added a comment that each block has its own ordering, so error/lint summaries are no longer sharing one mixed ordering.

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
johnynek: Implemented that separation: rendering now has distinct `RenderedError` and `RenderedLint` types with separate classifiers instead of one shared diagnostic bucket.

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
johnynek: Removed the helper and just use `.distinct` inline where the diagnostic lists are combined.

core/src/main/scala/dev/bosatsu/PackageCustoms.scala
johnynek: Removed the default-arg style from the customs helper path. Callers now pass the source statement option they want explicitly.

johnynek:
Addressed the thread-level issues as code changes. Warm-cache lint replay for optimize-enabled runs now recomputes package-level lint from parsed source (`PackageCustoms.lintDiagnosticsFromSource`) instead of the optimized cached let graph, so top-level lint stays stable on cache hits. `PackageMap` now preserves successful inferred packs alongside hard failures, which lets `--warn` keep replaying cached lint even when another package fails hard in the same run. Lint mode is also namespaced into the cache key via a lint-specific `InferPhases.id`, so strict runs do not reuse relaxed artifacts.

I also cleaned up the `library.Command` plumbing to use concrete `LintMode` values end-to-end, split error vs lint rendering in `CompilerApi`, and replaced the optimize/cache regression test with a direct `CompilerApi` test using `CompileOptions.Default` because the `lib test` memory-mode runtime failure path drops prior stderr writes in the `StateT[Either, ...]` harness.

Verified locally with:
- `sbt "coreJVM/testOnly dev.bosatsu.ToolAndLibCommandTest -- --log=failure"`
- `sbt "coreJS/testOnly dev.bosatsu.ToolAndLibCommandTest -- --log=failure"`
- `./test_cli.sh`
- `sbt "doc; paradox"`
- `scripts/test_basic.sh`

core/src/main/scala/dev/bosatsu/Package.scala
```diff
@@ -777,82 +791,76 @@ object Package {
...
               val topLevelDefs = TypedExprRecursionCheck.topLevelDefArgs(stmts)
 
-              val recursionCheck: ValidatedNel[PackageError, Unit] =
+              val recursionErrors: List[PackageError] =
                 TypedExprRecursionCheck
                 .checkLets(p, fullTypeEnv, typedLets, topLevelDefs)
-                .leftMap(
-                  _.map(err => PackageError.RecursionError(p, err): PackageError)
-                    .toNonEmptyList
+                .fold(
```
johnynek: Let's add a method:

```
def toErrs[A](v: ValidatedNel[A, Unit]): List[A]

def toErrsChain[A](v: ValidatedNec[A, Unit]): Chain[A]

```

And use that rather than fold since it's more clear.

core/src/main/scala/dev/bosatsu/Package.scala
```diff
@@ -595,6 +595,18 @@ object Package {
     }
   }
 
+  private def diagnosticsToIor[A](
```
johnynek, line 606: We use this twice with Validated and convert first.

What if we make another function to work directly on validated:

```
def toIorPostponable[F[_]: Foldable, A](
  v: Validated[F[A], Unit]): Ior[F[A], Unit] =
  v match {
    case Validated.Valid(b) => Ior.right(b)
    case Validated.Invalid(as) =>
      if (as.forall(isPostponable)) => Ior.both(as, ())
      else Ior.left(as)
  }
```

Then use that rather than convert to lists and then back to Ior.

core/src/main/scala/dev/bosatsu/PackageCustoms.scala
```diff
@@ -27,9 +27,54 @@ object PackageCustoms {
...
+      case Validated.Invalid(errs) =>
+        errs.toChain.toList
+    }
+
+  // Replay only the postponable customs on successful packages so cache hits
+  // surface the same lint diagnostics as cache misses. Use the source-level
+  // lets/import graph here so optimized cached packages do not change the
+  // lint set by dropping unused private top levels.
+  def lintDiagnosticsFromSource(
+      pack: Package.Inferred,
```
johnynek: This method take the Package.Inferred but it only uses the name, the main value and the test value. I think it would be clearer to just pass those in as arguments to make it clear we aren't using the package for anything else. This function takes so many arguments, it's unclear what it actually depends on.

Also, I think it is weird to do the work to produce all the errors but filter at the end just to postponable. Can't the caller do that?

It feels expensive to compute everything and then discard inside.

core/src/main/scala/dev/bosatsu/PackageCustoms.scala
```diff
@@ -383,104 +428,95 @@ object PackageCustoms {
...
 
-  private def noUselessBinds[A: HasRegion](
-      pack: Package.Typed[A]
+  private def noUselessBindsFromLets[A: HasRegion](
+      packName: PackageName,
+      exports: List[ExportedName[?]],
+      lets: List[(Bindable, RecursionKind, A)],
+      sourceStmts: List[Statement],
+      testRoots: Set[Bindable],
+      mainRoot: Option[Bindable],
```
johnynek: `mainRoot` and `testRoots` are only used to convert to lists and put into `roots`. Maybe it's clearer in this private def to just accept `roots: List[Bindable]` and the caller can set that up with testRoots + main root.

core/src/main/scala/dev/bosatsu/PackageError.scala
```diff
@@ -236,6 +236,22 @@ object PackageError {
       }
   }
 
+  def isPostponable(err: PackageError): Boolean =
```
johnynek, line 242: Add a comment to this method and explain this separates which errors are "lints" that don't need to be enforced for correctness and can be disabled with development mode with --lax or --warn

core/src/main/scala/dev/bosatsu/PackageMap.scala
```diff
@@ -787,27 +787,43 @@ object PackageMap {
...
-        dedupeErrors(resultMap.sequence.map { inferredByName =>
-          PackageMap(
-            SortedMap.from(
-              inferredByName.iterator.map { case (name, inferredPack) =>
-                name -> inferredPack.inferred
-              }
+        // Keep successfully inferred packages alongside hard failures so later
+        // callers can still replay lint for the warm-cache-success subset.
+        val (errsRev, inferredByName) =
+          resultMap.iterator.foldLeft(
```
johnynek: Isn't this just `.sequence` and then on `Ior.Left` making an empty SortedMap?

You can `.sequence` a SortedMap to get `Ior[Errors, SortedMap[PackageName.Inferred]]`. That would be better than reimplementing it with a complex fold.

core/src/main/scala/dev/bosatsu/PackageMap.scala
```diff
@@ -787,27 +787,43 @@ object PackageMap {
...
+                  packsAcc.updated(name, inferred.inferred)
+                )
+            }
+          }
+
+        val dedupedErrs = errsRev.reverse.distinct
+        val inferred = PackageMap(inferredByName)
+
+        NonEmptyList.fromList(dedupedErrs) match {
+          case Some(errs) if inferredByName.nonEmpty => Ior.Both(errs, inferred)
```
johnynek: Aren't we doing a custom sequence, then jumping out of the Ior, then going right back in?

Would a `sequence` with a dedup of the errors be the same thing and a lot less code?

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
```diff
@@ -221,24 +346,57 @@ object CompilerApi {
...
+      if (totalDiagnostics == 1) "warning" else "warnings"
+    Doc.text(s"$totalDiagnostics $summaryWord: ") +
       Doc.intercalate(Doc.text(", "), parts)
   }
 
   private def renderMultiErrorMessage(
-      diagnostics: List[RenderedDiagnostic]
+      diagnostics: List[RenderedError]
+  ): Doc = {
+    val divider = "-" * 72
```
johnynek: `val divider = Doc.char('-') * 72` is more efficient and already a doc.

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
```diff
@@ -221,24 +346,57 @@ object CompilerApi {
...
+    }
+    val entriesDoc = Doc.intercalate(entrySeparator, entries)
+    val summaryDoc = renderErrorSummary(diagnostics)
+    entriesDoc + Doc.hardLine + Doc.hardLine + dividerDoc + Doc.hardLine + summaryDoc
+  }
+
+  private def renderMultiWarningMessage(
+      diagnostics: List[RenderedLint]
   ): Doc = {
     val divider = "-" * 72
```
johnynek: `Doc.char('-') * 72` is better.

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
```diff
@@ -264,6 +422,301 @@ object CompilerApi {
...
+
+  private def localTypeNames(pack: Package.Inferred): Set[TypeName] =
+    pack.types.definedTypes.keysIterator.collect {
+      case (packageName, tn) if packageName == pack.name => tn
+    }.toSet
+
+  private def isScalaJsUndefinedBehavior(err: Throwable): Boolean =
+    err.getClass.getName == "org.scalajs.linker.runtime.UndefinedBehaviorError"
+
+  private def bestEffortTypedReplay(
```
johnynek: I don't understand why we need this and I really want to remove it. It is very fishy that the code should work differently on scalajs here.

If we cannot remove it we need a very detailed comment as to what is it about the code that makes it work on the JVM but not on scalajs. It should be a pure logic code. We shouldn't be doing any reflection or unsafe behavior that has different mechanics on scalajs.

core/src/main/scala/dev/bosatsu/Package.scala
johnynek: Added `toErrs` and `toErrsChain` in `Package` and switched the validated-to-list conversions for recursion, shadowed-binding, and totality diagnostics over to those helpers.

core/src/main/scala/dev/bosatsu/Package.scala
johnynek: Added `toIorPostponable` for `Validated[F[A], Unit]` and used it directly for the expr-dag bindable checks and unused-let path instead of converting through lists first.

core/src/main/scala/dev/bosatsu/PackageCustoms.scala
johnynek: Changed `lintDiagnosticsFromSource` to take only the pieces it actually uses: `packName`, explicit `roots`, imports, exports, and the parsed source program. It no longer filters to postponable diagnostics internally; the caller now does that after combining replayed diagnostics.

core/src/main/scala/dev/bosatsu/PackageCustoms.scala
johnynek: Updated the private helper to take `roots: List[Bindable]` directly, and the caller now constructs `testRoots ++ mainRoot` before calling it.

core/src/main/scala/dev/bosatsu/PackageError.scala
johnynek: Added a comment on `PackageError.isPostponable` explaining that it defines the development-mode lint set that `--warn` can demote and `--lax` can suppress, while everything else stays correctness-critical.

core/src/main/scala/dev/bosatsu/PackageMap.scala
johnynek: Reworked the `PackageMap` aggregation to use `sequence` instead of the custom fold. The only extra step is mapping a package-local `Ior.Left(errs)` to `Ior.Both(errs, None)` before sequencing so we can still preserve other successfully inferred packages for lint replay.

core/src/main/scala/dev/bosatsu/PackageMap.scala
johnynek: Yes, this is now a `sequence`-based flow with error dedup on the left side afterward, rather than custom-sequencing and then rebuilding `Ior` by hand.

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
johnynek: Switched the divider construction to `Doc.char('-') * 72` in the multi-error renderer.

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
johnynek: Switched the warning divider to `Doc.char('-') * 72` as well.

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
johnynek: I tried removing the helper entirely and reran the Scala.js command suite; it reproduced the cached replay failure in `lib check --warn reports lint warnings on cache hits` with `org.scalajs.linker.runtime.UndefinedBehaviorError: java.lang.ClassCastException: undefined cannot be cast to dev.bosatsu.Declaration`. I restored the helper as `bestEffortCachedTypedReplay` and added a detailed comment tying it to `CompileCache.get` deserializing cached packages with erased source tags and then casting them back to `Package.Inferred`. That makes the platform difference explicit: the logic is the same, but Scala.js surfaces the bad cached-tag read more eagerly. Since this path only replays optional lint on cache hits, the helper keeps those replay-only failures from turning a successful cached compile into a hard command failure.

johnynek:
Applied the review-driven cleanup across `Package`, `PackageCustoms`, `PackageError`, `PackageMap`, and `CompilerApi`: validated helpers are clearer, the source-lint replay API now takes explicit dependencies instead of a whole package, the package-map aggregation is sequence-based, and the divider rendering uses `Doc.char('-') * 72`. I also rechecked the cached typed replay behavior directly: removing the helper reproduces the Scala.js cache-hit failure, so the final code keeps the helper with a concrete cache-format comment rather than a vague platform workaround.

Verification run after the edits:
- `sbt "coreJVM/testOnly dev.bosatsu.ToolAndLibCommandTest -- --log=failure"`
- `sbt "coreJS/testOnly dev.bosatsu.ToolAndLibCommandTest -- --log=failure"`
- `scripts/test_basic.sh`

core/src/main/scala/dev/bosatsu/PackageMap.scala
```diff
@@ -787,27 +787,40 @@ object PackageMap {
...
+          resultMap.map { case (name, result) =>
+            name -> (result match {
+              case Ior.Left(errs)     =>
+                Ior.both(errs, Option.empty[(PackageName, Package.Inferred)])
+              case Ior.Right(inferred) =>
+                Ior.right(Some(name -> inferred.inferred))
+              case Ior.Both(errs, inferred) =>
+                Ior.both(errs, Some(name -> inferred.inferred))
+            })
+          }.sequence
```
johnynek: Map followed by sequence is the same thing as traverse.

Please make that change.

johnynek: Changed the `PackageMap` aggregation to use `resultMap.traverse` directly instead of `map(...).sequence`. Since `SortedMap` traverse preserves the keys and only traverses the values, I also adjusted the rebuild step to collect `(name, pack)` pairs from the traversed map.

johnynek:
Applied the requested `traverse` cleanup in `PackageMap` and reverified with `scripts/test_basic.sh` (passed: `Passed: Total 60, Failed 0, Errors 0, Passed 60`).

johnynek:
read the diff of PR #2218 which just merged to origin/main as 4239a890e137f7ee534975e67971395a5a28f0b9 -- it conflicts with this PR, but it resolves some design flaws in the previous code.

Adapt this PR to deal with the changes to the InferCache and the new Package.Compiled.

johnynek:
Adapted this PR to the `origin/main` changes from merged PR #2218. The lint-mode path now works against `Package.Compiled` and the new `InferCache` flow instead of cached inferred packages: `PackageMap` preserves successful compiled packages alongside hard failures for replay, `CompilerApi` replays typed/source lint from compiled artifacts, and the old Scala.js-specific cached replay workaround was removed in favor of the new replay-safe totality/shadowed-binding checks already on main. I also updated the test helper to the compiled-package return type and added stable diagnostic deduping by rendered identity so a cold compile plus replay no longer double-reports the same warning.

Verification:
- `sbt "coreJVM/testOnly dev.bosatsu.ToolAndLibCommandTest -- --log=failure"`
- `scripts/test_basic.sh`

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
```diff
@@ -80,175 +125,285 @@ object CompilerApi {
...
           ErrorCategory.ImportError,
           "import error",
           1,
           body
         )
       case _: PackageError.UnknownImportName[?, ?] =>
-        RenderedDiagnostic(
+        RenderedError(
           ErrorCategory.ImportError,
           "import error",
```
johnynek: we repeat a lot of these strings. Could we have these strings connected to `ErrorCategory.ImportError.title` or something so we don't keep rewriting this constant (with an opportunity for a typo in each one). Similarly with the other patterns like this (totality error, package error).

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
```diff
@@ -80,175 +125,285 @@ object CompilerApi {
...
+            ErrorCategory.PackageError.singularLabel,
+            ErrorCategory.PackageError.pluralLabel,
+            count
+          ),
           count,
           body
         )
     }
 
+  private def classifyWarning(
```
johnynek, line 201: add a comment that this method `requires that err is a lint-warning, not a general error`.

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
```diff
@@ -264,6 +419,302 @@ object CompilerApi {
         platformIO.moduleIOMonad.raiseError(ParseErrors(errs, color))
     }
 
+  private def partitionDiagnostics(
+      errors: List[PackageError]
+  ): (List[PackageError], List[PackageError]) =
+    errors.partitionMap { err =>
```
johnynek: we don't need `partitionMap` I think just `partition` will do: `errors.partition(PackageError.isPostponable)` but the postponable ones will be on the left side of the tuple.

Speaking of which I think maybe:
```scala
case class PartitionedErrors(postponable: List[PackageError], errors: List[PackageError])

(postpone, errs) = errors.partition(PackageError.isPostponable)
PartitionErrors(postpone, errs)
```

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
```diff
@@ -264,6 +419,302 @@ object CompilerApi {
...
+      .map(_.flatten)
+  }
+
+  private def phasesForLintMode(
+      lintMode: LintMode
+  ): InferPhases = {
+    val base = InferPhases.default
+    // Cache the strict and relaxed lint modes separately even though they share
+    // the same finishPackage steps. Strict runs must not reuse artifacts that
+    // only existed because warn/lax allowed postponable diagnostics through.
```
johnynek: warn and lax allow the same code though right? We could have a key that separates shares a key for warn and lax and uses a different one for strict. Maybe just `warn_or_lax` and `strict` We could add function like `def lintModeCacheKey(lm: LintMode): String`

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
```diff
@@ -325,45 +776,136 @@ object CompilerApi {
...
+              moduleIOMonad.raiseError(
+                PackageErrors(diagnostics.sourceMap, errs, errColor)
+              )
+            case None       =>
+              compiledOrError(diagnostics)
+          }
+        case LintMode.Warn   =>
+          diagnostics.hardDiagnostics match {
+            case _ :: _ =>
+              (
```
johnynek: we don't need these wrapping parens

core/src/main/scala/dev/bosatsu/Package.scala
```diff
@@ -785,82 +820,76 @@ object Package {
...
+                    .leftMap(
+                      _.map(err =>
+                        PackageError.ShadowedBindingTypeError(
+                          p,
+                          err,
+                          localTypeNames
+                        ): PackageError
+                      )
+                    )
+                ).toList
```
johnynek: instead of converting both `recursionErrors` and `shadowedBindingErrors` to list and then doing the concat, keep them as a chains and below do:
```scala
(recursionErrors ++ shadowedBindingErrors).toList ::: totalityErrors
```
concat on chain is O(1), and then we convert to list to prepend onto the totalityErrors (which are already a list).

core/src/main/scala/dev/bosatsu/PackageCustoms.scala
```diff
@@ -383,104 +426,95 @@ object PackageCustoms {
...
+
+  private def noUselessBinds[A: HasRegion](
+      pack: Package.Typed[A],
+      sourceStatements: Option[List[Statement]]
+  ): ValidatedNec[PackageError, Unit] = {
+    val sourceStmts: List[Statement] =
+      sourceStatements.getOrElse {
+        pack.program._1.from match {
+          case stmts: List[?] =>
+            stmts.collect { case stmt: Statement => stmt }
```
johnynek: this branch is matching on `from: Any`, and if it is `List[Statement]` we have the statments. But this only happens when we are compiling from the source, and in that case we would also have the sourceStatements wouldn't we?

I don't see that this actually works. I think we could have a method that reflectively gets statements via:

```scala
def maybeStatements[A](pack: Package.Typed[A]): Option[List[Statement]]
```
and then, we could call `noUselessBindsFromLets` if that is `Some` but we shouldn't silently convert that to a list with `.getOrElse(Nil)` and not comment that we aren't actually checking anything in that branch.

core/src/main/scala/dev/bosatsu/PackageMap.scala
```diff
@@ -802,15 +802,44 @@ object PackageMap {
...
         )
       } else {
-        dedupeErrors(resultMap.sequence.map { compiledByName =>
-          PackageMap(
-            SortedMap.from(
-              compiledByName.iterator.map { case (name, compiledPack) =>
-                name -> compiledPack.compiled
-              }
+        val sequenced =
+          resultMap.traverse { result =>
```
johnynek: we don't need `{ result => result match {` in scala. We can just put the cases at the top level:
```scala
resultMap.traverse {
  case Ior.Left(errs) =>
  case Ior.Right(compiledPack) =>
  case Ior.Both(errs, compiledPack) =>
}
```

core/src/main/scala/dev/bosatsu/PackageMap.scala
```diff
@@ -802,15 +802,44 @@ object PackageMap {
...
+              case Ior.Left(errs)          =>
+                Ior.both(errs, Option.empty[CompiledPack[CacheDepHash]])
+              case Ior.Right(compiledPack) =>
+                Ior.right(Some(compiledPack))
+              case Ior.Both(errs, compiledPack) =>
+                Ior.both(errs, Some(compiledPack))
+            }
+          }
+
+        dedupeErrors(sequenced) match {
```
johnynek: this would be clearer if we map first, and then filter Both with empty Map:
```
dedupeErrors(sequenced)
  .map { entries =>
              PackageMap(
                SortedMap.from(
                  entries.iterator.collect { case (name, Some(compiledPack)) =>
                    name -> compiledPack.compiled
                  }
                )
  } match {
  case Ior.Both(errs, compiled) if compiled.toMap.isEmpty => Ior.left(errs)
  case notEmpty => notEmpty
}
```

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
johnynek: Pulled the repeated singular/plural titles onto `ErrorCategory.title(count)` / `LintCategory.title(count)` and added small `renderedError` / `renderedLint` helpers, so the repeated `import error` / `package error` / `totality error` strings now come from the category definitions instead of being retyped at each match arm.

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
johnynek: Added that comment above `classifyWarning`: it now explicitly says this method assumes the input was already partitioned as a postponable lint warning, and that passing a hard error there is a bug.

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
johnynek: Replaced the `partitionMap` tuple with a small `PartitionedErrors(postponable, errors)` case class and now use `errors.partition(PackageError.isPostponable)` directly.

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
johnynek: Changed the cache-mode naming so warn/lax share the same key. `phasesForLintMode` now uses a `lintModeCacheKey` helper with `strict` vs `warn_or_lax`.

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
johnynek: Removed the extra wrapping parens around the `NonEmptyList.fromList(... ) match` in the warn branch.

core/src/main/scala/dev/bosatsu/Package.scala
johnynek: Kept `recursionErrors` and `shadowedBindingErrors` as `Chain[PackageError]` and now concatenate them as chains before the final `.toList ::: totalityErrors`.

core/src/main/scala/dev/bosatsu/PackageCustoms.scala
johnynek: Reworked that path to use a dedicated `maybeStatements(pack)` helper returning `Option[List[Statement]]`. `noUselessBinds` now only runs when source statements are actually available, and the `None` branch is explicit with a comment explaining why the lint is skipped for compiled/cache-only packages.

core/src/main/scala/dev/bosatsu/PackageMap.scala
johnynek: Changed the `resultMap.traverse` call to the direct `case ... =>` style instead of wrapping `result => result match`.

core/src/main/scala/dev/bosatsu/PackageMap.scala
johnynek: Restructured that block to map the traversed entries into a `PackageMap` first, then collapse `Ior.Both(errs, compiled)` to `Ior.Left(errs)` only when the compiled map is empty.

johnynek:
Applied the review cleanup across `CompilerApi`, `Package`, `PackageCustoms`, and `PackageMap`: category-driven diagnostic titles, explicit lint/hard partitioning, shared warn/lax cache keying, the `maybeStatements` source-availability helper, and the requested chain/traverse simplifications. Verified with `scripts/test_basic.sh`.

Result: `Passed: Total 60, Failed 0, Errors 0, Passed 60`.

PR #2218, opened 2026-03-19
Return region-tagged compiled artifacts from CompileCache

### LINKED ISSUE #2216: remove unsound cast in CompileCache
In CompileCache, there is a some code like:

```scala
            moduleIOMonad.fromTry {
              for {
                protoPackages <- Try(proto.Packages.parseFrom(bytes))
                decoded <- ProtoConverter
                  .packagesFromProto(Nil, protoPackages.packages, depIfaces)
              } yield decoded._2
            }.map {
              case pack :: Nil if pack.name == key.packageName =>
                // Serialized package artifacts are tag-erased to Unit.
                Some(pack.asInstanceOf[Package.Inferred])
              case _ =>
                statsUpdate { getCasDecodeMisses.incrementAndGet(); () }
                None
            }
```

The line:
```scala
                // Serialized package artifacts are tag-erased to Unit.
                Some(pack.asInstanceOf[Package.Inferred])
```
is unsound. The deserialized package really doesn't have the declarations and it really doesn't know what regions the code came from. If we cast it, elsewhere when we attempt to read the regions (which other code in later PRs have tried to do), we get runtime crashes.

We never should have merged the code with this unsound cast in place, but we have to fix it. We should have a near zero tolerance policy for `asInstanceOf` and each usage needs to have a clear explanation as to why it is impossible to have a runtime issue. This case clearly does have an issue.

We need to design a solution to the problem.

Possible solutions:

1. return `Package.Typed[Any]` and just not make any promise about what is in the tag. This will require downstream callers to be changed but may be the smallest change
2. serialize the `Region` into the full packages (and zero it out on the Interfaces) and potentially the blake3 hash of the source utf8 into the package object (but don't set it on Interfaces so we can have a minimal impact on caching that uses interfaces data hash as a key). Then change the return type to `Package.Typed[Region]` and make sure `HasRegion[Region]` as an identify function exists. Then any callers that need the region can get it.

Maybe some other solutions will also work.

Read the current code and consider solutions to this problem and propose the best solution to solve this issue. Keeping in mind we do have uses that are blocked on this working correctly (e.g. PR: #2206 is blocked on having some work around to this unsound cast and that cannot merge until we can solve this problem without the cast. It may be helpful to read that diff and see what that PR is needing from the returned value from the cache and see if we can actually easily provide it. If it really needs `Package.Typed[Declaration]` that will almost certainly be impossible. I don't want to reserialize the entire source code AST into the compiled package. I think we either need to do `Typed[Any]` or `Typed[Region]`. I can't see too many other possibilities at the moment, but maybe others will.

johnynek:
Implemented issue #2216 per the merged design doc by replacing the compile-cache boundary with truthful `Package.Compiled` (`Package.Typed[Region]`) artifacts instead of pretending cached packages are `Package.Inferred`. The change adds region serialization to `TypedAst.proto` and `ProtoConverter`, preserves typed-expression and match-branch pattern regions across package round-trips, bumps `CompileCache.schemaVersion` to invalidate old regionless cache entries, and updates `PackageMap`, `CompilerApi`, platform/package IO, library plumbing, and output types to return compiled region-tagged packages at the public compile boundary. Source-sensitive flows still stay on the cold path via `Package.Inferred`, while region-only replay paths were widened in `ShadowedBindingTypeCheck`, `TotalityCheck`, and `PackageError`; tests now cover warm-cache region access, schema misses, protobuf round-trips, and CLI/library cache-hit behavior. Also kept older non-cache serialized package artifacts readable by defaulting missing decoded regions to `Region.empty`, and made the new inferred-only test helper thread-safe with a `TrieMap`. Verification passed with `sbt "coreJVM/test"`, `sbt "cli/test"`, `scripts/test_basic.sh`, and `./test_cli.sh`.

Fixes #2216

Implements design doc: [docs/design/2216-remove-unsound-cast-in-compilecache.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/2216-remove-unsound-cast-in-compilecache.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/2217

### DIFF
--- a/core/.js/package-lock.json
+++ b/core/.js/package-lock.json
@@ -416,7 +416,6 @@
       "integrity": "sha512-j3fVLgvTo527anyYyJOGTYJbG+vnnQYvE0m5mmkc1TK+nxAppkCLMIL0aZ4dblVCNoGShhm+kzE4ZUykBoMg4g==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "fast-deep-equal": "^3.1.1",
         "fast-json-stable-stringify": "^2.0.0",
@@ -873,7 +872,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "caniuse-lite": "^1.0.30001688",
         "electron-to-chromium": "^1.5.73",
@@ -5478,7 +5476,6 @@
       "integrity": "sha512-B/gBuNg5SiMTrPkC+A2+cW0RszwxYmn6VYxB/inlBStS5nx6xHIt/ehKRhIMhqusl7a8LjQoZnjCs5vhwxOQ1g==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "fast-deep-equal": "^3.1.3",
         "fast-uri": "^3.0.1",
@@ -5892,7 +5889,6 @@
       "integrity": "sha512-x7lrWZ7wlWAdyKdML6YPvfVZkhD1ICuIZGODE5SzKJjqI9A4SpqGTjGJTc6CwaHqn19gGaoOR3ONJ46nYsn9rw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@types/eslint-scope": "^3.7.0",
         "@types/estree": "^0.0.46",
@@ -5940,7 +5936,6 @@
       "integrity": "sha512-wXg/ef6Ibstl2f50mnkcHblRPN/P9J4Nlod5Hg9HGFgSeF8rsqDGHJeVe4aR26q9l62TUJi6vmvC2Qz96YJw1Q==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@discoveryjs/json-ext": "^0.5.0",
         "@webpack-cli/configtest": "^1.0.1",
--- a/core/src/test/scala/dev/bosatsu/TestUtils.scala
+++ b/core/src/test/scala/dev/bosatsu/TestUtils.scala
@@ -8,6 +8,7 @@ import dev.bosatsu.rankn._
 import dev.bosatsu.tool.Output
 import munit.Assertions.{assertEquals, fail}
 import IorMethods.IorExtension
+import scala.collection.mutable
 import scala.collection.immutable.SortedMap
 
 import cats.syntax.all._
@@ -275,8 +276,8 @@ object TestUtils {
   ): Ior[NonEmptyList[PackageError], PackageMap.Inferred] = {
     import Par.F
 
-    val captured =
-      scala.collection.concurrent.TrieMap.empty[PackageName, Package.Inferred]
+    val capturedLock = new AnyRef
+    val captured = mutable.HashMap.empty[PackageName, Package.Inferred]
     val defaultPhases = InferPhases.default
     val capturePhases = new InferPhases {
       val id: String = defaultPhases.id
@@ -291,7 +292,9 @@ object TestUtils {
       ): Package.Inferred = {
         val finished =
           defaultPhases.finishPackage(pack, depIfaces, compileOptions)
-        captured.update(finished.name, finished)
+        capturedLock.synchronized {
+          captured.update(finished.name, finished)
+        }
         finished
       }
     }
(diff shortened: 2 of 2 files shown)


core/src/main/scala/dev/bosatsu/Region.scala
```diff
@@ -66,6 +66,12 @@ object HasRegion {
       def region(t: T) = fn(t)
     }
 
+  given HasRegion[Region] =
+    instance(identity)
+
+  given HasRegion[Unit] =
```
johnynek: Why do we need this? Looks dangerous. Can we comment where it is used and why it's safe.

core/src/main/scala/dev/bosatsu/ShadowedBindingTypeCheck.scala
```diff
@@ -204,31 +197,33 @@ object ShadowedBindingTypeCheck {
...
         argCheck *> branchCheck
     }
 
-  def checkLets(
+  def checkLets[A: HasRegion](
       pack: PackageName,
-      lets: List[(Bindable, RecursionKind, TypedExpr[Declaration])]
+      lets: List[(Bindable, RecursionKind, TypedExpr[A])]
   ): Res[Unit] = {
     val _ = pack
```
johnynek, line 212: Can we remove this?

johnynek:
Reproduced the reported CI failure locally and fixed the missed fallout from the compiled-artifact type change. `bench/src/main/scala/dev/bosatsu/TestBench.scala` now uses `PackageMap.Compiled`, which fixes the bench/docs compile break after the compile boundary moved from `Declaration` tags to `Region` tags. While validating the docs path, scaladoc also crashed on the new public `TypedExpr.Branch.copy` signature, so I made that helper `private[bosatsu]`; internal callers still preserve `patternRegion`, but the problematic signature is no longer part of the public docs surface. Local verification passed with `sbt "bench/compile"`, `scripts/test_basic.sh`, and `sbt "doc; paradox"`.

core/src/main/scala/dev/bosatsu/Region.scala
johnynek: Removed the `HasRegion[Unit]` instance instead of trying to justify it. That keeps production code from fabricating empty regions implicitly. The only default instance left here is `HasRegion[Region]`, which is just the identity for already-region-tagged compiled artifacts.

core/src/main/scala/dev/bosatsu/ShadowedBindingTypeCheck.scala
johnynek: Removed. `checkLets` now reads the top-level binding region directly with `HasRegion.region(expr.tag)` and passes `HasRegion.region[A]` into `checkExpr` instead of stashing a local function first.

johnynek:
Addressed the remaining feedback in code. `ProtoConverterTest` now has a ScalaCheck round-trip law over generated, well-typed source packages that compile to `Package.Compiled`, then encode/decode through protobuf and compare for equality. I also updated the interface/package-overlap proto test to use real regions now that the global `HasRegion[Unit]` fallback is gone. Validation passed with `sbt "coreJVM/testOnly dev.bosatsu.ProtoConverterTest"`, `sbt "coreJVM/testOnly dev.bosatsu.ShadowedBindingTypeCheckTest"`, and `scripts/test_basic.sh`.

core/src/main/scala/dev/bosatsu/PackageMap.scala
```diff
@@ -976,6 +979,66 @@ object PackageMap {
     )
   }
 
+  /** Internal cold-path helper for tests and source-sensitive compiler checks
+    * that still need Declaration-tagged packages after finish phases run.
+    */
+  private[bosatsu] def typeCheckParsedInferred[A: Show](
```
johnynek: let's move this into the test code where it is called. I really don't like the trick of getting the Package.Inferred from a side effect pass of recompiling the code. I'm worried someone will reach for it outside of tests. Let's move it to TestUtils or something.

johnynek: Moved this helper out of `PackageMap` and into `core/src/test/scala/dev/bosatsu/TestUtils.scala`, with a test-only comment explaining why it exists. `TypedExprRecursionCheckTest` now calls `TestUtils.typeCheckParsedInferred(...)`, and the production `PackageMap` surface no longer exposes the capture-via-finish-phase path.

johnynek:
Validation passed with `sbt "coreJVM/testOnly dev.bosatsu.TypedExprRecursionCheckTest"` and `scripts/test_basic.sh`.

johnynek:
Reproduced the `testJS` linker failure locally, fixed it, and revalidated. The break was the new test-only inferred helper in `core/src/test/scala/dev/bosatsu/TestUtils.scala` pulling in `scala.collection.concurrent.TrieMap`, which Scala.js then failed to link. I replaced that capture map with a synchronized `mutable.HashMap`, preserving JVM safety for the helper without dragging `scala.collection.concurrent` into the JS test bundle. Validation passed with `sbt "coreJS/testOnly dev.bosatsu.TypedExprRecursionCheckTest"`, `sbt "coreJS/testOnly * -- --log=failure; jsapiJS/compile"`, and `scripts/test_basic.sh`.

Issue #2222, opened 2026-03-20
Lift all lib options to top level


johnynek:
Generally, any user interacting with the bosatsu command will use the `lib` subcommands. This means almost always they do `bosatsu lib <cmd>`. This makes the lib feel like an annoying unneeded ceremony.

Design a change so all lib commands are lifted to the top level, then tool is still a sub command to access the compiler without using the default "build tool/package manager" mode of the current lib check/build/test commands.

This should improve discoverability since the main things users want will be top level on --help. It will also reduce the verbosity of the most common commands.

Review the current CLI and identify how we can make this change to improve ergonomics.

johnynek:
MergeXO assigned an agent and started design work for issue #2222.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/2223

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #2222.

johnynek:
Opened implementation PR: https://github.com/johnynek/bosatsu/pull/2226

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/2222#issuecomment-4101485907
- linked PR: #2226 (https://github.com/johnynek/bosatsu/pull/2226)

Comments on the source issue are no longer actioned after a PR exists. Please comment on the PR thread instead.

PR #2223, opened 2026-03-20
Design doc for #2222: Lift Library Commands to the Top-Level CLI

johnynek:
Design doc.

Refs #2222

### DIFF
--- a/docs/design/2222-lift-all-lib-options-to-top-level.md
+++ b/docs/design/2222-lift-all-lib-options-to-top-level.md
@@ -29,7 +29,7 @@ _Issue: #2222 (https://github.com/johnynek/bosatsu/issues/2222)_
 
 ## Summary
 
-Make library-mode commands canonical at the CLI root, keep `tool` as the explicit file-based/compiler namespace, preserve `lib` as a compatibility alias via argument normalization, and update help, docs, tests, and automation to use the new surface.
+Make library-mode commands canonical at the CLI root, keep `tool` as the explicit file-based/compiler namespace, remove `lib` from the documented/tested user surface, and update help, docs, tests, and automation to use the new top-level commands.
 
 ## Context
 
@@ -41,9 +41,9 @@ In practice, the common Bosatsu workflows are library-mode workflows: `check`, `
 
 ## Goals
 
-1. Make the current `bosatsu lib <cmd>` workflows available as canonical `bosatsu <cmd>` workflows.
+1. Make the current library workflows available as canonical `bosatsu <cmd>` workflows.
 2. Keep `tool` as the explicit low-level/file-based namespace.
-3. Preserve existing automation and scripts that still call `bosatsu lib <cmd>`.
+3. Remove `lib` usage from in-repo docs, tests, and automation.
 4. Improve `--help` discoverability by surfacing common library commands directly at the root.
 5. Keep command semantics, flags, and outputs stable.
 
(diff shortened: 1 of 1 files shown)


docs/design/2222-lift-all-lib-options-to-top-level.md
```diff
@@ -0,0 +1,219 @@
...
+3. Do not remove or deprecate `lib` in the same PR. If removal is ever desired, it should be a separate follow-up after the new surface has been in docs and releases for a while.
+4. Because command implementations are being reused rather than rewritten, rollback is straightforward: restore the old root wiring while leaving the command bodies unchanged.
+
+## Alternatives Considered
+
+1. Keep both lifted commands and a visible `lib` subcommand.
+   Rejected because it improves typing ergonomics but still leaves the old ceremony in the primary help surface.
+
+2. Remove `lib` compatibility entirely.
+   Rejected because the repo, docs, CI, and likely downstream scripts currently rely heavily on `lib`.
```
johnynek: We should fully remove any use of lib in the docs or the tests. We shouldn't reinforce in any way the usage of lib.

docs/design/2222-lift-all-lib-options-to-top-level.md
```diff
@@ -0,0 +1,219 @@
...
+
+| Invocation | Behavior |
+| --- | --- |
+| `bosatsu check`, `test`, `build`, `eval`, `json`, `show`, `doc`, `deps`, `fetch`, `init`, `list`, `assemble`, `publish` | Current library-mode implementation |
+| `bosatsu tool <subcommand>` | Current file-based/compiler implementation |
+| `bosatsu lib <subcommand>` | Compatibility alias to the lifted library command |
+
+Canonical usage becomes the top-level form. `tool` remains the escape hatch for compiler/file-based behavior. `lib` remains accepted for compatibility, but it is no longer the documented primary path.
+
+Top-level help should also be reordered so the most common workflows are first. A reasonable order is:
```
johnynek, line 78: I'm not sure decline allows you to control the ordering. Research that and add an explicit comment about how that is done in decline and if it can't just let decline use its default order.

Perhaps decline preserves the order that options are built. If so, we just need to be careful about the construction of the Opts values.

docs/design/2222-lift-all-lib-options-to-top-level.md
johnynek: Updated the design doc to make the steady-state surface top-level only. I removed the proposed `lib` compatibility/testing language, rewrote the testing/docs sections around converting existing in-repo usage to the top-level commands, and added acceptance criteria that docs/tests/automation stop using the old nested spelling.

docs/design/2222-lift-all-lib-options-to-top-level.md
johnynek: I checked Decline 2.6.1 directly and updated the doc with the concrete behavior. The doc now calls out that `Help.collectCommandHelp` and `Help.commandList` both traverse `Opts.App` / `Opts.OrElse` left-to-right with `++`, so subcommand help order follows the order the `Opts.subcommand(...)` values are constructed. The design now explicitly says to control ordering by the `combineAllK` / `orElse` construction order rather than assume custom Decline behavior.

johnynek:
Updated `docs/design/2222-lift-all-lib-options-to-top-level.md` to remove `lib` from the proposed documented/tested surface and to add an explicit Decline 2.6.1 ordering note based on the source implementation. `scripts/test_basic.sh` passed locally.

Issue #2224, opened 2026-03-20
Remove unneeded packages from exports


johnynek:
Review the list of packages we are exporting in core_alpha. We need certain packages like Bosatsu/Lazy and Bosatsu/Eval because the compiler knows about them. But we want to have core be as minimal as we can. Bosatsu/Prog is also a key package. Anything with an external def should probably be exported and any public dependencies of those.

What packages should be exported? 

How can we minimize but retain functionality?

johnynek:
MergeXO assigned an agent and started design work for issue #2224.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/2225

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #2224.

johnynek:
Opened implementation PR: https://github.com/johnynek/bosatsu/pull/2229

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/2224#issuecomment-4101753962
- linked PR: #2229 (https://github.com/johnynek/bosatsu/pull/2229)

Comments on the source issue are no longer actioned after a PR exists. Please comment on the PR thread instead.

PR #2225, opened 2026-03-20
Design doc for #2224: Issue #2224 Design: Minimize `core_alpha` Exports

johnynek:
Design doc.

Refs #2224

### DIFF
--- a/docs/design/2224-remove-unneeded-packages-from-exports.md
+++ b/docs/design/2224-remove-unneeded-packages-from-exports.md
@@ -3,6 +3,7 @@ issue: 2224
 priority: 3
 touch_paths:
   - docs/design/2224-remove-unneeded-packages-from-exports.md
+  - test_workspace/Bosatsu/Json.bosatsu
   - test_workspace/core_alpha_conf.json
   - core/src/test/scala/dev/bosatsu/library/LibConfigTest.scala
   - core/src/test/scala/dev/bosatsu/ToolAndLibCommandTest.scala
@@ -17,7 +18,7 @@ _Issue: #2224 (https://github.com/johnynek/bosatsu/issues/2224)_
 
 ## Summary
 
-Reduce `core_alpha`'s public surface to compiler-known packages, command-known or intentionally public utility packages, packages with runtime-backed `external` APIs, and the small local dependency closure required by those packages. Implement this as an explicit allowlist change in `core_alpha_conf.json`, treat it as a major-version break, and add regression coverage so the trimmed export set does not drift.
+Reduce `core_alpha`'s public surface to compiler-known packages, command-known or intentionally public utility packages, packages with runtime-backed `external` APIs, and the small local dependency closure required by those packages. Refactor `Bosatsu/Json` to replace `Nat`-based fuel with `Int`-based recursion so `Bosatsu/Num/Nat` does not remain exported solely as an implementation dependency, then lock the smaller allowlist in `core_alpha_conf.json` with regression coverage.
 
 ## Context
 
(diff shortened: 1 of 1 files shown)


docs/design/2224-remove-unneeded-packages-from-exports.md
```diff
@@ -0,0 +1,163 @@
...
+- `Bosatsu/IO/Error`
+  Reason: `Bosatsu/IO/Core` exposes `Prog[IOError, ...]` APIs.
+
+Notably absent from the closure:
+
+- `Bosatsu/List`, `Bosatsu/Option`, and `Bosatsu/Bool`
+  Reason: `List`, `Option`, `Bool`, `Char`, `Int`, `String`, and `Float64` types come from predef/type space; exporting helper packages is not required just because those types appear in signatures.
+- `Bosatsu/Rand` and `Bosatsu/Testing/Properties`
+  Reason: they are only used by test code inside packages such as `Bosatsu/Collection/Array`.
+- `Bosatsu/Json`, `Bosatsu/IO/Std`, `Bosatsu/Num/Nat`, `Bosatsu/Num/BinNat`, `Bosatsu/Num/Binary`, `Bosatsu/Dict`, `Bosatsu/Collection/Queue`, `Bosatsu/Collection/TreeList`, `Bosatsu/Nothing`
```
johnynek: I think Bosatsu/Json is known by the json command. Also Bosatsu/IO/Std has a lot of key functions. We should include that.

Verify my claim that the json command knows about Json adt. If it doesn't, it should.

johnynek: Verified and updated the doc. `core/src/main/scala/dev/bosatsu/ValueToJson.scala` has explicit `Bosatsu/Json::{Json, Optional, Nullable}` handling, so I changed the design to keep `Bosatsu/Json` exported. I also changed the proposal to keep `Bosatsu/IO/Std` as intentional public surface, which in turn pulls `Bosatsu/Num/Nat` back in via `Bosatsu/Json`'s implementation. The target allowlist is now 12 packages instead of 9, and the removed-package examples/tests were updated accordingly.

johnynek:
Updated `docs/design/2224-remove-unneeded-packages-from-exports.md` to reflect the verified `Bosatsu/Json` and `Bosatsu/IO/Std` requirements, including the resulting `Bosatsu/Num/Nat` dependency. Ran `scripts/test_basic.sh` locally and it passed.

docs/design/2224-remove-unneeded-packages-from-exports.md
```diff
@@ -0,0 +1,175 @@
...
+- `Bosatsu/Collection/Array`
+- `Bosatsu/Eval`
+- `Bosatsu/IO/Bytes`
+- `Bosatsu/IO/Core`
+- `Bosatsu/IO/Error`
+- `Bosatsu/IO/Std`
+- `Bosatsu/Json`
+- `Bosatsu/Lazy`
+- `Bosatsu/Num/Float64`
+- `Bosatsu/Num/Nat`
```
johnynek: Let's remove Num/Nat from the export set. Rewrite the JSON code that uses it to use Int recursion so we don't need Nat in the JSON code.

johnynek: Updated the design doc to take this path. It now treats `Bosatsu/Json`'s `Nat` usage as an implementation detail to remove, adds `test_workspace/Bosatsu/Json.bosatsu` to `touch_paths`, introduces a `Bosatsu/Json` refactor section describing the `Int` countdown rewrite, removes `Bosatsu/Num/Nat` from the retained export set, and updates the implementation plan/acceptance criteria to verify that `Bosatsu/Num/Nat` is no longer exported while `Bosatsu/Json` behavior stays the same.

johnynek:
Updated `docs/design/2224-remove-unneeded-packages-from-exports.md` to plan the `Bosatsu/Json` refactor that eliminates the `Bosatsu/Num/Nat` dependency and shrinks the proposed export set from 12 packages to 11. Ran `scripts/test_basic.sh` locally and it passed.

Issue #2227, opened 2026-03-21
Add a lint for using recur instead of loop


johnynek:
Using recur or loop when there is no recursion or using recur when the loop is actually a tail loop should both be postponable lints.

They should get warnings which can be suppressed with using --lax.

The idea is: the text of the code should be the true checked state of the code after the compiler runs. Reviewers should be able to trust recur means recursion without a tail loop. And a loop means it is definitely a while loop.

johnynek:
MergeXO assigned an agent and started design work for issue #2227.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/2228

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #2227.

johnynek:
Opened implementation PR: https://github.com/johnynek/bosatsu/pull/2230

johnynek:
MergeXO source-issue comment routing update:
- source comment: https://github.com/johnynek/bosatsu/issues/2227#issuecomment-4101906472
- linked PR: #2230 (https://github.com/johnynek/bosatsu/pull/2230)

Comments on the source issue are no longer actioned after a PR exists. Please comment on the PR thread instead.

PR #2228, opened 2026-03-21
Design doc for #2227: Add postponable recursion-form lints for `recur` and `loop`

johnynek:
Design doc.

Refs #2227

### DIFF
--- a/docs/design/2227-add-a-lint-for-using-recur-instead-of-loop.md
+++ b/docs/design/2227-add-a-lint-for-using-recur-instead-of-loop.md
@@ -161,8 +161,9 @@ Persisting the finalized lint result is the lowest-risk way to keep behavior cor
 6. Extend `PackageError.isPostponable` and `CompilerApi` warning rendering to classify the new lint cleanly in strict, warn, and lax modes.
 7. Persist recursion-form lint metadata through compiled-package serialization and bump the compile-cache schema version.
 8. Replay the stored recursion-form lints from cached compiled packages in `CompilerApi.replayTypedLintDiagnostics`.
-9. Add regression tests for strict, warn, lax, cache-hit replay, and serialization.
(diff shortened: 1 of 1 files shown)


docs/design/2227-add-a-lint-for-using-recur-instead-of-loop.md
```diff
@@ -0,0 +1,201 @@
...
+   Mitigation: only emit recursion-form lints for definitions that finished hard recursion checking successfully.
+
+4. Risk: warning text overpromises backend details.
+   Mitigation: phrase the new lint in terms of the checked `loop` guarantee and reuse the same tail-position classifier that `loop` already depends on.
+
+5. Risk: the new sidecar complicates package serialization.
+   Mitigation: keep the metadata small, source-anchored, and limited to finalized lint results rather than raw checker state.
+
+## Rollout Notes
+1. This change is source-breaking in strict modes for code that currently uses non-recursive `recur` / `loop` or tail-recursive `recur`. The immediate migration path is to rewrite to `match` or `loop`, or temporarily use `lib check --warn` / `--lax`.
```
johnynek, line 199: We should fix all the recur that are loop so we don't need to use warn or lax on any code in the repo that isn't exercising the warning code.

johnynek: Updated the design doc to make this explicit. The implementation plan now includes a mechanical cleanup pass that rewrites existing in-repo tail-recursive `recur` sites to `loop`, and the rollout notes now say the repo should stay strict-clean with warning-triggering examples limited to targeted tests/fixtures.

johnynek:
Updated the design doc to require the in-repo `recur` to `loop` cleanup pass, and ran `scripts/test_basic.sh` locally. It passed.

PR #2229, opened 2026-03-21
Remove unneeded packages from core_alpha exports

### LINKED ISSUE #2224: Remove unneeded packages from exports

johnynek:
Implemented issue #2224 per the merged design doc. `test_workspace/core_alpha_conf.json` now exports only the 11 retained public packages and bumps `core_alpha` from `5.1.0` to `6.0.0`. `test_workspace/Bosatsu/Json.bosatsu` no longer imports `Bosatsu/Num/Nat`; its parser/render budget helpers now use `Int` countdown recursion with Bosatsu-approved `recur` structure so behavior stays bounded without exporting `Nat`.

Added regression coverage in `LibConfigTest` for the exact allowlist, the absence of `Bosatsu/Num/Nat`, and the required major version bump. Added an integration test in `ToolAndLibCommandTest` that publishes a trimmed library, confirms retained packages remain importable through dependency interfaces, and verifies a removed package (`Bosatsu/Num/Nat`) is hidden.

Checks passed: `sbt "coreJVM/test; cli/test"` and `scripts/test_basic.sh`.

Fixes #2224

Implements design doc: [docs/design/2224-remove-unneeded-packages-from-exports.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/2224-remove-unneeded-packages-from-exports.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/2225

### DIFF
--- a/core/.jvm/src/test/scala/dev/bosatsu/library/LibConfigJvmTest.scala
+++ b/core/.jvm/src/test/scala/dev/bosatsu/library/LibConfigJvmTest.scala
@@ -1,0 +1,62 @@
+package dev.bosatsu.library
+
+import dev.bosatsu.{Json, PackageName}
+import java.nio.charset.StandardCharsets
+import java.nio.file.{Files, Paths}
+
+class LibConfigJvmTest extends munit.FunSuite {
+  private def decodeConfig(jsonStr: String): LibConfig = {
+    val json = Json.parserFile.parseAll(jsonStr) match {
+      case Right(value) => value
+      case Left(err)    => fail(s"failed to parse config json: $err")
+    }
+
+    Json.Reader[LibConfig].read(Json.Path.Root, json) match {
+      case Right(value)         => value
+      case Left((msg, got, jp)) =>
+        fail(s"failed to decode config json: $msg, got=$got at $jp")
+    }
+  }
+
+  private def packageName(str: String): PackageName =
+    PackageName.parse(str).getOrElse(fail(s"invalid package name: $str"))
+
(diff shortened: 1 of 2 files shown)


johnynek:
Reproduced the reported `testJS` failure locally. The break was real: `coreJS` link failed because the new shared `LibConfigTest` referenced `java.nio.file.Files/Paths`, which are not available on Scala.js. I moved the file-backed `core_alpha_conf.json` regression into a JVM-only suite at `core/.jvm/src/test/scala/dev/bosatsu/library/LibConfigJvmTest.scala` and restored `core/src/test/scala/dev/bosatsu/library/LibConfigTest.scala` to platform-neutral code.

Validation run:
- reproduced original failure with `sbt "coreJS/testOnly * -- --log=failure; jsapiJS/compile"`
- post-fix Scala.js check passed with `sbt "coreJS/testOnly dev.bosatsu.library.LibConfigTest -- --log=failure; jsapiJS/compile"`
- affected JVM suites passed with `sbt "coreJVM/testOnly dev.bosatsu.library.LibConfigTest dev.bosatsu.library.LibConfigJvmTest dev.bosatsu.ToolAndLibCommandTest -- --log=failure"`
- required pre-push check passed with `scripts/test_basic.sh`

PR #2230, opened 2026-03-21
Add recursion-form lints for non-recursive and tail-recursive recur/loop usage

### LINKED ISSUE #2227: Add a lint for using recur instead of loop

johnynek:
Implemented issue #2227 per the merged design doc. The typed recursion checker now separates hard recursion soundness errors from postponable recursion-form lints, emitting `NoRecursiveCall` for non-recursive `recur`/`loop` blocks and `TailRecursiveRecur` when a source `recur` is fully tail-recursive. These flow through `PackageError.RecursionLint`, get classified cleanly in strict/warn/lax modes, and remain fatal in strict mode while warning in `--warn` and suppressing in `--lax`.

To keep warm-cache behavior identical to cold compiles, finalized recursion-form lint metadata is now stored in `Program` metadata, serialized through `TypedAst.proto`/`ProtoConverter`, replayed from compiled packages in `CompilerApi`, and protected by a compile-cache schema bump. I added regression coverage for recursion lint detection, messages, strict/warn/lax behavior, cache-hit replay, and proto/cache round-tripping, then updated docs and converted in-repo tail-recursive `recur` sites to `loop` so the repository stays strict-clean.

Validation run: `sbt -batch "coreJVM/testOnly dev.bosatsu.TypedExprRecursionCheckTest"`, `sbt -batch "coreJVM/testOnly dev.bosatsu.ProtoConverterTest dev.bosatsu.ToolAndLibCommandTest"`, `sbt -batch "doc; paradox"`, and the required `scripts/test_basic.sh` (passed).

Fixes #2227

Implements design doc: [docs/design/2227-add-a-lint-for-using-recur-instead-of-loop.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/2227-add-a-lint-for-using-recur-instead-of-loop.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/2228

### DIFF
--- a/cli/src/test/scala/dev/bosatsu/PathModuleTest.scala
+++ b/cli/src/test/scala/dev/bosatsu/PathModuleTest.scala
@@ -110,19 +110,64 @@ class PathModuleTest extends munit.FunSuite {
         }
     }
 
-  test("root help lists top-level library commands before tool") {
-    PathModule.run(List("--help")) match {
-      case Left(help) =>
-        val msg = help.toString
-        val checkIdx = msg.indexOf("    check")
-        val toolIdx = msg.indexOf("    tool")
-        assert(checkIdx >= 0, msg)
-        assert(toolIdx >= 0, msg)
-        assert(checkIdx < toolIdx, msg)
-        assert(!msg.contains("\n  lib"), msg)
-      case Right(_) =>
-        fail("expected help output")
+  private def helpText(args: String*): String =
+    PathModule.run(args.toList) match {
+      case Left(help) => help.toString
+      case Right(_)   =>
+        fail(s"expected help output for command: ${args.toList.mkString(" ")}")
     }
+
+  private def subcommands(helpText: String): List[String] =
+    helpText.linesIterator
+      .dropWhile(_ != "Subcommands:")
+      .drop(1)
+      .collect { case line if line.startsWith("    ") && !line.startsWith("        ") =>
+        line.trim
+      }
+      .toList
+
+  test("root help lists top-level commands in the intended order") {
+    val msg = helpText("--help")
+    assertEquals(
+      subcommands(msg),
+      List(
+        "check",
+        "test",
+        "build",
+        "json",
+        "doc",
+        "publish",
+        "eval",
+        "fetch",
+        "deps",
+        "list",
+        "show",
+        "assemble",
+        "init",
+        "tool",
+        "version",
+        "c-runtime"
+      )
+    )
+    assert(!msg.contains("\n  lib"), msg)
+  }
+
+  test("tool help lists subcommands in the intended order") {
+    assertEquals(
+      subcommands(helpText("tool", "--help")),
+      List(
+        "check",
+        "test",
+        "transpile",
+        "json",
+        "doc",
+        "eval",
+        "deps",
+        "show",
+        "assemble",
+        "extract-iface"
+      )
+    )
   }
 
   test("top-level library commands use repo mode") {
--- a/core/src/main/scala/dev/bosatsu/library/Command.scala
+++ b/core/src/main/scala/dev/bosatsu/library/Command.scala
@@ -2741,16 +2741,16 @@ object Command {
       checkCommand ::
         testCommand ::
         buildCommand ::
-        evalCommand ::
         jsonCommand ::
-        showCommand ::
         docCommand ::
-        depsCommand ::
-        fetchCommand ::
-        initCommand ::
-        listCommand ::
-        assembleCommand ::
         publishCommand ::
+        evalCommand ::
+        fetchCommand ::
+        depsCommand ::
+        listCommand ::
+        showCommand ::
+        assembleCommand ::
+        initCommand ::
         Nil
     )
   }
--- a/core/src/main/scala/dev/bosatsu/tool_command/ToolCommand.scala
+++ b/core/src/main/scala/dev/bosatsu/tool_command/ToolCommand.scala
@@ -12,17 +12,19 @@ object ToolCommand {
   ): Opts[F[Output[Path]]] = {
     val commonOpts = new CommonOpts(platformIO)
 
+    // Decline help preserves the left-to-right construction order, so keep
+    // this list aligned with the intended tool workflow surface.
     MonoidK[Opts].combineAllK(
       CheckCommand.opts(platformIO, commonOpts) ::
-        DocCommand.opts(platformIO, commonOpts) ::
         TestCommand.opts(platformIO, commonOpts) ::
-        EvalCommand.opts(platformIO, commonOpts, evalPassthroughArgs) ::
-        JsonCommand.opts(platformIO, commonOpts) ::
         TranspileCommand.opts(platformIO, commonOpts) ::
-        ShowCommand.opts(platformIO, commonOpts) ::
+        JsonCommand.opts(platformIO, commonOpts) ::
+        DocCommand.opts(platformIO, commonOpts) ::
+        EvalCommand.opts(platformIO, commonOpts, evalPassthroughArgs) ::
         DepsCommand.opts(platformIO, commonOpts) ::
-        ExtractIfaceCommand.opts(platformIO) ::
+        ShowCommand.opts(platformIO, commonOpts) ::
         AssembleCommand.opts(platformIO) ::
+        ExtractIfaceCommand.opts(platformIO) ::
         Nil
     )
   }
--- a/docs/src/main/paradox/design-docs/index.md
+++ b/docs/src/main/paradox/design-docs/index.md
@@ -17,6 +17,7 @@ This section collects implementation design documents for Bosatsu.
 * [Issue 1732: Match-Site Struct/Tuple Scrutinee Elision in `Matchless`](issue_1732_match_site_struct_scrutinee_elision_design.md)
 * [Lexicographic `recur` Tuple Design](lexicographic_recur_design.md)
 * [Minimal IO Functions for Basic Tools (`core_alpha` + `Prog`)](minimal_prog_io_tools_design.md)
+* [Prog Concurrency API Design](prog_concurrency_design.md)
 * [`recur ... by int_decrease` Design](recur_int_decrease_design.md)
 @@@
 
@@ -34,6 +35,7 @@ This section collects implementation design documents for Bosatsu.
 * [Issue 1732: Match-Site Struct/Tuple Scrutinee Elision in `Matchless`](issue_1732_match_site_struct_scrutinee_elision_design.html)
 * [Lexicographic `recur` Tuple Design](lexicographic_recur_design.html)
 * [Minimal IO Functions for Basic Tools (`core_alpha` + `Prog`)](minimal_prog_io_tools_design.html)
+* [Prog Concurrency API Design](prog_concurrency_design.html)
 * [`recur ... by int_decrease` Design](recur_int_decrease_design.html)
 
 ## Notes
--- a/docs/src/main/paradox/design-docs/prog_concurrency_design.md
+++ b/docs/src/main/paradox/design-docs/prog_concurrency_design.md
@@ -1,0 +1,1639 @@
+# Prog Concurrency API Design
+
+Status: proposed  
+Date: 2026-03-21
+
+## Problem
+`Bosatsu/Prog` currently has a synchronous execution model on every backend in this repository:
+
+1. JVM evaluator: `run_prog` steps one `Prog` value inline.
+2. Python runtime: `ProgExt.run` steps one `Prog` value inline.
+3. C runtime: `bsts_Bosatsu_Prog_run` steps one `Prog` value inline.
+
+That is enough for sequencing effects, but it leaves two missing pieces:
+
+1. We cannot start a child `Prog` and later join or cancel it.
+2. We cannot move expensive pure work off the main thread, IO thread, or future event-loop thread.
+
+For a future libuv-backed C runtime, the second point is required: CPU-heavy work must not run on the event loop. For JVM and Python, the same API should still make sense even if the concrete scheduler is different.
+
+## Decision Summary
+1. Add `JoinHandle` and `JoinResult` to `Bosatsu/Prog`.
+2. Add `start`, `join`, `cancel`, and `compute` to `Bosatsu/Prog`.
+3. Make `join` return the child outcome as a value, not through the caller's `Prog` error channel.
+4. Define cancellation as cooperative and best-effort across all backends.
+5. Treat `compute` as the standard "shift CPU work off the main scheduler" primitive.
+6. Do not add `race`, `par_map`, `poll`, masking, or finalizers in this first API.
+
+## Proposed API
+
+```bosatsu
+package Bosatsu/Prog
+
+export (
+  ...,
+  JoinHandle,
+  JoinResult(),
+  start,
+  join,
+  cancel,
+  compute,
+)
+
+external struct JoinHandle[err: +*, a: +*]
+
+enum JoinResult[err: +*, a: +*]:
+  Succeeded(value: a)
+  Errored(error: err)
+  Canceled
+
+external def start[err, a](prog: Prog[err, a]) -> forall f. Prog[f, JoinHandle[err, a]]
+external def join[err, a](handle: JoinHandle[err, a]) -> forall f. Prog[f, JoinResult[err, a]]
+external def cancel[err, a](handle: JoinHandle[err, a]) -> forall f. Prog[f, Unit]
+
+# Run a pure, potentially expensive computation on the backend's compute pool
+# or worker threads rather than the main scheduler thread.
+external def compute[a](thunk: Unit -> a) -> forall err. Prog[err, a]
+```
+
+## Why `join` Returns `JoinResult[err, a]`
+The sketched type
+
+```bosatsu
+join(h: JoinHandle[e, a]) -> Prog[e, JoinResult[a]]
+```
+
+mixes two different concerns:
+
+1. did `join` itself fail?
+2. what happened in the child fiber?
+
+For this API, the important failure is the child outcome, not a second effect-level error channel for `join`. Returning `JoinResult[err, a]` makes the contract simpler:
+
+1. `join` itself is total in the typed `Prog` error channel.
+2. child success, child error, and cancellation all live in one value.
+3. repeated `join` calls can deterministically return the same terminal value.
+
+This is the closest Bosatsu analogue to cats-effect's `Outcome`, but simplified because Bosatsu does not yet have finalizers or a need to return `Prog` inside the success case.
+
+## Semantics
+
+### `start`
+1. `start(p)` is still lazy like every other `Prog` constructor; nothing happens until the returned `Prog` is executed.
+2. Once executed, the backend creates a child task and returns a `JoinHandle` immediately.
+3. The child may already be running, queued, or even completed by the time the parent receives the handle. That race is allowed.
+4. `JoinHandle` is opaque and may be shared freely.
+
+### `join`
+1. `join(h)` waits until `h` reaches a terminal state.
+2. The terminal states are `Succeeded(a)`, `Errored(e)`, and `Canceled`.
+3. `join` is idempotent: every successful call returns the same terminal `JoinResult`.
+4. Multiple fibers may join the same handle.
+
+### `cancel`
+1. `cancel(h)` is idempotent.
+2. If the child is already complete, `cancel` is a no-op.
+3. Otherwise, `cancel` records a cancellation request.
+4. Backends should stop scheduling the child at the next cancellation checkpoint.
+5. If a backend can interrupt a blocked operation, it should do so. If it cannot, work may continue in the background and its result is discarded.
+6. If cancellation wins the race against completion, `join(h)` returns `Canceled`.
+
+### Cancellation checkpoints
+Cancellation is not preemptive in the middle of arbitrary pure code. The shared contract is:
+
+1. before entering a backend effect, the runtime may observe cancellation
+2. after resuming from a backend wait, the runtime may observe cancellation
+3. `compute` is a cancellation boundary
+4. long pure Bosatsu loops are not required to notice cancellation until they reach an effect boundary
+
+This is the only semantics that works across JVM threads, Python threads, and libuv worker tasks without unsafe thread killing.
+
+### `compute`
+1. `compute(thunk)` runs `thunk(())` away from the main scheduler thread.
+2. `compute` is for pure, CPU-bound work.
+3. If the caller wants background compute, use `start(compute(thunk))`.
+4. `compute` does not promise parallel speedup on Python because of the GIL, but it still gives a portable "not on the main/event-loop thread" guarantee.
+5. On libuv, `compute` is the standard way to keep the event loop responsive while doing CPU work.
+6. This "pure only" restriction is not optional. Bosatsu functions are pure and do not throw exceptions, so a value of type `Unit -> a` has no typed failure mode. If a later API is needed for shifting blocking effectful work, that should be a separate combinator over `Prog`, not a broader meaning for `compute`.
+
+## Example Usage
+
+```bosatsu
+from Bosatsu/Prog import await, compute, join, start, JoinResult()
+
+background_hash =
+  handle <- compute(() -> expensive_hash(input_bytes)).start()
+  join(handle)
+```
+
+The result is a `Prog` returning a `JoinResult`, because child completion, child error, and cancellation are all explicit values. Higher combinators such as `race` or `par_map2` can be layered on top once the base semantics are stable.
+
+## Shared Runtime Shape
+Backends should implement `JoinHandle` with a small shared state machine:
+
+1. `Running(cancel_requested, waiters...)`
+2. `Succeeded(a)`
+3. `Errored(e)`
+4. `Canceled`
+
+`waiters` here means backend-owned continuations, condition variables, promises, or event-loop callbacks. The Bosatsu surface does not expose this representation.
+
+No new public `Prog` tag is required for the API itself. On JVM and Python, the new operations can still be exposed as ordinary `Prog` externals. A libuv runtime will likely need richer internal scheduler machinery, but that can stay behind the same Bosatsu API.
+
+## Shared `Bosatsu/IO/Core` Process Contract
+The existing `Bosatsu/IO/Core` API already commits us to a specific process model:
+
+```bosatsu
+spawn(cmd: String, args: List[String], stdio: StdioConfig) -> Prog[IOError, SpawnResult]
+wait(p: Process) -> Prog[IOError, Int]
+```
+
+To keep all backends aligned, this design commits to the following shared semantics:
+
+1. `spawn` uses argv semantics, not shell semantics. `cmd` is the program to execute and `args` are literal argument strings.
+2. `spawn` inherits the current working directory and environment of the host runtime unless a future Bosatsu API adds explicit `cwd` or `env` fields.
+3. `spawn` raises `IOError` only if process creation or requested stdio setup fails. A child process exiting non-zero is observed only through `wait`.
+4. `SpawnResult.stdin`, `SpawnResult.stdout`, and `SpawnResult.stderr` are `Some(handle)` exactly when the corresponding `Stdio` entry was `Pipe`. For `Inherit`, `Null`, and `UseHandle`, the returned field is `None`.
+5. `UseHandle(handle)` validates direction at runtime: `stdin` requires a readable handle; `stdout` and `stderr` require writable handles. Closed or invalid handles raise `IOError`.
+6. `wait(p)` is idempotent and may be called multiple times or concurrently. Every successful call returns the same cached exit code.
+7. `wait(p)` does not implicitly close or drain pipe handles returned from `spawn`. If the caller requested `Pipe`, those returned handles remain ordinary Bosatsu handles with their own lifetime.
+8. Canceling a Bosatsu fiber blocked in `wait(p)` never kills the external process. It only cancels that Bosatsu wait.
+9. If a backend has to emulate `UseHandle` with internal copy tasks rather than native OS-level stdio inheritance, `wait(p)` must not complete until those backend-owned bridge tasks have also settled. Otherwise `wait` could return before the requested redirection is actually complete.
+10. `wait(p)` returns a single `Int`, so this API intentionally collapses normal exit and signal termination into one value. On POSIX backends that expose signal termination separately or as negative codes, Bosatsu should normalize signal termination to `128 + signal_number`.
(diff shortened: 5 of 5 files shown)


core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -7,13 +7,31 @@ import dev.bosatsu.pattern.NameMap
...
 import dev.bosatsu.smt.{SExpr, SmtCommand, SmtExpr, SmtLibRender, SmtScript, SmtScriptScope, SmtSort, Z3Api}
+import scala.collection.mutable.ListBuffer
 
 import Identifier.Bindable
 
 object TypedExprRecursionCheck {
 
   type Res[+A] = ValidatedNec[RecursionCheck.Error, A]
 
+  final case class Result(
```
johnynek: I don't like this design at all.

What I would like to do instead is make:
```
type Res[+A] = ValidatedNec[RecursionCheck.Error | RecursionCheck.Lint, A]
```

This way we would process the lints and errors all together, then we could sort them out in the end where we are separating postponable in Package.inferBodyUnopt

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -3630,19 +3660,28 @@ object TypedExprRecursionCheck {
         topLevelLowerableAliases: TopLevelLowerableAliases,
         topLevelPredefAliases: TopLevelPredefAliases,
         totalityCheck: TotalityCheck
-    ): Res[Unit] = {
+    ): Result = {
+      given ListBuffer[RecursionCheck.Lint] = ListBuffer.empty
```
johnynek: this design is unacceptable. We cannot use a mutable accumulator as a given and pass it through for everyone to mutate.

That is the whole reason for the State monad design here. We can't have two separate styles and mechanisms for keeping track of errors and lints. This is a bad design that isn't in keeping with the style of this repo.

As I mentioned, you should change the current RecursionCheck.Error to be `RecursionCheck.Error | RecursionCheck.Lint` in this code, and then accumulate exactly as we have been in the past. Then access it in inferBodyUnopt to see which errors and lints we have.

proto/src/main/protobuf/bosatsu/TypedAst.proto
```diff
@@ -365,6 +390,7 @@ message Package {
   repeated ExportedName exports = 8;
   repeated Let lets = 9;
   repeated ExternalDef externalDefs = 10;
+  repeated RecursionLint recursionLints = 11;
```
johnynek: i don't like this design. I don't want to serialize lints into the packages. If we are going to serialize lints they need to be into a separate item in the cache, possible keyed on the same keys, but junking up the package definitions with serialized lints I don't like.

core/src/test/scala/dev/bosatsu/TypeValidator.scala
```diff
@@ -1197,9 +1197,9 @@ object TypeValidator {
       unoptimized: Program[
         TypeEnv[Kind.Arg],
         TypedExpr[Declaration],
-        List[Statement]
+        Any
```
johnynek: Any is a very suspect type in scala. When we use it in these highly parameterized types it was intended to mean "we no longer care what this is", not "we can squirrel away anything here and use reflection to get it later". That is not the intent. That is an unacceptable design for this project.

core/src/main/scala/dev/bosatsu/Program.scala
```diff
@@ -14,3 +14,24 @@ case class Program[+T, +D, +S](
 
   def getLet(name: Bindable): Option[(RecursionKind, D)] = letMap.get(name)
 }
+
+object Program {
+  sealed trait Metadata derives CanEqual
+  object Metadata {
+    final case class WithSource[+A](
+        source: A,
+        recursionLints: List[RecursionCheck.Lint]
```
johnynek: I don't understand why RecursionCheck.Lint is special here. We already have postponable errors. Why are we taking such a different approach on this set of errors?

core/src/main/scala/dev/bosatsu/PackageCustoms.scala
```diff
@@ -540,6 +538,11 @@ object PackageCustoms {
       pack: Package.Typed[A]
   ): Option[List[Statement]] =
     pack.program._1.from match {
```
johnynek, line 542: this design is unacceptable to me. We can't match on the `from: Any` and just use reflection to see whats there. We can't see statically what might be there, and I suspect we are violating assumptions about it and creating dead code.

If we are going to use `from`, then it has to have a real type, and we have to be clear that it is populated correctly and branches using it have good coverage. This design is no good.

core/src/main/scala/dev/bosatsu/Package.scala
```diff
@@ -157,9 +157,12 @@ object Package {
...
   def toCompiled[A: HasRegion](pack: Typed[A]): Compiled = {
     val withRegions = typedFunctor.map(pack)(tag => HasRegion.region(tag))
-    // Serialized compiled artifacts never retain the parsed-source payload in
-    // Program.from, so drop it before crossing the cache/compiler boundary.
-    setProgramFrom(withRegions, ())
+    // Compiled artifacts drop the source statements, but keep finalized
+    // recursion-form lint metadata so warn/lax cache hits can replay it.
+    setProgramFrom(
+      withRegions,
+      Program.Metadata.Compiled(recursionLints(withRegions))
```
johnynek: this design is no good. If we want the `Compiled` to have a certain value from `Program.from`, set that as the type. Don't make it `Any`.

Also, I continue to be very skeptical that of all the lints/postponable-errors we have in the code, somehow recursion lints need to be handled totally differently.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Reworked this. `TypedExprRecursionCheck` now uses a single `ValidatedNec[RecursionCheck.Issue, A]` stream, with `RecursionCheck.Issue = RecursionCheck.Error | RecursionCheck.Lint`, and `Package.inferBodyUnopt` does the final partitioning into `PackageError.RecursionError` vs `PackageError.RecursionLint`.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Removed the mutable lint accumulator entirely. The recursion checker is back to one issue stream and one state mechanism. For nested defs, I changed the def-like `let` path to combine the nested def result with the continuation explicitly, so lint-only nested defs no longer rely on side effects to be preserved.

proto/src/main/protobuf/bosatsu/TypedAst.proto
johnynek: Removed the protobuf/package lint serialization path. `TypedAst.proto` no longer carries recursion lint payloads, `ProtoConverter` no longer serializes/deserializes them, and the compile-cache schema bump was reverted. Cache replay now re-runs `Package.inferBodyUnopt` from parsed source plus the deserialized package imports to recover recursion-form lints.

core/src/test/scala/dev/bosatsu/TypeValidator.scala
johnynek: Removed the new `Any`-based metadata usage. `Program.from` is back to concrete `List[Statement]` / `Unit` use here, and the test helpers (`TypeValidator`, `TestUtils`) were restored to those concrete types as well.

core/src/main/scala/dev/bosatsu/Program.scala
johnynek: Addressed by removing the special metadata/proto path. Recursion lints now flow through the same recursion-check result stream as recursion errors and only become postponable at the existing `Package.inferBodyUnopt` boundary, like the other postponable package diagnostics.

core/src/main/scala/dev/bosatsu/PackageCustoms.scala
johnynek: Removed this design entirely. `PackageCustoms.maybeStatements` is back to handling only real `List[Statement]` payloads, and the `Program.from` reflection/match-on-`Any` path is gone.

core/src/main/scala/dev/bosatsu/Package.scala
johnynek: `Package.toCompiled` is back to dropping `Program.from` to `()`, with no `Any` metadata payload and no recursion-lint side channel. The recursion warnings are now replayed by recomputing them during cache replay instead of treating compiled packages as a special serialization container.

johnynek:
Reworked the PR to match the review direction: recursion checking now produces a single validated issue stream (`Error | Lint`), the `Program.from`/`Any` metadata path is removed, and recursion-form lints are no longer serialized into compiled packages or protobufs. Warm-cache replay now re-runs `Package.inferBodyUnopt` from parsed source plus the deserialized package imports to recover recursion lints, while shadowed-binding and totality replay stay on the compiled package path. I also removed the lint-metadata proto/cache tests that only existed for the rejected design. Validation run in this turn: `sbt -batch "coreJVM/test:compile"`, `sbt -batch "coreJVM/testOnly dev.bosatsu.TypedExprRecursionCheckTest dev.bosatsu.ErrorMessageTest dev.bosatsu.ProtoConverterTest dev.bosatsu.ToolAndLibCommandTest dev.bosatsu.tool.CompileCacheTest"`, and `scripts/test_basic.sh`.

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
```diff
@@ -343,12 +343,22 @@ object CompilerApi {
...
+      parsed: Package.Parsed
   ): List[PackageError] = {
+    // Recursion-form warnings depend on source `recur`/`loop` shape, so
+    // cache replay re-runs that typed check from parsed source instead of
+    // serializing finalized lint payloads into compiled packages.
     val recursionLints =
       Package
-        .recursionLints(pack)
-        .map(lint => PackageError.RecursionLint(pack.name, lint): PackageError)
+        .inferBodyUnopt(pack.name, pack.imports, Nil, parsed.program) match {
```
johnynek: if we rerun inference on the package, this destroys the utility of the cache entirely. If we have to rerun inference on every call, we shouldn't also pay the cost of the cache.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -799,21 +799,48 @@ object TypedExprRecursionCheck {
...
         fnname: Bindable,
         body: TypedExpr[Declaration],
-        recur: Declaration.Match
-    ): St[Unit] =
-      recur.kind match {
-        case Declaration.MatchKind.Loop =>
-          SelfCallKind(fnname, body) match {
-            case SelfCallKind.NonTailCall =>
-              failSt(RecursionCheck.LoopRequiresTailRecursion(fnname, recur.region))
+        finalState: State
```
johnynek: make: `finalState: InDef | InDefRecurred` if that's what we require, and push through the requirement to the callers, and maybe we can show that we only call finish when we are inside `InDef | InDefRecurred`.

Use unions of different states to keep track of the state to make our types more precise.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -3188,9 +3215,19 @@ object TypedExprRecursionCheck {
...
-              val nextRes = checkExpr(currentPackage, in, wrappers)
-              defn *> nextRes
+              val checkedDef =
+                checkDef(currentPackage, state, arg, ex, fromSource)
+              val checkedIn =
+                Validated.fromEither(
+                  checkExpr(currentPackage, in, wrappers).run(state).value
+                ).map(_._1)
+
+              (checkedDef, checkedIn).mapN((_, nextState) => nextState) match {
```
johnynek: `(checkedDef, checkedIn).mapN((_, nextState) => nextState)` is the same as `checkedDef *> checkedIn`. Use that.

core/src/main/scala/dev/bosatsu/tool/CompilerApi.scala
johnynek: Changed direction here. Cache-hit replay no longer re-runs inference. I took the match-annotation path: `TypedExpr.Match` now carries the source form (`match`/`recur`/`loop`), that survives proto round-tripping, and `CompilerApi` replays recursion-form lints directly from compiled typed lets using a replay analysis that understands lowered `Loop`/`Recur` nodes.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Made this precise. `successfulDefResult` now takes `FinalDefState = InDef | InDefRecurred`, and the callers narrow through `finalDefState(...)` before calling it, so we no longer accept arbitrary `State` there.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Updated that site to use `checkedDef *> checkedIn`.

johnynek:
I took option 1 rather than dropping cache-hit replay. The basic problem with option 2 is that strict/warn/lax behavior would diverge depending on cache state, and the design doc acceptance criteria call out stable warning behavior after cache hits. The implemented path is: preserve source recursion form on `TypedExpr.Match`, serialize that through proto, and replay recursion-form lints from compiled lets without re-inference. The replay logic now uses compiled-AST facts that survive lowering: it reads the stored match kind and classifies recursion with a replay-specific self-call analysis that treats outer lowered `Recur` nodes as tail self-calls while ignoring nested loops.

I also merged `origin/main`, resolved the conflicts in the PR area, updated the command tests for the current CLI surface, and cleaned up the remaining strict-workspace tail-recursive `recur` in `test_workspace/List.bosatsu` by converting it to a local `loop` helper form that the current recursion checker accepts. Validation in this turn: `sbt -batch "coreJVM/testOnly dev.bosatsu.ToolAndLibCommandTest dev.bosatsu.TypedExprRecursionCheckTest dev.bosatsu.ProtoConverterTest dev.bosatsu.tool.CompileCacheTest"` and `scripts/test_basic.sh`, both passed.

core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -967,12 +968,36 @@ object TypedExpr {
     )(using patternRegion: Region = this.patternRegion): Branch[T1] =
       Branch(pattern, guard, expr)(using patternRegion)
   }
-  case class Match[+T](
+  final case class MatchExpr[+T](
       arg: TypedExpr[T],
       branches: NonEmptyList[Branch[T]],
-      tag: T
+      tag: T,
+      matchKind: Declaration.MatchKind = Declaration.MatchKind.Match
```
johnynek: move this as the first argument, not the last argument. Also import `Declaration.MatchKind` in the file so we can just write `MatchKind`.

core/src/main/scala/dev/bosatsu/rankn/Infer.scala
```diff
@@ -2500,6 +2500,12 @@ object Infer {
           }
 
         case Match(term, branches, tag) =>
+          val matchKind: _root_.dev.bosatsu.Declaration.MatchKind =
```
johnynek: don't reach into the tag. Add MatchKind to `Expr` as well so we have it directly and don't need to search through the tag.

Also, don't do local full names, `_root_.dev.bosatsu.Declaration.MatchKind`. Import it at the top of the file.

core/src/main/scala/dev/bosatsu/tool/ShowEdn.scala
```diff
@@ -827,7 +828,13 @@ object ShowEdn {
               err[TypedExpr.Branch[Unit]](s"invalid branch: ${rendered(other)}")
           }
           nel <- NonEmptyList.fromList(branches).toRight("match branches cannot be empty")
-        } yield TypedExpr.Match(arg, nel, ())
+          matchKind <- matchKeyword match {
+            case "match" => Right(_root_.dev.bosatsu.Declaration.MatchKind.Match)
```
johnynek: import MatchKind. Don't use the full name at each use.

core/src/main/scala/dev/bosatsu/TypedExpr.scala
```diff
@@ -538,10 +538,11 @@ object TypedExpr {
             Eq[Lit].eqv(llit, rlit) &&
             eqType(lt, rt) &&
             eqA.eqv(ltag, rtag)
-          case (Match(larg, lbranches, ltag), Match(rarg, rbranches, rtag)) =>
+          case (lm @ Match(larg, lbranches, ltag), rm @ Match(rarg, rbranches, rtag)) =>
             loop(larg, rarg) &&
             eqBranches(lbranches, rbranches) &&
-            eqA.eqv(ltag, rtag)
+            eqA.eqv(ltag, rtag) &&
+            (lm.matchKind == rm.matchKind)
```
johnynek: compare matchKind first so if it fails we don't need to recurse.

core/src/main/scala/dev/bosatsu/TypedExpr.scala
johnynek: Updated `TypedExpr.MatchExpr` so `matchKind` is the first field, imported `Declaration.MatchKind` in the file, and rewired the `TypedExpr.Match`/`TypedExpr.Rho.Match` helpers plus the typed-AST rebuild sites to use that order.

core/src/main/scala/dev/bosatsu/rankn/Infer.scala
johnynek: Changed this to carry match kind directly on `Expr.Match` as well. `Infer` now reads `m.matchKind` instead of peeling it back out of the declaration tag, and the local fully-qualified `Declaration.MatchKind` references in that path were replaced with a top-level import.

core/src/main/scala/dev/bosatsu/tool/ShowEdn.scala
johnynek: Cleaned that up. `ShowEdn` now imports `MatchKind` and uses `MatchKind.Match` / `Recur` / `Loop` directly when decoding typed matches.

core/src/main/scala/dev/bosatsu/TypedExpr.scala
johnynek: Reordered the equality check so `matchKind` is compared before recursing into the match scrutinee and branches.

johnynek:
Applied the API cleanup pass from the review: `TypedExpr.Match` now takes `matchKind` first, `Expr.Match` also carries `matchKind` directly, infer no longer searches declaration tags for recursion form, and the typed/source AST rebuild sites were updated to preserve the field in the new order. I also kept the 3-argument `Match` patterns/constructors working through the companion/unapply path so the broader tree traversals stay readable.

Validation in this turn: `sbt -batch "coreJVM/test:compile"`, `sbt -batch "coreJVM/testOnly dev.bosatsu.ProtoConverterTest dev.bosatsu.TypedExprTest dev.bosatsu.ExprTest dev.bosatsu.SourceConverterTest dev.bosatsu.rankn.RankNInferTest"`, and the required `scripts/test_basic.sh` all passed.

core/src/main/scala/dev/bosatsu/Expr.scala
```diff
@@ -237,12 +239,34 @@ object Expr {
       guard: Option[Expr[T]],
       expr: Expr[T]
   )(using val patternRegion: Region)
-  case class Match[T](
+  final case class MatchExpr[T](
+      matchKind: MatchKind = MatchKind.Match,
```
johnynek: remove the default argument. Be explicit on all calls or make a separate apply argument.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -799,25 +805,363 @@ object TypedExprRecursionCheck {
...
+                    }
+                  acc.merge(branchCalls)
+              }
+            }
+        }
+
+      loop(expr, 0)
+    }
+
+    private def replayNestedBindingLints(
```
johnynek: why do we need a totally separate implementation of this?

Why can't we have a single call path. For instance, change checkExpr to something like:
```
    private def checkExpr[A: HasRegion](
        currentPackage: PackageName,
        expr: TypedExpr[A],
        wrappers: WrapperScope
    ): St[Unit] =
```
maybe with an additional function/trait passed to deal with the uses or `tag` other than getting the region. Then we can implement that one way given the original `TypedExpr[Declaration]` and another way given the `TypedExpr[Region]` that was deserialized from the disk.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -799,25 +805,363 @@ object TypedExprRecursionCheck {
...
+        currentPackage: PackageName,
+        lets: List[(Bindable, RecursionKind, TypedExpr[Region])]
+    ): List[RecursionCheck.Lint] =
+      lets.flatMap { case (name, rec, expr) =>
+        if (isReplayDefLike(rec, expr))
+          replayBindingLints(currentPackage, name, expr)
+        else replayNestedBindingLints(currentPackage, expr)
+      }
+
+    private def replayBindingLints(
```
johnynek: there is too much duplicaiton in this replay path. We need to unify to have a single implementation.

core/src/main/scala/dev/bosatsu/Expr.scala
johnynek: Removed the default argument from `Expr.MatchExpr`. The default `match` case now only comes from the separate 3-argument `Expr.Match(...)` apply, so call sites are explicit unless they intentionally use the plain-match helper.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Reworked this so replay is no longer a separate lint walker. I introduced a small `LintSyntax[A]` strategy and shared `bindingOwnLints` / `bindingAndNestedLints` / `nestedBindingLints` / `findRecursiveHeader` helpers that work for both `TypedExpr[Declaration]` and cached `TypedExpr[Region]`. The source path uses `SourceLintSyntax`, the cached replay path uses `ReplayLintSyntax`, and the only differences are the header/self-call/def-like hooks.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Addressed by collapsing the replay duplication into that shared generic lint collector. The cached replay path now goes through the same binding/header/called-name traversal structure as the source lint path instead of maintaining separate replay-specific binding/header walkers.

johnynek:
I removed the `Expr.Match` defaulted `matchKind` field and unified recursion-form lint collection around a single generic traversal. Source checking and cached replay now share the same lint-collection path via `LintSyntax[A]`; only the strategy pieces differ (`SourceLintSyntax` vs `ReplayLintSyntax`). I also wired the source-side `successfulDefResult` through the shared own-binding lint collector so the lint classification itself is no longer split between source and replay.

Validation in this turn: `sbt -batch "coreJVM/test:compile"`, `sbt -batch "coreJVM/testOnly dev.bosatsu.TypedExprRecursionCheckTest dev.bosatsu.ToolAndLibCommandTest dev.bosatsu.ProtoConverterTest dev.bosatsu.ExprTest dev.bosatsu.SourceConverterTest"`, and `scripts/test_basic.sh` all passed.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -799,25 +805,429 @@ object TypedExprRecursionCheck {
...
+          currentPackage: PackageName,
+          arg: TypedExpr[Region],
+          m: TypedExpr.MatchExpr[Region]
+      ): Option[LintHeader[Region]] =
+        if (m.matchKind.isRecursive)
+          Some(LintHeader(m.matchKind, m.tag, m.branches))
+        else None
+    }
+
+    private def mergeCallCounts(
```
johnynek: this is `cats.Monoid.combine(left, right)` no need to reimplement.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -799,25 +805,429 @@ object TypedExprRecursionCheck {
...
+                ) :: Nil
+            case SelfCallKind.TailCall
+                if header.kind == Declaration.MatchKind.Recur =>
+              RecursionCheck.TailRecursiveRecur(fnname, header.region) :: Nil
+            case _ =>
+              Nil
+          }
+        }
+
+      ownLint
```
johnynek: don't capture a value just to return it. Don't do `val x = <expr> ; x` just make `<expr>` the final expression.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -799,25 +805,429 @@ object TypedExprRecursionCheck {
...
+    private def bindingAndNestedLints[A](
+        syntax: LintSyntax[A],
+        currentPackage: PackageName,
+        fnname: Bindable,
+        expr: TypedExpr[A]
+    ): List[RecursionCheck.Lint] =
+      bindingOwnLints(syntax, currentPackage, fnname, expr) :::
+        nestedBindingLints(syntax, currentPackage, expr)
+
+    private def replaySelfCallKind(
```
johnynek: isn't this a reimplementation of `SelfCallKind(fnname, expr)`? What is the difference? Why do we need this here? Add a comment to make it very clear if we need this duplication. Ideally we can remove it.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Replaced the hand-rolled map merge with Cats combine operations. `calledNamesInBranches` / `calledNamesInExpr` now use `|+|` directly and the local `mergeCallCounts` helper is gone.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Cleaned this up by removing the extra `ownLint` temporary in `bindingOwnLints`; the flatMap is now returned directly.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Removed the duplicate implementation from `TypedExprRecursionCheck`. The replay-specific behavior now lives in `SelfCallKind.afterLoopRecurLowering`, which shares the same traversal as `SelfCallKind.apply` and only switches the `Loop`/`Recur` interpretation needed after loop/recur lowering. I added a comment there explaining why cached replay needs that variant.

johnynek:
Folded the cached replay self-call logic into `SelfCallKind` so the recursion checker no longer carries a separate walker, and simplified the lint helpers per review. Validation in this turn: `sbt -batch "coreJVM/test:compile"`, `sbt -batch "coreJVM/testOnly dev.bosatsu.SelfCallKindTest dev.bosatsu.TypedExprRecursionCheckTest dev.bosatsu.ToolAndLibCommandTest dev.bosatsu.ProtoConverterTest"`, `git diff --check`, and `scripts/test_basic.sh` all passed.

johnynek:
Merge origin main and make sure the tests still pass.

johnynek:
Merged `origin/main` as requested; current HEAD is `06670591c` (`Merge remote-tracking branch 'origin/main' into agent/impl/2227-add-a-lint-for-using-recur-instead-of-loop`). The worktree is clean after the merge, and the required `scripts/test_basic.sh` passed on the merged branch state. No additional file edits or follow-up commit were needed in this turn beyond the merge commit already created by the git op.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
```diff
@@ -799,25 +805,324 @@ object TypedExprRecursionCheck {
           None
       }
 
-    private def checkLoopTailRecursion(
+    private type FinalDefState = InDef | InDefRecurred
+
+    private def finalDefState(state: State): Res[FinalDefState] =
```
johnynek: there is no reason for this to return `Res`. Just return `FinalDefState`

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: this code calls `val st = setSt(state1) *> expr` and then `st.run(state)`. Isn't that the same as:

`expr.run(state1)`?

And doesn't that give an explanation as to why the result state is a final state? Because `state1` is a `state.inDef` returns `InDef` and once we start with `InDef` we always transition to `InDef` or `InDefRecurred`.

Is that right?

If so, let's make the change to simplify and avoid setSt. Maybe we can remove setSt entirely? Then we can add the explanation as to why it is FinalDefState

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Changed this to return `FinalDefState` directly. The helper is now just the narrowing match over `InDef | InDefRecurred`, with the impossible cases left as the existing internal error path.

core/src/main/scala/dev/bosatsu/TypedExprRecursionCheck.scala
johnynek: Made that simplification. The def body now runs directly as `checkExpr(...).run(state1)`, and I added a comment at the call site explaining why a successful run from `state1: InDef` must finish in `InDef` or `InDefRecurred`, which is exactly what `finalDefState` narrows to.

johnynek:
Simplified the def-state flow in `TypedExprRecursionCheck.scala` per review: `finalDefState` no longer returns a validated wrapper, and the def checker now runs the body directly from `state1` instead of reseeding through `setSt`. Validation in this turn: `sbt -batch "coreJVM/testOnly dev.bosatsu.TypedExprRecursionCheckTest dev.bosatsu.SelfCallKindTest"`, `git diff --check`, and the required `scripts/test_basic.sh` all passed.

Issue #2239, opened 2026-03-23
Type aliases


johnynek:
How could we implement type aliases in Bosatsu?

I imagine syntax like

Foo = Bar[Int]

Baz[a: +*] = List[a]

There are a few issues:

1. Are they transparent? When exported can everyone see the left and right are the same type?
2. How are they checked? I guess we need to ensure the kind is legal, and the variances match.
3. If variance isn't annotated, can we infer variance using the existing variance inference code?
4. How does it change type inference? If we have things like: Baz[a] = Quux[a, Int] then we can't represent unification of Baz without a type lambda [a] => Quux[a, Int]. Do we have to add type lambdas at the same time? If so does that make it easier to implement since all type aliases can be just direct fully applied types so in this case Baz = [a] => Quux[a, Int].

We need to explicit about all these especially how type aliases work with Infer.scala.

johnynek:
MergeXO assigned an agent and started design work for issue #2239.

johnynek:
MergeXO assigned an agent and started design work for issue #2239.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/2241

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #2239.

johnynek:
MergeXO pre-PR flow is waiting for source-issue follow-up.
- blocked reason: implementation flow blocked: codex invocation timed out after local changes were made: working tree has local changes
- checkpoint branch: `agent/impl/2239-type-aliases`
- checkpoint commit: `47bb1ee7407cde24c5b43e49142151896324bb03`
- checkpoint tree: https://github.com/johnynek/bosatsu/tree/47bb1ee7407cde24c5b43e49142151896324bb03
- compare vs `main`: https://github.com/johnynek/bosatsu/compare/main...47bb1ee7407cde24c5b43e49142151896324bb03

Next steps:
1. Review the checkpoint tree/compare links above.
2. Reply on this issue with clarifications or updated instructions.
3. MergeXO will resume from this checkpoint branch head.

johnynek:
please retry opening the PR. I think the diff is worth a PR.

johnynek:
MergeXO assigned an agent and started implementation PR work for issue #2239.

johnynek:
We eventually used `type Foo = Bar` due to ambiguities with parsing that came up. This matches an allowed syntax in Python 3.12 which also has a `type` statement like this.

PR #2241, opened 2026-03-24
Design doc for #2239: Add Transparent Type Aliases

johnynek:
Design doc.

Refs #2239

### DIFF
--- a/docs/design/2239-type-aliases.md
+++ b/docs/design/2239-type-aliases.md
@@ -61,7 +61,8 @@ That means aliases are not just a parser change. They affect name resolution, ki
 2. No value-level aliasing.
 3. No new constructor or value namespace entry for an alias name.
 4. No change to runtime data representation.
-5. No requirement that every user-facing rendered type preserve alias spelling if internal canonicalization prefers expanded types.
+5. No local type aliases inside `Declaration`; aliases are top-level statements only.
+6. No requirement that every user-facing rendered type preserve alias spelling if internal canonicalization prefers expanded types.
 
 ## Proposed Design
 ### 1. Add a dedicated top-level alias statement
@@ -72,6 +73,7 @@ Parsing rules:
 2. The right-hand side is parsed with `TypeRef.parser`, not `Declaration.parser`.
 3. The alias parser must run before the generic binding parser, because `Foo = ...` currently falls through to pattern binding.
 4. This intentionally reserves top-level uppercase `Foo = ...` for aliases rather than zero-argument constructor-pattern bindings. That is an acceptable compatibility tradeoff because the pattern form has an unambiguous `match` rewrite.
+5. Aliases are only added to `Statement`; this issue does not add a `Declaration`-level type binding form.
 
 Printing rules:
 1. Aliases render back to the same surface form.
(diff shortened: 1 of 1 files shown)


docs/design/2239-type-aliases.md
```diff
@@ -0,0 +1,249 @@
...
+A fully saturated use such as `Baz[String]` can be expanded by substituting `String` into the alias body. The difficult case is comparing the bare higher-kinded alias `Baz` itself against other higher-kinded types.
+
+The recommended design is:
+1. Add an alias table to `Infer.Env`.
+2. Add a helper that expands any saturated alias spine:
+   - collect `(head, args)` with `Type.unapplyAll`
+   - if `head` is an alias and it has received enough arguments for its declared arity, substitute those arguments into the alias body
+   - rebuild with any remaining arguments
+   - repeat until the head is no longer an expandable alias
+3. When `Infer` must compare two types whose kind is not `Type`, eta-expand both sides with fresh rigid arguments until the comparison reaches kind `Type`, then run the existing subsumption and unification logic on the instantiated result types.
```
johnynek: we need to be clearer here. If we eta expand a kind that is not Type, that makes sense, but the current algorithm would still recurse into the `TyApply` nodes and reach types that don't have kind `Type`, so how do we make progress.

For instance, if we are trying to see if `Monad[Baz]` type checks, how can we do that? If `struct Monad[f: * -> *](pure: forall a. a -> f[a], flat_map: forall a, b. (f[a], a -> f[b]) -> f[b])` then we want to be able to typecheck `m: Monad[Baz] = Monad(...` where we have `Quux[a, Int]` values inside the construction.

Be explicit. How will this type check?

docs/design/2239-type-aliases.md
```diff
@@ -0,0 +1,249 @@
...
+2. Aliases contribute constraints directly from their `rhs` type expression.
+3. Alias solving should reuse the existing arbitrary-type walkers already present in `Shape.shapeOfType` and `KindFormula.addTypeConstraints`-style logic, rather than pretending aliases are constructors.
+
+Variance behavior:
+1. Yes, variance can be inferred for aliases.
+2. The same variance lattice already used by `KindFormula` is sufficient.
+3. An alias parameter is covariant, contravariant, invariant, or phantom based on how it appears in the solved `rhs` kind formula.
+4. Explicit variance annotations remain checked, not trusted blindly.
+
+### 6. Reject recursive alias cycles explicitly
```
johnynek: bosatsu requires files to be topologically sorted. I think an alias can only reference prior types (not including itself) in the file. End of story. No recursive type definitions of any kind.

docs/design/2239-type-aliases.md
```diff
@@ -0,0 +1,249 @@
...
+
+After inference succeeds:
+1. Expand saturated aliases through typed expression types and the internal `TypeEnv`.
+2. Keep alias declarations in the package and interface environment for import/export and library serialization.
+3. Run totality and similar runtime-facing passes on the normalized view.
+
+This keeps alias support localized to the front end, type environment, and interface metadata instead of forcing every downstream pass to understand transparent synonyms.
+
+## Implementation Plan
+1. Add `Statement.TypeAlias` plus parser and document support in `core/src/main/scala/dev/bosatsu/Statement.scala`.
```
johnynek, line 186: are local type aliases allowed in Declaration? Basically a TypeBinding? Or to all type aliases have to be at the top level. All types have to be, so there is some argument for that.

docs/design/2239-type-aliases.md
johnynek: Section 7 is now explicit about the `Infer.scala` change. I replaced the vague eta-expansion description with an application-spine normalization step in `unifyTau`/`unifyRho`, and added a concrete `Monad[Baz]` walkthrough: instantiate `Monad` with `f := Baz`, normalize each saturated `Baz[a]`/`Baz[b]` to `Quux[a, Int]`/`Quux[b, Int]`, then typecheck `pure` and `flat_map` against those normalized field types.

docs/design/2239-type-aliases.md
johnynek: Updated section 6 to match Bosatsu's existing source-order rule. The doc now says aliases may reference imports and prior local type definitions only, with no recursive or forward type definitions, and I removed the separate alias-cycle/toposort proposal.

docs/design/2239-type-aliases.md
johnynek: Clarified that aliases are top-level only. The doc now states this in Non-goals, in the parser/design section, and in Acceptance Criteria, and explicitly says this issue does not add a `Declaration`-level type binding form.

johnynek:
Updated `docs/design/2239-type-aliases.md` to address the review feedback: aliases are now specified as top-level-only and source-order-only, and the higher-kinded inference section now explains the `Monad[Baz]` case concretely in terms of application-spine normalization before `TyApply` recursion. `scripts/test_basic.sh` passes locally.

PR #2242, opened 2026-03-24
Fix type-only package import usage tracking

### LINKED ISSUE #2240: Type-only packages should work
I've seen reports that you can't create type-only packages, because the import/export checker doesn't consider items used only in the types.

We should have tests that show creating a package with some types, then another package uses those types in another type-only package.

both value only and types only packages should be legal.

johnynek:
Resolved issue #2240 by teaching `PackageCustoms.allImportsAreUsed` to count imported types referenced from local type declarations, not just lets and externals. This makes type-only packages legal when they import types used only in exported/local structs. Added a `PackageTest` regression that builds a type-only provider, a type-only consumer, and a value-only downstream package, and asserts the resulting interface dependency visibility. Verified with `sbt "coreJVM/testOnly dev.bosatsu.PackageTest"` and `scripts/test_basic.sh`.

Fixes #2240

### DIFF
--- a/core/src/main/scala/dev/bosatsu/PackageCustoms.scala
+++ b/core/src/main/scala/dev/bosatsu/PackageCustoms.scala
@@ -314,15 +314,13 @@ object PackageCustoms {
       }
 
     val localDefinedTypes =
-      pack.program._1.types.allDefinedTypes.iterator
-        .collect { case dt if dt.packageName == pack.name => dt }
-        .flatMap { dt =>
-          dt.constructors.iterator.flatMap { constructor =>
-            constructor.args.iterator.flatMap { param =>
-              Type.constantsOf(param.tpe)
-            }
-          }
-        }
+      for {
+        dt <- pack.program._1.types.allDefinedTypes.iterator
+        if dt.packageName == pack.name
+        constructor <- dt.constructors.iterator
+        param <- constructor.args.iterator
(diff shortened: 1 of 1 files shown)


core/src/main/scala/dev/bosatsu/PackageCustoms.scala
```diff
@@ -298,6 +298,35 @@ object PackageCustoms {
...
+    val externalTypes =
+      pack.program._1.externalDefs.iterator.flatMap { nm =>
+        pack.program._1.types
+          .getExternalValue(pack.name, nm)
+          .iterator
+          .flatMap(Type.constantsOf(_))
+      }
+
+    val localDefinedTypes =
+      pack.program._1.types.allDefinedTypes.iterator
```
johnynek: Use a for comprehension here. Don't nest so deep.

johnynek: Updated `usedTypeConsts` in `PackageCustoms.scala` to use a `for`-comprehension for the local defined-type traversal, with no behavior change. Re-ran `scripts/test_basic.sh` after the change.

johnynek:
Applied the requested readability cleanup in `PackageCustoms.scala` and reran `scripts/test_basic.sh` successfully.

PR #2247, opened 2026-03-25
Implement type aliases

### LINKED ISSUE #2239: Type aliases

johnynek:
close #2239

### DIFF
--- a/core/src/main/scala/dev/bosatsu/rankn/Infer.scala
+++ b/core/src/main/scala/dev/bosatsu/rankn/Infer.scala
@@ -165,47 +165,52 @@ object Infer {
 
     @annotation.tailrec
     final def normalizeAliasHead(t: Type): Type =
-      Type.unapplyAll(t) match {
-        case (
-              Type.TyConst(tc @ Type.Const.Defined(_, _)),
-              args
-            ) =>
-          aliases.get(tc) match {
-            case Some(alias) =>
-              alias.expandWith(args) match {
-                case Some(t1) if t1 != t =>
-                  normalizeAliasHead(t1)
-                case _ =>
-                  t
-              }
-            case None =>
-              t
-          }
-        case _ =>
-          t
-      }
+      if (aliases.isEmpty) t
+      else
+        Type.unapplyAll(t) match {
+          case (
+                Type.TyConst(tc @ Type.Const.Defined(_, _)),
+                args
+              ) =>
+            aliases.get(tc) match {
+              case Some(alias) =>
+                alias.expandWith(args) match {
+                  case Some(t1) if t1 != t =>
+                    normalizeAliasHead(t1)
+                  case _ =>
+                    t
+                }
+              case None =>
+                t
+            }
+          case _ =>
+            t
+        }
 
     final def normalizeAliasesDeep(t: Type): Type = {
-      val head = normalizeAliasHead(t)
-      if (head != t) normalizeAliasesDeep(head)
-      else
-        t match {
-          case t0 @ Type.ForAll(vars, in) =>
-            val in1 = normalizeAliasesDeep(in).asInstanceOf[Type.Rho]
-            if (in1 eq in) t0 else Type.ForAll(vars, in1)
-          case t0 @ Type.Exists(vars, in) =>
-            val in1 =
-              normalizeAliasesDeep(in).asInstanceOf[Type.Leaf | Type.TyApply]
-            if (in1 eq in) t0 else Type.Exists(vars, in1)
-          case t0 @ Type.TyApply(on, arg) =>
-            val on1 =
-              normalizeAliasesDeep(on).asInstanceOf[Type.Leaf | Type.TyApply]
-            val arg1 = normalizeAliasesDeep(arg)
-            if ((on1 eq on) && (arg1 eq arg)) t0
-            else Type.TyApply(on1, arg1)
-          case other =>
-            other
-        }
+      if (aliases.isEmpty) t
+      else {
+        val head = normalizeAliasHead(t)
+        if (head ne t) normalizeAliasesDeep(head)
+        else
+          t match {
+            case t0 @ Type.ForAll(vars, in) =>
+              val in1 = normalizeAliasesDeep(in).asInstanceOf[Type.Rho]
+              if (in1 eq in) t0 else Type.ForAll(vars, in1)
+            case t0 @ Type.Exists(vars, in) =>
+              val in1 =
+                normalizeAliasesDeep(in).asInstanceOf[Type.Leaf | Type.TyApply]
+              if (in1 eq in) t0 else Type.Exists(vars, in1)
+            case t0 @ Type.TyApply(on, arg) =>
+              val on1 =
+                normalizeAliasesDeep(on).asInstanceOf[Type.Leaf | Type.TyApply]
+              val arg1 = normalizeAliasesDeep(arg)
+              if ((on1 eq on) && (arg1 eq arg)) t0
+              else Type.TyApply(on1, arg1)
+            case other =>
+              other
+          }
+      }
     }
 
     private val kindCache: Type => Either[Region => Error, Kind] =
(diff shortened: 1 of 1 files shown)


core/src/main/scala/dev/bosatsu/rankn/Infer.scala
```diff
@@ -572,6 +604,127 @@ object Infer {
       GetEnv.map(env => tpe => env.getKind(tpe, emptyRegion).toOption)
     }
 
+    inline def assertTau(
+        t: Type,
+        inline context: => String,
+        inline region: Region
+    ): Infer[Type.Tau] =
+      if (Type.Tau.isTau(t)) pure(t.asInstanceOf[Type.Tau])
```
johnynek: this should be:

```scala
t match {
  case Tau(tau) => pure(tau)
  case _ => fail(Error.ExpectedRho(t, context, region))
}
```

core/src/main/scala/dev/bosatsu/rankn/Infer.scala
```diff
@@ -572,6 +604,127 @@ object Infer {
...
+        t: Type.Rho,
+        region: Region,
+        context: String
+    ): Infer[Type.Rho] =
+      normalizeAliasType(t).flatMap(assertRho(_, context, region))
+
+    def normalizeAliasTau(
+        t: Type.Tau,
+        region: Region,
+        context: String
```
johnynek: this should be `inline context: => String` or there is no point in making assertTau take an inline context.

core/src/main/scala/dev/bosatsu/rankn/Infer.scala
```diff
@@ -572,6 +604,127 @@ object Infer {
...
+      if (Type.Tau.isTau(t)) pure(t.asInstanceOf[Type.Tau])
+      else fail(Error.ExpectedRho(t, context, region))
+
+    def normalizeAliasType(t: Type): Infer[Type] =
+      GetEnv.map(_.normalizeAliasHead(t))
+
+    def normalizeAliasRho(
+        t: Type.Rho,
+        region: Region,
+        context: String
```
johnynek: this should be `inline context: => String` or there is no point in making assertRho take an inline context.

core/src/main/scala/dev/bosatsu/rankn/Infer.scala
```diff
@@ -572,6 +604,127 @@ object Infer {
...
+        k2: Kind,
+        left: Region,
+        right: Region,
+        direction: Error.Direction
+    ): Infer[TypedExpr.CoerceRho] =
+      if (!Kind.shapeMatch(k1, k2))
+        fail(Error.KindMismatch(t, k1, rho, k2, left, right))
+      else
+        for {
+          et1 <- etaExpandToType(t, k1)
```
johnynek: I'm confused how we seem to be unconditionally etaExpandingToType and then recursing. I don't see what prevents an infinite loop here.

Won't we add new variables, then get back down to the kinds that aren't type, then add more variables, and loop forever?

core/src/main/scala/dev/bosatsu/KindFormula.scala
```diff
@@ -31,6 +32,12 @@ sealed abstract class KindFormula derives CanEqual
 object KindFormula {
   given cats.Eq[KindFormula] = cats.Eq.fromUniversalEquals
 
+  sealed trait Source
```
johnynek: we should use
```scala
enum Source derives CanEqual {
  case CounstructorFn(cfn: ConstructorFn[?])
  case AliasBody
}
```

core/src/main/scala/dev/bosatsu/Shape.scala
```diff
@@ -45,6 +53,12 @@ object Shape {
       ref: Ref[UnknownState]
   ) extends NotKnownShape
 
+  sealed trait Source derives CanEqual
```
johnynek: this is duplicated in Kind inference. We should use the enum syntax here and share. Since Kind inference depends on Shape, I think, it should use this version.

core/src/main/scala/dev/bosatsu/SourceConverter.scala
```diff
@@ -1176,16 +1167,74 @@ final class SourceConverter(
...
+      case Type.Const.Predef(cons)
+          if legacyDefaultHashPredefTypes(const) &&
+            !localTypeNames(cons) &&
+            !importedTypes.contains(cons) =>
+        Type.Const.Defined(thisPackage, TypeName(cons))
+      case _ =>
+        const
+    }
+
+  private def legacyDefaultHashType(tpe: Type): Type = {
```
johnynek: let's remove this. Let's break the hashes if we had a true bug. This legacy behavior isn't safe and shouldn't have been there.

I don't think there are a lot of examples of default values anyway.

core/src/main/scala/dev/bosatsu/rankn/Infer.scala
```diff
@@ -145,13 +147,65 @@ object Infer {
...
+
+    @annotation.tailrec
+    final def normalizeAliasHead(t: Type): Type = {
+      val next =
+        Type.unapplyAll(t) match {
+          case (
+                Type.TyConst(tc @ Type.Const.Defined(_, _)),
+                args
+              ) =>
+            aliases.get(tc).flatMap(_.expandWith(args))
```
johnynek: use a pattern match here to avoid an allocation. We care about the performance of this code and inference is already the most expensive part of the code

core/src/main/scala/dev/bosatsu/rankn/Infer.scala
```diff
@@ -841,10 +913,15 @@ object Infer {
         _ <- a2s.zip(a1s).parTraverse { case (a2, a1) =>
           subsCheck(a2, a1, right, left, direction.flip)
         }
+        outArgs <- a1s.zip(a2s).traverse { case (a1, a2) =>
+          aliasEquivalent(a1, a2).map { same =>
+            if (same) a2 else a1
```
johnynek: add a comment to the motivation here. Why if they are not aliasEquivalent do we choose a1? Why not a1 unconditionally? We need to explain.

core/src/main/scala/dev/bosatsu/rankn/ParsedTypeEnv.scala
```diff
@@ -20,7 +31,22 @@ case class ParsedTypeEnv[+A](
...
+  sealed abstract class TypeStatement[+A] {
+    def toTypeConst: Type.Const.Defined
+  }
+  object TypeStatement {
+    final case class Defined[+A](definedType: DefinedType[A])
+        extends TypeStatement[A] {
+      def toTypeConst: Type.Const.Defined = definedType.toTypeConst
+    }
+
+    final case class Alias[+A](typeAlias: dev.bosatsu.rankn.TypeAlias[A])
```
johnynek: import `TypeAlias` don't fully address it here.

core/src/main/scala/dev/bosatsu/rankn/ParsedTypeEnv.scala
```diff
@@ -20,7 +31,22 @@ case class ParsedTypeEnv[+A](
 }
 
 object ParsedTypeEnv {
-  val Empty: ParsedTypeEnv[Nothing] = ParsedTypeEnv(Nil, Nil)
+  sealed abstract class TypeStatement[+A] {
```
johnynek: use enum syntax:
```scala
enum TypedStatement[+A] {
  case Defined(...)
  case Alias(...)

  def toTypedConst: Type.Const.Defined = this match {
    ...
  }
}
```

core/src/test/scala/dev/bosatsu/PackageTest.scala
```diff
@@ -469,4 +480,57 @@ main = 1
...
+  test("exported type aliases can be imported transparently") {
+    val exported = parse("""
+package Alias/Export
+export Foo, mkFoo
+
+struct Box(item)
+
+type Foo = Box[Int]
+
+mkFoo: Foo = Box(1)
```
johnynek, line 493: This is a weird example. Here `Box` isn't exported, but `Foo` is. Does that imply that `Foo` gets exported or does `Foo` become basically an opaque type at that point?

This is very confusing. I don't think we have specified this well.

PR #2251, opened 2026-03-26
Implement type aliases with native-image timeout overrides

### LINKED ISSUE #2239: Type aliases

johnynek:
close #2239

Adds the native-image compiler timeout overrides:
- -H:CompilationExpirationPeriod=0
- -H:CompilationNoProgressPeriod=0

This keeps the type alias changes and makes the release native-image build less likely to fail on long method compilation.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -356,7 +356,9 @@ lazy val cli = (project in file("cli"))
         "-O2",
         s"-J-Xmx$nativeImageBuilderXmx",
         "-H:IncludeResources=dev/bosatsu/scalawasiz3/aot/.*\\.meta",
-        "-H:+RemoveUnusedSymbols"
+        "-H:+RemoveUnusedSymbols",
+        "-H:CompilationExpirationPeriod=0",
+        "-H:CompilationNoProgressPeriod=0"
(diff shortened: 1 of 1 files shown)


johnynek:
Closing this duplicate. The intended standalone timeout-only PR is #2252.

johnynek:
dup again of #2247

PR #2256, opened 2026-03-26
Design doc for #2254: Design required `exposes` declarations for exported dependency visibility

johnynek:
Design doc.

Refs #2254

PR #2257, opened 2026-03-26
Require checked exposes declarations for exported dependency visibility

### LINKED ISSUE #2254: Design required `exposes` declarations for exported dependency visibility
## Summary

Bosatsu currently aims for typechecked source to make a package's important interface properties visible to a human reader. We can already see package/import/export structure directly in the file, and after typechecking we can tell which values and types escape via `export`. What we still cannot see directly in source is which package dependencies are exposed by the exported API.

That matters for libraries with public vs private dependencies: if a package from a private dependency appears in an exported API, that dependency has effectively leaked into the public surface. The compiler already computes this information and library assembly already checks some of it, but the source language does not currently require authors to declare it.

This issue is to request a detailed design for a new required `exposes` feature that makes exposed package dependencies explicit in source.

## Motivation

We want the package header to tell the truth about the package's public surface.

Today, after typechecking, a reader can see:

- what the file depends on via `from ... import ...`
- what escapes via `export`
- recursion properties on functions

But the reader still cannot see which package dependencies are visible through the exported API.

That is a gap in the current source-locality story. In library code, it is especially important because some package dependencies should remain private implementation details. If an exported type or value mentions a type from such a package, that dependency is no longer private in any meaningful sense.

## Existing Compiler Context

Bosatsu already has most of the semantic machinery needed:

- package typing computes the package set visible from exported referants
- library assembly validates that exported package APIs only mention packages from exported packages or public dependencies, and rejects packages from private dependencies

What is missing is a source-level declaration and a source-level check that the declaration matches reality.

## Request

Please design a required source feature for declaring exposed package dependencies. This is **not** meant to be an opt-in lint or optional annotation. It is meant to become part of the required source interface for exported packages.

## Required Properties Of The Design

1. If a package has any `export` declarations, it must also declare its exposed package dependencies in source.
2. This must be a checked language feature, not a documentation-only convention.
3. The check must run after the compiler has enough information to know the actual exposed dependency set of the exported API.
4. Missing declarations and extra declarations must both be errors.
5. The design must treat this as a backwards-incompatible language change and discuss migration.

### 1. Syntax

Discuss candidate syntaxes and choose one. For example:

- `exposes Foo/Bar, Baz/Qux`
- a block form such as `exposes:`
- an import modifier such as `from Foo/Bar visibly import ...`
- any other syntax that fits Bosatsu's header style

The design should explain why the chosen syntax matches Bosatsu's principles better than the alternatives.

### 2. Where The Declaration Lives

The design should explicitly discuss placement.

Questions to settle:

- Is this a new top-level header declaration next to `package`, `import`, and `export`?
- Is it tied to imports?
- Is it tied to exports?
- If multiple `export` lines exist, is there one `exposes` declaration or several?
- What is the required spelling when the exposed set is empty?

This matters because the goal is to make the package interface recoverable from the top of the file.

### 3. Exact Meaning Of "Exposed"

Define exactly which packages count as exposed by a package's API.

At minimum the design should discuss:

- exported value types
- exported type definitions
- exported constructors vs opaque exported types
- foreign packages that appear only through inference from exported values
- `external` values/types
- renamed imports and aliases
- whether the package should count as exposing itself
- how `Bosatsu/Predef` should be treated
- whether same-library packages count, or whether `exposes` is only about external/package-dependency boundaries

The design should be explicit about whether the source declaration is intended to reflect the compiler's current visible package computation exactly, or a normalized/user-facing subset of it.

### 4. Checking Rules

Specify the compiler checks in detail.

For example:

- compute the actual exposed package set from the typed exported API
- compare it to the declared set
- fail on missing packages
- fail on extra packages
- produce diagnostics that make it easy to see the mismatch

The design should also discuss whether diagnostics should point at the exported names that cause a package to escape.

### 5. Interaction With Library Public/Private Dependencies

Explain how this source-level feature interacts with existing library assembly checks around public and private dependencies.

Questions to settle:

- Does the package-level `exposes` declaration become the source-visible counterpart of the existing library-level public/private dependency checks?
- Does it affect how we think about exported packages inside the same library?
- Does it require any changes to library assembly validation, semver checks, or docs generation?

### 6. Backwards Compatibility And Migration

This should be treated as a breaking language change.

The design should discuss:

- migration strategy for existing Bosatsu source
- whether auto-fix or tooling support is needed
- how this affects existing libraries and package headers
- versioning/release implications

## Desired Outcome

After this feature exists, a reader of a typechecked Bosatsu source file should be able to look at the package header and know not only:

- what the file imports
- what it exports

but also:

- which package dependencies are exposed by that exported API

That would close an important gap in the current source-locality/design story.

johnynek:
Implements issue #2254 by adding header-level `exposes` declarations to package parsing and source checking. The compiler now carries parsed `exposes` declarations through package resolution, normalizes actual exported dependency visibility, treats omitted `exposes` as the empty set, reports duplicate or mismatched declarations with canonical fixes and exported-name causes, replays postponable mismatches under `--warn` and `--lax`, and reuses the same normalized helper in `LibConfig` visibility validation while keeping compiled/interface artifacts unchanged. The change also updates predef, docs, CLI/path-module fixtures, and in-repo test workspace packages with explicit `exposes` headers, and adds parser, package semantics, error-message, library, cache-hit warning, and integration coverage. Verified with `sbt "coreJVM/test; cli/test"`, `sbt "doc; paradox"`, and `scripts/test_basic.sh`.

Fixes #2254

Implements design doc: [docs/design/2254-design-required-exposes-declarations-for-exported-dependency-visibility.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/2254-design-required-exposes-declarations-for-exported-dependency-visibility.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/2256

### DIFF
--- a/core/src/main/resources/bosatsu/predef.bosatsu
+++ b/core/src/main/resources/bosatsu/predef.bosatsu
@@ -109,9 +109,6 @@ export (
   trace,
   xor,
 )
-
-exposes ()
-
 struct Unit
 struct Tuple1[a: +*](item1: a)
 struct Tuple2[a: +*, b: +*](item1: a, item2: b)
--- a/core/src/main/scala/dev/bosatsu/Package.scala
+++ b/core/src/main/scala/dev/bosatsu/Package.scala
@@ -176,7 +176,7 @@ object Package {
     val withRegions = typedFunctor.map(pack)(tag => HasRegion.region(tag))
     // Serialized compiled artifacts never retain the parsed-source payload in
     // Program.from, so drop it before crossing the cache/compiler boundary.
-    setProgramFrom(withRegions, ()).copy(exposes = Nil)
+    dropSourceOnlyExposes(setProgramFrom(withRegions, ()))
   }
 
   def normalizeExposedDepPackages(
@@ -379,8 +379,16 @@ object Package {
   def fromStatements(pn: PackageName, stmts: List[Statement]): Package.Parsed =
     Package(pn, Nil, Nil, stmts)
 
+  // `exposes` is a checked source declaration, but compiled/interfaces already
+  // carry the exported typed API it is derived from. Dropping it keeps those
+  // artifacts unchanged and avoids storing redundant header spelling.
+  private def dropSourceOnlyExposes[A, B, C, D](
+      pack: Package[A, B, C, D]
+  ): Package[A, B, C, D] =
+    pack.copy(exposes = Nil)
+
   def interfaceOf[A](inferred: Typed[A]): Interface =
-    inferred.mapProgram(_ => ()).replaceImports(Nil).copy(exposes = Nil)
+    dropSourceOnlyExposes(inferred.mapProgram(_ => ()).replaceImports(Nil))
 
   def setProgramFrom[A, B](t: Typed[A], newFrom: B): Typed[A] =
     t.copy(program = (t.program._1.copy(from = newFrom), t.program._2))
@@ -417,18 +425,23 @@ object Package {
         val x = exposes match {
           case Nil               => Doc.empty
           case exposeDecls       =>
-            def exposeDoc(deps: List[PackageName]): Doc =
-              if (deps.isEmpty) Doc.text("exposes ()")
-              else {
-                Doc.text("exposes ") +
-                  Doc.intercalate(
-                    Doc.text(", "),
-                    deps.map(Document[PackageName].document)
+            val rendered =
+              exposeDecls.flatMap { deps =>
+                if (deps.isEmpty) Nil
+                else {
+                  List(
+                    Doc.text("exposes ") +
+                      Doc.intercalate(
+                        Doc.text(", "),
+                        deps.map(Document[PackageName].document)
+                      )
                   )
+                }
               }
-
+            if (rendered.isEmpty) Doc.empty
+            else
             Doc.line +
-              Doc.intercalate(Doc.line, exposeDecls.map(exposeDoc)) +
+              Doc.intercalate(Doc.line, rendered) +
               Doc.line
         }
         val b = statments.map(Document[Statement].document(_))
--- a/core/src/main/scala/dev/bosatsu/PackageError.scala
+++ b/core/src/main/scala/dev/bosatsu/PackageError.scala
@@ -157,8 +157,11 @@ object PackageError {
           .grouped
     }
 
-  private def exposesLineDoc(packages: List[PackageName]): Doc =
-    Doc.text("exposes ") + exposesSetDoc(packages)
+  private def exposesLineDoc(packages: List[PackageName]): Option[Doc] =
+    packages match {
+      case Nil      => None
+      case nonEmpty => Some(Doc.text("exposes ") + exposesSetDoc(nonEmpty))
+    }
 
   private def suggestedName(
       ident: Identifier,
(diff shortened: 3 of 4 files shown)


core/src/main/resources/bosatsu/predef.bosatsu
```diff
@@ -110,6 +110,8 @@ export (
   xor,
 )
 
+exposes ()
```
johnynek: Don't add empty exposes. Just let the empty set rule work.

core/src/main/scala/dev/bosatsu/Package.scala
```diff
@@ -360,15 +380,15 @@ object Package {
     Package(pn, Nil, Nil, stmts)
 
   def interfaceOf[A](inferred: Typed[A]): Interface =
-    inferred.mapProgram(_ => ()).replaceImports(Nil)
+    inferred.mapProgram(_ => ()).replaceImports(Nil).copy(exposes = Nil)
```
johnynek: Add a comment why we drop the exposes here? Exposes is a strict function of exports so it seems like it is computable on interfaces. Why not keep it?

core/src/main/scala/dev/bosatsu/PackageError.scala
```diff
@@ -144,6 +144,22 @@ object PackageError {
...
+        Doc
+          .intercalate(
+            Doc.text(",") + Doc.lineOrSpace,
+            nonEmpty.map(pn => Doc.text(pn.asString))
+          )
+          .grouped
+    }
+
+  private def exposesLineDoc(packages: List[PackageName]): Doc =
+    Doc.text("exposes ") + exposesSetDoc(packages)
```
johnynek: Don't print this line for empty sets.

core/src/main/scala/dev/bosatsu/PackageError.scala
```diff
@@ -1981,6 +1998,98 @@ object PackageError {
...
+      pack: PackageName,
+      declarations: NonEmptyList[List[PackageName]]
+  ) extends PackageError {
+    def message(
+        sourceMap: Map[PackageName, (LocationMap, String)],
+        errColor: Colorize
+    ) = {
+      val prefix = sourceMap.headLine(pack, None)
+      val declCount = declarations.length
+      val declWord = if (declCount == 1) "declaration" else "declarations"
```
johnynek: 1 is not an error. We don't need this handling. The error is only triggered for 2 or more.

core/src/main/scala/dev/bosatsu/Package.scala
```diff
@@ -394,8 +414,25 @@ object Package {
               ) +
               Doc.line
         }
+        val x = exposes match {
+          case Nil               => Doc.empty
+          case exposeDecls       =>
+            def exposeDoc(deps: List[PackageName]): Doc =
+              if (deps.isEmpty) Doc.text("exposes ()")
```
johnynek: Don't print the empty set.

core/src/main/resources/bosatsu/predef.bosatsu
johnynek: Removed the explicit `exposes ()` from predef and let omission stand for the empty set.

core/src/main/scala/dev/bosatsu/Package.scala
johnynek: Added a shared helper comment explaining why compiled/interface artifacts drop `exposes`: it is a checked source declaration that is derivable from the typed exports, so we keep artifacts unchanged and avoid storing redundant header spelling.

core/src/main/scala/dev/bosatsu/PackageError.scala
johnynek: Adjusted the error rendering so empty `exposes` declarations are no longer printed in the duplicate-declaration output.

core/src/main/scala/dev/bosatsu/PackageError.scala
johnynek: Updated the duplicate-declaration message so it only talks about multiple declarations; there is no longer any singular-case handling.

core/src/main/scala/dev/bosatsu/Package.scala
johnynek: Updated the package pretty-printer to skip empty `exposes` declarations instead of rendering `exposes ()`.

johnynek:
Applied the review fixes and reran `scripts/test_basic.sh` locally; it passes.

PR #2264, opened 2026-03-27
Implement Matchless global inlining phase

### LINKED ISSUE #2185: Design a matchless global inlining phase

johnynek:
Added a post-lowering Matchless global inlining pass that keeps raw lowering parallel, publishes per-binding inline summaries, and rewrites eligible cross-package calls before codegen and evaluation. The change refactors Matchless cleanup into explicit stages, adds `inlineApplyArgs` plus demand/weight analysis, wires `CompilationSource`, `DecodedLibraryWithDeps`, and `LibraryEvaluation` through the optimized pipeline, always inlines `Bosatsu/Predef.and` and `Bosatsu/Predef.or`, keeps while-lowered helpers eligible with conservative higher-order loop gating, and adds structural, interface-only, while-loop, and cross-library codegen regressions. Verification included the focused Matchless and clang dependency suites and `scripts/test_basic.sh`.

Fixes #2185

Implements design doc: [docs/design/2185-design-a-matchless-global-inlining-phase.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/2185-design-a-matchless-global-inlining-phase.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/2187

### DIFF
--- a/core/src/main/scala/dev/bosatsu/Matchless.scala
+++ b/core/src/main/scala/dev/bosatsu/Matchless.scala
@@ -1605,12 +1605,18 @@ object Matchless {
         .collect {
           case ((argName, argExpr), demand)
               if !demand.unused &&
-                // CheapExpr positions such as EqualsNat/GetStructElement cannot
-                // directly hold nullary constructors like ZeroNat or MakeStruct(0),
-                // even though they are semantically trivial. Memoize any non-CheapExpr
-                // argument once so those sites can still reference a cheap local.
                 !argExpr.isInstanceOf[CheapExpr[?]] &&
-                ((demand.cheapPositionUses > 0) || isTriviallyCheap(argExpr)) =>
+                (
+                  // Any eager use means the original call would have evaluated the
+                  // argument before entering the helper, so binding it once at the
+                  // call site preserves that behavior while avoiding duplication.
+                  ((demand.eagerUses > 0) &&
+                    ((demand.totalUses > 1) || (demand.cheapPositionUses > 0))) ||
+                  // CheapExpr positions such as EqualsNat/GetStructElement cannot
+                  // directly hold nullary constructors like ZeroNat or
+                  // MakeStruct(0), even though they are semantically trivial.
+                  ((demand.cheapPositionUses > 0) && isTriviallyCheap(argExpr))
+                ) =>
             (argName, argExpr)
         }
       val argMemoNames =
--- a/core/src/main/scala/dev/bosatsu/MatchlessGlobalInlining.scala
+++ b/core/src/main/scala/dev/bosatsu/MatchlessGlobalInlining.scala
@@ -385,30 +385,25 @@ object MatchlessGlobalInlining {
             case (demand, arg) =>
               demand.lambdaCalleeOnly && resolvesToLambda(arg).nonEmpty
           }
-      val nonCheapCheapPositionArgument =
+      val deferredCheapPositionArgument =
         summary.paramDemand.iterator.zip(argList.iterator).exists {
           case (demand, arg) =>
-            (demand.cheapPositionUses > 0) && !isCheapArgument(arg)
+            (demand.eagerUses == 0) &&
+            (demand.cheapPositionUses > 0) &&
+            !isCheapArgument(arg)
         }
-      val eagerNonCheapArgument =
+      val duplicatedDeferredExpensiveArg =
         summary.paramDemand.iterator.zip(argList.iterator).exists {
           case (demand, arg) =>
-            !isCheapArgument(arg) &&
-            !demand.unused &&
-            !demand.deferrable &&
-            !demand.lambdaCalleeOnly
-        }
-      val duplicatedExpensiveArg =
-        summary.paramDemand.iterator.zip(argList.iterator).exists {
-          case (demand, arg) =>
-            (demand.totalUses > MaxExpensiveArgUses) && !isCheapArgument(arg)
+            (demand.eagerUses == 0) &&
+            (demand.totalUses > MaxExpensiveArgUses) &&
+            !isCheapArgument(arg)
         }
 
       if (
         whileLambdaArgument ||
-        nonCheapCheapPositionArgument ||
-        eagerNonCheapArgument ||
-        duplicatedExpensiveArg
+        deferredCheapPositionArgument ||
+        duplicatedDeferredExpensiveArg
       ) false
       else {
         val hasUnusedBenefit =
--- a/cli/src/test/scala/dev/bosatsu/codegen/clang/ClangGenTest.scala
+++ b/cli/src/test/scala/dev/bosatsu/codegen/clang/ClangGenTest.scala
@@ -42,7 +42,7 @@ class ClangGenTest extends munit.FunSuite {
       To inspect the code, change the hash, and it will print the code out
      */
     testFilesCompilesToHash("test_workspace/Ackermann.bosatsu")(
-      "edde15d44a15bd861424b98c10232cad"
+      "a213c02bffb91cfaa72981b6a2992614"
     )
   }
 }
--- a/core/src/test/scala/dev/bosatsu/MatchlessApplyArgsTest.scala
+++ b/core/src/test/scala/dev/bosatsu/MatchlessApplyArgsTest.scala
@@ -76,6 +76,109 @@ class MatchlessApplyArgsTest extends munit.FunSuite {
     }
   }
 
+  test("Matchless.inlineApplyArgs directly substitutes single eager non-cheap arguments") {
+    val arg = Identifier.Name("arg")
+    val consume = Identifier.Name("consume")
+    val expensive = Identifier.Name("expensive")
+    val actual: Matchless.Expr[Unit] =
+      Matchless.App(
+        Matchless.Local(expensive),
+        NonEmptyList.one(Matchless.Literal(Lit.fromInt(1)))
+      )
+    val lam: Matchless.Expr[Unit] =
+      Matchless.Lambda(
+        captures = Nil,
+        recursiveName = None,
+        args = NonEmptyList.one(arg),
+        body = Matchless.App(
+          Matchless.Local(consume),
+          NonEmptyList.one(Matchless.Local(arg))
+        )
+      )
+
+    Matchless.inlineApplyArgs(lam, NonEmptyList.one(actual)) match {
+      case Matchless.App(Matchless.Local(`consume`), appliedArgs) =>
+        assertEquals(appliedArgs.toList, actual :: Nil)
+      case Matchless.Let(_, `actual`, _) =>
+        fail("expected single eager use to substitute directly")
+      case other =>
+        fail(s"expected direct substitution for the eager argument, found: $other")
+    }
+  }
+
+  test("Matchless.inlineApplyArgs memoizes repeated eager non-cheap arguments once") {
+    val arg = Identifier.Name("arg")
+    val consume = Identifier.Name("consume")
+    val expensive = Identifier.Name("expensive")
+    val actual: Matchless.Expr[Unit] =
+      Matchless.App(
+        Matchless.Local(expensive),
+        NonEmptyList.one(Matchless.Literal(Lit.fromInt(1)))
+      )
+    val lam: Matchless.Expr[Unit] =
+      Matchless.Lambda(
+        captures = Nil,
+        recursiveName = None,
+        args = NonEmptyList.one(arg),
+        body = Matchless.App(
+          Matchless.Local(consume),
+          NonEmptyList.of(Matchless.Local(arg), Matchless.Local(arg))
+        )
+      )
+
+    Matchless.inlineApplyArgs(lam, NonEmptyList.one(actual)) match {
+      case Matchless.Let(
+            Right(tmp),
+            `actual`,
+            Matchless.App(Matchless.Local(`consume`), appliedArgs)
+          ) =>
+        assertEquals(
+          appliedArgs.toList,
+          List(Matchless.Local(tmp), Matchless.Local(tmp))
+        )
+      case other =>
+        fail(s"expected repeated eager use to memoize once, found: $other")
+    }
+  }
+
+  test("Matchless.inlineApplyArgs memoizes eager non-cheap args for cheap positions") {
+    val arg = Identifier.Name("arg")
+    val expensive = Identifier.Name("expensive")
+    val actual: Matchless.Expr[Unit] =
+      Matchless.App(
+        Matchless.Local(expensive),
+        NonEmptyList.one(Matchless.Literal(Lit.fromInt(1)))
+      )
+    val lam: Matchless.Expr[Unit] =
+      Matchless.Lambda(
+        captures = Nil,
+        recursiveName = None,
+        args = NonEmptyList.one(arg),
+        body = Matchless.If(
+          Matchless.EqualsNat(Matchless.Local(arg), rankn.DataRepr.ZeroNat),
+          Matchless.Literal(Lit.fromInt(1)),
+          Matchless.Literal(Lit.fromInt(2))
+        )
+      )
+
+    Matchless.inlineApplyArgs(lam, NonEmptyList.one(actual)) match {
+      case Matchless.Let(
+            Right(tmp),
+            `actual`,
+            Matchless.If(
+              Matchless.EqualsNat(Matchless.Local(tmpRef), rankn.DataRepr.ZeroNat),
+              Matchless.Literal(ifTrue),
+              Matchless.Literal(ifFalse)
+            )
+          ) =>
+        assertEquals(tmpRef, tmp)
+        assertEquals(ifTrue, Lit.fromInt(1))
+        assertEquals(ifFalse, Lit.fromInt(2))
+      case other =>
+        fail(s"expected eager cheap-position use to memoize once, found: $other")
+    }
+  }
+
   test(
     "Matchless.inlineApplyArgs memoizes non-CheapExpr nullary constructors in cheap positions"
   ) {
--- a/core/src/test/scala/dev/bosatsu/MatchlessTests.scala
+++ b/core/src/test/scala/dev/bosatsu/MatchlessTests.scala
@@ -5146,6 +5146,51 @@ def seg_final_literal_char(s):
     }
   }
 
+  test("optimized Matchless inlines mixed eager and deferrable helpers") {
+    val helperPack = PackageName.parts("Helper", "Mixed")
+    val callerPack = PackageName.parts("Caller", "Mixed")
+    val chooseName = Identifier.Name("choose")
+    val useName = Identifier.Name("use")
+
+    checkOptimizedMatchlessPackages(
+      NonEmptyList.of(
+        """package Helper/Mixed
+          |
+          |export choose
+          |
+          |def choose(flag: Bool, eager: Int, on_true: Int) -> Int:
+          |  if flag:
+          |    on_true
+          |  else:
+          |    eager.add(eager)
+          |""".stripMargin,
+        """package Caller/Mixed
+          |
+          |from Helper/Mixed import choose
+          |
+          |def expensive(i: Int) -> Int:
+          |  if False:
+          |    0
+          |  else:
+          |    i
+          |
+          |def use(i: Int) -> Int:
+          |  choose(True, expensive(i), 42)
+          |""".stripMargin
+      )
+    ) { compiled =>
+      val useExpr = compiled(callerPack).toMap.apply(useName)
+      assertEquals(containsGlobal(useExpr, helperPack, chooseName), false)
+
+      Matchless.recoverTopLevelLambda(useExpr) match {
+        case Matchless.Lambda(Nil, None, _, Matchless.Literal(lit)) =>
+          assertEquals(lit, Lit.fromInt(42))
+        case other =>
+          fail(s"expected mixed helper to collapse to 42, found: $other")
+      }
+    }
+  }
+
   test("optimized Matchless exposes beta-reduction for lambda arguments") {
     val helperPack = PackageName.parts("Helper", "Lambda")
     val callerPack = PackageName.parts("Caller", "Lambda")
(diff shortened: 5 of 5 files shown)


johnynek:
Updated the brittle core clang goldens to stop depending on exact temporary ids after the shared anon/mutable-anon namespace fix. `core/src/test/scala/dev/bosatsu/codegen/clang/ClangGenTest.scala` now canonicalizes generated local temp names for exact-string predef checks and uses id-agnostic structural assertions in the list-search and matrix-scoring tests. Validation passed with `sbt --batch -Dsbt.server.autostart=false "coreJVM/testOnly dev.bosatsu.codegen.clang.ClangGenTest"` and the required `scripts/test_basic.sh`.

core/src/main/scala/dev/bosatsu/library/DecodedLibraryWithDeps.scala
```diff
@@ -54,6 +55,18 @@ case class DecodedLibraryWithDeps[A](
     MatchlessFromTypedExpr.compile(nameVersion, lib.implementations)
   }
 
+  def compileRaw(implicit
+      ec: Par.EC
+  ): MatchlessFromTypedExpr.Compiled[(Name, Version)] = {
+    given Order[(Name, Version)] = Order.fromOrdering(using
```
johnynek: is this needed? Surely the default `Order` for tuple2 will work without anything extra won't it?

Is one of `Name` or `Version` missing a default `Order`? If so, that's the fix. Put the given on their companion object.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -591,6 +591,348 @@ object Matchless {
       .take(count)
       .toList
 
+  private[bosatsu] def allLocalAnonIds[A](expr: Expr[A]): Set[Long] = {
+    def loopExpr(ex: Expr[A], acc: Set[Long]): Set[Long] =
```
johnynek: the result I don't think depends on the order or traversal, so let's use a `List[Expr[A]]` stack to make sure this is stack safe.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -591,6 +591,348 @@ object Matchless {
...
+          loopExpr(ge.arg, acc)
+        case gs: GetStructElement[?] =>
+          loopExpr(gs.arg, acc)
+        case Local(_) | Global(_, _, _) | ClosureSlot(_) | LocalAnonMut(_) |
+            Literal(_) | MakeEnum(_, _, _) | MakeStruct(_) | SuccNat |
+            ZeroNat =>
+          acc
+      }
+
+    def loopBool(ex: BoolExpr[A], acc: Set[Long]): Set[Long] =
```
johnynek: same comment about stack safety here. Maybe what we need is `List[Expr[A] | BoolExpr[A]]` and have one big match.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -591,6 +591,348 @@ object Matchless {
...
+        case LetMutBool(_, in) =>
+          loopBool(in, acc)
+        case TrueConst =>
+          acc
+      }
+
+    loopExpr(expr, Set.empty)
+  }
+
+  private[bosatsu] def allLocalAnonMutIds[A](expr: Expr[A]): Set[Long] = {
```
johnynek, line 673: same comments about stack safety.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -591,6 +591,348 @@ object Matchless {
...
+          loopBool(in, acc + name.ident)
+        case TrueConst =>
+          acc
+      }
+
+    loopExpr(expr, Set.empty)
+  }
+
+  private def maxLocalAnonId[A](expr: Expr[A]): Long =
+    allLocalAnonIds(expr).foldLeft(-1L)(_ max _)
```
johnynek, line 734: if we are often doing max and min, maybe use `SortedSet[Long]` as the type for anonIds, so it would be faster.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -591,6 +591,348 @@ object Matchless {
...
+
+  private def maxLocalAnonMutId[A](expr: Expr[A]): Long =
+    allLocalAnonMutIds(expr).foldLeft(-1L)(_ max _)
+
+  private[bosatsu] def refreshAnonBinders[A](expr: Expr[A]): Expr[A] = {
+    case class RenameEnv(
```
johnynek: add a comment what the key and value types mean here.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -679,41 +1021,614 @@ object Matchless {
...
+  private def substituteBindables[A](
+      expr: Expr[A],
+      subst: Map[Bindable, Expr[A]]
+  ): Expr[A] = {
+    def loopExpr(ex: Expr[A], env: Map[Bindable, Expr[A]]): Expr[A] =
+      ex match {
+        case Local(name) =>
+          env.getOrElse(name, ex)
+        case Lambda(captures, recName, args, body) =>
+          val bodyEnv = env -- recName.toSet -- args.toList
```
johnynek: if `bodyEnv.isEmpty` we can avoid the captures.map. We should check that before recursing.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -679,41 +1021,614 @@ object Matchless {
...
+        case WhileExpr(cond, effectExpr, result) =>
+          WhileExpr(loopBool(cond, env), loopExpr(effectExpr, env), result)
+        case App(fn, appArgs) =>
+          App(loopExpr(fn, env), appArgs.map(loopExpr(_, env)))
+        case Let(arg, value, in) =>
+          val env1 =
+            arg match {
+              case Right(name) => env - name
+              case Left(_)     => env
+            }
```
johnynek: if `env1.isEmpty` we can just use `in` and we don't need to loop. We should do that check.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -679,41 +1021,614 @@ object Matchless {
...
+        case CheckVariantSet(arg, expect, size, famArities) =>
+          CheckVariantSet(loopCheap(arg, env), expect, size, famArities)
+        case SetMut(target, value) =>
+          SetMut(target, loopExpr(value, env))
+        case LetBool(arg, value, in) =>
+          val env1 =
+            arg match {
+              case Right(name) => env - name
+              case Left(_)     => env
+            }
```
johnynek: if `env1.isEmpty` we don't need to recurse, check that.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -679,41 +1021,614 @@ object Matchless {
...
+                args = lamArgs,
+                body = body
+              )
+            Let(name, recLam, bodyWithArgs)
+          case None =>
+            bodyWithArgs
+        }
+
+      val withArgTmps =
+        argTmpNames.zip(args.toList).foldRight(bodyWithRec) {
```
johnynek: can't this use `Let.bindNamed` or another `object Let` function?

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -679,41 +1021,614 @@ object Matchless {
...
+        )
+      val (captureTmpNames, argTmpNames) = tmpNames.splitAt(captureCount)
+      val captureTmpLocals: List[CheapExpr[A]] =
+        captureTmpNames.map(Local(_))
+
+      val bodyWithCaptures =
+        substituteClosureSlots(captureTmpLocals.toVector, body)
+
+      val bodyWithArgs =
+        lamArgs.toList.zip(argTmpNames).foldRight(bodyWithCaptures) {
```
johnynek: can't this use `Let.bindNamed` or another `object Let` function?

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -679,41 +1021,614 @@ object Matchless {
...
+          case ((argName, argTmp), in) =>
+            Let(argName, Local(argTmp), in)
+        }
+
+      val bodyWithRec =
+        recName match {
+          case Some(name) =>
+            val recLam =
```
johnynek, line 1252: add a comment here. I'm confused how we can betareduce a recursive function. I guess what's happening is that we are making a let of the function, then doing the let binding after, but it's always going to increase the size of the code. Let's add a comment about why this is beneficial.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -679,41 +1021,614 @@ object Matchless {
...
-  def applyArgs[A](fn: Expr[A], args: NonEmptyList[Expr[A]]): Expr[A] = {
-    def betaReduce(lam: Lambda[A]): Expr[A] = {
-      val Lambda(captures, recName, lamArgs, body) = lam
-      val captureCount = captures.length
-      val argCount = args.length
-      val usedNames =
+  def inlineApplyArgs[A](fn: Expr[A], args: NonEmptyList[Expr[A]]): Expr[A] = {
+    val argNames = allNamesMany(args.iterator)
+    val argAnonIds =
+      args.iterator.foldLeft(Set.empty[Long]) { case (acc, arg) =>
```
johnynek: I think this is `args.foldMap(allLocalAnonIds)` isn't it? Can we do that?

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -679,41 +1021,614 @@ object Matchless {
...
+        acc ++ allLocalAnonIds(arg)
+      }
+    val argMutIds =
```
johnynek: I think this is `args.foldMap(allLocalAnonMutsIds)` isn't it? Can we do that?

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -679,41 +1021,614 @@ object Matchless {
...
+            bodyWithArgs
+        }
+
+      val withArgTmps =
+        argTmpNames.zip(args.toList).foldRight(bodyWithRec) {
+          case ((argTmp, argExpr), in) =>
+            Let(argTmp, argExpr, in)
+        }
+
+      captureTmpNames.zip(captures).foldRight(withArgTmps) {
```
johnynek: can't this use Let.bindNamed or another object Let function?

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -679,41 +1021,614 @@ object Matchless {
...
+    val argAnonIds =
+      args.iterator.foldLeft(Set.empty[Long]) { case (acc, arg) =>
+        acc ++ allLocalAnonIds(arg)
+      }
+    val argMutIds =
+      args.iterator.foldLeft(Set.empty[Long]) { case (acc, arg) =>
+        acc ++ allLocalAnonMutIds(arg)
+      }
+
+    case class RenameEnv(
```
johnynek, line 1326: this feels duplicative. Earlier in this file we have added a very similar data-structure just without the `bindables`.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -679,41 +1021,614 @@ object Matchless {
...
+        (next, copy(nextId = nextId + 1L))
+      }
+
+      def freshMut: (LocalAnonMut, RenameState) = {
+        val next = LocalAnonMut(nextId)
+        (next, copy(nextId = nextId + 1L))
+      }
+    }
+
+    def renameBinders(lam0: Lambda[A]): Lambda[A] = {
```
johnynek, line 1358: adda  comment why we need this. Is this analogous to the `unshadow` methods we have elsewhere (in TypedExpr and Expr I think, maybe also in the python.Code)?

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -679,41 +1021,614 @@ object Matchless {
...
-        lamArgs.toList.zip(argTmpNames).foldRight(bodyWithCaptures) {
-          case ((argName, argTmp), in) =>
-            Let(argName, Local(argTmp), in)
+      val argMemos = lamArgs.toList
+        .zip(args.toList)
+        .zip(paramDemand)
+        .collect {
+          case ((argName, argExpr), demand)
+              if (demand.cheapPositionUses > 0) &&
+                !argExpr.isInstanceOf[CheapExpr[?]] =>
```
johnynek: add a comment why this shouldn't be `!argExpr.isTriviallyCheap` which may be closer to what we want.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -793,18 +1694,217 @@ object Matchless {
...
+    val cheapPositionUses = Array.fill(args.length)(0)
+
+    def record(
+        name: Bindable,
+        branchOnly: Boolean,
+        directCallee: Boolean,
+        insideLambda: Boolean,
+        cheapContext: Boolean
+    ): Unit =
+      indexOf.get(name).foreach { idx =>
```
johnynek: let's use `indexOf.get(name) match` and avoid allocating a closure

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -793,18 +1694,217 @@ object Matchless {
...
+    val args = lambda.args.toList
+    val indexOf = args.zipWithIndex.toMap
+    val totalUses = Array.fill(args.length)(0)
+    val eagerUses = Array.fill(args.length)(0)
+    val branchOnlyUses = Array.fill(args.length)(0)
+    val directCalleeUses = Array.fill(args.length)(0)
+    val nonDirectCalleeUses = Array.fill(args.length)(0)
+    val cheapPositionUses = Array.fill(args.length)(0)
+
+    def record(
```
johnynek: I think `inline def` would help here since a lot of the boolean params are known and will reduce in compilation.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -793,18 +1694,217 @@ object Matchless {
...
+      indexOf.get(name).foreach { idx =>
+        totalUses(idx) += 1
+        if (branchOnly && !insideLambda) branchOnlyUses(idx) += 1
+        else eagerUses(idx) += 1
+        if (directCallee && !insideLambda) directCalleeUses(idx) += 1
+        else nonDirectCalleeUses(idx) += 1
+        if (cheapContext) cheapPositionUses(idx) += 1
+      }
+
+    def loopExpr(
```
johnynek: I think making a stack of `List[LoopState]` where we have `case class LoopState(arg: Expr[A] | BoolExpr[A], branchOnly: Boolean, directCallee: Boolean, insideLambda: Boolean, cheapContext: Boolean, shadowed: Set[Bindable]))`

we could make this traversal stack safe.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -3356,18 +4389,34 @@ object Matchless {
       case NonEmptyList(h0, h1 :: t)   => h0 :: stopAt(NonEmptyList(h1, t))(fn)
     }
 
-  // same as fromLet below, but uses RefSpace
+  private[bosatsu] def postLoweringCleanup[A: Order](expr: Expr[A]): Expr[A] =
+    reuseConstructors(hoistInvariantLoopLets(expr))
+
+  // same as fromLetRaw below, but uses RefSpace
```
johnynek: it's not the same. It uses fromLetRaw and then runs postLoweringCleanup. Please fix the comment.

core/src/test/scala/dev/bosatsu/MatchlessInterfaceTest.scala
```diff
@@ -56,4 +124,60 @@ class MatchlessInterfaceTest extends munit.FunSuite {
...
+
+    val helperPm = typeCheck(helperSrc, Nil)
+    val helperIface = Package.interfaceOf(helperPm.toMap(helperPack))
+
+    val callerSrc =
+      """package Caller/Bool
+        |
+        |from Helper/Bool import choose
+        |
+        |def use(i: Int) -> Int:
```
johnynek, line 154: we should check that this becomes the same as `def use(i: Int) -> Int: 0` That's the idea right? But the tests don't assert it.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -730,37 +1645,22 @@ object Matchless {
             bodyWithArgs
         }
 
-      val withArgTmps =
-        argTmpNames.zip(args.toList).foldRight(bodyWithRec) {
-          case ((argTmp, argExpr), in) =>
-            Let(argTmp, argExpr, in)
+      val withArgMemos =
+        argMemoNames.zip(argMemos.map(_._2)).foldRight(bodyWithRec) {
```
johnynek: should use `Let.bindNames` or whatever it is. We already have functions for making Let chains. Use them.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -730,37 +1645,22 @@ object Matchless {
...
+          case ((memoName, memoExpr), in) =>
+            Let(memoName, memoExpr, in)
         }
 
-      captureTmpNames.zip(captures).foldRight(withArgTmps) {
+      captureTmpNames.zip(captures).foldRight(withArgMemos) {
```
johnynek: should use `Let.bindNames` or whatever it is. We already have functions for making Let chains. Use them.

core/src/main/scala/dev/bosatsu/MatchlessGlobalInlining.scala
```diff
@@ -0,0 +1,597 @@
...
+  private val MaxExpensiveArgUses = 1
+
+  def optimize[K: Order](
+      rawCompiled: SortedMap[K, MatchlessFromTypedExpr.Compiled[K]],
+      topoSort: Toposort.Result[(K, PackageName)],
+      depFor: (K, PackageName) => K
+  )(implicit ec: Par.EC): SortedMap[K, MatchlessFromTypedExpr.Compiled[K]] = {
+    val cleanedBase = SortedMap.from(rawCompiled.iterator.map { case (scope, packs) =>
+      val packs1 = packs.iterator.map { case (pack, lets) =>
+        pack -> lets.map { case (name, expr) =>
```
johnynek: seems like we could use `Par.start(lets.map( ...))` and start all the work, then go through again and await each one, so we can in principle do the lowering of each package in parallel.

core/src/main/scala/dev/bosatsu/MatchlessGlobalInlining.scala
```diff
@@ -0,0 +1,597 @@
...
+      scope: K,
+      pack: PackageName,
+      lets: List[(Bindable, Expr[K])],
+      depFor: (K, PackageName) => K
+  ): List[Bindable] = {
+    case class LocalNode(index: Int, name: Bindable)
+    given Ordering[LocalNode] = Ordering.by((node: LocalNode) =>
+      (node.index, node.name)
+    )(using
+      Ordering.Tuple2(using
```
johnynek: we shouldn't need this. The default should just work. If it isn't add the correct given Orderings to the right objects, don't keep manually passing givens around with two layers of `using`.

core/src/main/scala/dev/bosatsu/MatchlessGlobalInlining.scala
```diff
@@ -0,0 +1,597 @@
...
+      )
+    )
+
+    val nodeByName = lets.zipWithIndex.iterator.map { case ((name, _), idx) =>
+      name -> LocalNode(idx, name)
+    }.toMap
+    val exprByName = lets.toMap
+
+    val topo = Toposort.sort(nodeByName.values) { node =>
+      collectGlobals(exprByName.getOrElse(node.name, Matchless.UnitExpr))
```
johnynek: how can using `UnitExpr` be correct here? This is a strong codesmell. Can you add a comment or remove the `getOrElse` and instead handle the missing node case (maybe it never happens, if not, give a sys.error with an comment explaining why that is some bug in the compiler if that is ever reached).

core/src/main/scala/dev/bosatsu/MatchlessGlobalInlining.scala
```diff
@@ -0,0 +1,597 @@
...
+    }
+
+    val ordered = topo.layers.iterator.flatMap(_.toList).map(_.name).toList
+    val missing =
+      lets.iterator.map(_._1).filterNot(ordered.toSet).toList
+
+    ordered ::: missing
+  }
+
+  private def collectGlobals[A](
```
johnynek, line 184: let's do one big stack safe loop on `List[Expr[A] | BoolExpr[A]]`

core/src/main/scala/dev/bosatsu/MatchlessGlobalInlining.scala
```diff
@@ -0,0 +1,597 @@
...
+
+    loopExpr(expr, Set.empty)
+  }
+
+  private def rewriteExpr[K](
+      expr: Expr[K],
+      summaries: Map[SummaryKey[K], InlineSummary[K]],
+      depFor: (K, PackageName) => K
+  ): Expr[K] = {
+    def builtinLambda(
```
johnynek: this is cheating. Remove this. We need to get this behavior automatically from the default inlining rules, not a hand written special case. We need to make sure our default rules would work for and and or as written, or likely our optimizations are too brittle. We know these should be the kinds of functions that will work.

johnynek:

- error type: `DirectFlowBlockedError`
- detail: `feedback flow blocked: codex invocation timed out after local changes were made: working tree has local changes`

johnynek:
MergeXO operator command result:
- command: `/mergexo unblock`
- status: `applied`
- detail: Reset blocked state for PR #2264.

Source command: https://github.com/johnynek/bosatsu/pull/2264#issuecomment-4148577248

johnynek:
please reread the previous feedback and make sure it was actioned. I think there was a lost feedback round that didn't get pushed.

johnynek:

- error type: `DirectFlowBlockedError`
- detail: `feedback flow blocked: codex invocation timed out after local changes were made: working tree has local changes`

core/src/main/scala/dev/bosatsu/MatchlessGlobalInlining.scala
```diff
@@ -0,0 +1,566 @@
...
+            !demand.deferrable &&
+            !demand.lambdaCalleeOnly
+        }
+      val duplicatedExpensiveArg =
+        summary.paramDemand.iterator.zip(argList.iterator).exists {
+          case (demand, arg) =>
+            (demand.totalUses > MaxExpensiveArgUses) && !isCheapArgument(arg)
+        }
+
+      if (
```
johnynek, line 403: if I understand correctly, these rules are assuming we never do any let binding of non-cheap arguments to the function. Is that right? Why would we have that limitation? For instance, if we have `fn(a, b)` where `a` is used eagerly but b is not, and `b`, it seems like we might want to inline that. Even if `a` is used more than once, we could just do: `let a = <a_arg> in subs(b = <b_arg>, fn-body)`. Couldn't we?

Am I misunderstanding this?

PR #2267, opened 2026-03-28
Improve exposes mismatch diagnostics

### LINKED ISSUE #2266: Bad error message on bad "exposes"
Here's an example:
```
in file: /Users/oscar/code/zafu/src/Zafu/Cli/Args.bosatsu, package Zafu/Cli/Args
declared `exposes` does not match the exported API.
declared: Zafu/Abstract/Applicative, Zafu/Cli/Args/Types,
Zafu/Collection/HashMap, Zafu/Collection/NonEmptyChain,
Zafu/Collection/NonEmptyList, Zafu/Control/Result, Zafu/Text/Pretty
actual: Zafu/Abstract/Applicative, Zafu/Cli/Args/Types, Zafu/Collection/HashMap,
Zafu/Collection/NonEmptyChain, Zafu/Collection/NonEmptyList,
Zafu/Control/Result, Zafu/Text/Parse/Internal/Core, Zafu/Text/Pretty
canonical fix: exposes Zafu/Abstract/Applicative, Zafu/Cli/Args/Types,
Zafu/Collection/HashMap, Zafu/Collection/NonEmptyChain,
Zafu/Collection/NonEmptyList, Zafu/Control/Result,
Zafu/Text/Parse/Internal/Core, Zafu/Text/Pretty.
missing declarations:
Zafu/Text/Parse/Internal/Core escapes via export `value_from_parse`
```
It would look nice with more better spacing:
```
in file: /Users/oscar/code/zafu/src/Zafu/Cli/Args.bosatsu, package Zafu/Cli/Args
declared `exposes` does not match the exported API.

declared: Zafu/Abstract/Applicative, Zafu/Cli/Args/Types,
    Zafu/Collection/HashMap, Zafu/Collection/NonEmptyChain,
    Zafu/Collection/NonEmptyList, Zafu/Control/Result, Zafu/Text/Pretty

actual: Zafu/Abstract/Applicative, Zafu/Cli/Args/Types, Zafu/Collection/HashMap,
    Zafu/Collection/NonEmptyChain, Zafu/Collection/NonEmptyList,
    Zafu/Control/Result, Zafu/Text/Parse/Internal/Core, Zafu/Text/Pretty

canonical fix: exposes Zafu/Abstract/Applicative, Zafu/Cli/Args/Types,
    Zafu/Collection/HashMap, Zafu/Collection/NonEmptyChain,
    Zafu/Collection/NonEmptyList, Zafu/Control/Result,
    Zafu/Text/Parse/Internal/Core, Zafu/Text/Pretty

missing declarations:
    Zafu/Text/Parse/Internal/Core escapes via export `value_from_parse`
```

So we need to use some `nested(4)` from Doc. We need to add additional hardLine between the groups.

Also, instead of `declared` and `actual` I think it would be better to show the actual region from the file. If we have to reparse the file to find the exposes region and then `showContext` for that region. I think "actual" and "canonical fix" is duplicative. I don't think we need both. Lastly, to make the canonical fix more copy/paste friendly we should use the `()` around the exposes if there is more than one package.

johnynek:
Updates `PackageError.ExposesMismatch` to render clearer blank-line-separated sections, reparse the package header to show the declared `exposes` source region, remove the redundant `declared`/`actual` list blocks, and render multi-package canonical fixes as `exposes (...)`. Adds regression coverage in `ErrorMessageTest`, updates the related CLI warning assertions in `ToolAndLibCommandTest`, and verifies the change with `sbt 'coreJVM/testOnly dev.bosatsu.ErrorMessageTest dev.bosatsu.ToolAndLibCommandTest'` plus `scripts/test_basic.sh`.

Fixes #2266

### DIFF
--- a/core/src/main/scala/dev/bosatsu/PackageError.scala
+++ b/core/src/main/scala/dev/bosatsu/PackageError.scala
@@ -1,14 +1,13 @@
 package dev.bosatsu
 
 import cats.data.{Chain, NonEmptyList, NonEmptyMap}
-import cats.parse.{Parser0 => P0, Parser => P}
+import cats.parse.{Parser => P}
 import cats.syntax.all._
 import org.typelevel.paiges.{Doc, Document}
 import scala.collection.immutable.LongMap
 
 import rankn._
 import LocationMap.Colorize
-import Parser.Combinators
 
 sealed abstract class PackageError {
   def message(
@@ -190,45 +189,11 @@ object PackageError {
   private def sectionDoc(title: String, body: Doc): Doc =
     Doc.text(title) + (Doc.hardLine + body).nested(4)
 
-  private val exposesRegionParser: P0[List[Region]] = {
-    val spaceComment: P0[Unit] =
-      (Parser.spaces.? ~ CommentStatement.commentPart.?).void
-
-    val eol = spaceComment <* Parser.termination
-    val parsePack = Padding
-      .parser(
-        (P.string("package").soft ~ Parser.spaces) *> PackageName.parser <* eol,
-        spaceComment
-      )
-      .void
-    val im = Padding.parser(Import.parser <* eol, spaceComment).void.rep0
-    val ex = Padding
-      .parser(
-        (P.string("export")
-          .soft ~ Parser.spaces) *> ExportedName.parser.itemsMaybeParens
-          .map(_._2) <* eol,
-        spaceComment
-      )
-      .void
-    val exposeItems =
-      PackageName.parser.parensLines0Cut.backtrack.orElse(
-        PackageName.parser.nonEmptyListOfWs(Parser.maybeSpace).map(_.toList)
-      )
-    val exposes = Padding
-      .parser(
-        (
-          (P.index.with1 ~
-            ((P.string("exposes").soft ~ Parser.spaces) *> exposeItems <* eol)) ~
-            P.index
-        )
-          .map { case ((start, _), end) => Region(start, end) },
-        spaceComment
-      )
-      .map(_.padded)
-
-    ((((parsePack ~ im) ~ ex.rep0) ~ exposes.rep0)
-      .map { case (((_, _), _), xs) => xs }) <* P.anyChar.rep0.void
-  }
+  private val exposesRegionParser =
+    (Package.headerPackageNameParser.void *>
+      Package.headerImportsParser.void *>
+      Package.headerExportsParser.void *>
+      Package.headerExposeRegionsParser) <* P.anyChar.rep0.void
 
   private def declaredExposesRegion(source: String): Option[Region] =
     Parser
@@ -2137,9 +2102,8 @@ object PackageError {
           canonicalExposesLineDoc(actual).get + Doc.char('.')
 
       val missingDoc =
-        if (missing.isEmpty) Doc.empty
-        else {
-          val lines = missing.flatMap { dep =>
+        NonEmptyList.fromList(missing).map { missingNel =>
+          val lines = missingNel.toList.flatMap { dep =>
             missingCauses.get(dep).map { exports =>
               val exportWord =
                 if (exports.tail.isEmpty) "export" else "exports"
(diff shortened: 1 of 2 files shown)


core/src/main/scala/dev/bosatsu/PackageError.scala
```diff
@@ -157,12 +158,84 @@ object PackageError {
...
+            ((P.string("exposes").soft ~ Parser.spaces) *> exposeItems <* eol)) ~
+            P.index
+        )
+          .map { case ((start, _), end) => Region(start, end) },
+        spaceComment
+      )
+      .map(_.padded)
+
+    ((((parsePack ~ im) ~ ex.rep0) ~ exposes.rep0)
+      .map { case (((_, _), _), xs) => xs }) <* P.anyChar.rep0.void
```
johnynek: yes this is better as:
```
(parsePack *> im *> ex.rep0 *> exposes.rep0) <* P.anyChar.rep0.void
```
not only for readability, but for performance: non-capturing parsers don't allocate as much.

core/src/main/scala/dev/bosatsu/PackageError.scala
```diff
@@ -157,12 +158,84 @@ object PackageError {
...
+    val exposeItems =
+      PackageName.parser.parensLines0Cut.backtrack.orElse(
+        PackageName.parser.nonEmptyListOfWs(Parser.maybeSpace).map(_.toList)
+      )
+    val exposes = Padding
+      .parser(
+        (
+          (P.index.with1 ~
+            ((P.string("exposes").soft ~ Parser.spaces) *> exposeItems <* eol)) ~
+            P.index
```
johnynek: there is a `.region` combinator in Parser.scala in this repo I beleive that does this so you don't need to reimplement it.

core/src/main/scala/dev/bosatsu/PackageError.scala
```diff
@@ -2074,25 +2156,31 @@ object PackageError {
...
         if (extra.isEmpty) Doc.empty
         else
-          Doc.hardLine + Doc.text("extra declarations: ") + exposesSetDoc(extra)
+          sectionDoc("extra declarations:", exposesSetDoc(extra))
+
+      val sections =
+        List(
+          Some(sectionDoc("declared here:", declaredDoc)),
+          Some(sectionDoc("canonical fix:", fixDoc)),
+          Option.when(missing.nonEmpty)(missingDoc),
```
johnynek: rather than using `missing.nonEmpty` and `missingDoc` just make `missingDoc: Option[Doc]` since it is already checking emptiness of the missing option. Same with extra.

core/src/main/scala/dev/bosatsu/PackageError.scala
```diff
@@ -157,12 +158,84 @@ object PackageError {
...
+    val im = Padding.parser(Import.parser <* eol, spaceComment).void.rep0
+    val ex = Padding
+      .parser(
+        (P.string("export")
+          .soft ~ Parser.spaces) *> ExportedName.parser.itemsMaybeParens
+          .map(_._2) <* eol,
+        spaceComment
+      )
+      .void
+    val exposeItems =
```
johnynek: this is duplicative with the parsers in `headerParser` in Package.scala. I think what would be better is just to lift the parts of the header out as individual parsers in that file (like `def importsParser: P0[List...` etc... and not copy two implementations of all of these.

core/src/main/scala/dev/bosatsu/PackageError.scala
johnynek: Updated this to use non-capturing combinators. `PackageError` now reuses the lifted header parsers from `Package.scala` with `*>` and only keeps the exposes-region results before consuming the rest of the file.

core/src/main/scala/dev/bosatsu/PackageError.scala
johnynek: Switched to the existing `.region` combinator. The new shared `headerExposeRegionsParser` is built from `headerExposeDeclParser.region` in `Package.scala`, so there is no manual `P.index`/`Region` reconstruction here anymore.

core/src/main/scala/dev/bosatsu/PackageError.scala
johnynek: Changed `missingDoc` and `extraDoc` to `Option[Doc]` using `NonEmptyList.fromList(...)`, and the section list now flattens those options directly instead of re-checking emptiness separately.

core/src/main/scala/dev/bosatsu/PackageError.scala
johnynek: Lifted the header pieces into shared parsers in `Package.scala` (`headerPackageNameParser`, `headerImportsParser`, `headerExportsParser`, `headerExposesParser`, and `headerExposeRegionsParser`) and removed the duplicated header parsing logic from `PackageError.scala`.

johnynek:
Addressed the parser-refactor review feedback by centralizing the header sub-parsers in `Package.scala`, reusing the existing `.region` combinator, and simplifying the optional mismatch sections in `PackageError.scala`. Re-ran `sbt 'coreJVM/testOnly dev.bosatsu.ErrorMessageTest dev.bosatsu.ToolAndLibCommandTest'` and `scripts/test_basic.sh`; both passed.

PR #2272, opened 2026-03-28
[codex] Add JVM profiling harness and cache compiler hashes

### LINKED ISSUE #2271: Reduce allocation in the PackageMap resolve/infer pipeline by replacing transformer-heavy orchestration with builder-backed loops
## Summary

After the recent hash-caching wins, `zafu` warm `check` time improved a lot, but the remaining hot path is still spending too much work in allocation-heavy orchestration inside `PackageMap.resolvePackages` and `PackageMap.inferAllSourceUnits`.

This no longer looks like a hashing problem. It looks like a control-flow and data-shape problem.

The main issue is not one single slow algorithm. It is the amount of intermediate structure we build while resolving and inferring packages:

- `ReaderT` over `Either` in `resolvePackages`
- `Memoize.memoizeDag[F, ...]` plus promise-based recursion in `inferAllSourceUnits`
- repeated `foldLeft + parMapN` accumulation over imports
- `IorT` / `Validated` / `Either` wrapping at fine granularity
- `ImportMap.parTraverse` plus `ImportMap.toList` regrouping
- repeated list / tuple / closure allocation around small steps

We should keep the public API pure, but move more of the hot implementation into private, specialized helpers that use direct loops and local builders.

## Evidence

Warm CPU profiles after the hash fixes no longer show structural hashing as the main wall. The remaining signal is mostly GC, allocation, and cats/collection machinery.

Warm alloc profile highlights:

- `scala.collection.immutable.$colon$colon`: about 11.7%
- `scala.util.Right`: about 4.3%
- `cats.data.Validated$Valid`: about 4.1%
- `java.lang.Object[]`: about 3.9%
- `scala.Tuple2`: about 3.3%
- `cats.Later`: about 3.2%
- `cats.data.Kleisli`: about 2.7%
- `cats.Eval` node types combined: several additional percent
- `cats.kernel.instances.ListOrder`: about 2.4%

Warm CPU profile also still shows:

- GC evacuation work
- `DirectMethodHandle.allocateInstance`
- `cats.Eval$.loop$1`
- `cats.data.Kleisli.mapF`
- `Growable.addAll` / `ArrayBuffer.addAll`
- package-name equality / ordering work

That is the profile signature of too much intermediate functional plumbing in a hot loop.

## Hot code paths

The current hot orchestration is concentrated in two places.

`resolvePackages` currently builds package resolution through:

- `ReaderT[Either[NonEmptyList[PackageError], *], List[PackageName], PackageFix]`
- `Memoize.memoizeDagHashed`
- `ReaderT.ask`
- `flatMapF`
- `local`
- `parTraverse`

This is elegant, but it allocates a lot of closures and list wrappers just to carry the import path for cycle reporting.

`inferAllSourceUnits` currently builds package inference through:

- `Memoize.memoizeDag[F, PackageName, Ior[...]]`
- nested `imports.foldLeft(...)`
- repeated `(acc, next).parMapN(...)`
- `IorT`
- `ImportMap.parTraverse`
- `impMap.toList(using Package.orderByName)`
- another `parTraverse` over all packages at the top level

This is pure, but it is layering several abstractions over a workflow that is already known to be a DAG and already has package-level structure.

## Main hypothesis

We are paying a lot of runtime and allocation to express a fairly structured algorithm in a very generic way.

In particular:

- we are representing dependency traversal as general recursive monadic composition instead of a specialized DAG walk
- we are representing small local accumulations as repeated `Either` / `Validated` / `Ior` node construction instead of direct builders
- we are repeatedly converting between import representations instead of keeping the representation needed by the next step
- we are using generic transformers where a private specialized loop would be cheaper and still referentially transparent from the outside

## Proposal

I think we should attack this in three passes.

## 1. Replace `resolvePackages` `ReaderT` with a specialized resolver

We do not need `ReaderT` to keep cycle-path information.

We can write a private resolver that keeps:

- a current DFS stack for error reporting
- a `visiting` set for cycle detection
- a memo table for already-resolved packages

The public function can stay pure and still return `ValidatedNel[PackageError, PackageMap[...]]`.

The implementation can be private and locally imperative:

- `mutable.ArrayBuffer[PackageName]` or small custom stack for the current path
- `mutable.HashSet[PackageName]` for `visiting`
- `mutable.AnyRefMap[PackageName, PackageFix]` or similar for memoized resolved packages
- `ListBuffer[PackageError]` or direct `NonEmptyList` assembly only on failure
- direct loops over imports instead of `ReaderT.ask.flatMapF.local.parTraverse`

This should remove a lot of:

- `ReaderT` instances
- closure allocation
- `pname :: _` path copies
- repeated `List.contains` scans
- tiny `Either` allocations on the happy path

If we want to stay fully immutable even internally, a hand-written recursive function with explicit arguments is still likely cheaper than `ReaderT`. But I suspect the best cost/benefit is a private mutable implementation with a pure boundary.

## 2. Replace `inferAllSourceUnits` recursive memoized `F` graph with topo-order execution

We already prove the graph is a DAG with `Toposort.sort(...)`.

Once we know that, using recursive `Memoize.memoizeDag[F, ...]` with promises is probably the wrong shape for the hot path.

A better structure is:

1. compute a topological order once
2. iterate through that order
3. for each package, read already-computed dependency results from an index
4. compute cache key, cache hit/miss, inference, assembly, finish package
5. store the compiled result for later dependents

This can still be pure at the API boundary. Internally it can use:

- a dense package index or map from `PackageName` to slot
- an array or mutable map of compiled results
- direct loops to gather dependency interfaces and dependency compiled packages
- one `F` boundary per actual effectful step, instead of many fine-grained combinators

This avoids:

- `Memoize.memoizeDag[F, ...]`
- promise allocation and waiting
- nested `parMapN` over each import
- `IorT`
- a lot of `Kleisli` / `Eval` / lambda churn coming from parallel typeclass composition

Parallelism does not have to disappear. If package-level parallelism still matters, we can compute by DAG layers or Kahn frontiers and run each frontier in parallel. The important part is to stop expressing the whole algorithm as recursive monadic graph plumbing.

## 3. Stop rebuilding import structures in the hot path

`ImportMap` is currently optimized for lookup by local name, but later we also need grouped imports by package order.

Right now `ImportMap.toList` effectively:

- iterates the map
- builds a list
- groups by package
- rebuilds `Import` values
- sorts / orders again at the call site

That is extra work in a hot path.

I think we should add a faster representation or at least a faster one-pass conversion.

Options:

- add a private one-pass `toGroupedImports` builder that avoids `toList.groupByNel(...).map(...).toList`
- store a second grouped view inside `ImportMap` when it is constructed
- introduce a specialized `ResolvedImports` type for the inference path that stores exactly what `inferBodyUnopt` and `assemble` need

If we choose a custom data structure, this is the first place I would do it. Not a giant general-purpose replacement, just a structure that preserves:

- lookup by local name
- iteration grouped by package
- stable order for downstream inference

A dual-view `ImportMap` may be worth the extra memory if it saves repeated regrouping and intermediate list allocation.

## What I would implement first

I would not start with a brand new public data structure.

I would start with private helper methods that hide the imperative parts and keep the API pure:

- `resolvePackagesFast(...)`
- `inferAllSourceUnitsTopo(...)`
- `collectDepResults(...)`
- `collectDepInterfaces(...)`
- `resolveImportsFast(...)`
- `groupImportsFast(...)`

Each of these can return the same immutable result types we use today, but use local mutation internally.

This gives us the best chance of preserving style and correctness while removing the specific allocations the profiler is showing.

## Possible custom data structures if the private-helper pass is not enough

If the simpler pass is not enough, the next structural ideas I would consider are:

- `PackageId`
- a dense per-run integer id for each package name
- arrays or vectors indexed by `PackageId` instead of repeated `Map[PackageName, ...]` lookups

and:

- a dual-index import structure
- one index by local imported name
- one grouped-by-package sequence for inference / assembly

`PackageId` is probably the highest-ceiling optimization, but it is more invasive. I would treat it as phase 2.

## Purity / style

I do not think this requires abandoning pure FP style.

The important distinction is:

- pure public interface
- locally optimized private implementation

It is still referentially transparent if a private method uses a builder, `ArrayBuffer`, or mutable map internally and only returns an immutable result.

That is very different from spreading mutation through the compiler. We can keep the API and semantics clean while using lower-overhead machinery in the hot path.

The current profile suggests that generic transformer layering is costing us more than it is helping here.

## Success criteria

We should consider this successful if warm profiles show a meaningful drop in:

- `scala.collection.immutable.$colon$colon`
- `scala.Tuple2`
- `scala.util.Right`
- `cats.data.Validated$Valid`
- `cats.data.Kleisli`
- `cats.Eval` node allocations
- `DirectMethodHandle.allocateInstance`
- GC evacuation time

and if steady-state warm checks on a large repo like `zafu` become less GC-sensitive and more stable.

## Suggested implementation order

1. Add a one-pass fast grouped-import builder and remove `ImportMap.toList` regrouping from the inference path.
2. Replace the `imports.foldLeft(...).parMapN(...)` accumulation in `inferAllSourceUnits` with direct private helpers and builders.
3. Replace `IorT` orchestration in the inference path with direct pattern matching and a single return wrapper.
4. Replace `resolvePackages` `ReaderT` with a specialized resolver.
5. If needed after reprofiling, introduce topo-order inference execution and then consider `PackageId`.

## Open question

The biggest design question is whether we want to keep package-level parallelism exactly as-is, or whether we are willing to trade some of that generality for much lower allocation. My guess is that a topo-level or frontier-based executor will be a better balance than the current recursive promise-based graph.

johnynek:
## Summary

This PR adds a dev-only JVM profiling harness for `check` and lands the compiler changes that came out of profiling `zafu`.

The main wins in the compiler were from reducing structural hash work on hot compiler objects and shifting `PackageName` equality and ordering onto its cached canonical string. The harness made it possible to get repeatable warm JVM profiles with `async-profiler` against the in-process compiler.

Closes #2271.

## What Changed

- added `dev.bosatsu.CheckBenchmarkMain` as a dev-only JVM entrypoint for repeated `check` runs and `async-profiler` attachment
- cached hash codes on `ImportedName`, `Import`, `ImportMap`, `SourceUnit`, and moved `Package` hash caching off eager construction
- switched `PackageMap` memoizers to key by `PackageName` instead of full `Package` values
- updated `PackageName` equality, hash, and ordering to use its cached canonical string
- added a property test proving the new `PackageName` ordering agrees with the legacy parts-based ordering for valid package names

## Why

The warm profiles on `zafu` showed that the compiler was spending a large fraction of time in structural hashing and immutable collection traversal before we even got to the real inference work. After the hash-caching changes, the hot path shifted toward allocation-heavy orchestration in `PackageMap`, which is captured in #2271.

This PR lands the profiler tooling plus the low-risk wins we already validated from that profiling pass.

## Validation

- `./scripts/test_basic.sh`
- `sbt cli/assembly`
- manual JVM harness runs against `zafu`
- manual `async-profiler` captures on the warm loop harness

### DIFF
--- a/core/src/test/scala/dev/bosatsu/TypedExprTest.scala
+++ b/core/src/test/scala/dev/bosatsu/TypedExprTest.scala
@@ -3944,28 +3944,74 @@ x = Foo
     (unoptimizedExpr, normalizedExpr)
   }
 
-  test("if matches normalizes to same code as equivalent match") {
-    def normalizedFromPackage(packSrc: String): TypedExpr[Unit] =
-      Par.withEC {
-        var out: Option[TypedExpr[Unit]] = None
-        TestUtils.testInferred(
-          List(packSrc),
-          "Test",
-          { (pm, mainPack) =>
-            val pack = pm.toMap(mainPack)
-            val fExpr = pack.lets.find(_._1 == Identifier.Name("f")) match {
-              case Some((_, _, te)) => te
-              case None => fail(s"missing let f in ${pack.lets.map(_._1)}")
-            }
-            val normalized =
-              TypedExprNormalization.normalize(fExpr).getOrElse(fExpr).void
-            out = Some(normalized)
-          }
-        )
-        out.getOrElse(fail("failed to infer normalized expression for f"))
+  def inferLoweredAndNormalizedExpr(
+      statement: String,
+      letName: String
+  ): (TypedExpr[Declaration], TypedExpr[Declaration]) = {
+    val stmts = Parser.unsafeParse(Statement.parser, statement)
+    val (fullTypeEnv, unoptProgram) =
+      Package.inferBodyUnopt(
+        TestUtils.testPackage,
+        predefResolvedImport :: Nil,
+        Nil,
+        stmts
+      ) match {
+        case cats.data.Ior.Right(res) =>
+          res
+        case cats.data.Ior.Both(errs, _) =>
+          fail(s"inference failure:\n${errs.toList.mkString("\n")}")
+        case cats.data.Ior.Left(errs) =>
+          fail(s"inference failure:\n${errs.toList.mkString("\n")}")
       }
 
-    val ifNormalized = normalizedFromPackage(
+    val targetName = Identifier.Name(letName)
+    val loweredLets = TypedExprLoopRecurLowering.lowerAll(unoptProgram.lets)
+    val loweredExpr = loweredLets.find(_._1 == targetName) match {
+      case Some((_, _, te)) => te
+      case None             =>
+        fail(s"missing lowered let: $letName in ${loweredLets.map(_._1)}")
+    }
+    val normalizedLets =
+      TypedExprNormalization.normalizeAll(
+        TestUtils.testPackage,
+        loweredLets,
+        fullTypeEnv
+      )
+    val normalizedExpr = normalizedLets.find(_._1 == targetName) match {
+      case Some((_, _, te)) => te
+      case None             =>
+        fail(s"missing normalized let: $letName in ${normalizedLets.map(_._1)}")
+    }
+    (loweredExpr, normalizedExpr)
+  }
+
+  private def normalizedLetFromPackage(
+      packSrc: String,
+      letName: String
+  ): TypedExpr[Unit] =
+    Par.withEC {
+      var out: Option[TypedExpr[Unit]] = None
+      TestUtils.testInferred(
+        List(packSrc),
+        "Test",
+        { (pm, mainPack) =>
+          val pack = pm.toMap(mainPack)
+          val target = Identifier.Name(letName)
+          val te = pack.lets.find(_._1 == target) match {
+            case Some((_, _, expr)) => expr
+            case None               =>
+              fail(s"missing let $letName in ${pack.lets.map(_._1)}")
+          }
+          out = Some(TypedExprNormalization.normalize(te).getOrElse(te).void)
+        }
+      )
+      out.getOrElse(
+        fail(s"failed to infer normalized expression for $letName")
+      )
+    }
+
+  test("if matches normalizes to same code as equivalent match") {
+    val ifNormalized = normalizedLetFromPackage(
       """
 package Test
 
@@ -3974,10 +4020,11 @@ enum E: Left(l), Right(r)
 def f(x):
   if x matches Left(_): 1
   else: -1
-"""
+""",
+      "f"
     )
 
-    val matchNormalized = normalizedFromPackage(
+    val matchNormalized = normalizedLetFromPackage(
       """
 package Test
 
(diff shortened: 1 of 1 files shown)


core/src/main/scala/dev/bosatsu/PackageName.scala
```diff
@@ -10,7 +10,16 @@ import org.typelevel.paiges.{Doc, Document}
...
+  final override def equals(that: Any): Boolean =
+    that match {
+      case pn: PackageName =>
+        (this eq pn) || (asString == pn.asString)
+      case _ => false
+    }
+
+  final override val hashCode: Int = asString.hashCode
 
   def isPredef: Boolean = parts === PackageName.PredefName.parts
```
johnynek: shouldn't this just be `equals(PackageName.PredefName)`?

Issue #2274, opened 2026-03-29
Add Matchless support to show


johnynek:
Currently `show` command cannot show the matchless version of the code. It would be nice for debugging purposes to be able to see what packages lower to in the Matchless AST and access that via show.

We should have the same options: edn or json output, and the edn should look as good as possible. Ideally we could make the syntax of the matchless written as edn as close to clojure or something as we can, within the constraint that it also look like the lowered typedexpr. Of course, the matchless code intentionally has no types, so we wouldn't see that.

This should be an option to show to select what we are are outputting. It should continue to default to the TypeExpr nodes, but we should be able to opt into matchless.

Also, we should be able to turn off individually any TypedExpr optimizations as AND matchless optimizations. This will help us debug the optimizers running at different layers of the code.

johnynek:
MergeXO assigned an agent and started design work for issue #2274.

johnynek:
MergeXO design flow could not push because the required pre-push test `scripts/test_basic.sh` failed.
Failure summary: Command failed cmd: /Users/oscar/.local/share/mergexo/checkouts/johnynek/bosatsu/worker-03/scripts/test_basic.sh exit: 1 stdout: [info] welcome to sbt 1.12.8 (Homebrew Java 25.0.2) [info] loading settings for project worker-03-build from...

johnynek:
MergeXO pre-PR flow is waiting for source-issue follow-up.
- blocked reason: required pre-push tests failed before pushing design branch
- checkpoint branch: `agent/design/2274-add-matchless-support-to-show`
- checkpoint commit: `582ab5c59ab680e0fc0a9c807de46717f72b3b6c`
- checkpoint tree: https://github.com/johnynek/bosatsu/tree/582ab5c59ab680e0fc0a9c807de46717f72b3b6c
- compare vs `main`: https://github.com/johnynek/bosatsu/compare/main...582ab5c59ab680e0fc0a9c807de46717f72b3b6c

Next steps:
1. Review the checkpoint tree/compare links above.
2. Reply on this issue with clarifications or updated instructions.
3. MergeXO will resume from this checkpoint branch head.

johnynek:
/mergexo unblock

please re-run the tests, I suspect there was a flakey test

johnynek:
Please retry, I think we have fixed the bugs.

johnynek:
MergeXO assigned an agent and started design work for issue #2274.

johnynek:
[...]

PR #2276, opened 2026-03-29
Design doc for #2274: Design matchless support and pass controls for `show`

johnynek:
Design doc.

Refs #2274

### DIFF
--- a/docs/design/2274-add-matchless-support-to-show.md
+++ b/docs/design/2274-add-matchless-support-to-show.md
@@ -10,6 +10,7 @@ touch_paths:
   - core/src/main/scala/dev/bosatsu/cache/InferPhases.scala
   - core/src/main/scala/dev/bosatsu/cache/CompileCache.scala
   - core/src/main/scala/dev/bosatsu/Matchless.scala
+  - core/src/main/scala/dev/bosatsu/MatchlessGlobalInlining.scala
   - core/src/main/scala/dev/bosatsu/MatchlessFromTypedExpr.scala
   - core/src/test/scala/dev/bosatsu/ToolAndLibCommandTest.scala
   - core/src/test/scala/dev/bosatsu/tool/ShowEdnRoundTripTest.scala
@@ -32,6 +33,8 @@ Add `--ir matchless` plus explicit typedexpr and Matchless pass toggles to both 
 
 `show` currently renders selected `Package.Typed[Any]` values through `ShowEdn`. Library `show` exposes only a coarse `--no-opt` switch, `tool show` has no equivalent, and both commands can only emit typedexpr output. Matchless lowering already exists in `Matchless.fromLet` and `MatchlessFromTypedExpr.compile`, and the lowering path already applies two cleanup passes by default: `hoistInvariantLoopLets` and `reuseConstructors`.
 
+The Matchless roadmap now also includes a namespace-level global inlining phase. If `show` is meant to debug Matchless optimizations rather than only raw lowering, it needs to be able to show packages before and after that inliner as well.
+
 Typed post-typecheck work is similarly coarse. `InferPhases.default.finishPackage` currently treats optimization as one boolean and, when enabled, runs three ordered stages:
 
 1. `TypedExprLoopRecurLowering.lowerProgram`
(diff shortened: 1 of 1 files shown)


docs/design/2274-add-matchless-support-to-show.md
```diff
@@ -0,0 +1,201 @@
...
+3. `Package.discardUnused`
+
+Because compiled artifacts are cached, cache identity must distinguish typed-pass profiles. `CompileCache.keyHashValue` cannot continue to hash only `optimize` once the pass model expands.
+
+### 3. Make Matchless lowering and cleanup selectable
+
+Refactor Matchless lowering into explicit stages:
+
+1. raw typedexpr-to-Matchless lowering
+2. ordered Matchless cleanup passes
```
johnynek: there is also the new global matchless inliner. Seeing the result of packages after that has run is very interesting as well, since it can help us debug that inliner. Please discuss how we can support this.

docs/design/2274-add-matchless-support-to-show.md
```diff
@@ -0,0 +1,201 @@
...
+
+This keeps `--package`, `--type`, `--value`, and `--externals` semantics stable even when later Matchless passes inline or restructure expression bodies. The existing imported-interface fallback in `MatchlessFromTypedExpr.compile` is sufficient for selected-package lowering, so `show` does not need a separate runtime-only namespace model.
+
+### 5. Extend the `show` schema rather than inventing a second serializer
+
+Keep one logical `show` schema and make it self-describing. The top-level wrapper should continue to be `show`, but add IR metadata:
+
+1. `:ir` with value `typedexpr` or `matchless`
+2. `:typed-passes` with the enabled typed pass names
+3. `:matchless-passes` with the enabled Matchless pass names
```
johnynek: this should only be present when we are emitting matchless.

docs/design/2274-add-matchless-support-to-show.md
```diff
@@ -0,0 +1,201 @@
...
+4. the existing `:interfaces` and `:packages`
+
+Package-level metadata such as imports, exports, defined types, and externals should stay structurally familiar. The main IR-dependent field is `:defs`.
```
johnynek: Matchless doesn't have types or interfaces exactly. Comment on this. For matchless we only really have values which are untyped. We can sometimes see that they are functions when they are top level lambdas, but not always.

docs/design/2274-add-matchless-support-to-show.md
```diff
@@ -0,0 +1,201 @@
...
+
+For Matchless EDN rendering:
+
+1. prefer the same s-expression style already used by `ShowEdn`
+2. omit typedexpr-only type annotations
+3. render anonymous and mutable locals with stable synthetic names so output is readable and diffable
```
johnynek: how will we do this exactly? They share the same namespace on allocation, so the numbering can be used to distinguish them.

docs/design/2274-add-matchless-support-to-show.md
johnynek: Added this to the design. The Matchless pass model now includes an optional namespace-level `global-inlining` stage, and the selection section now explains how `show` can support it: build a temporary namespace rooted at the selected defs, include reachable implementation packages for the inliner, run the inliner there, then trim rendering back to the original requested packages/defs. I also added `MatchlessGlobalInlining.scala` to `touch_paths` and a regression item for pre/post-inline `show --ir matchless` output.

docs/design/2274-add-matchless-support-to-show.md
johnynek: Updated the schema section so `:matchless-passes` is only present when `:ir matchless`.

docs/design/2274-add-matchless-support-to-show.md
johnynek: Clarified this in the schema section. The doc now says the Matchless view is explicitly a value-only, untyped view: typed `:interfaces` and `:types` are omitted, and top-level defs may sometimes reveal lambda/function shape structurally but not type/interface information.

docs/design/2274-add-matchless-support-to-show.md
johnynek: Made the naming rule explicit. The doc now proposes rendering `LocalAnon(n)` as `anon$n` and `LocalAnonMut(n)` as `mut$n`, reusing the shared allocator id directly rather than inventing a second numbering scheme.

johnynek:
Updated the design doc to cover Matchless global inlining as a debuggable `show` stage, narrow the Matchless schema to a value-only view, and specify the exact allocator-based naming rule for Matchless locals. `scripts/test_basic.sh` passed locally.

Issue #2278, opened 2026-03-29
Remove and repurpose literal dict syntax


johnynek:
Currently we support literal dict syntax mostly because we wanted to be more Python like. That said because we need to be parametric, the literal syntax requires String keys. As a result they are very rarely used in practice (almost never). Typed programming prefers custom records to dictionaries.

Thus we have syntax that isn't really "paying its way".

What else could we do with the `{}` syntax? Make a plan for removing dict and dict comprehensions.

One idea could be to support anonymous records, but that would be type system addition that may be considerable.

Another idea would be to infer the constructor name if it is fully unambiguous from the field names given. Evaluating the options give a recommendation in the design.

johnynek:
MergeXO assigned an agent and started design work for issue #2278.

johnynek:
Opened design PR: https://github.com/johnynek/bosatsu/pull/2279

johnynek:
reusing the syntax now seems like a bad idea. I think we should either remove and not use the syntax or leave it as is. Closing for now.

PR #2279, opened 2026-03-29
Design doc for #2278: Design: Remove Literal Dict Syntax And Reuse Bare Braces For Inferred Records

johnynek:
Design doc.

Refs #2278


docs/design/2278-remove-and-repurpose-literal-dict-syntax.md
```diff
@@ -0,0 +1,258 @@
...
+
+Anonymous records are much larger. `rankn.Type` and `TypeRef` currently model nominal type constructors, type application, and quantifiers. They do not have a structural record type, row variable, or field-projection concept. Adding anonymous records would therefore spill into parsing, type checking, interface serialization, pretty-printing, and likely pattern semantics.
+
+There is also a migration concern. The repo already has old dict code that uses bindable keys such as `{k: v}`, not only string literals. A one-step change from “dict literal” to “inferred record” would reparse some existing programs into a different syntax class and produce confusing failures. The safest plan is staged: remove and migrate first, repurpose second.
+
+## Goals
+1. Remove literal dict and dict-comprehension surface syntax from Bosatsu source.
+2. Keep `Dict` as a supported library type with no runtime or `ValueToJson` behavior change.
+3. Reuse bare braces for a feature that aligns with Bosatsu's nominal typing model.
+4. Provide an explicit migration surface and targeted diagnostics for old dict syntax.
```
johnynek: We don't need this. We have few enough users that they can manage all in one go when updating the compiler.

docs/design/2278-remove-and-repurpose-literal-dict-syntax.md
```diff
@@ -0,0 +1,258 @@
...
+2. No empty `{}` form.
+3. No pattern form; `Foo { ... }` remains the only record-pattern spelling.
+4. No default-based omission during inference.
+
+### 6. Resolve inferred records in `SourceConverter` by exact field-set match
+Inference should happen in `SourceConverter`, where constructor metadata already exists.
+
+Resolution algorithm:
+1. Build a field-set index over visible constructors in `localTypeEnv`.
+2. Use only constructors whose full declared field-name set exactly matches the explicit field-name set written in the literal.
```
johnynek: This conflicts with default arguments. I think we should allow giving partial arguments if they are still unambiguous.

docs/design/2278-remove-and-repurpose-literal-dict-syntax.md
```diff
@@ -0,0 +1,258 @@
...
+3. If there is exactly one candidate, synthesize the equivalent explicit `RecordConstructor` and reuse the current lowering path.
+4. If there are zero candidates, emit a new source-converter error that no visible constructor matches the field set.
+5. If there are multiple candidates, emit a new ambiguity error and require the user to write the constructor name explicitly.
+
+Why exact field-set match:
+1. It keeps inference purely name-based, with no type-directed behavior.
+2. It avoids hidden interaction with constructor defaults.
+3. It keeps the feature easy to explain: bare braces mean “the unique visible constructor with exactly these fields”.
```
johnynek: I think "bare braces are allowed if the given fields are only defined by one in scope constructor".

That said, default arguments are pretty rare currently so maybe requiring all of the fields in practice isn't a problem.

docs/design/2278-remove-and-repurpose-literal-dict-syntax.md
johnynek: Updated the design to remove the staged rollout and dedicated migration-surface requirement. The doc now treats this as a single compiler change and reframes the goal around keeping the brace-inference rule simple with clear errors.

docs/design/2278-remove-and-repurpose-literal-dict-syntax.md
johnynek: Adjusted the inference rule here. Bare braces now select the unique visible constructor compatible with the provided fields, and then reuse the existing explicit-record validation so omitted fields can still be filled from defaults.

docs/design/2278-remove-and-repurpose-literal-dict-syntax.md
johnynek: Reworded the recommendation and resolution algorithm around that criterion: the provided fields must identify exactly one visible constructor. The doc no longer requires exact field-set equality, and the testing/acceptance sections now cover partial arguments plus default handling.

johnynek:
One serious concern I have with constructor name inference is that currently we go to lengths to make it possible to see where everything is coming from when we read a source file. We have no glob imports. You can see what the source of every name is.

But if we have constructor inference for constructors defined outside the current file, then someone can write a record and it not be clear why it compiles or where the fields are defined. The compiler will know, but a reader of a local file won't.

That's against the design of bosatsu packages. They have a strong local context property.

Now, we could allow inference only for local constructors. But that then creates a difference between how records are treated. Maybe that's okay. I don't know. But if we allow inference of non-local constructors, why not allow global imports of all constructor names of a type. The reason we don't is to make the file more readable. This works against that.

Of course, type inference in general works against that but labeling every type is also painful. So there is some balance. Maybe constructor inference for local constructors is that balance. I don't know yet. It would be good to explore the pro and con arguments.

johnynek:
Updated the design doc to address the local-context concern directly. It now adds a local-only vs imported-constructor analysis, recommends limiting bare-brace inference to constructors defined in the current source file, and updates the recommendation, algorithm, acceptance criteria, risks, rollout notes, and test plan around that rule. `scripts/test_basic.sh` passed locally (`Passed: Total 67, Failed 0, Errors 0, Passed 67`).

johnynek:
Blocked on the required pre-push check after the merge-only refresh. The checkout is clean, but `scripts/test_basic.sh` still fails in the full suite with `dev.bosatsu.codegen.clang.ClangGenTest.test_workspace/Ackermann.bosatsu` timing out after the test's 30s limit (`40.189s` wall time in the failing run). I reran that test in isolation on the warmed checkout and it passed (`Passed: Total 1, Failed 0, Errors 0`), but the full required script still reproduces the timeout, so I cannot return a ready-to-push commit message from this turn.

johnynek:
Blocked on the required pre-push gate after the merge-only refresh. The checkout is clean, but `scripts/test_basic.sh` has not completed successfully on this worker: it reaches the CLI test phase, emits the expected `fib(20) = 10946` lines, and then never returns a final SBT footer. The generated CLI XML reports are all passing, but there is no corresponding `coreJVM` test-report/stream output yet, so the hang appears to be before or during the handoff from `cli/testOnly` to `coreJVM/testOnly`. Because the required script has not actually finished, I cannot return a ready-to-push commit message from this turn.

johnynek:
Blocked on the required pre-push gate after the merge-only refresh. The authoritative diff remains within the allowed design-doc artifact set (`docs/design/2278-remove-and-repurpose-literal-dict-syntax.md` only), but `scripts/test_basic.sh` failed locally in `coreJVM/testOnly` with four 30s timeout failures unrelated to the design doc: `dev.bosatsu.codegen.clang.ClangGenLibraryDepsTest.clang gen resolves globals across library deps`, `dev.bosatsu.Issue2087Test.issue 2087: lazy thunks keep delayed opaque work inside Unit lambdas`, `dev.bosatsu.ToolAndLibCommandTest.tool assemble and tool extract-iface`, and `dev.bosatsu.ToolAndLibCommandTest.eval/json/show use library context`. Final summary: `Failed: Total 2100, Failed 4, Errors 0, Passed 2096, Ignored 2`. The selected gate only permits the design doc path, so I cannot repair these test timeout issues in this turn.

johnynek:
Blocked on the required pre-push gate after the merge-only refresh. The authoritative diff remains within the allowed design-doc artifact set (`docs/design/2278-remove-and-repurpose-literal-dict-syntax.md` only), but `scripts/test_basic.sh` failed locally in `coreJVM/testOnly` with a 30s timeout in `dev.bosatsu.MatchlessInterfaceTest.Matchless can compile constructors from imported interfaces` (`31.487s`). Final summary: `Failed: Total 2100, Failed 1, Errors 0, Passed 2099, Ignored 2`. The selected gate only permits the design doc path, so I cannot repair this test timeout in this turn.

johnynek:
Blocked on the required pre-push gate after the merge-only refresh. The authoritative diff remains within the allowed design-doc artifact set (`docs/design/2278-remove-and-repurpose-literal-dict-syntax.md` only), but `scripts/test_basic.sh` failed locally in `coreJVM/testOnly` with two 30s timeout failures unrelated to the design doc: `dev.bosatsu.MatchlessRegressionTest.polymorphic recursion lowers to while in Matchless without self-calls` (`45.276s`) and `dev.bosatsu.ShadowedBindingTypeCheckTest.top-level shadowing by locals is allowed` (`40.039s`). Final summary: `Failed: Total 2100, Failed 2, Errors 0, Passed 2098, Ignored 2`. The selected gate only permits the design doc path, so I cannot repair these test timeout issues in this turn.

PR #2280, opened 2026-03-29
Add Matchless IR support and pass controls to show

### LINKED ISSUE #2274: Add Matchless support to show

johnynek:
Implemented issue #2274 per the merged design doc across both library and tool `show` flows. The change adds `--ir typedexpr|matchless`, repeatable typed and Matchless pass-disable flags, a shared show request/output model, Matchless EDN/JSON rendering with value-only output and pass metadata, explicit typed-pass compile options with cache-key separation, and optional Matchless global inlining for `show` while preserving existing defaults. It also updates docs and adds regressions for typed-pass toggles, Matchless pass toggles, global inlining, EDN/JSON round-tripping, cache identity, and the affected CLI/UI call sites. Validation completed with focused suites, `sbt "cli/test"`, `sbt "doc; paradox"`, `./test_cli.sh`, and the required `scripts/test_basic.sh`.

Fixes #2274

Implements design doc: [docs/design/2274-add-matchless-support-to-show.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/2274-add-matchless-support-to-show.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/2276

### DIFF
--- a/core/src/main/scala/dev/bosatsu/library/Command.scala
+++ b/core/src/main/scala/dev/bosatsu/library/Command.scala
@@ -1836,44 +1836,40 @@ object Command {
         "show fully type-checked packages from this library or dependency tree (EDN by default; JSON with --json)"
       ) {
         import ShowSelection.{typeArgument, valueArgument}
+        import ShowSupport.{matchlessPassArgument, showIrArgument, typedPassArgument}
 
         val irOpt =
           Opts
-            .option[String](
+            .option[Output.ShowIr](
               "ir",
               help = "which IR to render: typedexpr or matchless"
             )
-            .mapValidated(ShowSupport.parseIr)
             .withDefault(Output.ShowIr.TypedExpr)
 
         val disableTypedPassOpt =
           Opts
-            .options[String](
+            .options[CompileOptions.TypedPass](
               "disable-typed-pass",
               help =
                 "disable a typed pass: loop-recur-lowering, normalize, discard-unused"
             )
             .orEmpty
-            .mapValidated(ShowSupport.parseDisabledTypedPasses)
+            .map(_.toSet)
 
         val disableMatchlessPassOpt =
           Opts
-            .options[String](
+            .options[dev.bosatsu.Matchless.Pass](
               "disable-matchless-pass",
               help =
                 "disable a Matchless pass: hoist-invariant-loop-lets, reuse-constructors, global-inlining"
             )
             .orEmpty
-            .mapValidated(ShowSupport.parseDisabledMatchlessPasses)
+            .map(_.toSet)
 
-        def defsInOriginalOrder(
+        def defsInOriginalOrder[A](
             pack: dev.bosatsu.Package.Typed[Any],
-            lets: List[
-              (dev.bosatsu.Identifier.Bindable, dev.bosatsu.Matchless.Expr[?])
-            ]
-        ): List[
-          (dev.bosatsu.Identifier.Bindable, dev.bosatsu.Matchless.Expr[?])
-        ] = {
+            lets: List[(dev.bosatsu.Identifier.Bindable, A)]
+        ): List[(dev.bosatsu.Identifier.Bindable, A)] = {
           val byName = lets.toMap
           pack.lets.flatMap { case (name, _, _) =>
             byName.get(name).map(name -> _)
--- a/core/src/main/scala/dev/bosatsu/tool/ShowSupport.scala
+++ b/core/src/main/scala/dev/bosatsu/tool/ShowSupport.scala
@@ -2,6 +2,7 @@ package dev.bosatsu.tool
 
 import cats.data.{Validated, ValidatedNel}
 import cats.syntax.all._
+import com.monovore.decline.Argument
 import dev.bosatsu.{
   CompileOptions,
   ExportedName,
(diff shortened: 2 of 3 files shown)


johnynek:
Reproduced the failing `testJS` check locally from the GitHub Actions context. The break was not in `coreJS` or `jsapiJS`; it was a stale matcher in `jsui/src/test/scala/dev/bosatsu/jsui/StoreTest.scala` still expecting the old `Output.ShowOutput(_, _, _)` shape after the `ShowOutput(show, outputPath)` refactor. Updated that test to the new two-argument form and revalidated with `sbt "jsuiJS/testOnly * -- --log=failure"` and `scripts/test_basic.sh`, both passing locally.

johnynek:
please read and action all the comments made by gemini. These are good recommendations.

johnynek:
Applied the actionable Gemini recommendations on this branch. `core/src/main/scala/dev/bosatsu/Matchless.scala` now defines both `LocalPass` and `Pass` as Scala 3 enums, removing the manual `values`/`CanEqual` boilerplate while preserving the existing CLI-name helpers. In `core/src/main/scala/dev/bosatsu/tool/ShowEdn.scala`, the import encoder now has a one-argument overload so the default `includeTypes = true` path can use the method-group form `normalized.imports.map(encodeImportForShow)` without breaking the explicit `includeTypes = false` call site. Verified with `sbt "coreJVM/test:compile" "coreJVM/testOnly dev.bosatsu.MatchlessTests dev.bosatsu.tool.ShowEdnRoundTripTest -- --log=failure"` and the required `scripts/test_basic.sh`, both passing locally.

johnynek:
Remove all use of default arguments. Just explicitly pass arguments to any methods and functions. This helps us more precisely deal with refactors.

core/src/main/scala/dev/bosatsu/library/Command.scala
```diff
@@ -1833,6 +1837,92 @@ object Command {
       ) {
         import ShowSelection.{typeArgument, valueArgument}
 
+        val irOpt =
+          Opts
+            .option[String](
+              "ir",
+              help = "which IR to render: typedexpr or matchless"
+            )
+            .mapValidated(ShowSupport.parseIr)
```
johnynek: Don't use mapValidated. It doesn't give help. Use an Argument and define it for the two cases we have.

core/src/main/scala/dev/bosatsu/library/Command.scala
```diff
@@ -1833,6 +1837,92 @@ object Command {
...
+
+        val disableTypedPassOpt =
+          Opts
+            .options[String](
+              "disable-typed-pass",
+              help =
+                "disable a typed pass: loop-recur-lowering, normalize, discard-unused"
+            )
+            .orEmpty
+            .mapValidated(ShowSupport.parseDisabledTypedPasses)
```
johnynek: Use Argument, not mapValidated. It gives better help.

core/src/main/scala/dev/bosatsu/library/Command.scala
```diff
@@ -1833,6 +1837,92 @@ object Command {
...
+              help =
+                "disable a Matchless pass: hoist-invariant-loop-lets, reuse-constructors, global-inlining"
+            )
+            .orEmpty
+            .mapValidated(ShowSupport.parseDisabledMatchlessPasses)
+
+        def defsInOriginalOrder(
+            pack: dev.bosatsu.Package.Typed[Any],
+            lets: List[
+              (dev.bosatsu.Identifier.Bindable, dev.bosatsu.Matchless.Expr[?])
```
johnynek: Don't use ? Here. Use a type parameter on the method since we preserve the type parameter.

johnynek:
Removed default-argument usage from the PR’s Matchless/show changes and from the remaining helper/test code in the files touched by this turn. The updated call sites now pass `Matchless.LocalPassOptions.Default`, `CompileOptions.Mode.Emit`, `packageNamesOnly = false`, `quiet = false`, and explicit helper arguments directly, so the modified files no longer rely on default params. Revalidated with `sbt "coreJVM/test:compile" "coreJVM/testOnly dev.bosatsu.tool.ShowEdnRoundTripTest dev.bosatsu.tool.CompileCacheTest -- --log=failure"` and the required `scripts/test_basic.sh`, both passing locally.

core/src/main/scala/dev/bosatsu/library/Command.scala
johnynek: Replaced the `--ir` string parsing path with a shared `Argument[Output.ShowIr]` in `ShowSupport`, then switched both show entry points to `Opts.option[Output.ShowIr]`. That keeps the accepted values visible to Decline help instead of hiding them behind `mapValidated`.

core/src/main/scala/dev/bosatsu/library/Command.scala
johnynek: Switched both `--disable-typed-pass` and `--disable-matchless-pass` to `Opts.options[...]` backed by `Argument` instances in `ShowSupport`, then convert the parsed lists to sets with `.map(_.toSet)`. This removes the `mapValidated` handling and gives better help/error output.

core/src/main/scala/dev/bosatsu/library/Command.scala
johnynek: Made `defsInOriginalOrder` generic as `defsInOriginalOrder[A](..., lets: List[(Bindable, A)])`, so the method preserves the value type directly and no longer uses `?`.

johnynek:
Updated `show` option parsing to use Decline `Argument` instances shared through `ShowSupport`, and made `defsInOriginalOrder` generic in both command paths. Verified with `sbt "coreJVM/test:compile"`, `sbt "coreJVM/testOnly dev.bosatsu.ToolAndLibCommandTest -- --log=failure"`, and the required `scripts/test_basic.sh`, all passing locally.

PR #2319, opened 2026-04-09
Teach ClangGen to direct-call top-level function aliases

### LINKED ISSUE #2318: C backend misses direct-call optimization for top-level function aliases (operator wrappers emit read_or_build/call_fn in hot loops)
## Summary

The C backend is missing a direct-call optimization for top-level function aliases.

In practice, this means wrappers like:

```bosatsu
def operator +(a, b): add_Int64(a, b)
def addf(a, b): addf_predef(a, b)
```
or
```bosatsu
operator + = add_Int64
operator - = sub_Int64
```

can survive into generated C as first-class function values instead of direct calls. Once that happens, the generated code goes through:

- `static _Atomic BValue ...`
- `read_or_build(...)`
- `alloc_boxed_pure_fn$arity(...)`
- `call_fn$arity(...)`

instead of directly calling the external/null-capture function.

This shows up very clearly in `bosatsu_c` on numeric kernels like spectral norm, where the hot loop still spends a lot of time in `call_fn2`, `call_fn3`, and `read_or_build` even after using the new Array APIs.

## Why This Matters

I was profiling `bosatsu_c` spectral norm after upgrading `zafu` to Bosatsu `0.0.65` and using the new Array helpers (`foldl_with_index_Array`, `dotf_Array`, `sumsqf_Array`, etc.).

That improved the benchmark materially, but the current `bosatsu_c` binary is still much slower than the checked-in C baseline because the O(N^2) inner kernel is still dispatch-heavy.

On my local run:

- `bosatsu_c` spectral norm median: `33.850s`
- checked-in C median: `1.286s`

The relevant `sample` output for `bosatsu_c` shows the top hot frames are still:

- `call_fn2`
- `read_or_build`
- `Bosatsu.Num.Int64.add__Int64`
- `Bosatsu.Num.Int64.mul__Int64`
- `Bosatsu.Num.Int64.div__Int64`
- `Bosatsu.Collection.Array.foldl_with_index_Array`

The final `dotf_Array` / `sumsqf_Array` reductions are already compiled as tight vectorized loops, so they are not the main problem.

### 1. `operator +` is already reduced to a direct alias in Matchless

Using:

```sh
./bosatsu show \
  --package Bosatsu/Num/Int64 \
  --value 'Bosatsu/Num/Int64::operator +' \
  --ir matchless \
  --json \
  --disable-matchless-pass global-inlining
```

I get:

```json
["def", "operator +", ["global", "Bosatsu/Num/Int64", "add_Int64"]]
```

So the source-level wrapper lambda is already gone by this point. The IR is a top-level function alias.

### 2. Consumer code still calls the alias, not the underlying external

In a profiled consumer (`Zafu/Benchmark/Game/SpectralNorm::eval_A`), Matchless still contains:

```json
["app", ["global", "Bosatsu/Num/Int64", "operator +"], "i", "j"]
```

and similarly for `operator *`, `operator /`, and `Bosatsu/Num/Float64.addf/mulf/divf`.

So the call site is not being rewritten from:

```text
App(Global(operator +), args)
```

to:

```text
App(Global(add_Int64), args)
```

even though the wrapper has already collapsed to a direct alias.

### 3. Generated C materializes these aliases as cached function values

In generated C for spectral norm, I see patterns like:

```c
static _Atomic BValue ___bsts_s___core__alpha...Bosatsu_l_Num_l_Int64_l_operator_W__h_ = 0;

static BValue ___bsts_c___core__alpha...Bosatsu_l_Num_l_Int64_l_operator_W__h_() {
    return alloc_boxed_pure_fn2(___bsts_g_Bosatsu_l_Num_l_Int64_l_add__Int64);
}

static BValue ___bsts_g___core__alpha...Bosatsu_l_Num_l_Int64_l_operator_W__h_() {
    return read_or_build(&___bsts_s___core__alpha...Bosatsu_l_Num_l_Int64_l_operator_W__h_,
        ___bsts_c___core__alpha...Bosatsu_l_Num_l_Int64_l_operator_W__h_);
}
```

and then:

```c
BValue __bsts_b_ij0 = call_fn2(___bsts_g___core__alpha...Bosatsu_l_Num_l_Int64_l_operator_W__h_(),
    __bsts_b_i0,
    __bsts_b_j0);
```

Likewise, `Bosatsu/Num/Float64.addf/mulf/divf` are also emitted as getter + `read_or_build` + `call_fn2`.

That is exactly the wrong shape for hot numeric loops.

### 4. The raw externals themselves are already tiny

The actual Int64 external implementation is already basically raw arithmetic:

```c
BValue ___bsts_g_Bosatsu_l_Num_l_Int64_l_add__Int64(BValue a, BValue b) {
  return bsts_int64_from_bits(bsts_int64_to_bits(a) + bsts_int64_to_bits(b));
}
```

and `Int64` is stored inline in a `BValue` word.

So the problem is not that the external is intrinsically expensive. The problem is that the compiler is not getting to a direct call site.

## Likely Root Cause

This looks like a mismatch between Matchless global inlining and C codegen:

### Matchless side

`MatchlessGlobalInlining` has two relevant behaviors:

1. It can inline direct lambdas in callee position via `inlineGlobal`.
2. It can inline cheap references via `inlineReferenceExpr`.

But `inlineGlobal` only handles `LambdaSummary`, not `ValueSummary`.

Relevant code:

- `core/src/main/scala/dev/bosatsu/MatchlessGlobalInlining.scala`

Specifically:

- `rewriteExpr`: around the `Matchless.App(fn, args)` handling
- `inlineGlobal`: only rewrites when the callee summary is a `LambdaSummary`
- `summarizeValue`: aliases like `operator + = global add_Int64` become `ValueSummary`, not `LambdaSummary`

So once a wrapper is normalized to a top-level alias:

```text
operator + = global add_Int64
```

the inliner no longer treats it as a direct callee rewrite opportunity.

### C backend side

`ClangGen` only treats top-level values as directly callable when they are:

- externals, or
- top-level lambdas with no captures

Relevant code:

- `core/src/main/scala/dev/bosatsu/codegen/clang/ClangGen.scala`

Specifically:

- `innerApp`: direct-calls `App(Global(...), args)` only when `directFn(...)` succeeds
- `directFn`: recognizes externals or null-capture top-level lambdas
- `renderTop`: any top-level value that is not a null-capture lambda is emitted via `read_or_build`

So a top-level alias to a direct function is currently treated as an arbitrary top-level value rather than as another direct function.

## Likely Fix

I think the best fix is in Matchless:

### Primary fix

Teach `MatchlessGlobalInlining` to chase direct function aliases in callee position.

In other words, when rewriting:

```text
App(Global(from, pack, name), args)
```

if `name` has a `ValueSummary` whose value is:

- `Global(...)`
- or another trivially direct function alias chain

then rewrite the callee to the final direct target before codegen.

Example desired transformation:

```text
App(Global(Bosatsu/Num/Int64::operator +), [a, b])
```

to:

```text
App(Global(Bosatsu/Num/Int64::add_Int64), [a, b])
```

and similarly for:

- `operator *`
- `operator /`
- `Bosatsu/Num/Float64.addf`
- `Bosatsu/Num/Float64.mulf`
- `Bosatsu/Num/Float64.divf`
- other wrapper/alias functions of the same form

This should let `ClangGen.innerApp` take the `directFn` path and emit a direct C call.

### Secondary / belt-and-suspenders fix

Also consider teaching `ClangGen.directFn` to recognize top-level aliases to direct functions/externals.

That would make the backend more robust even if Matchless does not canonicalize the callee.

## Relevant Files

- `core/src/main/scala/dev/bosatsu/MatchlessGlobalInlining.scala`
- `core/src/main/scala/dev/bosatsu/codegen/clang/ClangGen.scala`
- `test_workspace/Int64.bosatsu`
- `test_workspace/Float64.bosatsu`
- `c_runtime/bosatsu_ext_Bosatsu_l_Num_l_Int64.c`

## Nice Follow-Up Test Cases

I think this should get a codegen regression test or IR-level test around something like:

```bosatsu
package Repro

from Bosatsu/Num/Int64 import (
  Int64,
  int_low_bits_to_Int64 as i64,
  operator +,
)

def add1(x: Int64) -> Int64:
  x + i64(1)
```

Expected C shape:

- direct call to `___bsts_g_Bosatsu_l_Num_l_Int64_l_add__Int64(...)`
- no `read_or_build` / `alloc_boxed_pure_fn2` / `call_fn2` on the `operator +` path

Likewise for:

```bosatsu
from Bosatsu/Num/Float64 import addf
def f(a, b): addf(a, b)
```

which should not need to go through a function-value getter either.

## User-Visible Impact

This is not just a cosmetic codegen issue. It has a large runtime effect on numeric kernels because the missed optimization happens inside hot inner loops.

The new Array APIs in `0.0.65` help a lot, but without alias-to-direct-callee rewriting, wrapper-heavy numeric code still pays unnecessary indirect call overhead at every element of the kernel.

johnynek:
Updated `ClangGen.directFn` to follow top-level global alias chains until they resolve to an external or a null-capture lambda, so C codegen no longer routes those calls through cached function values and `call_fn$arity`. Added a focused `ClangGenTest` regression covering an `Int64` alias chain and verified the generated helper calls `add_Int64` directly. Validation: `sbt "coreJVM/testOnly dev.bosatsu.codegen.clang.ClangGenTest"` and `scripts/test_basic.sh` both passed.

Fixes #2318

### DIFF
--- a/core/src/main/scala/dev/bosatsu/codegen/clang/ClangGen.scala
+++ b/core/src/main/scala/dev/bosatsu/codegen/clang/ClangGen.scala
@@ -1449,21 +1449,35 @@ class ClangGen[K](ns: CompilationNamespace[K]) {
               b: Bindable
           ): T[Option[(Code.Ident, Int)]] =
             StateT { s =>
-              val depKey = ns.depFor(k, pack)
-              val key = (depKey, pack, b)
-              s.allValues.get(key) match {
-                case Some((fn: Matchless.Lambda[K], ident)) if fn.captures.isEmpty =>
-                  result(s, Some((ident, fn.arity)))
-                case None =>
-                  // this is external
-                  s.externals(depKey, pack, b) match {
-                    case Some((incl, ident, arity)) if arity > 0 =>
-                      val withIncl = s.include(incl)
-                      result(withIncl, Some((ident, arity)))
-                    case _ => result(s, None)
+              def loop(
+                  state: State,
+                  key: (K, PackageName, Bindable),
+                  seen: Set[(K, PackageName, Bindable)]
+              ): EitherT[Eval, Error, (State, Option[(Code.Ident, Int)])] = {
+                val (depKey, pn, bn) = key
+                if (seen(key)) result(state, None)
+                else
+                  state.allValues.get(key) match {
+                    case Some((fn: Matchless.Lambda[K], ident))
+                        if fn.captures.isEmpty =>
+                      result(state, Some((ident, fn.arity)))
+                    case Some((Matchless.Global(nextK, nextPack, nextB), _)) =>
+                      val nextKey = (ns.depFor(nextK, nextPack), nextPack, nextB)
+                      loop(state, nextKey, seen + key)
+                    case None =>
+                      // this is external
+                      state.externals(depKey, pn, bn) match {
+                        case Some((incl, ident, arity)) if arity > 0 =>
+                          val withIncl = state.include(incl)
+                          result(withIncl, Some((ident, arity)))
+                        case _ => result(state, None)
+                      }
+                    case _ =>
+                      result(state, None)
                   }
-                case _ => result(s, None)
               }
+
+              loop(s, (ns.depFor(k, pack), pack, b), Set.empty)
             }
 
           def directFn(b: Bindable): T[Option[DirectFnRef]] =
--- a/core/src/main/scala/dev/bosatsu/codegen/clang/ClangGen.scala
+++ b/core/src/main/scala/dev/bosatsu/codegen/clang/ClangGen.scala
@@ -1454,7 +1454,6 @@ class ClangGen[K](ns: CompilationNamespace[K]) {
                   key: (K, PackageName, Bindable),
                   seen: Set[(K, PackageName, Bindable)]
               ): EitherT[Eval, Error, (State, Option[(Code.Ident, Int)])] = {
-                val (depKey, pn, bn) = key
                 if (seen(key)) result(state, None)
                 else
                   state.allValues.get(key) match {
@@ -1466,6 +1465,7 @@ class ClangGen[K](ns: CompilationNamespace[K]) {
                       loop(state, nextKey, seen + key)
                     case None =>
                       // this is external
+                      val (depKey, pn, bn) = key
                       state.externals(depKey, pn, bn) match {
                         case Some((incl, ident, arity)) if arity > 0 =>
                           val withIncl = state.include(incl)
--- a/core/src/test/scala/dev/bosatsu/MatchlessTests.scala
+++ b/core/src/test/scala/dev/bosatsu/MatchlessTests.scala
@@ -5356,6 +5356,109 @@ ${tmpLines}
     }
   }
 
+  test("optimized Matchless rewrites same-package callee alias chains") {
+    val pack = PackageName.parts("Alias", "Same")
+    val targetPack = PackageName.parts("Target", "Fn")
+    val targetName = Identifier.Name("target")
+    val aliasName = Identifier.Name("alias")
+    val alias2Name = Identifier.Name("alias2")
+    val useName = Identifier.Name("use")
+    val arg = Matchless.Literal(Lit.fromInt(1))
+
+    val rawCompiled =
+      SortedMap(
+        () -> Map(
+          pack -> List(
+            (aliasName, Matchless.Global((), targetPack, targetName)),
+            (alias2Name, Matchless.Global((), pack, aliasName)),
+            (
+              useName,
+              Matchless.App(
+                Matchless.Global((), pack, alias2Name),
+                NonEmptyList.one(arg)
+              )
+            )
+          )
+        )
+      )
+    val topoSort =
+      dev.bosatsu.graph.Toposort.sort(List(((), pack)))(_ => Nil)
+
+    val optimized =
+      Par.withEC {
+        MatchlessGlobalInlining.optimize(
+          rawCompiled,
+          topoSort,
+          (_, _) => (),
+          Matchless.LocalPassOptions.Default
+        )
+      }
+    val useExpr = optimized(())(pack).toMap.apply(useName)
+
+    assertEquals(
+      useExpr,
+      Matchless.App(
+        Matchless.Global((), targetPack, targetName),
+        NonEmptyList.one(arg)
+      )
+    )
+  }
+
+  test("optimized Matchless rewrites cross-package callee alias chains") {
+    val helperPack = PackageName.parts("Alias", "Cross", "Helper")
+    val callerPack = PackageName.parts("Alias", "Cross", "Caller")
+    val targetPack = PackageName.parts("Target", "Fn")
+    val targetName = Identifier.Name("target")
+    val aliasName = Identifier.Name("alias")
+    val alias2Name = Identifier.Name("alias2")
+    val useName = Identifier.Name("use")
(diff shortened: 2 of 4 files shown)


core/src/main/scala/dev/bosatsu/codegen/clang/ClangGen.scala
```diff
@@ -1449,21 +1449,35 @@ class ClangGen[K](ns: CompilationNamespace[K]) {
...
-                    case Some((incl, ident, arity)) if arity > 0 =>
-                      val withIncl = s.include(incl)
-                      result(withIncl, Some((ident, arity)))
-                    case _ => result(s, None)
+              def loop(
+                  state: State,
+                  key: (K, PackageName, Bindable),
+                  seen: Set[(K, PackageName, Bindable)]
+              ): EitherT[Eval, Error, (State, Option[(Code.Ident, Int)])] = {
+                val (depKey, pn, bn) = key
```
johnynek: let's move this down to where it is used. I think only on line `state.externals(depKey, pn, bn) `

johnynek (commented):
This is good, but this only optimizes the C code. We want to keep this optimization because it seems more correct in code generation.

But also inlining should remove all alias type functions:
```
foo = bar

y = foo(x)
```
should *always* be `y = bar(x)` inside or outside of the package foo is defined in. This will benefit python as well as the scala interpreter.

So, we still need to unconditionally inline such alias cases so avoid indirection cost in the code.

Don't remove the Clang change you have, because we want to have good c-code generation no matter what.

core/src/main/scala/dev/bosatsu/codegen/clang/ClangGen.scala
johnynek: Moved the tuple destructure down into the external-resolution branch so `depKey` is only unpacked where it is actually used.

johnynek:
Addressed the follow-up by keeping the Clang direct-call fallback and adding the broader Matchless fix: callee alias chains now canonicalize through top-level global aliases before any lambda-inlining decision, so `foo = bar; foo(x)` rewrites through the final target within or across packages. Added Matchless regression coverage for same-package and cross-package alias chains, kept the Clang regression, and reran `sbt "coreJVM/test:compile"`, `sbt "coreJVM/testOnly dev.bosatsu.MatchlessTest"`, `sbt "coreJVM/testOnly dev.bosatsu.codegen.clang.ClangGenTest"`, and `scripts/test_basic.sh` successfully.

core/src/test/scala/dev/bosatsu/MatchlessTests.scala
```diff
@@ -5356,6 +5356,109 @@ ${tmpLines}
     }
   }
 
+  test("optimized Matchless rewrites same-package callee alias chains") {
```
johnynek, line 5395: we also need a test that exercises cases like:
```
operator + = add

# in another package

foo = operator +

baz = [1, 2, 3].foldl_List(0, foo)
```
that should get inlined by matchless global inlining: `baz = foldl_List([1, 2, 3], 0, add)` So, we should be able to inline aliases like this even when they aren't call targets.

johnynek: Added the cross-package regression you asked for in `MatchlessTests`: `operator + = add` in one package, `foo = operator +` in another, and `foldl_List(0, foo)` at the use site. The test now asserts Matchless removes both the imported operator alias and the local `foo` binding, and that `foldl_List` still devirtualizes to the loop form.

PR #2321, opened 2026-04-09
Design doc for #2320: Design: lower common Int and Int64 comparison idioms to direct Matchless predicates

johnynek:
Design doc.

Refs #2320

PR #2324, opened 2026-04-10
Lower common numeric comparison observations in Matchless

### LINKED ISSUE #2320: improve compilation of common int expressions
We very often have code like:

```
cmp_Int(x, 0) matches LT | EQ

cmp_Int(a, b) matches GT | EQ

eq_Int(x, 0)
```
and `Int64` versions of these. When asking `cmp_Int(x, 0) matches LT | EQ` we could check if the sign is negative or the value is zero. We don't want this to compile into something like 
```
let r = cmp_Int(x, 0) in
  if get_variant(r) == 0
    then enum(1)
    else
      if get_variant(r) == 1
      then enum(1)
      else enum(0)
```
we should be able to compile it into something like `enum(bsts_Int_lteq_zero(x))`

Similarly, `eq_Int(x, 0)` should be able to compile to `enum(bsts_Int_eq_zero(x))` which should basically be a an `==`.

We should have a list of Int and Int64 comparisons that we want to be able to lower into platform specific handling.

It may be that we need to do this in Matchless first by adding comparison nodes to Matchless. Maybe we could add Int64 to Matchless Literals. Then add comparison to Lit (LT, LTEQ, GT, GTEQ, NOTEQ, EQ), and then add direct code generation of those comparisons on all the platform backend.

Since we know the core comparison functions are built ins: `eq_Int, cmp_Int, eq_String, cmp_String, ....` we could lift those into Matchless LitComp when one side of the comparison is a literal. When both sides are literals, of course, we should evaluate the function.

johnynek:
Implements the #2320 Matchless comparison lowering work and addresses the approval-blocking review fixes. `Matchless.compareLiteralValues` and `MatchlessToValue` now handle `Lit.Chr` explicitly, the Comparison-subset-to-relation mapping is centralized in `Matchless.comparisonObservation` and covered by a property test, and the five affected test suites now traverse `CompareLit`, `CompareInt`, `CompareInt64`, `CompareFloat64`, and `LitInt64` explicitly instead of suppressing exhaustiveness warnings. Verification passed with `sbt "coreJVM/test:compile"`, focused suites (`dev.bosatsu.MatchlessRegressionTest`, `dev.bosatsu.MatchlessInterfaceTest`, `dev.bosatsu.Issue1633Test`), and the required `scripts/test_basic.sh`.

Fixes #2320

Implements design doc: [docs/design/2320-improve-compilation-of-common-int-expressions.md](https://github.com/johnynek/bosatsu/blob/main/docs/design/2320-improve-compilation-of-common-int-expressions.md)

Design source PR: https://github.com/johnynek/bosatsu/pull/2321

### DIFF
--- a/core/src/main/scala/dev/bosatsu/Matchless.scala
+++ b/core/src/main/scala/dev/bosatsu/Matchless.scala
@@ -3575,6 +3575,22 @@ object Matchless {
       case DataRepr.ZeroNat => 0
       case DataRepr.SuccNat => 1
     }
+    private def compareExprRelExpr[A: Order](
+        leftL: Expr[A],
+        relL: CompareRel,
+        rightL: Expr[A],
+        leftR: Expr[A],
+        relR: CompareRel,
+        rightR: Expr[A]
+    ): Int = {
+      val c1 = Order[Expr[A]].compare(leftL, leftR)
+      if (c1 != 0) c1
+      else {
+        val c2 = Order[CompareRel].compare(relL, relR)
+        if (c2 != 0) c2
+        else Order[Expr[A]].compare(rightL, rightR)
+      }
+    }
 
     given [A: Order]: Order[BoolExpr[A]] with {
       def compare(left: BoolExpr[A], right: BoolExpr[A]): Int =
@@ -3589,37 +3605,19 @@ object Matchless {
             }
 
           case (CompareInt(leftL, relL, rightL), CompareInt(leftR, relR, rightR)) =>
-            val c1 = Order[Expr[A]].compare(leftL, leftR)
-            if (c1 != 0) c1
-            else {
-              val c2 = Order[CompareRel].compare(relL, relR)
-              if (c2 != 0) c2
-              else Order[Expr[A]].compare(rightL, rightR)
-            }
+            compareExprRelExpr(leftL, relL, rightL, leftR, relR, rightR)
 
           case (
                 CompareInt64(leftL, relL, rightL),
                 CompareInt64(leftR, relR, rightR)
               ) =>
-            val c1 = Order[Expr[A]].compare(leftL, leftR)
-            if (c1 != 0) c1
-            else {
-              val c2 = Order[CompareRel].compare(relL, relR)
-              if (c2 != 0) c2
-              else Order[Expr[A]].compare(rightL, rightR)
-            }
+            compareExprRelExpr(leftL, relL, rightL, leftR, relR, rightR)
 
           case (
                 CompareFloat64(leftL, relL, rightL),
                 CompareFloat64(leftR, relR, rightR)
               ) =>
-            val c1 = Order[Expr[A]].compare(leftL, leftR)
-            if (c1 != 0) c1
-            else {
-              val c2 = Order[CompareRel].compare(relL, relR)
-              if (c2 != 0) c2
-              else Order[Expr[A]].compare(rightL, rightR)
-            }
+            compareExprRelExpr(leftL, relL, rightL, leftR, relR, rightR)
 
           case (EqualsNat(exprL, natL), EqualsNat(exprR, natR)) =>
             val c1 = Order[Expr[A]].compare(exprL, exprR)
--- a/core/src/main/scala/dev/bosatsu/Matchless.scala
+++ b/core/src/main/scala/dev/bosatsu/Matchless.scala
@@ -3534,7 +3534,7 @@ object Matchless {
       case (lhs: Lit.Chr, rhs: Lit.Chr) =>
         Some(compareRelHolds(rel, Integer.compare(lhs.toCodePoint, rhs.toCodePoint)))
       case (lhs: Lit.StringMatchResult, rhs: Lit.StringMatchResult) =>
-        Some(compareRelHolds(rel, lhs.asStr.compareTo(rhs.asStr)))
+        Some(compareRelHolds(rel, StringUtil.codePointCompare(lhs.asStr, rhs.asStr)))
       case (lhs: Lit.Float64, rhs: Lit.Float64) =>
         Some(compareFloat64Values(lhs.toDouble, rel, rhs.toDouble))
       case _ =>
--- a/core/src/main/scala/dev/bosatsu/Matchless.scala
+++ b/core/src/main/scala/dev/bosatsu/Matchless.scala
@@ -3810,27 +3810,11 @@ object Matchless {
   object EqualsLit {
     def apply[A](expr: CheapExpr[A], lit: Lit): BoolExpr[A] =
       CompareLit(expr, CompareRel.Eq, lit)
-
-    def unapply[A](boolExpr: BoolExpr[A]): Option[(CheapExpr[A], Lit)] =
-      boolExpr match {
-        case CompareLit(expr, CompareRel.Eq, lit) =>
-          Some((expr, lit))
-        case _ =>
-          None
-      }
   }
 
   object LtEqLit {
     def apply[A](expr: CheapExpr[A], lit: Lit): BoolExpr[A] =
       CompareLit(expr, CompareRel.Lte, lit)
-
-    def unapply[A](boolExpr: BoolExpr[A]): Option[(CheapExpr[A], Lit)] =
-      boolExpr match {
-        case CompareLit(expr, CompareRel.Lte, lit) =>
-          Some((expr, lit))
-        case _ =>
-          None
-      }
   }
   case class EqualsNat[A](expr: CheapExpr[A], nat: DataRepr.Nat)
       extends BoolExpr[A]
@@ -5599,6 +5583,9 @@ object Matchless {
         right: Expr[B],
         value: Boolean
     ): F[Expr[B]] =
+      // A total Comparison observation can be constant, but evaluating the
+      // original cmp_* application still has to force both operands exactly
+      // once in source order because those expressions can contain effects.
       memoizeBinaryExpr(left, right) { (_, _) =>
         if (value) TrueExpr else FalseExpr
       }
--- a/core/src/main/scala/dev/bosatsu/Matchless.scala
+++ b/core/src/main/scala/dev/bosatsu/Matchless.scala
@@ -5578,17 +5578,8 @@ object Matchless {
         If(comparePredicate(domain, leftCheap, rel, rightCheap), TrueExpr, FalseExpr)
       }
 
-    def lowerConstantBoolWithOperands(
-        left: Expr[B],
-        right: Expr[B],
-        value: Boolean
-    ): F[Expr[B]] =
-      // A total Comparison observation can be constant, but evaluating the
-      // original cmp_* application still has to force both operands exactly
-      // once in source order because those expressions can contain effects.
-      memoizeBinaryExpr(left, right) { (_, _) =>
-        if (value) TrueExpr else FalseExpr
-      }
+    def lowerConstantBool(value: Boolean): F[Expr[B]] =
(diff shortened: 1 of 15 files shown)


core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -3358,15 +3579,47 @@ object Matchless {
...
+                CompareFloat64(leftL, relL, rightL),
+                CompareFloat64(leftR, relR, rightR)
+              ) =>
+            val c1 = Order[Expr[A]].compare(leftL, leftR)
+            if (c1 != 0) c1
+            else {
+              val c2 = Order[CompareRel].compare(relL, relR)
+              if (c2 != 0) c2
+              else Order[Expr[A]].compare(rightL, rightR)
+            }
```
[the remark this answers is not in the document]
johnynek: Extracted the shared `Expr`/`CompareRel`/`Expr` ordering logic into a private helper in `Matchless.BoolExpr` and reused it for `CompareInt`, `CompareInt64`, and `CompareFloat64`.

core/src/test/scala/dev/bosatsu/Issue1633Test.scala
```diff
@@ -92,13 +92,22 @@ main = parse_value(" null")
...
+          loopExpr(expr)
+        case Matchless.CompareInt(left, _, right) =>
+          loopExpr(left) || loopExpr(right)
+        case Matchless.CompareInt64(left, _, right) =>
+          loopExpr(left) || loopExpr(right)
+        case Matchless.CompareFloat64(left, _, right) =>
+          loopExpr(left) || loopExpr(right)
         case Matchless.EqualsLit(expr, _) =>
           loopExpr(expr)
         case Matchless.LtEqLit(expr, _) =>
```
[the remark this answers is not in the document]
johnynek: Removed the unreachable `EqualsLit` and `LtEqLit` match arms from the affected bool-walker helpers. The cleanup is applied in `Issue1633Test.scala`, `MatchlessInterfaceTest.scala`, `MatchlessRegressionTest.scala`, `MatchlessTests.scala`, and `ToolAndLibCommandTest.scala`; the remaining `LtEqLit` references in `MatchlessTests.scala` are direct assertions, not shadowed helper cases.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -3329,19 +3475,86 @@ object Matchless {
...
+      rel: CompareRel,
+      right: Lit
+  ): Option[Boolean] =
+    (left, right) match {
+      case (Lit.Integer(lhs), Lit.Integer(rhs)) =>
+        Some(compareRelHolds(rel, lhs.compareTo(rhs)))
+      case (lhs: Lit.Chr, rhs: Lit.Chr) =>
+        Some(compareRelHolds(rel, Integer.compare(lhs.toCodePoint, rhs.toCodePoint)))
+      case (lhs: Lit.StringMatchResult, rhs: Lit.StringMatchResult) =>
+        Some(compareRelHolds(rel, lhs.asStr.compareTo(rhs.asStr)))
```
johnynek: Bosatsu string ordering is `StringUtil.codePointCompare` (`Predef.cmp_String` uses that), but this now folds `CompareLit` with JVM `String.compareTo`. Those orders diverge for astral Unicode strings, e.g. `"\uE000"` vs `"\U00010000"`: `codePointCompare` says LT while `String.compareTo` says GT. That lets constant-folding produce the opposite boolean from runtime evaluation. Please switch this branch to `StringUtil.codePointCompare(lhs.asStr, rhs.asStr)` and add a regression covering that counterexample.

core/src/main/scala/dev/bosatsu/MatchlessToValue.scala
```diff
@@ -245,71 +245,99 @@ object MatchlessToValue {
...
+                  rel,
+                  Integer.compare(s.codePointAt(0), rhs.toCodePoint)
+                )
+              case _ =>
+                false
+            }
+          case rhs: Lit.StringMatchResult =>
+            left match {
+              case s: String =>
+                Matchless.compareRelHolds(rel, s.compareTo(rhs.asStr))
```
johnynek: Same semantic issue in the evaluator: Bosatsu string comparison is defined by `StringUtil.codePointCompare`, but `CompareLit` over strings is using JVM `String.compareTo` here. The IR interpreter can therefore disagree with `cmp_String` and with the code generators on astral Unicode strings. Please use `StringUtil.codePointCompare(s, rhs.asStr)` in this branch and cover it with the same regression.

core/src/main/scala/dev/bosatsu/codegen/python/PythonGen.scala
```diff
@@ -3227,38 +3268,101 @@ object PythonGen {
...
-                loop(expr, slotName, inlineSlots)
-                  .flatMap(Env.onLast(_)(ex => !(literal :< ex)))
-              case _ =>
-                // $COVERAGE-OFF$
-                throw new IllegalStateException(
-                  s"LtEqLit only supports Int and Char literals, found: $lit"
-                )
-              // $COVERAGE-ON$
-            }
+              .flatMap(Env.onLast(_)(ex => compareRelExpr(ex, rel, literal)))
```
johnynek, line 3367: This generic `CompareLit` fast path now sends string literal relations through raw Python `<`/`<=`/`>` instead of the existing `cmp_String` helper. The helper in this file (`buildPyStringCmpHelper`) exists specifically because some Python runtimes do not order strings by Bosatsu's Unicode-scalar semantics. Bypassing it here reintroduces that cross-runtime mismatch for string literal comparisons, so string literals should keep using the helper and compare its `Comparison` result instead of going through `compareRelExpr` directly.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Switched `Matchless.compareLiteralValues` to `StringUtil.codePointCompare(...)` for string literals, so constant folding now matches Bosatsu string semantics even on the `"\uE000"` vs `"\U00010000"` counterexample. I also added a regression in `MatchlessTests` that checks both the folded result and the corresponding `CompareLit` evaluation across all six relations for that exact pair.

core/src/main/scala/dev/bosatsu/MatchlessToValue.scala
johnynek: Updated the Matchless evaluator to use `StringUtil.codePointCompare(...)` in the `CompareLit` string branch as well. The new `MatchlessTests` regression exercises `evalBoolExpr` on the astral-Unicode counterexample, so it now guards both the constant-folding path and the IR interpreter path.

core/src/main/scala/dev/bosatsu/codegen/python/PythonGen.scala
johnynek: Routed ordered non-empty `CompareLit` string relations through the existing Python `cmp_String` helper path and compare its `Comparison` result instead of using raw Python ordering. I kept `Eq`/`Ne` and empty-string literals on direct operators because those cases are semantically identical and avoiding the helper there prevents the Jython module-size regression in `PredefTests`. Added a `PythonGenTest` regression that checks the astral-literal ordered case still emits the semantic probe/helper and does not fall back to a raw `<` literal comparison.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -3525,8 +3790,48 @@ object Matchless {
...
+  case class CompareFloat64[A](
+      left: CheapExpr[A],
+      rel: CompareRel,
+      right: CheapExpr[A]
+  ) extends BoolExpr[A]
+  object EqualsLit {
+    def apply[A](expr: CheapExpr[A], lit: Lit): BoolExpr[A] =
+      CompareLit(expr, CompareRel.Eq, lit)
+
+    def unapply[A](boolExpr: BoolExpr[A]): Option[(CheapExpr[A], Lit)] =
```
johnynek: let's remove the unapply.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -3525,8 +3790,48 @@ object Matchless {
...
+        case _ =>
+          None
+      }
+  }
+
+  object LtEqLit {
+    def apply[A](expr: CheapExpr[A], lit: Lit): BoolExpr[A] =
+      CompareLit(expr, CompareRel.Lte, lit)
+
+    def unapply[A](boolExpr: BoolExpr[A]): Option[(CheapExpr[A], Lit)] =
```
johnynek: let's remove the unapply.

core/src/main/scala/dev/bosatsu/Matchless.scala
```diff
@@ -5177,6 +5484,205 @@ object Matchless {
...
+      memoizeBinaryExpr(left, right) { (leftCheap, rightCheap) =>
+        If(comparePredicate(domain, leftCheap, rel, rightCheap), TrueExpr, FalseExpr)
+      }
+
+    def lowerConstantBoolWithOperands(
+        left: Expr[B],
+        right: Expr[B],
+        value: Boolean
+    ): F[Expr[B]] =
+      memoizeBinaryExpr(left, right) { (_, _) =>
```
johnynek: why do we need the `memoizeBinazyExpr` here and not just `Monad[F].pure(if (value) TrueExpr else FalseExpr)`?

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Removed the `EqualsLit.unapply` extractor and kept `EqualsLit` as an `apply`-only compatibility constructor. The remaining code now matches `CompareLit(..., Eq, ...)` directly instead of going back through an extractor.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Removed the `LtEqLit.unapply` extractor as well. I updated the remaining matcher-based assertions to look for `CompareLit(..., Lte, ...)` directly, so there’s no extractor dependency left.

core/src/main/scala/dev/bosatsu/Matchless.scala
johnynek: Kept `memoizeBinaryExpr` here and added an inline comment explaining why: even when the observed boolean result is constant, lowering still has to evaluate both original `cmp_*` operands exactly once in source order because those Matchless expressions can contain local effects. Returning a bare `TrueExpr`/`FalseExpr` would skip that evaluation and change semantics.

johnynek: I don't think this is right. The only effect in `F[A]` is the allocation of a new unique numeric identifier. If we aren't using the thing that allocated the numbers, we don't need the identifier to be allocated. Can you double check and read the fact that the only thing `F[_]` has is `Monad[F]` and the ability to allocate a new long: `F[Long]`? It doesn't model any side effects.

If I am missing something, it's obviously subtle, so add a comment to help the next reader. If I am right, refactor this to avoid evaluating `left` and `right` and just return the pure value.

johnynek: Re-read the lowering with that constraint in mind and you were right: this `F` only carries fresh-name allocation, so the constant-observation path does not need to memoize or evaluate `left`/`right`. I replaced `lowerConstantBoolWithOperands` with a pure `lowerConstantBool` helper and updated the two call sites to return `TrueExpr`/`FalseExpr` directly.

Issue #2340, opened 2026-04-28
Add basic support for libuv in C backend


johnynek:
Using the same approach as libgc, we should build libuv on the target platform, not assume or require that it exists already.

Then we should replace the existing C functions that we have with equivalent libuv functions. To do this, we would need to improve the Prog loop and the main function that runs Prog instances be inside of an uv_loop I think.

I think the main challenge is that when we do IO that returns a `Prog` value, how do we interpret that? I think the way to do it would be to suspend the stack as a data value and attach it to the call to the IO. Then in the completion of that IO, we restore the stack and start interpreting again. Something like that.

We need to take special care and be 100% sure we integrate libgc with libuv safely. We need to make sure libgc is built with threading support which libuv uses.

We don't want any new bosatsu functions or surface area ideally, but we do need to support all old Main and Test code.

johnynek:
MergeXO assigned an agent and started roadmap planning work for issue #2340.

johnynek:
Opened roadmap PR: https://github.com/johnynek/bosatsu/pull/2341

johnynek:
MergeXO roadmap sizing recommendation:
- detected nodes: 8
- recommended max: 7
Suggestion: split this into a roadmap-of-roadmaps where practical.

johnynek:
MergeXO roadmap sizing recommendation:
- detected nodes: 8
- recommended max: 7
This is non-blocking; consider splitting into a roadmap-of-roadmaps.

johnynek:
MergeXO adjustment gate opened a same-roadmap revision PR before issuing the next frontier.
- ready frontier: io_files
- revision PR: https://github.com/johnynek/bosatsu/pull/2355
- revision branch: `agent/roadmap/2340-revision-v2`
- target graph version: `2`
- summary: `io_files` is a sound next executable slice, but the pending validation node has a direct-handoff gap that should be fixed before more child work is issued.

The current ready frontier is `io_files`. Its direct inputs are satisfied: `design` has produced `docs/design/2342-document-the-libuv-c-runtime-integration-contract.md`, `async_resume` is implemented, and `io_phase1` is implemented. Repository inspection matches the planned slice: libuv is vendored, the Prog runtime has loop/suspend-resume infrastructure, time/sleep/env have moved onto libuv runtime plumbing, and file/directory operations remain the next migration area in `c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Core.c`.

[...]

PR #2355, opened 2026-04-30
Roadmap revision for #2340 (v2)

johnynek:
Automated same-roadmap revision.

- target graph version: `2`
- roadmap doc path: `docs/roadmap/2340-add-basic-support-for-libuv-in-c-backend.md`
- graph path: `docs/roadmap/2340-add-basic-support-for-libuv-in-c-backend.graph.json`
- summary: `io_files` is a sound next executable slice, but the pending validation node has a direct-handoff gap that should be fixed before more child work is issued.

The current ready frontier is `io_files`. Its direct inputs are satisfied: `design` has produced `docs/design/2342-document-the-libuv-c-runtime-integration-contract.md`, `async_resume` is implemented, and `io_phase1` is implemented. Repository inspection matches the planned slice: libuv is vendored, the Prog runtime has loop/suspend-resume infrastructure, time/sleep/env have moved onto libuv runtime plumbing, and file/directory operations remain the next migration area in `c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Core.c`.

The graph still needs revision because the unstarted `validation` node explicitly needs the implemented vendored-libuv state to validate link flags, `GC_THREADS`, and generated C runtime install behavior, but it does not directly depend on `vendored_libuv`. Under the roadmap dependency handoff contract, relying on `loop_core` to transitively carry that satisfied state is not valid. This is a missing direct artifact/state handoff, not an unsafe `io_files` frontier. The revision preserves started nodes and the next two frontiers (`io_files`, then `process_stdio`) and replaces only the unstarted validation node with an equivalent node that directly receives `vendored_libuv` as implemented input.

Refs #2340

### DIFF
--- a/docs/roadmap/2340-add-basic-support-for-libuv-in-c-backend.graph.json
+++ b/docs/roadmap/2340-add-basic-support-for-libuv-in-c-backend.graph.json
@@ -104,7 +104,7 @@
       "title": "Implement libuv-backed process spawn and wait for C IO/Core"
     },
     {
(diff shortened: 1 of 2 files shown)


johnynek:
No repository edit was needed. I read the issue-tree context and candidate diff, re-read both roadmap artifacts, and verified the graph parses as JSON. The v2 revision already contains the direct `vendored_libuv` implemented dependency and matching worker handoff text for `validation_with_vendored`; adding explicit `promised_artifact_path: null` fields would be canonicalized away and is not a meaningful change. I did not run `scripts/test_basic.sh` because I am not returning a commit_message.

PR #2370, opened 2026-05-02
Implement portable process stop and status APIs

### LINKED ISSUE #2369: Implement portable terminate, kill, poll, and wait_timeout APIs
Parent roadmap: #2361
Roadmap node: low_level_api

Dependency handoff:
- direct dependencies only; MergeXO resolved them ahead of time.
- contract [reference_doc] requires=planned satisfied_by=planned
  child issue: #2365 (https://github.com/johnynek/bosatsu/issues/2365)
  PR: #2366 (https://github.com/johnynek/bosatsu/pull/2366)
  primary artifact: docs/design/2365-specify-the-portable-process-stop-and-status-contract.md
  changed files: docs/design/2365-specify-the-portable-process-stop-and-status-contract.md
  issue notes: MergeXO assigned an agent and started reference document work for issue #2365.

 | Opened reference doc PR: https://github.com/johnynek/bosatsu/pull/2366
- wait_state [small_job] requires=implemented satisfied_by=implemented
  child issue: #2367 (https://github.com/johnynek/bosatsu/issues/2367)
  PR: #2368 (https://github.com/johnynek/bosatsu/pull/2368)
  changed files: c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Core.c, c_runtime/test.c, core/src/main/scala/dev/bosatsu/Predef.scala, core/src/test/scala/dev/bosatsu/EvaluationTest.scala, docs/code-plans/2367-prepare-process-state-for-idempotent-wait-and-status-queries.json
  issue notes: MergeXO assigned an agent and started small-job PR work for issue #2367.

 | Opened small-job PR: https://github.com/johnynek/bosatsu/pull/2368

Direct inputs: receive the merged reference document from `contract` as a `planned` dependency artifact, and receive the implemented stable process state from `wait_state` as a satisfied `implemented` dependency state. Use both directly; do not reconstruct these requirements from roadmap history.

Add the low-level public API to `test_workspace/Bosatsu/IO/Core.bosatsu`: `StopResult` with `StopSent` and `AlreadyExited`, plus external `terminate`, `kill`, `poll`, and `wait_timeout` functions. Export the new type and functions consistently with existing `spawn` and `wait` exports.

Implement the corresponding runtime externals for JVM in `core/src/main/scala/dev/bosatsu/Predef.scala`, Python in `test_workspace/ProgExt.py` and `test_workspace/Prog.bosatsu_externals`, and C/libuv in `c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Core.c` and `.h`. JVM should map to `Process.destroy()` and `destroyForcibly()`. Python should map to `Popen.terminate()` and `Popen.kill()`. C/libuv should operate on the `uv_process_t` handle and use the strongest reviewed portable mapping from the reference document; do not expose raw signal names or pid-based public behavior.

Low-level behavior must satisfy these contracts:

- `terminate` and `kill` return `AlreadyExited` for an already-recorded exit.
- `terminate` and `kill` return `StopSent` when a stop request is sent.
- `poll` returns `Some(exit_code)` only after a recorded exit.
- `poll` returns `None` while the process is still running.
- `wait_timeout` returns `None` on timeout without consuming the final result.
- Subsequent `wait` calls still return the stable exit code after `wait_timeout` times out.
- Stop operations do not automatically close returned stdin/stdout/stderr handles.

Add focused tests for each backend where the repository already exercises process externals. Include idempotent stop on an already-exited child, stop followed by wait, kill escalation for a long-running child, poll before/after exit, and wait_timeout timeout followed by successful wait. Keep platform-sensitive commands guarded or factored through existing test helpers.

Completion outcome: the portable low-level stop/status API is publicly available and implemented across JVM, Python, and C/libuv with tests passing.

johnynek:
Adds the low-level `StopResult`, `terminate`, `kill`, `poll`, and `wait_timeout` API to `Bosatsu/IO/Core`, with external wiring across JVM, Python, and C/libuv runtimes. The implementations preserve stable final process status across `wait`, `poll`, and timeout observations, return `AlreadyExited` for completed children, send direct-child terminate/kill requests for running children, and keep timeout observations non-consuming.

The branch also adds shared JVM/Python process coverage and focused C runtime tests for poll before/after exit, timeout followed by wait, terminate/kill followed by wait, already-exited stop behavior, and oversized timeout conversion clamping in Python and C/libuv.

Fixes #2369


c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Core.c
```diff
@@ -2887,11 +3106,33 @@ static BValue bsts_core_now_mono_effect(BValue unit)
...
   {
+    return 0ULL;
+  }
+
+  BValue divmod = bsts_integer_div_mod(
+      nanos_value,
+      bsts_integer_from_int(1000000));
+  BValue millis_value = get_struct_index(divmod, 0);
+  BValue nanos_remainder = get_struct_index(divmod, 1);
+  BValue max_millis_value = bsts_integer_from_uint64(UINT64_MAX);
```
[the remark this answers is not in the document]
johnynek: Addressed by caching the `UINT64_MAX` Bosatsu integer in a function-local static inside `bsts_timeout_millis_from_nanos_value`, so repeated timeout conversions reuse the same value instead of allocating it each call. `scripts/test_basic.sh` passes.

johnynek:
Addressed the C runtime timeout conversion allocation feedback and verified with `scripts/test_basic.sh` passing.

c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Core.c
```diff
@@ -95,6 +96,8 @@ typedef struct
 
 #define BSTS_CORE_PROCESS_MAGIC 0x42505350u
 
+typedef struct BSTS_Core_Wait_Request BSTS_Core_Wait_Request;
```
johnynek, line 101: please add a comment why we have this. I think it is because BSTS_Core_Wait_Request is recursive. Is that right?

c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Core.c
```diff
@@ -2856,13 +2917,171 @@ static BValue bsts_core_wait_effect(BValue process)
...
+  if (process_state->exited)
+  {
+    return ___bsts_g_Bosatsu_l_Prog_l_pure(
+        bsts_core_process_exit_option(process_state));
+  }
+
+  BValue zero = bsts_integer_from_int(0);
+  if (bsts_integer_cmp(duration, zero) <= 0)
+  {
+    GC_reachable_here(args2);
```
johnynek, line 3064: why do we need this here if we have the `GC_reachable_here(args2)` right before the return?

core/src/main/scala/dev/bosatsu/Predef.scala
```diff
@@ -1750,15 +1751,22 @@ object PredefImpl {
   private final class ProcessValue(val process: java.lang.Process) {
     private var cachedExitCode: Option[Int] = None
```
johnynek, line 1755: does this not need to be an `AtomicReference` or something? Add a comment as to why not.

core/src/main/scala/dev/bosatsu/Predef.scala
```diff
@@ -1750,15 +1751,22 @@ object PredefImpl {
...
 
+    private def recordExitCode(code: Int): Int = {
```
johnynek, line 1759: I think if this were `private inline def recordExitCode(inline code: Int): Int ` We could simplify the code below and just do: `synchronized { try Right(recordExitCode(process.waitFor()) ... ` 

If we want to be clearer that it's a function:
```scala
private inline def recordExitCode(inline code: () => Int): Int 
```

Then the call sites are clearer, but due to inline it should be just as efficient.

In any case, add a comment as to the motivation.

c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Core.c
johnynek: Added a comment above the forward declaration explaining that `BSTS_Core_Wait_Request` is recursive with `BSTS_Core_Process`: requests link to the process, and the process stores the pending request.

c_runtime/bosatsu_ext_Bosatsu_l_IO_l_Core.c
johnynek: Added a comment on the non-positive timeout branch. That branch returns before the later `GC_reachable_here(args2)`, so it needs its own keep-alive before returning `None`.

core/src/main/scala/dev/bosatsu/Predef.scala
johnynek: Added a comment explaining why `cachedExitCode` does not need an `AtomicReference`: every read and write is guarded by the `ProcessValue` monitor through synchronized methods.

core/src/main/scala/dev/bosatsu/Predef.scala
johnynek: Added a motivation comment for `recordExitCode`: multiple operations can be first to observe process exit, so the helper centralizes the write-once cache used by wait, poll, and timeout.

johnynek:
Addressed the review feedback by documenting the recursive C wait-request declaration, the early-return GC keep-alive, and the JVM process-exit cache synchronization/motivation. `scripts/test_basic.sh` passes.

